PR #3, opened 2015-03-31
Fix the rotation animation on the thumbnails and object.

tflynnt:
Original had +=1 which was incrementing the angle on each frame.  Ended
    up being crazy.  Just needs to be =1 for the animation to look proper.
    Oh, and slowed down the thumbnail animation so one can see the thumbnail
    long enough before clicking on it.

flynnt@acm.org


samsungax:
make comment to see if you see it

JonShemitz:
Looks sensible; works fine.

chenchao1407:
Looks much better now.

PR #8, opened 2015-04-09
Port to the Oculus 0.5.0 sdk.

thomasflynn:
The 0.5.0 sdk has some minor API changes.  Mostly adding some extra
arguments to some methods.  This change addresses that.  (rebased edition)

tom.flynn@samsung.com

### DIFF
--- a/GVRf/Framework/jni/oculus/activity_jni.cpp
+++ b/GVRf/Framework/jni/oculus/activity_jni.cpp
@@ -16,6 +16,7 @@
 #include "activity_jni.h"
 #include <jni.h>
 #include <glm/gtc/type_ptr.hpp>
+#include <VrApi/VrApi_Helpers.h>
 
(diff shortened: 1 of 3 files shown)


thomasflynn:
Ok, this is rebased and the spacing should be fixed.  Once this goes in, we'll need to update the build instructions in the Dev Guide.  I can take care of that.

JonShemitz:
The code looks fine - will switch to new VRLib and verify, after lunch. (If I verify now, there may be no food by the time I get down there.)

JonShemitz:
Verified. We can pull the change as soon as we've updated the Developer's Guide.

chenchao1407:
Verified + 1. Code Review + 2 :-)

thomasflynn:
ok, changed the build instructions on the wiki.  The pdf of the Dev Guide still needs to be updated, re-generated and uploaded to the wiki, but that's a lengthy process that can be done tomorrow.

PR #11, opened 2015-04-13
Fix another couple of object-identity bugs.

JonShemitz:
1. When an app creates a scene object and doesn't immediately
   add it to the scene graph (or when the object is repeatedly
   removed and re-added) the object-identity code may see that
   there are no native references to the object, and remove
   the wrapper from the list of 'registered' wrappers. The
   object is still 'alive' and still perfectly usable - but
   object identity has been lost: eye picking, for example,
   will return a 'generic' GVRSceneObject, not the actual
   instance that was put in (which will often be a sub-classed
   type, containing important app information.)

   This patch adds a keep-wrapper flag, that will preserve
   object identity, even for objects that have not been
   added to the scene graph.

2. The 'dereference thread' scans for objects that have
   been removed from the scene graph every N allocations.
   In large, complex apps, it's not at all impossible that
   it will 'wake up' two or more times during onInit() or
   onStep(). This patch adds a generation counter to each
   GVRHybridObject, so that no object will be 'dereferenced'
   until the app has had a chance to either add it to the
   scene graph or set the keep-wrapper flag.

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/GVRHybridObject.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRHybridObject.java
@@ -13,7 +13,6 @@
  * limitations under the License.
  */
 
-
 package org.gearvrf;
 
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


jason2kim:
This can also work as a solution to the known issue of gvr video sample: 

signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 00000000 
#00  pc 00172a5c  /data/app-lib/com.samsung.android.gvr.video-3/libgvrf.so (gvr::Texture::getId() const+4)

    virtual GLuint getId() const {
        return gl_texture_->id();
    }
   Where, gl_texture_ == 0 as it was recycled already.

chenchao1407:
I thought this would fix that crash but actually not. However I am still considering this patch is related to the crash since it changes the timing of crash in gvrvideo.

chenchao1407:
Anyway, it works well with other apps. I think we can merge it.

PR #12, opened 2015-04-13
Fix data check on GVRMesh.setVecXVector(), also put check on all setX().

chenchao1407:
Add this exception mechanism to make sure it won't crash at native level without any useful information or call stack.

c.chen@samsung.com

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/GVRMesh.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRMesh.java
@@ -15,6 +15,8 @@
 
 package org.gearvrf;
 
+import org.gearvrf.utility.Exceptions;
+
 /**
  * This is one of the key GVRF classes: It holds GL meshes.
  * 
@@ -64,6 +66,8 @@ public class GVRMesh extends GVRHybridObject {
      *            Array containing the packed vertex data.
      */
     public void setVertices(float[] vertices) {
+        isNonNull(vertices);
+        isValidData(vertices.length, 3);
         NativeMesh.setVertices(getPtr(), vertices);
     }
 
@@ -89,6 +93,8 @@ public class GVRMesh extends GVRHybridObject {
      *            Array containing the packed normal data.
      */
     public void setNormals(float[] normals) {
+        isNonNull(normals);
+        isValidData(normals.length, 3);
         NativeMesh.setNormals(getPtr(), normals);
     }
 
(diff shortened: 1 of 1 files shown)


chenchao1407:
Hi, Jon.

I modified the setVecXVector based on the Mihail's comment on my previous one.

You can see that in mesh_jni.cpp, the logic of setVec2Vector is as the following:
    std::vector<glm::vec2> native_vec2_vector;
    for (int i = 0; i < vec2_vector_length; ++i) {
        native_vec2_vector.push_back(vec2_vector_pointer[i]);
    }

It is similar in other setVecXVector, so that this is why I check the input like that.

JonShemitz:
Yes. So your test will keep that code from breaking. But what happens if you pass an array 4 floats (two vec2-s) to a mesh with 64 vertices?

chenchao1407:
And also, the reason why I use switch-case is mostly for null pointer check. Otherwise we need to do another null pointer check as you point out.

I can use the mechanism you indicate. But I think the logic of this patch should be fine.

chenchao1407:
OK. I got your point. I am going to modify the codes and update this pull request.

chenchao1407:
Pull request updated.

chenchao1407:
Hi Jon
This is updated based on your comments. Can you help check again?

Thanks in advance.

GVRf/Framework/src/org/gearvrf/GVRMesh.java
```diff
@@ -272,13 +291,31 @@ public GVRMesh getBoundingBox() {
...
+            throw Exceptions
+                    .IllegalArgument(
+                            "The input array should be an array of %d-component elements whose length is non-zero and divisible by %d. But current data length is %d.",
+                            expectedComponents, expectedComponents, dataLength);
+        }
+    }
+
+    private void verticesCheck(int dataLength, int expectedComponents) {
+        int verticesNumber = getVertices().length / 3;
+        int numberOfElements = dataLength / expectedComponents;
```
JonShemitz: What if there are 12 vertices; dataLength is 37; and expectedComponents is 3?

JonShemitz:
This looks pretty good. The error check is not 100% accurate but I'm not sure that's 100% necessary ...

GVRf/Framework/src/org/gearvrf/GVRMesh.java
chenchao1407: Hi Jon.

I think for this issue, the isValidData has already taken care of that so it should be fine.

JonShemitz:
Yes, thanks - I missed the way setValidData() has already done the modulus check.

PR #14, opened 2015-04-15
Post effect: add a way to set uniform mat4.

KolorCompany:


JonShemitz:
This looks fine to me. Not 100% sure how I'd test it. (Tom, do you have any ideas?)

chenchao1407:
Code looks fine.  Source built and compiled. May need to find a way to test those additional APIs.

chenchao1407:
Tested with apps which use post-effect and everything runs well.

PR #15, opened 2015-04-15
Add 2D screenshot APIs for left eye, right eye, and center eye.

rongguodong:
rongguodong@hotmail.com

### DIFF
--- a/GVRf/Framework/jni/oculus/view_manager.cpp
+++ b/GVRf/Framework/jni/oculus/view_manager.cpp
@@ -18,41 +18,67 @@
 #include <jni.h>
 #include "../engine/renderer/renderer.h"
 #include "../objects/components/camera.h"
+
 namespace gvr {
 
-extern "C"
-{
+extern "C" {
 
-void Java_org_gearvrf_GVRViewManager_renderCamera(
-        JNIEnv * jni, jclass clazz, jlong appPtr,
-        jlong jscene, jlong jcamera, jlong jrender_texture,
-        jlong jshader_manager, jlong jpost_effect_shader_manager,
-        jlong jpost_effect_render_texture_a,
-        jlong jpost_effect_render_texture_b
-        )
-{
-    GVRActivity *activity = (GVRActivity*)((App *)appPtr)->GetAppInterface();
+void Java_org_gearvrf_GVRViewManager_renderCamera(JNIEnv * jni, jclass clazz,
+		jlong appPtr, jlong jscene, jlong jcamera, jlong jrender_texture,
+		jlong jshader_manager, jlong jpost_effect_shader_manager,
+		jlong jpost_effect_render_texture_a,
+		jlong jpost_effect_render_texture_b) {
+	GVRActivity *activity = (GVRActivity*) ((App *) appPtr)->GetAppInterface();
 
-    std::shared_ptr<Scene> scene =
-            *reinterpret_cast<std::shared_ptr<Scene>*>(jscene);
-    std::shared_ptr<Camera> camera =
-            *reinterpret_cast<std::shared_ptr<Camera>*>(jcamera);
-    std::shared_ptr<RenderTexture> render_texture =
-            *reinterpret_cast<std::shared_ptr<RenderTexture>*>(jrender_texture);
-    std::shared_ptr<ShaderManager> shader_manager =
-            *reinterpret_cast<std::shared_ptr<ShaderManager>*>(jshader_manager);
-    std::shared_ptr<PostEffectShaderManager> post_effect_shader_manager =
-            *reinterpret_cast<std::shared_ptr<PostEffectShaderManager>*>(jpost_effect_shader_manager);
-    std::shared_ptr<RenderTexture> post_effect_render_texture_a =
-            *reinterpret_cast<std::shared_ptr<RenderTexture>*>(jpost_effect_render_texture_a);
-    std::shared_ptr<RenderTexture> post_effect_render_texture_b =
-            *reinterpret_cast<std::shared_ptr<RenderTexture>*>(jpost_effect_render_texture_b);
+	std::shared_ptr<Scene> scene =
+			*reinterpret_cast<std::shared_ptr<Scene>*>(jscene);
+	std::shared_ptr<Camera> camera =
+			*reinterpret_cast<std::shared_ptr<Camera>*>(jcamera);
+	std::shared_ptr<RenderTexture> render_texture =
+			*reinterpret_cast<std::shared_ptr<RenderTexture>*>(jrender_texture);
+	std::shared_ptr<ShaderManager> shader_manager =
+			*reinterpret_cast<std::shared_ptr<ShaderManager>*>(jshader_manager);
+	std::shared_ptr<PostEffectShaderManager> post_effect_shader_manager =
+			*reinterpret_cast<std::shared_ptr<PostEffectShaderManager>*>(jpost_effect_shader_manager);
+	std::shared_ptr<RenderTexture> post_effect_render_texture_a =
+			*reinterpret_cast<std::shared_ptr<RenderTexture>*>(jpost_effect_render_texture_a);
+	std::shared_ptr<RenderTexture> post_effect_render_texture_b =
+			*reinterpret_cast<std::shared_ptr<RenderTexture>*>(jpost_effect_render_texture_b);
 
-    activity->viewManager->renderCamera(activity->Scene,
-            scene, camera, render_texture, shader_manager,
-            post_effect_shader_manager, post_effect_render_texture_a,
-            post_effect_render_texture_b,
-            activity->viewManager->mvp_matrix);
+	activity->viewManager->renderCamera(activity->Scene, scene, camera,
+			render_texture, shader_manager, post_effect_shader_manager,
+			post_effect_render_texture_a, post_effect_render_texture_b,
+			activity->viewManager->mvp_matrix);
+}
+
+void Java_org_gearvrf_GVRViewManager_readRenderResult(JNIEnv * jni,
+		jclass clazz, jlong jrender_texture, jobject jreadback_buffer) {
+
+	uint8_t *pReadbackBuffer = (uint8_t *) jni->GetDirectBufferAddress(
+			jreadback_buffer);
+	std::shared_ptr<RenderTexture> render_texture =
+			*reinterpret_cast<std::shared_ptr<RenderTexture>*>(jrender_texture);
+	int width = render_texture->width();
+	int height = render_texture->height();
+
+	// remember current FBO bindings
+	GLint currentReadFBO, currentDrawFBO;
+	glGetIntegerv(GL_READ_FRAMEBUFFER_BINDING, &currentReadFBO);
+	glGetIntegerv(GL_DRAW_FRAMEBUFFER_BINDING, &currentDrawFBO);
+
+	// blit the multisampled FBO to a normal FBO and read from it
+	GLuint renderTextureFBO = render_texture->getFrameBufferId();
+	glBindFramebuffer(GL_DRAW_FRAMEBUFFER, renderTextureFBO);
+	glBlitFramebuffer(0, 0, width, height, 0, 0, width, height,
+			GL_COLOR_BUFFER_BIT, GL_NEAREST);
+	glBindFramebuffer(GL_READ_FRAMEBUFFER, renderTextureFBO);
+
+	glReadPixels(0, 0, width, height, GL_RGBA, GL_UNSIGNED_BYTE,
+			pReadbackBuffer);
+
+	// recover FBO bindings
+	glBindFramebuffer(GL_READ_FRAMEBUFFER, currentReadFBO);
+	glBindFramebuffer(GL_DRAW_FRAMEBUFFER, currentDrawFBO);
 }
 
 } // extern "C"
@@ -61,30 +87,28 @@ void Java_org_gearvrf_GVRViewManager_renderCamera(
 //                             GVRViewManager
 //=============================================================================
 
-GVRViewManager::GVRViewManager(JNIEnv & jni_, jobject activityObject_)
-{
-    // initial
+GVRViewManager::GVRViewManager(JNIEnv & jni_, jobject activityObject_) {
+	// initial
 	m_frameRendered = 0;
-    m_fps = 0.0;
-    m_startTime = m_currentTime = 0.0f;
-    gNumFrame = 0;
+	m_fps = 0.0;
+	m_startTime = m_currentTime = 0.0f;
+	gNumFrame = 0;
 
-    LOG("GVRViewManager::GVRViewManager");
+	LOG("GVRViewManager::GVRViewManager");
 }
 
 GVRViewManager::~GVRViewManager() {
-    LOG( "GVRViewManager::~GVRViewManager()");
+	LOG("GVRViewManager::~GVRViewManager()");
 }
 
 void GVRViewManager::renderCamera(OvrSceneView &ovr_scene,
-            std::shared_ptr<Scene> scene,
-            std::shared_ptr<Camera> camera,
-            std::shared_ptr<RenderTexture> render_texture,
-            std::shared_ptr<ShaderManager> shader_manager,
-            std::shared_ptr<PostEffectShaderManager> post_effect_shader_manager,
-            std::shared_ptr<RenderTexture> post_effect_render_texture_a,
-            std::shared_ptr<RenderTexture> post_effect_render_texture_b,
-            glm::mat4 mvp) {
+		std::shared_ptr<Scene> scene, std::shared_ptr<Camera> camera,
+		std::shared_ptr<RenderTexture> render_texture,
+		std::shared_ptr<ShaderManager> shader_manager,
+		std::shared_ptr<PostEffectShaderManager> post_effect_shader_manager,
+		std::shared_ptr<RenderTexture> post_effect_render_texture_a,
+		std::shared_ptr<RenderTexture> post_effect_render_texture_b,
+		glm::mat4 mvp) {
 #ifdef GVRF_FBO_FPS
 	// starting to collect rendering time
 	// first flash GPU tasks
@@ -97,20 +121,20 @@ void GVRViewManager::renderCamera(OvrSceneView &ovr_scene,
 	gettimeofday(&start, NULL);
 #endif
 
-    if( camera->render_mask() == 1){
-        glClearColor( 0.0f, 1.0f, 0.0f, 1.0f );
-    }else{
-        glClearColor( 1.0f, 0.0f, 0.0f, 1.0f );
-    }
-    glClear( GL_COLOR_BUFFER_BIT );
+	if (camera->render_mask() == 1) {
+		glClearColor(0.0f, 1.0f, 0.0f, 1.0f);
+	} else {
+		glClearColor(1.0f, 0.0f, 0.0f, 1.0f);
+	}
+	glClear (GL_COLOR_BUFFER_BIT);
 
-    Renderer::renderCamera(scene, camera, render_texture, shader_manager,
-                post_effect_shader_manager, post_effect_render_texture_a,
-                post_effect_render_texture_b, mvp);
+	Renderer::renderCamera(scene, camera, render_texture, shader_manager,
+			post_effect_shader_manager, post_effect_render_texture_a,
+			post_effect_render_texture_b, mvp);
 
 #ifdef GVRF_FBO_FPS
-    // finish rendering
-	glFinish(); // force rendering
+	// finish rendering
+	glFinish();// force rendering
 
 	gettimeofday(&end, NULL);
 	t1 = start.tv_sec + (start.tv_usec / 1000000.0);
@@ -121,7 +145,7 @@ void GVRViewManager::renderCamera(OvrSceneView &ovr_scene,
 	fps = (float)gNumFrame/gTotalSec;
 	if(!(gNumFrame%50) )
 	{
-	    LOGI("FPS is %.2f gNumFrame=%d, gTotalSec = %.2f", fps, gNumFrame, gTotalSec);
+		LOGI("FPS is %.2f gNumFrame=%d, gTotalSec = %.2f", fps, gNumFrame, gTotalSec);
 	}
 #endif
 
--- a/GVRf/Framework/src/org/gearvrf/GVRContext.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRContext.java
@@ -899,4 +899,53 @@ public abstract class GVRContext {
     abstract GVRRenderBundle getRenderBundle();
 
     abstract GVRRecyclableObjectProtector getRecyclableObjectProtector();
+
+    /**
+     * Capture a 2D screenshot from the position in the middle of left eye and
+     * right eye.
+     * 
+     * The screenshot capture is done asynchronously -- the function does not
+     * return the result immediately. Instead, it registers a callback function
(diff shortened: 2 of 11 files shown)


GVRf/Framework/src/org/gearvrf/GVRActivity.java
```diff
@@ -143,6 +143,18 @@ void oneTimeShutDown() {
     void beforeDrawEyes() {
         mGVRViewManager.beforeDrawEyes();
     }
+
```
JonShemitz: These are public methods in a public class - they need JavaDoc.

GVRf/Framework/src/org/gearvrf/GVRRenderTexture.java
```diff
@@ -63,10 +63,22 @@ protected boolean registerWrapper() {
         // deference thread to recycle() these.
         return false;
     }
+    
+    public int getWidth() {
```
JonShemitz: These also need JavaDoc.

GVRf/Framework/src/org/gearvrf/GVRScene.java
```diff
@@ -29,22 +29,30 @@
     public GVRScene(GVRContext gvrContext) {
         super(gvrContext, NativeScene.ctor());
 
+        GVRCamera centerCamera = new GVRPerspectiveCamera(gvrContext);
```
JonShemitz: Do we really want to always have a Center Camera? Would it make sense to lazy-create one, if and only if the app does a `captureScreenCenter()`?

GVRf/Framework/src/org/gearvrf/GVRScreenshotCallback.java
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+package org.gearvrf;
+
+import android.graphics.Bitmap;
+
+public interface GVRScreenshotCallback {
```
JonShemitz: A public interface needs JavaDoc.

GVRf/Framework/src/org/gearvrf/GVRScreenshotCallback.java
```diff
@@ -0,0 +1,23 @@
...
+    public void onScreenCaptured(Bitmap bitmap);
```
JonShemitz: This API needs JavaDoc.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -94,11 +98,20 @@
     GVRActivity mActivity;
     private int mCurrentEye;
 
+	private boolean mScreenshotCenterEnabled = false;
```
JonShemitz: You could probably do away with the booleans, and simply test for a non-null callback.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -94,11 +98,20 @@
...
+	private GVRScreenshotCallback mScreenshotLeftCallback = null;
+	private GVRScreenshotCallback mScreenshotRightCallback = null;
+
     private native void renderCamera(long appPtr, long scene, long camera,
             long renderTexture, long shaderManager,
             long postEffectShaderManager, long postEffectRenderTextureA,
             long postEffectRenderTextureB);
-
+    
+	private native void readbackResult(long renderTexture, Object readbackBuffer);
```
JonShemitz: This is a relatively minor issue, but "readbackResult" isn't a very descriptive name. "getScreenCapture" or "readRender" might be better.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -246,26 +259,116 @@ void beforeDrawEyes() {
...
+	void captureScreenCenter(GVRScreenshotCallback callback) {
+		mScreenshotCenterEnabled = true;
+		mScreenshotCenterCallback = callback;
+	}
+
+	Bitmap generateBitmap(byte[] byteArray, int width, int height) {
+		int[] pixels = new int[width * height];
+		for (int row = 0; row < height; row++) {
+			for (int col = 0; col < width; col++) {
+				int position = row * width + col;
```
JonShemitz: `row * width` is a loop invariant - could/should be moved outside the `col` loop.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -246,26 +259,116 @@ void beforeDrawEyes() {
...
+				int r = byteArray[position * 4] & 0xff;
```
JonShemitz: You could/should do this `position * 4` *once* and then just ++ the value.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -246,26 +259,116 @@ void beforeDrawEyes() {
...
+				int g = byteArray[position * 4 + 1] & 0xff;
+				int b = byteArray[position * 4 + 2] & 0xff;
+				// flip the image vertically
+				pixels[(height - 1 - row) * width + col] = Color.rgb(r, g, b);
```
JonShemitz: This `(height - 1 - row) * width` is also a loop invariant that should be calculated once per row, not on every pixel.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -246,26 +259,116 @@ void beforeDrawEyes() {
...
+			}
+		}
+		return Bitmap.createBitmap(pixels, width, height,
+				Bitmap.Config.ARGB_8888);
+	}
+
+	void returnScreenshotToUI(final GVRScreenshotCallback callback,
+			ByteBuffer readbackBuffer, int width, int height) {
+		final Bitmap capturedBitmap = generateBitmap(readbackBuffer.array(),
```
JonShemitz: This expensive operation should be in the Runnable, not on the GL thread!

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -246,26 +259,116 @@ void beforeDrawEyes() {
...
+				width, height);
+		mActivity.runOnUiThread(new Runnable() {
```
JonShemitz: Is there a good reason to put this on the main (GUI) thread, instead of a background (thread pool, `Threads.spawn()`) thread?

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -246,26 +259,116 @@ void beforeDrawEyes() {
...
+					|| mScreenshotRightEnabled) {
+				width = mRenderBundle.getRightRenderTexture().getWidth();
+				height = mRenderBundle.getRightRenderTexture().getHeight();
+				readbackBuffer = ByteBuffer.allocateDirect(width * height * 4);
+				readbackBuffer.order(ByteOrder.nativeOrder());
+			}
+
+			// if mScreenshotCenterEnabled is true, capture center eye
+			if (mScreenshotCenterEnabled) {
+				mainCameraRig.predict(4.0f / 60.0f);
```
JonShemitz: Shouldn't the `predict()` value be halfway between the left and right eye values?

GVRf/Sample/gvr-sample/src/org/gearvrf/sample/SampleActivity.java
```diff
@@ -13,17 +13,74 @@
...
-    }
+	@Override
+	protected void onCreate(Bundle icicle) {
+		super.onCreate(icicle);
+		setScript(new SampleCubeScript(), "gvr_note4.xml");
+	}
+
+	@Override
+	public boolean onTouchEvent(MotionEvent event) {
+		captureScreenLeft(newScreenshotCallback("screenshot_left.png"));
```
JonShemitz: I _guess_ this is why you put the API in `GVRActivity`, not `GVRContext`, but it still seems to me like the API belongs in `GVRContext` - the `GVRActivity` creates the `GVRScript`, which can always expose the `GVRContext` it gets in `onInit()`, if necessary.

GVRf/Sample/gvr-sample/src/org/gearvrf/sample/SampleActivity.java
```diff
@@ -13,17 +13,74 @@
...
+							Environment.getExternalStorageDirectory(),
+							filename);
+					try {
+						file.createNewFile();
+					} catch (IOException e) {
+						e.printStackTrace();
+					}
+					FileOutputStream outputStream = null;
+					try {
+						outputStream = new FileOutputStream(file);
```
JonShemitz: This code is more complex than it needs to be: The `FileOutputStream (File file)` docs say "The file will be ... created if it doesn't exist."

GVRf/Sample/gvr-sample/src/org/gearvrf/sample/SampleActivity.java
```diff
@@ -13,17 +13,74 @@
...
+					} catch (FileNotFoundException e) {
+						e.printStackTrace();
+					}
+					bitmap.compress(Bitmap.CompressFormat.PNG, 100,
+							outputStream);
+					try {
+						outputStream.flush();
```
JonShemitz: `close()` implies `flush()`.

JonShemitz:
Overall: looks like it should work, but needs a little cleanup and some documentation.

rongguodong:
I have updated the code according to Jon's comments. Please review it again. Thanks!

rongguodong:
For the prediction of camera rig: 

Oculus is alternatively calling onDrawEyeView for left eye and right eye. We can render-and-capture the centre eye in either call. I now included the centre eye code in the left-eye branch. So we do not need to update the camera rig for centre eye separately.

GVRf/Framework/jni/oculus/view_manager.cpp
```diff
@@ -18,41 +18,49 @@
...
+}
+
+void Java_org_gearvrf_GVRViewManager_readRenderResult(JNIEnv * jni, jclass clazz,
+		jlong jrender_texture, jobject jreadback_buffer) {
+
+	uint8_t *pReadbackBuffer = (uint8_t *) jni->GetDirectBufferAddress(
+			jreadback_buffer);
+	std::shared_ptr<RenderTexture> render_texture =
+			*reinterpret_cast<std::shared_ptr<RenderTexture>*>(jrender_texture);
+	int mWidth = render_texture->width();
```
JonShemitz: The 'm' prefix means this is a "member variable" - to the extent that it has any point at all, it is that it is NOT local; NOT declared nearby, but in the class declaration.

Please use "width" and "height".

GVRf/Framework/src/org/gearvrf/GVRActivity.java
```diff
@@ -143,7 +143,7 @@ void oneTimeShutDown() {
     void beforeDrawEyes() {
         mGVRViewManager.beforeDrawEyes();
     }
-
+
```
JonShemitz: Generally shouldn't check-in a file if all you did was change the whitespace.

(The exception would be if the whole point of the check-in was to apply uniform formatting to the whole project.)

GVRf/Framework/src/org/gearvrf/GVRContext.java
```diff
@@ -899,4 +899,41 @@ protected synchronized static void resetOnRestart() {
...
 
     abstract GVRRecyclableObjectProtector getRecyclableObjectProtector();
+
+    /**
+     * Capture a 2D screenshot from the position in the middle of left eye and
+     * right eye.
+     * 
+     * The screenshot capture is done asynchronously -- the function does not
+     * return the result immediately. Instead, it registers a callback function
+     * and pass the result (when it is available) to the callback function.
```
JonShemitz: This text is fine, as far as it goes, but I think it probably should include something about threading, something like

> The callback will happen on a background thread: It will probably not be the same thread that calls this method, and it will *not* be either the GUI or the GL thread.

GVRf/Framework/src/org/gearvrf/GVRContext.java
```diff
@@ -899,4 +899,41 @@ protected synchronized static void resetOnRestart() {
...
+     * 
+     * @param callback
+     *            handle a the callback function to process the capture result
```
JonShemitz: I think we could probably lose the "handle a the" part. You probably should mention that this is a required parameter, that may not be `null`.

GVRf/Framework/src/org/gearvrf/GVRRenderTexture.java
```diff
@@ -55,6 +55,20 @@ public GVRRenderTexture(GVRContext gvrContext, int width, int height,
...
         super(gvrContext, ptr);
     }
+    
+    /**
+     * Return the width of GVRRenderTexture (FBO)
+     */
+    public int getWidth() {
+    	return NativeRenderTexture.getWidth(getPtr());
+    }
+
```
JonShemitz: Do we really need these methods (and the new JNI code)? Could you just save the height and width in the constructor? (It doesn't look like these textures are being created by the native code.)

GVRf/Framework/src/org/gearvrf/GVRScreenshotCallback.java
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gearvrf;
+
+import android.graphics.Bitmap;
+
+/**
+ * Callback interface for asynchronous screenshot capture. Refer to user manual
```
JonShemitz: *No!* The JavaDoc is the authoritative documentation.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -246,27 +260,144 @@ void beforeDrawEyes() {
         mFrameHandler.beforeDrawEyes();
     }
 
+    @Override
+    public void captureScreenCenter(GVRScreenshotCallback callback) {
+        mScreenshotCenterCallback = callback;
```
JonShemitz: I guess it's 'harmless' to not check for `null` since you are using the callback as a flag, but I think you really should throw an `IllegalArgumentException` when `callback == null`

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -246,27 +260,144 @@ void beforeDrawEyes() {
...
             GVRCameraRig mainCameraRig = mMainScene.getMainCameraRig();
+
+            // if screenshot is enabled, initialize the readback buffer
+            ByteBuffer readbackBuffer = null;
+            int width = 0, height = 0;
+            if (mScreenshotCenterCallback != null
+                    || mScreenshotLeftCallback != null
+                    || mScreenshotRightCallback != null) {
+                width = mRenderBundle.getRightRenderTexture().getWidth();
+                height = mRenderBundle.getRightRenderTexture().getHeight();
```
JonShemitz: This is a small detail, but: that `mRenderBundle.getRightRenderTexture()` is not free, and it would be better to call it once, saving the result to a local variable.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -246,27 +260,144 @@ void beforeDrawEyes() {
...
                 renderCamera(mActivity.appPtr, mMainScene, rightCamera,
                         mRenderBundle.getRightRenderTexture(), mRenderBundle);
+
+                // if mScreenshotRightCallback is not null, capture right eye
+                if (mScreenshotRightCallback != null) {
+                    readRenderResult(mRenderBundle
+                            .getPostEffectRenderTextureA().getPtr(),
+                            readbackBuffer);
+                    returnScreenshotToCaller(mScreenshotRightCallback,
+                            Arrays.copyOf(readbackBuffer.array(),
```
JonShemitz: You have three copies of this `Arrays.copyOf(readbackBuffer.array(), readbackBuffer.array().length)` code. I'd move it to `returnScreenshotToCaller()`, where you could call it just before the `spawn()` call.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -246,27 +260,144 @@ void beforeDrawEyes() {
...
+
+                // if mScreenshotCenterCallback is not null, capture center eye
+                if (mScreenshotCenterCallback != null) {
+
+                    // temporarily create a center camera
+                    GVRCamera centerCamera = new GVRPerspectiveCamera(this);
+                    centerCamera.setRenderMask(GVRRenderMaskBit.Left
+                            | GVRRenderMaskBit.Right);
+                    GVRSceneObject centerCameraObject = new GVRSceneObject(this);
+                    centerCameraObject.attachCamera(centerCamera);
```
JonShemitz: You never detach this camera - what if you call `captureScreenCenter()` twice?

GVRf/Sample/gvr-sample/src/org/gearvrf/sample/SampleCubeScript.java
```diff
@@ -172,4 +180,52 @@ public void onStep() {
...
+            @Override
+            public void onScreenCaptured(Bitmap bitmap) {
+                if (bitmap != null) {
+                    File file = new File(
+                            Environment.getExternalStorageDirectory(), filename);
+                    FileOutputStream outputStream = null;
+                    try {
+                        outputStream = new FileOutputStream(file);
+                    } catch (FileNotFoundException e) {
+                        e.printStackTrace();
```
JonShemitz: If you get an exception, the "file cannot be opened for writing" - so you shouldn't continue!

JonShemitz:
Thanks, looks better. As you can see, though, I do still have some comments.

GVRf/Framework/src/org/gearvrf/GVRRenderTexture.java
rongguodong: My understanding is that GVRF is designed in the way that we do not store anything in Java side and everything is stored in native side. So we should get information from native side when needed.

This is the reason I implemented them in this "wired" way.

rongguodong:
I have just pushed an updated version. Please review again.

GVRf/Sample/gvr-sample/src/org/gearvrf/sample/SampleCubeScript.java
```diff
@@ -172,4 +180,53 @@ public void onStep() {
...
+                            outputStream);
+                    try {
+                        outputStream.flush();
+                    } catch (IOException e) {
+                        e.printStackTrace();
+                    }
+                    try {
+                        outputStream.close();
+                    } catch (IOException e) {
+                        e.printStackTrace();
```
chenchao1407: I have several concerns here.

1. You should combine all these try blocks together, not separately like this.
2. outputStream.close() should be called in a finally block. It means no matter what error occurs, the filestream should be closed.
3.I believe in this case, outputStream.flush() is not necessary.

chenchao1407: It should be as the following

            @Override
            public void onScreenCaptured(Bitmap bitmap) {
                if (bitmap != null) {
                    File file = new File(
                            Environment.getExternalStorageDirectory(), filename);
                    FileOutputStream outputStream = null;
                    try {
                        outputStream = new FileOutputStream(file);
                        bitmap.compress(Bitmap.CompressFormat.PNG, 100,
                                outputStream);
                    } catch (FileNotFoundException e) {
                        e.printStackTrace();
                        return;
                    } finally {
                        try {
                            outputStream.close();
                        } catch (IOException e) {
                            e.printStackTrace();
                        }
                    }
                } else {
                    Log.e("SampleActivity", "Returned Bitmap is null");
                }
            }

GVRf/Framework/src/org/gearvrf/GVRRenderTexture.java
JonShemitz: Fair enough, but I think it may be giving too much deference to the original ‘design’. I can’t see a lot of point to adding so much code – and burning so many cycles – to ask the jni side for data that came from Java in the first place.

rongguodong: I have just removed the native part of codes and stored width and height in Java part.

We can make similar changes in many other places, i.e. storing useful information in Java side and do not need to spend time calling native methods to retrieve them. This way is in fact the correct way in my opinion, comparing with the current "wired" design of GVRF. :)

rongguodong:
I have just posted modified version following comments from Jon and Chao. Thanks for your comments and please review again.

rongguodong:
Some explanation for the latest changes in "Java_org_gearvrf_GVRViewManager_readRenderResult":

My previous code works on devices with KitKat but gives totally black results on devices with Lollipop. The reason is KitKat has some bugs in the OpenGL driver which happened making my old code work. Details are as below:

According to OpenGL specification, glReadPixels will generate GL_INVALID_OPERATION if "GL_READ_FRAMEBUFFER_BINDING is non-zero, the read framebuffer is complete, and the value of GL_SAMPLE_BUFFERS for the read framebuffer is greater than zero". Basically, this means we cannot read pixels from a multisampled FBO.

It seems the FBO that Oculus provided to us is a multisampled FBO, and we should get error when using glReadPixels on it. However, I do not get this error when running on KitKat. This is the reason why I can successfully generate correct screen shots on those devices. On Lollipop, I do get the (expected) error, and do not get any read back results, thus the generated screen shot is totally black.

The solution is to blit (copy) the content from the multisampled FBO to a normal FBO and read pixels from thenormal FBO. This solution is implemented in my newly submitted code.

chenchao1407:
Code looks good to me. And it works fine now.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -246,27 +260,154 @@ void beforeDrawEyes() {
...
+                            centerCameraObject);
+
+                    readRenderResult(mRenderBundle
+                            .getPostEffectRenderTextureA().getPtr(),
+                            readbackBuffer);
+                    returnScreenshotToCaller(mScreenshotCenterCallback,
+                            readbackBuffer, width, height);
+
+                    mScreenshotCenterCallback = null;
+                }
```
chenchao1407: Actually capture center's logic lies in when we draw left eye buffer, so that the result may be a little bit left. I tried to move this codes block where we draw right eye buffer, and the result will be a little bit right. I talked to Guodong and he said it may be related to the difference on the predict value on mainCameraRig.predict(). 

Anyway, I think this is actually acceptable. Just to see if there is any better solution to get camera angle rather than directly setRenderMask and attach to mainCameraRig.

rongguodong: In fact, even if we do not predict at all (i.e. passing 0 to predict), we would still have the same problem.

This is because we render left eye and right at different time, the head orientation may be different. The better way should be render both eyes in one function call, but it is out of our control.

In the current implementation, the center camera already has spacial interpolation between left and right cameras. But it lacks temporal interpolation. We may store the orientation of the left camera when we render left eye, and use the stored value when we render right eye to compute the "temporal interpolation". This will of course lead to more computations.

I think the current spacial interpolation should be good enough for this screen shot.

thomasflynn:
so, what's the status on this commit?  I did my own cherry-pick.  I verified the sample and the screenshots it takes.  Are there further comments that need to be addressed?

chenchao1407:
Yeah. I think it is good to merge now. Last week guodong said he is waiting for your verification. And if everything is fine, then I am going to merge it.

PR #17, opened 2015-04-17
Simplified async code.

JonShemitz:
The two big complaints about the async code in v1.6.1 and 1.6.2
have been that it's hard to handle async callbacks, and that it
feels wrong to have specify load-bitmap-texture vs load-compressed-
texture. This change addresses both those issues.

* It adds Future<> textures and Future<> meshes, and APIs that
accept them directly. These APIs are much more convenient than
handling async callbacks ...and not all THAT much more expensive.

* It also adds a new 'unified sniffer' so you can simply
loadTexture() without having to specify Bitmap or Compressed.
This reduces the chance of error, while also letting you replace
bitmaps with compressed textures without having to change your code.
(This is probably how it should have been implemented in the first
place - the Bitmap/Compressed distinction was basically a Premature
Optimization.)

Also adds a small change to GVRScript.onInit() - the user
method doesn't HAVE TO catch exceptions: GVRF will catch
them and shut down gracefully. (This is not the behavior
you want if you are expecting possible excxeptions due to
(say) network issues ... but it's fine for handling develop
time issues like mispelled assets filenames.)

Fixes some JavaDoc errors, too.

j.shemitz@samsung.com


thomasflynn:
my initial skim of the code looks good.  i still need to cherry-pick and try to use it.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -312,7 +311,18 @@ public void beforeDrawEyes() {
                 getMainScene().addSceneObject(mSplashScreen);
             }
 
-            mScript.onInit(GVRViewManager.this);
+            try {
+                mScript.onInit(GVRViewManager.this);
+            } catch (Throwable t) {
+                t.printStackTrace();
```
chenchao1407: I am thinking if you don't want to crash the app at least we need to print the error message to the console with any reasonable tag.  I was thinking printStackTrace doesn't show anything in logcat and I just tried and that it is. So if there is any kind of exception occurs in onInit, there is just quick quit of the app but no error message shown to developers. I think it may make developers a little bit confused and hard to find any clue to debug.

I am considering it is better to the getStackTrace() to get stack trace info and use Log.e to print such information.

chenchao1407:
Everything looks fine except the exception handling with GVRScript.onInit() in GVRViewManager. 

And it compiles well and runs well with sample apps.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
JonShemitz: Hmm? printStackTrace() gives the full stack trace, with the right pid. Maybe you can miss it if you are searching for tag=="GVRF" (but why would you do that?) but it's there when you search by process id.

chenchao1407: Yep. I think I made a mistake. I will output the stack trace with tag System.err.

chenchao1407:
Just want to know if there is any concern or objection,  otherwise I am going to merge this.

Thanks

jason2kim:
I didn't have time to cherry pick test, but codes look good for me.

PR #18, opened 2015-04-17
Support for loading the full hierarchy in a model in case it exists

soumyajitdeb:
This patch supports loading the complete hierarchy if present in a model. The major changes that have  taken place are in GVRf/Framework/src/org/gearvrf/GVRAssimpImporter.java and GVRf/Framework/jni/engine/importer/assimp_importer.cpp to allow loading of the entire hierarchy via recursion.

We've also added a sample app that loads the public domain "Crytek Sponza" model that is available from the Crytek website. The sample app is located in GVRf/Sample/gvrmodelloader. 

Please let us know if this complies with your coding standards and requirements. Since this is our first contribution, please do pardon us for any inadvertent errors. We welcome constructive criticism. Once this one is checked in, we'll start creating pull requests for other features that we have ready.

Thanks
Soumyajit & Lokesh


soumyajitdeb:
Please ignore the commit comment that mentions frustum culling. That has been removed from this pull request.

thomasflynn:
Awesome!  Thanks for the pull request.  We're in the process of reviewing.  It's a lot of files, so it may take some time to review.  Thanks!

GVRf/Framework/.cproject
```diff
@@ -5,6 +5,7 @@
 			<storageModule buildSystemId="org.eclipse.cdt.managedbuilder.core.configurationDataProvider" id="com.android.toolchain.gcc.1319470856" moduleId="org.eclipse.cdt.core.settings" name="Default">
 				<externalSettings/>
 				<extensions>
+					<extension id="org.eclipse.cdt.core.ELF" point="org.eclipse.cdt.core.BinaryParser"/>
```
JonShemitz: This looks like a 'sideways change' - simply rearranging the XML should be harmless, but it's something we generally don't check in.

GVRf/Framework/.settings/org.eclipse.cdt.codan.core.prefs
```diff
@@ -0,0 +1,67 @@
+eclipse.preferences.version=1
```
JonShemitz: This is a file that hasn't previously been checked in - is it really necessary?

GVRf/Framework/jni/engine/importer/assimp_importer.cpp
```diff
@@ -68,4 +71,176 @@ std::shared_ptr<Mesh> AssimpImporter::getMesh(int index) {
 
     return std::shared_ptr < Mesh > (mesh);
 }
+
+void AssimpImporter::sceneRecursion(aiNode* node, const aiScene* aScene, std::shared_ptr<Scene> scnPtr, JNIEnv * env, jobject obj, jobject bitmap, aiMatrix4x4 accumulatedTransform)
+{
```
JonShemitz: We have been careful to use consistent (automatic) formatting throughout GVRF, so that reviewers can focus on 'real' changes, not whitespace changes.

In Eclipse, we use very slightly modified versions of the "K & R" built-in (for C++) and the "Java Conventions" built-in (for, obviously, Java). In both cases, the only change we've made is to set the "Tab policy" to "Spaces only" (from the default "Mixed".)

I will keep on reviewing, but will not merge this checkin until you've applied the standard formatting and updated the patch.

GVRf/Framework/jni/engine/importer/assimp_importer.cpp
```diff
@@ -68,4 +71,176 @@ std::shared_ptr<Mesh> AssimpImporter::getMesh(int index) {
 
     return std::shared_ptr < Mesh > (mesh);
 }
+
+void AssimpImporter::sceneRecursion(aiNode* node, const aiScene* aScene, std::shared_ptr<Scene> scnPtr, JNIEnv * env, jobject obj, jobject bitmap, aiMatrix4x4 accumulatedTransform)
```
JonShemitz: Pretty much all of the JNI code that I've read or touched uses C naming conventions, like `scene_pointer`, not Java conventions like `scenePointer`; the only C++ code I've seen use camelCaps is the contrib/assimp code. I can't say that *all* the GVR JNI code uses the _ `naming_convention`, but enough does that this really stands out. I would very much prefer it if such a large change followed the existing conventions. (I know this sort of feedback can seem petty, but inconsistencies do slow people down - "Why is it different _here_ than _there_?" - and massively inconsistent code is very hard to read.

Second, I'm consistently - almost to the point of fetish - opposed to the use of abbreviations. In a Globish-speaking world, the abbreviation that's obvious to one person might not be at all obvious to another person. It's just not worth slowing down and/or confusing your readers when long names like `scene_pointer` make your code all the clearer and auto-complete makes them just as easy to type. Plus, abbreviations can make bugs harder to spot: I once found a bug where a programmer had used `i` and `j` for loop variables, and coded `array[i, j]` where they meant `array[j, i]`.

GVRf/Framework/jni/engine/importer/assimp_importer.cpp
```diff
@@ -68,4 +71,176 @@ std::shared_ptr<Mesh> AssimpImporter::getMesh(int index) {
...
+
+            // Set the mesh to the render data.
+            renderData->set_mesh(mesh);
+
+            // Set material.
+            aiMesh* ai_mesh = aScene->mMeshes[node->mMeshes[i]];
+            aiMaterial* ai_material = aScene->mMaterials[ai_mesh->mMaterialIndex];
+
+            // Defines a shader with shader type 0. UNLIT_SHADER = 0
+            std::shared_ptr<Material> material(new Material(static_cast<Material::ShaderType>(0)));
```
JonShemitz: Why not just use `ShaderType.UNLIT_SHADER` instead of `static_cast<Material::ShaderType>(0)`?

GVRf/Framework/jni/engine/importer/assimp_importer.cpp
```diff
@@ -68,4 +71,176 @@ std::shared_ptr<Mesh> AssimpImporter::getMesh(int index) {
...
+            // Actual texture image
+            aiString textureImage;
+            ai_material->GetTexture(aiTextureType_DIFFUSE, i, &textureImage);
+            jstring texFileName = env->NewStringUTF(textureImage.C_Str());
+            jclass texClass = env->FindClass("org/gearvrf/GVRContext");
+
+            if (texClass == NULL) {
+                // Couldn't find the class GVRContext.
+            } else {
+                jmethodID mID = env->GetStaticMethodID(texClass, "loadBitmap", "(Ljava/lang/String;)Landroid/graphics/Bitmap;");
```
JonShemitz: Please follow GVRF JNI conventions, and do this (comparatively expensive) FindClass / Get_Method stuff once, at startup, not on every call.

GVRf/Framework/jni/engine/importer/assimp_importer.cpp
```diff
@@ -68,4 +71,176 @@ std::shared_ptr<Mesh> AssimpImporter::getMesh(int index) {
...
+                if (mID == NULL) {
```
JonShemitz: This does not seem like the right way to handle this. If the JNI code and the Java code get out of sync, it would probably be better to fail at startup than to have fallback code scattered throughout the system.

GVRf/Framework/jni/engine/importer/assimp_importer.cpp
```diff
@@ -68,4 +71,176 @@ std::shared_ptr<Mesh> AssimpImporter::getMesh(int index) {
...
+                    // No interaction with Java side so apply default texture
+                    std::shared_ptr<Texture> texture(new BaseTexture(env, bitmap));
+                    material->setTexture("main_texture", texture);
+                    // Sets the default main texture
+
+                } else {
+                    if ((ai_material->GetTextureCount(aiTextureType_DIFFUSE)) <= 0) {
+                        // No texture file found so apply default texture
+                        std::shared_ptr<Texture> texture(new BaseTexture(env, bitmap));
+                        material->setTexture("main_texture", texture);
```
JonShemitz: The 3 repetitions of `material->setTexture("main_texture", texture);` are 2 too many - better to declare `std::shared_ptr<Texture> texture;` before the decision tree, and then do `setTexture()` after ...

GVRf/Framework/jni/engine/importer/assimp_importer.cpp
```diff
@@ -68,4 +71,176 @@ std::shared_ptr<Mesh> AssimpImporter::getMesh(int index) {
...
+                    } else {
+                        // About to enter Java side.
+                        jobject mBitmap = env->CallStaticObjectMethod(texClass, mID, texFileName);
```
JonShemitz: 1. This code is assuming that `texFileName` contains the name of a file in the `assets` directory. Is this always true? What if the mesh file is in a subdirectory of the `assets` directory? What if the  mesh file is in `res/raw`?

2. `GVRContext.loadBitmap(String)` can crash the app, if the bitmap is too large, or if memory is fragmented. `GVRContext.loadTexture(GVRAndroidResource)` is safer and will often be faster.

GVRf/Framework/jni/engine/importer/assimp_importer.cpp
```diff
@@ -68,4 +71,176 @@ std::shared_ptr<Mesh> AssimpImporter::getMesh(int index) {
...
+                    // No such method in Java side
+                    // No interaction with Java side so apply default texture
+                    std::shared_ptr<Texture> texture(new BaseTexture(env, bitmap));
+                    material->setTexture("main_texture", texture);
+                    // Sets the default main texture
+
+                } else {
+                    if ((ai_material->GetTextureCount(aiTextureType_DIFFUSE)) <= 0) {
+                        // No texture file found so apply default texture
+                        std::shared_ptr<Texture> texture(new BaseTexture(env, bitmap));
```
JonShemitz: `bitmap` s/b `default_bitmap`?

GVRf/Framework/jni/engine/importer/assimp_importer.cpp
```diff
@@ -68,4 +71,176 @@ std::shared_ptr<Mesh> AssimpImporter::getMesh(int index) {
...
+
+
+std::shared_ptr<Scene> AssimpImporter::loadScene(JNIEnv* env, jobject obj, jobject bitmap)
+{
+    //  Scene
+    std::shared_ptr<Scene> scenePtr(new Scene());
+
+    // Left Camera
+    std::shared_ptr<Camera> leftCamera(new PerspectiveCamera());
+    leftCamera->set_render_mask(0x1);
```
JonShemitz: Please use RenderData::RenderMaskBit.Left

GVRf/Framework/jni/engine/importer/assimp_importer.cpp
```diff
@@ -68,4 +71,176 @@ std::shared_ptr<Mesh> AssimpImporter::getMesh(int index) {
...
+    //  Right Camera
+    std::shared_ptr<Camera> rightCamera(new PerspectiveCamera());
+    rightCamera->set_render_mask(0x2);
+
+    //  Left Camera object and its transform
+    std::shared_ptr<SceneObject> leftCameraObject(new SceneObject());
+    std::shared_ptr<Transform> transformLeftCameraObject(new Transform());
+    transformLeftCameraObject->set_owner_object(leftCameraObject);
+
+    float camera_separation_distance_ = 0.062f;
```
JonShemitz: Where does this 'magic number' come from? It really ought to be a symbolic constant, with a comment that explains the value and/or an expression that derives it from more fundamental constants.

GVRf/Framework/jni/engine/importer/assimp_importer.cpp
```diff
@@ -68,4 +71,176 @@ std::shared_ptr<Mesh> AssimpImporter::getMesh(int index) {
...
+
+    //  Sets transform for the Left and the Right Camera on the Camera Rig
+    glm::vec3 leftCameraPosition = glm::vec3(-camera_separation_distance_ * 0.5f, 0.0f, 0.0f);
```
JonShemitz: Fwiw, `constant / 2` is a generally a bit more readable than `constant * 0.5` ... and if it *is* a constant expression, the math is done at compile time, so there's no 'multiplication is faster than division' issue.

GVRf/Framework/jni/engine/importer/assimp_importer.h
```diff
@@ -44,7 +66,8 @@ class AssimpImporter: public HybridObject {
     unsigned int getNumberOfMeshes() {
         return assimp_importer_->GetScene()->mNumMeshes;
     }
-
+    void sceneRecursion(aiNode* node, const aiScene* aScene, std::shared_ptr<Scene> scnPtr, JNIEnv * env, jobject obj, jobject bitmap, aiMatrix4x4 accumulatedTransformation);
```
JonShemitz: Please leave a blank line between method bodies. (I'm fairly sure that Shift+Ctrl+F will do this for you.)

GVRf/Framework/jni/gl/gl_texture.h
```diff
@@ -30,8 +30,8 @@ class GLTexture {
             target_(target) {
         glGenTextures(1, &id_);
         glBindTexture(target, id_);
-        glTexParameteri(target, GL_TEXTURE_WRAP_S, GL_CLAMP_TO_EDGE);
-        glTexParameteri(target, GL_TEXTURE_WRAP_T, GL_CLAMP_TO_EDGE);
+        glTexParameteri(target, GL_TEXTURE_WRAP_S, GL_REPEAT);
```
JonShemitz: Why? This seems like a pretty far-reaching change.

GVRf/Framework/src/org/gearvrf/GVRAssimpImporter.java
```diff
@@ -31,7 +33,15 @@
     int getNumberOfMeshes() {
         return NativeAssimpImporter.getNumberOfMeshes(getPtr());
     }
-
+    
+    /**
+     * @return Loads the 3D model.
```
JonShemitz: JavaDoc on internal methods is a lot less critical than JavaDoc on public methods ... but this isn't very well-formed: "Loads the 3D model." makes sense as a method summary, not an `@return`, and it probably would be a very good idea to document the `@param bitmap`.

GVRf/Framework/src/org/gearvrf/GVRAssimpImporter.java
```diff
@@ -31,7 +33,15 @@
...
+     */
+    GVRScene loadScene(Bitmap bitmap)
```
JonShemitz: Shouldn't this method be returning a `GVRSceneObject`, not a `GVRScene`? Scenes contain scene objects, which may in turn contain scene objects. Most scenes will be composed of multiple scene objec ts, loaded from multiple model files.

GVRf/Framework/src/org/gearvrf/GVRContext.java
```diff
@@ -52,7 +53,7 @@
 public abstract class GVRContext {
     private static final String TAG = Log.tag(GVRContext.class);
 
-    private final Context mContext;
+    private static Context mContext = null;
```
JonShemitz: Why?

GVRf/Framework/src/org/gearvrf/GVRContext.java
```diff
@@ -296,6 +297,52 @@ public GVRMesh createQuad(float width, float height) {
     }
 
     /**
+     * Generates default color for missing textures in the modelling file.
```
JonShemitz: This JavaDoc will be 'attached to' `WIDTH`, not `createColors()`

GVRf/Framework/src/org/gearvrf/GVRContext.java
```diff
@@ -296,6 +297,52 @@ public GVRMesh createQuad(float width, float height) {
...
+     *
+     */
+    private static final int WIDTH = 50;
+    private static final int HEIGHT = 50;
+    private static final int STRIDE = 64;    // must be >= WIDTH
+
+    private static int[] createColors() 
+    {
```
JonShemitz: As in the JNI code, please Shift+Ctrl+F with the proper formatting rules.

GVRf/Framework/src/org/gearvrf/GVRContext.java
```diff
@@ -296,6 +297,52 @@ public GVRMesh createQuad(float width, float height) {
...
+        int[] colors = new int[STRIDE * HEIGHT];
+        for (int y = 0; y < HEIGHT; y++)
+        {
+            for (int x = 0; x < WIDTH; x++)
+            {
+                int r = x * 255 / (WIDTH - 1);
+                int g = y * 255 / (HEIGHT - 1);
```
JonShemitz: Loop invariant ...

GVRf/Framework/src/org/gearvrf/GVRContext.java
```diff
@@ -296,6 +297,52 @@ public GVRMesh createQuad(float width, float height) {
...
+     *            of subdirectories; the file name can specify any location in
+     *            or under the assets directory.
+     */
+    public void loadScene(String fileName)
+    {
+        GVRAssimpImporter assimpImporter = GVRImporter.readFileFromAssets(this, fileName);
+        //Default bitmap for objects whose texture are not present or there is some texture error
+        int[] mColors = createColors();
+        int[] colors = mColors;
+        Bitmap defaultBitmap = Bitmap.createBitmap(colors, 0, STRIDE, WIDTH, HEIGHT, Bitmap.Config.RGB_565);
```
JonShemitz: Please use an `res/drawable-xxxhdpi` bitmap for this `defaultBitmap` instead of embedding it in code. This will let use replace it with a designer-supplied graphic if necessary.

GVRf/Framework/src/org/gearvrf/GVRContext.java
```diff
@@ -296,6 +297,52 @@ public GVRMesh createQuad(float width, float height) {
...
+                int b = 255 - Math.min(r, g);
+                int a = Math.max(r, g);
+                colors[y * STRIDE + x] = (a << 24) | (r << 16) | (g << 8) | b;
+            }
+        }
+        return colors;
+    }   
+
+    /**
+     * Loads the complete scene from the imported 3D modelling file.
```
JonShemitz: Again, this is probably not right: a scene will often contain multiple model.

GVRf/Framework/src/org/gearvrf/GVRContext.java
```diff
@@ -296,6 +297,52 @@ public GVRMesh createQuad(float width, float height) {
...
+     * 
+     * @param fileName
+     *            The name of a 3D model file, relative to the assets directory.
+     *            The assets directory may contain an arbitrarily complex tree
+     *            of subdirectories; the file name can specify any location in
+     *            or under the assets directory.
+     */
+    public void loadScene(String fileName)
```
JonShemitz: Please use `GVRAndroidResource`, not `assets`-relative filenames.

GVRf/Framework/src/org/gearvrf/GVRContext.java
```diff
@@ -318,7 +365,7 @@ public GVRMesh createQuad(float width, float height) {
      * @return The file as a bitmap, or {@code null} if file path does not exist
      *         or the file can not be decoded into a Bitmap.
      */
-    public Bitmap loadBitmap(String fileName) {
+    public static Bitmap loadBitmap(String fileName) {
```
JonShemitz: I don't think it's worth making `mContext` static just so that this method becomes `static`. Not only does this generate a warning on every call like `mContext.loadBitmap(whatever)` it makes every reference to `mContext` a bit slower: static variables generally have to be referenced with a full 32 (or 64!) bit address, instead of a 8 or 16 bit offset to a register-resident `this` pointer.

GVRf/Sample/gvrmodelloader/.classpath
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
JonShemitz: Probably should not have been checked-in.

GVRf/Sample/gvrmodelloader/.project
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
JonShemitz: Probably should not have been checked-in.

GVRf/Sample/gvrmodelloader/assets/lens.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
JonShemitz: Why isn't this `gvr_s4.xml`?

(Does GVRF even run on an S4?)

GVRf/Sample/gvrmodelloader/AndroidManifest.xml
```diff
@@ -0,0 +1,45 @@
...
+    android:versionCode="1"
+    android:versionName="1.0" >
+
+    <uses-sdk
+        android:minSdkVersion="8"
+        android:targetSdkVersion="19" />
+
+    <application
+        android:allowBackup="true"
+        android:icon="@drawable/ic_launcher"
```
JonShemitz: There seem to be a lot of redundant ic_launcher files - all GVRF-supported devices are `xxxhdpi`

GVRf/Sample/gvrmodelloader/src/org/gearvrf/gvrmodelloader/ModelActivity.java
```diff
@@ -0,0 +1,30 @@
...
+import android.os.Bundle;
+
+import org.gearvrf.GVRActivity;
+
+public class ModelActivity extends GVRActivity {
+	
+    @Override
+    protected void onCreate(Bundle icicle) {
+        super.onCreate(icicle);
+        setScript(new ModelViewManager(), "gvr_note4.xml");
```
JonShemitz: Please remove the unused distortion files.

JonShemitz:
Thanks for the submission. I have a number of stylistic issues - we have tried pretty hard to maintain a consistent style in the code base, and this doesn't match in a number of ways - and a smaller number of semantic issues. I've called them out in in-line notes, but the most important are probably

1. I think we need to be loading scene objects, not scenes.
2. GVRContext.loadBitmap() is dangerous and expensive, and using it makes possibly invalid assumptions about where assets are located.

Beyond that: This is a *huge* submission. It took me several hours to review it; because GitHub doesn't seem to support the notion of 'patch sets' or have any way to show the changes between an initial `push` and a revised `push`, it will take me just as long to re-review it. I see that this Pull Request is actually composed of 18 discrete commits; I would much prefer more granular  Pull Requests, one per commit, that will make it much easier to track changes.

GVRf/Sample/gvrmodelloader/assets/lens.xml
thomasflynn: no, it doesn't run on S4.  Runs on Note4 and S6.

thomasflynn:
With regard to code formatting/styling, I thought we had something up on the wiki before the release, but i can't seem to find it now.  Let's get with Patrick on Monday and make sure we put something up there including the eclipse file to import.

GVRf/Framework/jni/engine/importer/assimp_importer.cpp
```diff
@@ -68,4 +71,176 @@ std::shared_ptr<Mesh> AssimpImporter::getMesh(int index) {
...
+        sceneRecursion(node->mChildren[i], aScene, scnPtr, env, obj, bitmap, accumulatedTransform);
+    }
+}
+
+
+std::shared_ptr<Scene> AssimpImporter::loadScene(JNIEnv* env, jobject obj, jobject bitmap)
+{
+    //  Scene
+    std::shared_ptr<Scene> scenePtr(new Scene());
+
```
jason2kim: These camera attachment requests blocks below after this new Scene()  are also made in GVRScence() and camera_rig.cpp. Is there a way to adjust or reuse codes between the two?

GVRf/Framework/jni/engine/importer/assimp_importer.h
jason2kim: The parameter jobject obj is not used, but is it here for future use?

GVRf/Framework/jni/engine/importer/assimp_importer.cpp
```diff
@@ -68,4 +71,176 @@ std::shared_ptr<Mesh> AssimpImporter::getMesh(int index) {
...
 }
+
+void AssimpImporter::sceneRecursion(aiNode* node, const aiScene* aScene, std::shared_ptr<Scene> scnPtr, JNIEnv * env, jobject obj, jobject bitmap, aiMatrix4x4 accumulatedTransform)
+{
+    for(int i=0; i < node->mNumMeshes; i++)
+    {
+        std::shared_ptr<SceneObject> scnObject(new SceneObject());
+
+        // Creates a scene object for the node children meshes.
+        if(scnObject->render_data() == NULL)
```
jason2kim: Is there any case that render_data() is not null, then making a comment in the source here would be helpful?

GVRf/Framework/jni/engine/importer/assimp_importer.h
JonShemitz: That’s just the jni calling convention – I think that would be the this pointer in a non-static method.

GVRf/Framework/jni/engine/importer/assimp_importer.cpp
lokeshguddu: Sorry for this section. It was just a flag to check the render data associated with the scene object. But the part I missed is that this was a check for a newly created scene object which, by default, won't have any render data. So this check won't be of any use as this flag will always be true. I fixed this issue in the latest commit. Thanks for pointing this out.

lokeshguddu:
We've made the following changes/corrections based on the comments received

- All the tabs has been converted to spaces and the codes written has followed the convention for GVRF.

- The naming covention followed at present is the snake_case type. I have also removed the abbreviations with more descriptive name.

- Now ShaderType.UNLIT_SHADER is used instead of static_cast<Material::ShaderType>(0). At the begining I did this without going for the first option without any preference.

- All magic numbers were an oversight. They have been removed/replaced.

- The FindClass has been replaces by using the gvr_context class directly though its class. The repeated call problem has also been fixed.

- Issues with GL_CLAMP_TO_EDGE/GL_REPEAT in gl_texture.h have been modified to take into account the actual texture coordinates in the model. The code now switches to GL_REPEAT when required.

- We've removed the reference to loadBitmap with a different customized function that uses GVRAndroidResource

- Other minor issues pointed out have been fixed.

GVRf/Framework/jni/engine/importer/assimp_importer.h
jason2kim: Actually sceneRecursion() is used as a local c++ function that can be freely stripped off any unnecessary parameters.

JonShemitz:
Thanks! I'm out of town on a family emergency, so I don't think we will thoroughly review & merge this until next week, but Chao, Jason, or Tom may have time to do a first pass before I get back

.gitignore
```diff
@@ -1,4 +1,6 @@
 bin/
 gen/
 obj/
+ovr_mobile_sdk/
```
chenchao1407: I am not sure if we should add "ovr_mobile_sdk" as gitignore. It can be a convenience but it seems a little weird to me.

GVRf/Framework/jni/engine/importer/assimp_importer.cpp
```diff
@@ -68,4 +76,178 @@ std::shared_ptr<Mesh> AssimpImporter::getMesh(int index) {
...
+                    else
+                    {
+                        // Back to native side from Java side
+                        // Applying actual texture
+                        std::shared_ptr<Texture> actual_texture(new BaseTexture(env, actual_texture_bitmap, repeat_flag));
+                        gvr_material->setTexture("main_texture", actual_texture);
+                    }
+                }
+            }
+
```
chenchao1407: Why there is space for this  if-else block. And also, I am a little bit confused about the use of default texture. It seems to me that if any error occurs while loading texture from resource, you will use the default texture instead. Should we rather quit the app with some error report? Just my personal opinion.

chenchao1407:
Just leave some comments there. I just tried the model loader app. I am not quite sure if GVRF has difficulty handling complicated model or there is any other issue. The side effect is not that significant without goggle. But if you wear a goggle and test with the app. You can see there is something like lagging or frames jumping. It looks really weird to me.

bonniemathew:
Hi Chao, 
I was looking into the performance issue in the model loading app. I could see some frame skips in the surfaceflinger and at those times there were increased load on the GPU. Are you aware of an profiling tool which we can use for profiling the gvrf apps? This is the screenshot we got from Systrace. 

image

Also we will try loading various 3d models and check the performance. We will update you with the results.

JonShemitz:
Good old Traceview does a decent job ...

lokeshguddu:
I tried to take the gl trace and found out that the glCompileShader(...) is being called more than once. This might be causing the lag and frame drop.


gl_trace

soumyajitdeb:
It appears that glCompileShader is being called at the mesh level for each mesh multiple times. This regular shader recompilation might end up slowing down the GPU as it takes a significant amount of time. We are still investigating why this occurs so frequently. It also seems to be happening for each mesh in the hierarchy. Ideally this should only happen once per geometry that has materials assigned at model load time. 

-Soumyajit

.gitignore
JonShemitz: > I am not sure if we should add "ovr_mobile_sdk" as gitignore. It can be a convenience but it seems a little weird to me.

Weird in what way? It certainly seems sensible to me - we require `ovr_mobile_sdk`, then have to keep ignoring its presence in `git status` reports.

GVRf/Framework/jni/engine/importer/assimp_importer.cpp
JonShemitz: >  if any error occurs while loading texture from resource, you will use the default texture instead. Should we rather quit the app with some error report?

Crashing with an error report is certainly 'the GVRF way' but I think you can argue that it's not 100% appropriate here: If we crash with a message like "Couldn't find texture 'bricks_and_mortar'" all the user will really know is that they were loading some model; the JNI code doesn't have any way to tell the user which model is being loaded. We certainly _could_ pass strings on down, but that's a lot of overhead for a mistake that (hopefully!) will get flushed during the very first trials. Showing a default texture will help the developer figure out which model is out of sync with the texture resource names.

That is, yes, this is a bit inconsistent, but that doesn't seem so horrible. Overall, I have no strong opinions, here, and think either approach is OK.

JonShemitz:
> glCompileShader is being called at the mesh level for each mesh multiple times. 

This is strange, as GVRF seems to do a pretty good job of managing compiled shaders. I'll refrain from re-reviewing until you find the issue and re-submit.
_____________________________________________

> We've removed the reference to loadBitmap with a different customized function that uses GVRAndroidResource

Good. Imho, you should have an overload that takes a texture loading callback. (The current method(s) could just use a default callback that maps names to asset-relative files.) This would let the app handle any mapping from `"texture_name"` to `R.drawable.texture_name` *and* would let the app maintain its own cache, so that (say) a set of models that all use the `"bricks_and_mortar"` texture all use the same `GVRTexture`.

> We now have two functions - loadScene() and loadModelAsSceneObject() for loading the 3d model as a scene or a scene object.

I'm all for nice long, descriptive names, but a method that returns a `GVRSceneObject` probably doesn't need to be called `loadModelAsSceneObject()` - `GVRSceneObject loadModel(GVRAndroidResource);` would seem to be nice and clear.

JonShemitz:
>  `loadScene()` and `loadModelAsSceneObject()`

Sorry, should have mentioned this sooner: What's your async strategy?

Loading a simple mesh is already pretty slow; recursively loading *and* loading textures is bound to be even slower. Seems to me like you should (at least) be using the new `GVRContext. loadFutureTexture( GVRAndroidResource )` API; it would probably make sense to support both synchronous and asynchronous APIs, the way we do elsewhere:

    GVRScene loadScene(...);
    Future<GVRScene> loadFutureScene(...);
    GVRSceneObject loadModel(...);
    Future<GVRSceneObject> loadFutureModel();

lokeshguddu:
Thanks @PatrickTW for your valuable suggestions . I will be more careful in the future, about the comments. :-)

PatrickTW:
Lokesh,

I was asked to review comments for code submittals.
Yours are in the A++ category.  The things I mentioned are well within the range of normal communication.
Adhering to those guidelines makes good writing even better.


Patrick

JonShemitz:
@soumyajitdeb, are you still working on the re-compilation bug? Is there anything we can do to help, here in Mountain View?

soumyajitdeb:
Many thanks for the offer of help. Yes we are still looking at this. We found that some of the shader recompilations are happening within oculus sdk itself. 

SRI-B was closed today because of a public holiday. We shall check and get back to you on Monday regarding a resolution and help if required. Many apologies for the delay.

Thanks
Soumyajit

soumyajitdeb:
We finally managed to fix the issue of stuttering. It turns out shader compilation was not the issue. The stuttering was strangely due to lack of mipmapping and too many large textures being pulled over the memory bus. We've enabled mipmaps in GVRF via the glGenerateMipmap call and also modified the corresponding glTexParameter calls to do trilinear filtering. Now the demo runs absolutely smoothly.

-Soumyajit

lokeshguddu:
1.	The name of the method has been changed from loadModelAsSceneObject(…) to loadModel(…).
2.	The new method to load the bitmap handles both the resource filename as well as the filename from the asset folder.
3.	The rendering delay has been fixed by the use of mipmap in the base texture.
4.  The loadModel(...) and loadScene(...) loads all the resources and textures asynchronously since internally textures & resources are loaded via the GVRAsynchronousResourceLoader.

- Soumyajit & Lokesh

JonShemitz:
Sounds good! @chenchao1407, would you please re-test and re-read the code? If it looks good to you, I'll make the time for a line-by-line re-review.

lokeshguddu:
Hi,

We have created a separate group for changes made by SRI-B. So once this pull request is confirmed to be working, we can then make all these changes through just one commit from the https://github.com/SRIB-GRAPHICS/GearVRf. It will keep things simple and well organized. For all future feature addition we will be using https://github.com/SRIB-GRAPHICS/ account.

chenchao1407:
Hi,
Actually  I don't think you have to re-org 24 commits into 1. In my opinion it may be cleaner but it may make us lose the capability to trace back each change.
I agree that you can re-org the commits using something like git rebase -i and then combine or we can say eliminate some commit like testing push, testing over. But I think it will be better to keep the valid commits information which may help us debug in the future.

GVRf/Framework/jni/engine/importer/assimp_importer.cpp
chenchao1407: My thinking is like from the aspect from new coming developers, if they see some textures which are not supposed to look like, they are not going to know what is happening. I mean, replacing bad textures with default textures is like an implicit way to tell developers what is going wrong but they may not get the point. If you don't want to crash the app, at least we should provide something like log information or even mention the default texture in JavaDoc to tell developers why there are some strange textures they may not know anything about it.

chenchao1407:
Hi

Reviewed the three new commits and they look good to me. Left a comment on the default texture. Tried and everything looks much better now.

Thanks

JonShemitz:
Thanks, @chenchao1407 - I'll do a line-by-line first thing tomorrow.

chenchao1407:
@JonShemitz 

Hi Jon, maybe it's your turn to review the codes now. Just FYI, you can just click into the three new commits to see what they changed recently to avoid reading the large amount of codes again. :-)

GVRf/Framework/jni/engine/importer/assimp_importer.cpp
jason2kim: This was my major concern from your previous patch.
I am already seeing these code re-uses in the new patch, including this:
+ float camera_separation_distance_ = camera_rig-default_camera_separation_distance(); 

May be we can further do code modularizartions in the future as the features grow. Thanks.

JonShemitz:
I'll go through the new commits and add some line notes (this afternoon, after I run some errands) but I thought it would be better to add a global response, first.

1. Loading bitmaps is slow; loading models is even slower. Doing either on the GL thread is a recipe for long startup times or animation stutters ... or both. We've added high and low level async methods for loading meshes and textures; they improve startup times by parallelizing loads, and they avoid stutter by doing expensive operations on background threads. Your new methods are even more expensive than the synchronous methods we have warned can take a very long time ... yet you have no async versions of the model loading methods.
2. It's quite possible that an app will want to load several models that share textures. (For example, a robot army might all share the same `carbon-fiber`, `stainless-steel` and `glowing-eye` textures.) This code would use a different texture in each model (incurring load costs for each). 

These are big issues, and I've mentioned them before. Either one would be enough to for me to block this; I would like to see both addressed.

There's also a new issue with this code. As _per_ issue https://github.com/Samsung/GearVRf/issues/24, we keep finding problems with the GVRF reference counting mechanism; Tom has asked me to go ahead and replace it with a more conventional `finalize()` based mechanism, where the Java objects own the native objects and the native objects are only deleted when the Java objects are garbage collected. This will eliminate a whole class of bugs that happen in 'real' apps, where objects may be removed from the scene graph and later added back; it will also eliminate the need for 'aliasing' tricks (see `GVRShaders.setMainTexture()`) or the `GVRHybridObject.setKeepWrapper()` hack.

This affects your code, because it's a huge wodge of JNI code that basically returns a single `GVRScene` object: there are no Java objects corresponding to the meshes or textures or even scene objects. (Yes, I do see that the `loadModel()` method creates wrappers for the scene objects.) As far as I can see, the only way your model loading work and my memory management work are going to be compatible is if your scene-recursion code is moved out of the JNI and into the Java, creating Java classes that own each GVRF-internal object in the 'normal' way.

Someone's going to have to do this port. If your scene loading code were fine otherwise, I'd see this as just another piece of my huge task; given the other issues with your scene loading code, I'd sure like to ask you guys to do the port.

soumyajitdeb:
Thank you for the comments! Regarding 1&2, we agree that the ideal scenario would be to have the entire code block moved to Java so that the parsing of the hierarchy creates java objects for each node. This would allow the new memory management system to function correctly. Unfortunately this means a complete rewrite of the entire code. 

This pull request was originally meant for the version of GVRF which was present at the repo at the time of authoring and as such was never designed to follow an asynchronous loading strategy. Now that the framework has evolved and is more efficient, we'd definitely want to support the changes. However, the primary problem we are facing is that we'd need to be able to access the references of the assimp objects on the java side to be able to parse the complete hierarchy. Currently no such functionality exists within the java code of GVRf. This involves either writing JNI wrappers for all required critical functionality (rather time consuming) or use jassimp (https://github.com/assimp/assimp/tree/master/port/jassimp) for making the functionality visible at the java layer.

We wish to know if it is okay to include jassimp with our codebase - this will be a much faster process and since jassimp is an open source project, this layer would get updated alongwith any changes to assimp itself. If we can't use jassimp, we'd end up spending a lot of time writing JNI glue code.

Thanks
Soumyajit

thomasflynn:
Jassimp sounds interesting and seems like the way to go.  Is it up to date with Assimp 3.1?  Brazil (SIDIA) needs fbx support, which is included in 3.1.  Also, they were having trouble loading some fbx files.  Is that something you could diagnose?  If so, I’ll privately send you the file.

Btw, when can we expect a pull-request for frustum culling?  I expect that can be done separate from the Jassimp work.
-tom

soumyajitdeb:
We'll go ahead and use jassimp in that case for model loading. The frustum culling update has been sent as a pull request by Lokesh. I shall be done with the more efficient frustum+occlusion culling system by tomorrow. 

As for the fbx issue, please pass us the file. We'll see what we can do at our end. I've handled fbx files before but because of their proprietory nature, they can be a bit painful to load at times.

thomasflynn:
so, how is it going?

lokeshguddu:
Hi,

To make the full use of async implementation in GVRF, we started with jassimp to get all the assimp data in the Java side and handle it with existing Asynchronous implementation. The initial integration is already done(jassimp-integration branch  of  https://github.com/SRIB-GRAPHICS/GearVRf). The functionality is working fine for models with less number of materials but crashes for larger models with the error [ Failed adding to JNI local ref table (has 512 entries) ]. We have tried all the solutions available to fix this issue without any luck. Any help regarding this issue will speed things up as I personally do not have any solution as of now. I'm attaching the logcat file for the test when sponza.dae model was used. The integration is working fine with the sample application [gvrjassimptest].

Thanks.




.
.
.

E/gvrf    (11560):    converting property $clr.diffuse ...

E/gvrf    (11560):    converting property $clr.specular ...

E/gvrf    (11560):    converting property $clr.emissive ...

E/gvrf    (11560):    converting property $clr.transparent ...

E/gvrf    (11560):    converting property $clr.reflective ...

E/gvrf    (11560):    converting property $mat.shininess ...

E/gvrf    (11560):    converting property $mat.reflectivity ...

E/gvrf    (11560):    converting property $mat.refracti ...

E/dalvikvm(11560): JNI ERROR (app bug): local reference table overflow (max=512)

W/dalvikvm(11560): JNI local reference table (0x7e93e1e8) dump:

W/dalvikvm(11560):   Last 10 entries (of 512):

W/dalvikvm(11560):       511: 0x418babb8 java.lang.Class<java.lang.Float>

W/dalvikvm(11560):       510: 0x4230c2c0 java.lang.String "$mat.refracti"

W/dalvikvm(11560):       509: 0x4230c278 org.util.jassimp.AiMaterial$Property

W/dalvikvm(11560):       508: 0x4230c268 java.lang.Float

W/dalvikvm(11560):       507: 0x422cc820 java.util.ArrayList

W/dalvikvm(11560):       506: 0x4230c210 java.lang.String "$mat.reflectivit... (17 chars)

W/dalvikvm(11560):       505: 0x4230c1c8 org.util.jassimp.AiMaterial$Property

W/dalvikvm(11560):       504: 0x423ec8a8 java.lang.Float

W/dalvikvm(11560):       503: 0x422cc820 java.util.ArrayList

W/dalvikvm(11560):       502: 0x423ec858 java.lang.String "$mat.shininess"

W/dalvikvm(11560):   Summary:

W/dalvikvm(11560):        20 of java.lang.Class (3 unique instances)

W/dalvikvm(11560):       110 of java.lang.String (110 unique instances)

W/dalvikvm(11560):       156 of java.nio.DirectByteBuffer (156 unique instances)

W/dalvikvm(11560):        14 of java.lang.Float (14 unique instances)

W/dalvikvm(11560):        21 of java.lang.Integer (21 unique instances)

W/dalvikvm(11560):        67 of java.util.ArrayList (4 unique instances)

W/dalvikvm(11560):        38 of java.nio.ByteBuffer[] (8 elements) (38 unique instances)

W/dalvikvm(11560):        18 of org.util.jassimp.AiColor (18 unique instances)

W/dalvikvm(11560):        64 of org.util.jassimp.AiMaterial$Property (64 unique instances)

W/dalvikvm(11560):         1 of org.util.jassimp.AiScene

W/dalvikvm(11560):         3 of org.util.jassimp.AiMaterial (3 unique instances)

E/dalvikvm(11560): Failed adding to JNI local ref table (has 512 entries)

.
.
.

JonShemitz:
1.       http://stackoverflow.com/questions/17033438/android-expand-jni-local-reference-table may help with your JNI issue.

2.       I don’t think AiScene getAssimpScene(GVRAndroidResource androidResource) is the right prototype for your new GVRContext method:

a.       It should return a GVRSceneObject, not an AiScene.

b.      Imho, there still needs to be a resource loading callback. I’m adding caching (so that if you load R.drawable.foo and it’s still in memory, it won’t get reloaded) but there still needs to be a way for apps to intercept an attempt to load (say) the “bricksAndMortar” texture and map that to R.drawable.bricks_and_mortar or to “assets/models/textures/bricksAndMortar.png”.

lokeshguddu:
Hi,
1.       http://stackoverflow.com/questions/17033438/android-expand-jni-local-reference-table may help with your JNI issue.

         >> I already tried those hacks and it still fails.

2.       I don’t think AiScene getAssimpScene(GVRAndroidResource androidResource) is the right prototype for your new GVRContext method:

a.       It should return a GVRSceneObject, not an AiScene.

b.      Imho, there still needs to be a resource loading callback. I’m adding caching (so that if you load R.drawable.foo and it’s still in memory, it won’t get reloaded) but there still needs to be a way for apps to intercept an attempt to load (say) the “bricksAndMortar” texture and map that to R.drawable.bricks_and_mortar or to “assets/models/textures/bricksAndMortar.png”.

    >>  About point 2, that is already in pipeline once we get around this JNI issue. This was just a initial commit to at least get things working and later we can modify it according to our need as per the requirements which have been mentioned by you.

JonShemitz:
Perhaps, then, you should try to do less in the JNI code.

thomasflynn:
So, given the jassimp pull-request, this one could be closed, correct?

lokeshguddu:
Closing this pull request. Please refer to [https://github.com/Samsung/GearVRf/pull/138] for the same functionality.

soumyajitdeb:
Closing this request as there is another one for jassimp loader [#138]

PR #23, opened 2015-04-17
Fix for (GVRF-14) Crash in video sample

jason2kim:
GearVRf-DCO-1.0-Signed-off-by: Jason Kim
jason2.kim@samsung.com

### DIFF
--- a/GVRf/Framework/jni/objects/textures/texture.h
+++ b/GVRf/Framework/jni/objects/textures/texture.h
@@ -22,7 +22,6 @@
 #define TEXTURE_H_
 
 #include "gl/gl_texture.h"
-
 #include "objects/recyclable_object.h"
 
 namespace gvr {
@@ -41,6 +40,10 @@ public:
     }
 
     virtual GLuint getId() const {
+        if (gl_texture_ == 0) {
+            // must be recycled already. The caller will handle error.
+            return 0;
+        }
         return gl_texture_->id();
     }
 
--- a/GVRf/Framework/jni/shaders/material/unlit_shader.cpp
+++ b/GVRf/Framework/jni/shaders/material/unlit_shader.cpp
@@ -85,6 +85,11 @@ void UnlitShader::render(const glm::mat4& mvp_matrix,
         throw error;
     }
 
+    if (texture->getId() == 0) {
+        std::string error = "UnlitShader::render : texture with invalid Id";
+        throw error;
+    }
+
 #if _GVRF_USE_GLES3_
     mesh->setVertexLoc(a_position_);
     mesh->setTexCoordLoc(a_tex_coord_);
--- a/GVRf/Sample/gvr-video/src/org/gearvrf/video/VideoScript.java
+++ b/GVRf/Sample/gvr-video/src/org/gearvrf/video/VideoScript.java
@@ -410,6 +410,12 @@ public class VideoScript extends GVRScript {
                     mGVRContext, "button/play-active.png"));
             mPlayPauseButton = new GVRSceneObject(gvrContext,
                     gvrContext.createQuad(0.7f, 0.7f), mInactivePause);
+            mPlayPauseButton.getRenderData().getMaterial().setTexture("active_play", mActivePlay);
+            mPlayPauseButton.getRenderData().getMaterial()
+                    .setTexture("inactive_play", mInactivePlay);
+            mPlayPauseButton.getRenderData().getMaterial().setTexture("active_pause", mActivePause);
+            mPlayPauseButton.getRenderData().getMaterial()
+                    .setTexture("inactive_pause", mInactivePause);
             mPlayPauseButton.getTransform().setPosition(0.0f, -0.8f, -8.0f);
             mPlayPauseButton.getRenderData().setRenderingOrder(
                     GVRRenderingOrder.TRANSPARENT + 1);
@@ -429,6 +435,8 @@ public class VideoScript extends GVRScript {
                     mGVRContext, "button/front-active.png"));
             mFrontButton = new GVRSceneObject(gvrContext,
                     gvrContext.createQuad(0.7f, 0.7f), mInactiveFront);
+            mFrontButton.getRenderData().getMaterial().setTexture("active_front", mActiveFront);
+            mFrontButton.getRenderData().getMaterial().setTexture("inactive_front", mInactiveFront);
             mFrontButton.getTransform().setPosition(1.2f, -0.8f, -8.0f);
             mFrontButton.getRenderData().setRenderingOrder(
                     GVRRenderingOrder.TRANSPARENT + 1);
@@ -448,6 +456,8 @@ public class VideoScript extends GVRScript {
                     mGVRContext, "button/back-active.png"));
             mBackButton = new GVRSceneObject(gvrContext, gvrContext.createQuad(
                     0.7f, 0.7f), mInactiveBack);
+            mBackButton.getRenderData().getMaterial().setTexture("active_back", mActiveBack);
+            mBackButton.getRenderData().getMaterial().setTexture("inactive_back", mInactiveBack);
             mBackButton.getTransform().setPosition(-1.2f, -0.8f, -8.0f);
             mBackButton.getRenderData().setRenderingOrder(
                     GVRRenderingOrder.TRANSPARENT + 1);
@@ -466,6 +476,8 @@ public class VideoScript extends GVRScript {
                     mGVRContext, "button/imaxselect.png"));
             mImaxButton = new GVRSceneObject(gvrContext, gvrContext.createQuad(
                     0.9f, 0.35f), mInactiveImax);
+            mImaxButton.getRenderData().getMaterial().setTexture("active_imax", mActiveImax);
+            mImaxButton.getRenderData().getMaterial().setTexture("inactive_imax", mInactiveImax);
             mImaxButton.getTransform().setPosition(2.5f, -0.9f, -7.5f);
             mImaxButton.getRenderData().setRenderingOrder(
                     GVRRenderingOrder.TRANSPARENT + 1);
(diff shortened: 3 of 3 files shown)


GVRf/Framework/jni/objects/textures/texture.h
```diff
@@ -41,6 +41,11 @@ class Texture: public RecyclableObject {
     }
 
     virtual GLuint getId() const {
+        if (gl_texture_ == 0) {
```
JonShemitz: We generally don't check in code for debugging a particular issue: do you think this is going to be useful in other contexts?

GVRf/Framework/jni/shaders/material/unlit_shader.cpp
```diff
@@ -84,6 +84,11 @@ void UnlitShader::render(const glm::mat4& mvp_matrix,
         std::string error = "UnlitShader::render : texture with wrong target";
         throw error;
     }
+    if (texture->getId() == 0)
+    {
+        std::string error = "UnlitShader::render : texture with invalid Id";
+        throw error;
```
JonShemitz: Doesn't this crash the app? That's what happens when other JNI code throws exceptions ....

GVRf/Sample/gvr-video/src/org/gearvrf/video/VideoScript.java
```diff
@@ -632,6 +632,38 @@ public void run() {
             mainScene.getMainCameraRig().getOwnerObject()
                     .addChildObject(mPassThroughObject);
 
+            /*
+             * Selective objects to keep
```
JonShemitz: I think it would probably be better to set these extra reference counts at the point where mPlayPlauseButton / mFrontButton / &c are each initialized, not all in a 'add reference' block at the end.

GVRf/Sample/gvr-video/src/org/gearvrf/video/VideoScript.java
```diff
@@ -632,6 +632,38 @@ public void run() {
...
+             */
+            mPlayPauseButton.getRenderData().getMaterial().setTexture("active_play", mActivePlay);
```
JonShemitz: I _know_ this sample is full of repeated calls like `mPlayPauseButton.getRenderData().getMaterial()`, but they really aren't free and I don't think we should be encouraging what amounts to a bad idea.

That is, I'd sure prefer it if you changed code like

            mPlayPauseButton = new GVRSceneObject(gvrContext,
                    gvrContext.createQuad(0.7f, 0.7f), mInactivePause);
            mPlayPauseButton.getTransform().setPosition(0.0f, -0.8f, -8.0f);
            mPlayPauseButton.getRenderData().setRenderingOrder(
                    GVRRenderingOrder.TRANSPARENT + 1);
            mPlayPauseButton.getRenderData().setOffset(true);
            mPlayPauseButton.getRenderData().setOffsetFactor(-1.0f);
            mPlayPauseButton.getRenderData().setOffsetUnits(-1.0f);

to 

            mPlayPauseButton = new GVRSceneObject(gvrContext,
                    gvrContext.createQuad(0.7f, 0.7f), mInactivePause);
            mPlayPauseButton.getTransform().setPosition(0.0f, -0.8f, -8.0f);
            GVRRenderData playPauseRenderData = mPlayPauseButton.getRenderData();
            playPauseRenderData.setRenderingOrder(
                    GVRRenderingOrder.TRANSPARENT + 1);
            playPauseRenderData.setOffset(true);
            playPauseRenderData.setOffsetFactor(-1.0f);
            playPauseRenderData.setOffsetUnits(-1.0f);

throughout `onInit()`. Otoh, that does go far beyond fixing a bug, so I won't insist on it.

GVRf/Framework/jni/shaders/material/unlit_shader.cpp
jason2kim: As a note,
the if (gl_texture_ == 0)  block is not a debugging code but an error check routine to avoid app crash. Along with “throw error”, it will be consistent to egvrf’s error handling convention.

GVRf/Framework/jni/objects/textures/texture.h
JonShemitz: OK, I misread this - you're preventing a null deref.

GVRf/Framework/jni/objects/textures/texture.h
```diff
@@ -41,6 +41,11 @@ class Texture: public RecyclableObject {
...
+            // must be recycled already. The caller will handle error.
+            LOGW("getId() but gl_texture_ == 0 Texture=%d", this);
```
JonShemitz: %p might be better than %d ...

GVRf/Framework/jni/shaders/material/unlit_shader.cpp
JonShemitz: Consistent. yes ... with code that Chao has had to fix ....

GVRf/Sample/gvr-video/src/org/gearvrf/video/VideoScript.java
jason2kim: I also thought about this, it would be code purpose versus code neatness. This object ref count mechanism even can be changed in the future. 
For now, neatness seems more valued, and let me change like that as Jon’s suggestion.

jason2kim:
updated

GVRf/Framework/jni/shaders/material/unlit_shader.cpp
```diff
@@ -84,6 +84,11 @@ void UnlitShader::render(const glm::mat4& mvp_matrix,
         std::string error = "UnlitShader::render : texture with wrong target";
         throw error;
     }
+    if (texture->getId() == 0)
+    {
```
JonShemitz: Inconsistent formatting ... please use Shift+Ctrl+F with the "K & R" builtin, modified so the "Tab policy" is "Spaces only".

GVRf/Framework/jni/shaders/material/unlit_shader.cpp
JonShemitz: Chao agrees: this will crash the app.

JonShemitz: OK - we're wrong! *This* code is called within a try/catch block.

GVRf/Framework/jni/shaders/material/unlit_shader.cpp
jason2kim: Let me format before submitting, using the Eclipse imported formatter. Thanks for pointing out.

chenchao1407:
Patch verified. The gvrvideo app doesn't get crashed any more.

GVRf/Framework/jni/shaders/material/unlit_shader.cpp
```diff
@@ -85,6 +85,11 @@ void UnlitShader::render(const glm::mat4& mvp_matrix,
         throw error;
     }
 
+    if (texture->getId() == 0) {
+        std::string error = "UnlitShader::render : texture with invalid Id";
```
liaxim: Isn't this misleading? The actual thing that happened is that the object was recycled.

GVRf/Sample/gvr-video/src/org/gearvrf/video/VideoScript.java
```diff
@@ -410,6 +410,12 @@ public void run() {
                     mGVRContext, "button/play-active.png"));
             mPlayPauseButton = new GVRSceneObject(gvrContext,
                     gvrContext.createQuad(0.7f, 0.7f), mInactivePause);
+            mPlayPauseButton.getRenderData().getMaterial().setTexture("active_play", mActivePlay);
```
liaxim: Although I have been using this "pattern" to keep textures from recycling (most of the time until recently), isn't it more appropriate to use setKeepWrapper(true) on the texture instead of this roundabout way to add a reference?

GVRf/Framework/jni/shaders/material/unlit_shader.cpp
JonShemitz: In this case - are you so sure that this may not happen any other way?

GVRf/Sample/gvr-video/src/org/gearvrf/video/VideoScript.java
JonShemitz: No, I don't think so. It's `setKeepWrapper()` not `setKeepAlive()` precisely because we did already have a mechanism for artificially inflating reference counts ... and because `setKeepWrapper()` _is_ rather hacky and not something we should be encouraging.

GVRf/Framework/jni/shaders/material/unlit_shader.cpp
liaxim: Based on my experience as a user of GVRF, it is helpful to be able to get diagnostics about (incorrect) usage of recycled objects. It could direct the application developer's attention to the right thing. I know something like this would have saved me time - I had the same problem as the one in VideoScript.java.

And to answer your question: Texture::getId() proxies to GLTexture::id(); so looking at GLTexture, glGenTextures is not going to write a texture name of 0, after assignment id_ is not mutated; if glGenTextures fails, then id_ could be anything; including 0, I guess.


---
Should this "if getId() == 0" check be applied to other renderers?

JonShemitz: > Should this "if getId() == 0" check be applied to other renderers?

I think we'd be better off looking into ways to keep early-recycloing from happening at all.

GVRf/Sample/gvr-video/src/org/gearvrf/video/VideoScript.java
chenchao1407: I think it may be an issue for us to write a sample that uses setKeepWrapper API(when necessary) or otherwise, developers may be confused with this.

liaxim: It is no less confusing to me than the current sample that sets the texture to key "active_play" in the material, and when actually needed for rendering - sets it as the main texture. What "active_play" is for is completely opaque and pretty much anyone who looks at the code for the first time will not get it. While I am anything but advocate of setKeepWrapper, its use would raise the awareness of how things work and how the framework does its own reference counting.

It may be a kludge but it is a highly visible method. Too late to try hiding it.

jason2kim: Either way is not clean for me. Discussion on changing the ref count mechanism as a permanent solution is going on a separate thread, where I put my ideas there.

Issue #30, opened 2015-04-23
GVRSceneObject::setName(String) - native crash


nathan-almeida:
When setting a null String in **GVRSceneObject::setName(String)**, a native crash occurs.

```java
public void testSceneObjectSetGetNameNull() {
        GVRMesh mesh = TestDefaultGVRViewManager.mGVRContext.loadMesh(UtilResource.androidResource(TestDefaultGVRViewManager.mGVRContext,R.mipmap.cylinder));
        GVRSceneObject sceneObject = new GVRSceneObject(TestDefaultGVRViewManager.mGVRContext, mesh);
        try {
            sceneObject.setName(null);
            fail();
        } catch (NullPointerException e) {
        }
    }
```

liaxim:
Will close in few days as it is just too outdated.

Issue #29, opened 2015-04-23
GVRMaterial::getVec2(String) - native crash


nathan-almeida:
When passing a not inatialized key to **GVRMaterial::getVec2(String)**, a native crash occurs.

```java
public void testGetVec2WithAbsentKey() {
        mColorShader = new ColorShader(TestDefaultGVRViewManager.mGVRContext);
        GVRMaterial material = new GVRMaterial(TestDefaultGVRViewManager.mGVRContext, mColorShader.getShaderId());
        assertNull(material.getVec2("radio_r"));
    }
```

If we add the line
```java
material.setVec2("radio_r", 0, 0);
```

before calling the **get** method, the error doesn`t happen anymore.

This also happens for **getVec3** and **getVec4**

liaxim:
Will close in few days as it is just too outdated.

Issue #28, opened 2015-04-22
GVRCameraRig::getFloat(String) - native crash


nathanalmeida:
When passing a not inatialized  key to **GVRCameraRig::getFloat(String)**, a native crash occurs.

``` java
public void testGetFloatWithoutSettingItBefore() {
        mGVRContext.getMainScene().getMainCameraRig().getFloat("ratio_r");
}
```

If we add the line

``` java
mGVRContext.getMainScene().getMainCameraRig().setFloat("ratio_r", 1.0f);
```

before calling the **get** method, the error doesn`t happen anymore.

nathan-almeida:
In fact, maybe just a better feedback for the user, telling in which line or which object made this issue happen, could be enough.

liaxim:
Will close in few days as it is just too outdated.

PR #31, opened 2015-04-24
fix for some similar issues of,

jason2kim:
GVRSceneObject::setName(String) - native crash (#30)
GVRMaterial::getVec2(String) - native crash (#29)
GVRCameraRig::getFloat(String) - native crash (#28)

GearVRf-DCO-1.0-Signed-off-by: Jason Kim
	jason2.kim@samsung.com


JonShemitz:
Seems like it would be better to do the tests on the Java side, where you can throw an `IllegalArgumentException` - I don't think we have any problem with JNI code passing bad parameters. The issue is more that we want to fail gracefully when a Java app programmer does something careless.

chenchao1407:
I am going to make a new patch for pre-condition check, maybe you can update this pull request based on that.

chenchao1407:
Hi, Jason.

I think you can continue working on this since my pull request is merged.

GVRf/Framework/jni/objects/components/camera_rig_jni.cpp
```diff
@@ -197,16 +195,22 @@ Java_org_gearvrf_NativeCameraRig_setCameraSeparationDistance(
...
-    jfloat camera_rig_float = camera_rig->getFloat(native_key);
     env->ReleaseStringUTFChars(key, char_key);
-    return camera_rig_float;
+    try {
+        jfloat value = camera_rig->getFloat(native_key);
+        jfloatArray result_float = env->NewFloatArray(1);
+        env->SetFloatArrayRegion(result_float, 0, 1, &value);
+        return result_float;
+    } catch (const std::string & err) {
+        return NULL;
```
chenchao1407: I think, overall is fine. One small concern is about performance. I know that you may not want to use NaN or Infinity as illegal state flag. But I think this way may be a little bit worse in terms of performance since you are going to create an array each time.

I mean, once if this one or similar methods get called so many times, it may cause a little performance degradation. Of course this is not the case in our sample apps right now and I don't see if there is any such case in the future. However compared to using NaN, this is not necessarily safer but still a little bit worse in performance.

GVRf/Framework/src/org/gearvrf/GVRCameraRig.java
```diff
@@ -145,7 +145,11 @@ public void setCameraSeparationDistance(float distance) {
      * @return The {@code float} value associated with {@code key}.
      */
     public float getFloat(String key) {
-        return NativeCameraRig.getFloat(getPtr(), key);
+        float[] value = NativeCameraRig.getFloat(getPtr(), key);
+        if (value == null) {
```
chenchao1407: Imho, I don't think it is necessary, but for code consistency, maybe you'd better use Preconditions.checkNotNull(). And also, I think it will help reduce the duplicate codes for every getX().

Thanks

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
```diff
@@ -239,15 +239,18 @@ public String getName() {
...
-     * Scene object names are not needed: they are only for the application's
-     * convenience.
-     * 
-     * @param name
-     *            Name of the object.
+     * @param name Name of the object.
+     * @throws IllegalArgumentException if input parameter is null or empty.
      */
     public void setName(String name) {
+        if (name == null || name.trim().isEmpty()) {
```
chenchao1407: Imho, it won't be cheap if we do trim() check on every input string since it doesn't only check for space but also check for some other characters, and even finally generate a new String. We may discuss if it is necessary to do such check and if necessary we can change Preconditions.checkStringNotNullOrEmpty().  We should unify the standards for each similar check and finally make code consistent and reduce duplicate codes.

GVRf/Framework/jni/objects/components/camera_rig_jni.cpp
jason2kim: As you mentioned mixing up error codes (object not found) with data value (NaN or whatever) is not a good idea. We can easily think of many many reasons why it is bad. We have to avoid faking error codes wherever possible. I believe using NULL is one of most natural ways in handling object-not-found, either it is vector or float, and consistent across our cods too. Performance difference on this must be trivial.

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
jason2kim: Yes, we must check the given string for white spaces by suppressing them all, either leading or trailing. The trimmed string only can be used as a valid key in the map. Unfortunately, as I pointed out previously, Preconditions.checkStringNotNullOrEmpty() is not suitable for this case, as it just check null and strlen zero. Considering usage of this setName(), performance is not a problem.

GVRf/Framework/jni/objects/components/camera_rig_jni.cpp
JonShemitz: > As you mentioned mixing up error codes (object not found) with data value (NaN or whatever) is not a good idea. We can easily think of many many reasons why it is bad. We have to avoid faking error codes wherever possible. I believe using NULL is one of most natural ways in handling object-not-found, either it is vector or float, and consistent across our cods too. Performance difference on this must be trivial.

Sorry, @jason2kim , I'm with Chao on this one: Creating an array to return a single `float` is adding a lot of overhead to the normal case. It would be reasonable if every possible `float` were a valid result, but that's not the case here. A `NaN` is "not a number" - and can't be passed in - so it does make sense to use it as an error code.

What you have coded is akin to a `find()` method returning an `Integer` so it can return `null` on not-found, instead of simply returning a `-1`. You wouldn't do that, because of the overhead of creating the reference type and then the cost of garbage collecting it ... and the same reasoning applies here.

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
JonShemitz: > Yes, we must check the given string for white spaces by suppressing them all, either leading or trailing. The trimmed string only can be used as a valid key in the map. Unfortunately, as I pointed out previously, Preconditions.checkStringNotNullOrEmpty() is not suitable for this case, as it just check null and strlen zero. Considering usage of this setName(), performance is not a problem.

Why are we even checking `GVRSceneObject.setName()` at all? We should check for `null` if (and _only_ if) the JNI crashes on `env->GetStringUTFChars(name, 0)` when name is `null` ... but we really shouldn't care if an app wants to name a scene object `""` or `" "` or even `" \t \n \t "`. The name is _not_ a key; the name is just stored.

jason2kim: No,  with map, I didn’t mean anything related to <SceneObject>*. I was saying possible uses of maps in the applications themselves, where they will use the name as a key with logical names.

GVRf/Framework/jni/objects/components/camera_rig_jni.cpp
jason2kim: More correct way would be using a separate parameter as error code and another one as return value, like float and vec arrary. But, considering JNI, I think the current implementation is the optimal, with minimum confusion.

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
JonShemitz: I agree that all-whitespace names would be easy to confuse and/or hard to spot in a log; I don’t see how it is GVRF’s responsibility to guard against that. The point of these parameter checks is to avoid native crashes, not to keep app developers from doing something that seems foolish to us but may actually have some sound reason in the context of the app.

GVRf/Framework/jni/objects/components/camera_rig_jni.cpp
JonShemitz: Jason, you’ve got two programmers telling you that they disagree. NULL is just as much a special pointer value as NaN is a special float value … and creating an array adds cost.

chenchao1407:
I think we need to have a face-to-face discussion tomorrow about this issue and make a standard policy throughout the whole GVRF. And if necessary  will update Preconditions class.

Thanks in advance.
Sincerely
-Chao Chen

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
chenchao1407: Hi All.

I think we may need to make an agreement on these cases. Let's discuss about it tomorrow and if it is really necessary, I will write a new String check API which is similar to String.trim() but without generate a new sub String which I think is redundant.

chenchao1407:
I will make another patch for these issues.

PR #33, opened 2015-04-29
Introduce class Preconditions and restructure some parameter checks in G...

chenchao1407:
...VRMesh, GVRMaterial, and GVRCameraRig

One of our goal is to provide Java side exception as possible as we can while in the illegal states.
Previously I impelment a GVRJniException class which can simulate Java stack trace and throw a JNI crash finally.I am thinking if we should avoid using this class as possible as we can. So for the readability and further robustness in GVRF, I want to introduce the class Preconditions and provide some convenient pre-condition check at Java level so that we can avoid using GVRJniException when crash happens in JNI side.

But I think the use of Preconditions may need to be very careful since unnecessary checks will eventually turn out to be a significant overhead. For example, in GVRMaterial and GVRCameraRig I finally decided to only add NaN check in setFloat is because only in the case of single float parameter may need NaN as a special flag to inform Java that an illegal state happens. Corresponsively, in the case of float vector/array, we can simply return null pointer back from JNI to Java as an illegal state identifier. I mean, we probably can't check all illegal states using Preconditions. But in necessary cases, I think we'd better make use of it and in the future extend it in a proper way.

c.chen@samsung.com

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/GVRCameraRig.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRCameraRig.java
@@ -15,6 +15,9 @@
 
 package org.gearvrf;
 
+import static org.gearvrf.utility.Preconditions.checkStringNotNullOrEmpty;
+import static org.gearvrf.utility.Preconditions.checkFloatNotNaNOrInfinity;
+
 /** Holds the GVRCameras. */
 public class GVRCameraRig extends GVRComponent {
     /** Ways to use the rotation sensor data. */
@@ -156,6 +159,8 @@ public class GVRCameraRig extends GVRComponent {
      *            The {@code float} value to map.
      */
     public void setFloat(String key, float value) {
+        checkStringNotNullOrEmpty("key", key);
+        checkFloatNotNaNOrInfinity("value", value);
         NativeCameraRig.setFloat(getPtr(), key, value);
     }
 
@@ -180,6 +185,7 @@ public class GVRCameraRig extends GVRComponent {
      *            'Y' component of vector.
      */
     public void setVec2(String key, float x, float y) {
+        checkStringNotNullOrEmpty("key", key);
         NativeCameraRig.setVec2(getPtr(), key, x, y);
     }
 
@@ -206,6 +212,7 @@ public class GVRCameraRig extends GVRComponent {
      *            'Z' component of vector.
      */
     public void setVec3(String key, float x, float y, float z) {
+        checkStringNotNullOrEmpty("key", key);
         NativeCameraRig.setVec3(getPtr(), key, x, y, z);
     }
 
@@ -234,6 +241,7 @@ public class GVRCameraRig extends GVRComponent {
      *            'W' component of vector.
      */
     public void setVec4(String key, float x, float y, float z, float w) {
+        checkStringNotNullOrEmpty("key", key);
         NativeCameraRig.setVec4(getPtr(), key, x, y, z, w);
     }
 
@@ -348,6 +356,7 @@ public class GVRCameraRig extends GVRComponent {
     public float[] getLookAt() {
         return NativeCameraRig.getLookAt(getPtr());
     }
+
 }
 
 class NativeCameraRig {
--- a/GVRf/Framework/src/org/gearvrf/GVRMaterial.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRMaterial.java
@@ -19,6 +19,7 @@ import java.util.concurrent.Future;
 
 import org.gearvrf.utility.Colors;
 import org.gearvrf.utility.Threads;
+import static org.gearvrf.utility.Preconditions.*;
 
 import android.graphics.Color;
 
@@ -267,6 +268,8 @@ public class GVRMaterial extends GVRHybridObject implements
     }
 
     public void setTexture(String key, GVRTexture texture) {
+        checkStringNotNullOrEmpty("key", key);
+        checkNotNull("texture", texture);
         NativeMaterial.setTexture(getPtr(), key, texture.getPtr());
     }
 
@@ -289,6 +292,8 @@ public class GVRMaterial extends GVRHybridObject implements
     }
 
     public void setFloat(String key, float value) {
+        checkStringNotNullOrEmpty("key", key);
+        checkFloatNotNaNOrInfinity("value", value);
         NativeMaterial.setFloat(getPtr(), key, value);
     }
 
@@ -297,6 +302,7 @@ public class GVRMaterial extends GVRHybridObject implements
     }
 
     public void setVec2(String key, float x, float y) {
+        checkStringNotNullOrEmpty("key", key);
         NativeMaterial.setVec2(getPtr(), key, x, y);
     }
 
@@ -305,6 +311,7 @@ public class GVRMaterial extends GVRHybridObject implements
     }
 
     public void setVec3(String key, float x, float y, float z) {
+        checkStringNotNullOrEmpty("key", key);
         NativeMaterial.setVec3(getPtr(), key, x, y, z);
     }
 
@@ -313,6 +320,7 @@ public class GVRMaterial extends GVRHybridObject implements
     }
 
     public void setVec4(String key, float x, float y, float z, float w) {
+        checkStringNotNullOrEmpty("key", key);
         NativeMaterial.setVec4(getPtr(), key, x, y, z, w);
     }
 
@@ -325,9 +333,11 @@ public class GVRMaterial extends GVRHybridObject implements
     public void setMat4(String key, float x1, float y1, float z1, float w1,
             float x2, float y2, float z2, float w2, float x3, float y3,
             float z3, float w3, float x4, float y4, float z4, float w4) {
+        checkStringNotNullOrEmpty("key", key);
         NativeMaterial.setMat4(getPtr(), key, x1, y1, z1, w1, x2, y2, z2, w2,
                 x3, y3, z3, w3, x4, y4, z4, w4);
     }
+
 }
 
 class NativeMaterial {
--- a/GVRf/Framework/src/org/gearvrf/GVRMesh.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRMesh.java
@@ -17,6 +17,9 @@ package org.gearvrf;
 
 import org.gearvrf.utility.Exceptions;
 
+import android.util.Log;
+import static org.gearvrf.utility.Preconditions.*;
+
 /**
  * This is one of the key GVRF classes: It holds GL meshes.
  * 
@@ -66,8 +69,7 @@ public class GVRMesh extends GVRHybridObject {
      *            Array containing the packed vertex data.
      */
     public void setVertices(float[] vertices) {
-        isNonNull(vertices);
-        isValidData(vertices.length, 3);
+        checkValidFloatArray("vertices", vertices, 3);
         NativeMesh.setVertices(getPtr(), vertices);
     }
 
@@ -93,8 +95,7 @@ public class GVRMesh extends GVRHybridObject {
      *            Array containing the packed normal data.
      */
     public void setNormals(float[] normals) {
-        isNonNull(normals);
-        isValidData(normals.length, 3);
+        checkValidFloatArray("normals", normals, 3);
         NativeMesh.setNormals(getPtr(), normals);
     }
 
@@ -120,8 +121,7 @@ public class GVRMesh extends GVRHybridObject {
      *            Array containing the packed texture coordinate data.
      */
     public void setTexCoords(float[] texCoords) {
-        isNonNull(texCoords);
-        isValidData(texCoords.length, 2);
+        checkValidFloatArray("texCoords", texCoords, 2);
         NativeMesh.setTexCoords(getPtr(), texCoords);
     }
 
@@ -153,8 +153,8 @@ public class GVRMesh extends GVRHybridObject {
      *            Array containing the packed triangle index data.
      */
     public void setTriangles(char[] triangles) {
-        isNonNull(triangles);
-        isValidData(triangles.length, 3);
+        checkNotNull("triangles", triangles);
+        checkDivisibleDataLength("triangles", triangles.length, 3);
         NativeMesh.setTriangles(getPtr(), triangles);
     }
 
@@ -180,9 +180,7 @@ public class GVRMesh extends GVRHybridObject {
      *            Data to bind to the shader attribute.
      */
     public void setFloatVector(String key, float[] floatVector) {
-        isNonNull(floatVector);
-        isValidData(floatVector.length, 1);
-        verticesCheck(floatVector.length, 1);
+        checkValidFloatVector("key", key, "floatVector", floatVector, 1);
         NativeMesh.setFloatVector(getPtr(), key, floatVector);
     }
 
@@ -209,9 +207,7 @@ public class GVRMesh extends GVRHybridObject {
      *            attribute.
      */
     public void setVec2Vector(String key, float[] vec2Vector) {
-        isNonNull(vec2Vector);
-        isValidData(vec2Vector.length, 2);
-        verticesCheck(vec2Vector.length, 2);
+        checkValidFloatVector("key", key, "vec2Vector", vec2Vector, 2);
         NativeMesh.setVec2Vector(getPtr(), key, vec2Vector);
     }
 
@@ -238,9 +234,7 @@ public class GVRMesh extends GVRHybridObject {
      *            shader attribute.
      */
     public void setVec3Vector(String key, float[] vec3Vector) {
-        isNonNull(vec3Vector);
-        isValidData(vec3Vector.length, 3);
-        verticesCheck(vec3Vector.length, 3);
+        checkValidFloatVector("key", key, "vec3Vector", vec3Vector, 3);
         NativeMesh.setVec3Vector(getPtr(), key, vec3Vector);
     }
 
@@ -267,9 +261,7 @@ public class GVRMesh extends GVRHybridObject {
      *            attribute.
      */
     public void setVec4Vector(String key, float[] vec4Vector) {
-        isNonNull(vec4Vector);
-        isValidData(vec4Vector.length, 4);
-        verticesCheck(vec4Vector.length, 4);
+        checkValidFloatVector("key", key, "vec4Vector", vec4Vector, 4);
         NativeMesh.setVec4Vector(getPtr(), key, vec4Vector);
     }
 
@@ -291,31 +283,31 @@ public class GVRMesh extends GVRHybridObject {
         return new GVRMesh(getGVRContext(), NativeMesh.getBoundingBox(getPtr()));
     }
 
-    private static void isNonNull(Object data) {
-        if (data == null) {
-            throw new IllegalArgumentException(
-                    "The input array should not be null.");
-        }
+    private void checkValidFloatVector(String keyName, String key,
+            String vectorName, float[] vector, int expectedComponents) {
+        checkStringNotNullOrEmpty(keyName, key);
+        checkNotNull(vectorName, vector);
+        checkDivisibleDataLength(vectorName, vector.length, expectedComponents);
+        checkVectorLengthWithVertices(vectorName, vector.length,
+                expectedComponents);
     }
 
-    private static void isValidData(int dataLength, int expectedComponents) {
-        if (dataLength == 0 || dataLength % expectedComponents != 0) {
-            throw Exceptions
-                    .IllegalArgument(
-                            "The input array should be an array of %d-component elements whose length is non-zero and divisible by %d. But current data length is %d.",
-                            expectedComponents, expectedComponents, dataLength);
-        }
+    private void checkValidFloatArray(String parameterName, float[] data,
+            int expectedComponents) {
+        checkNotNull(parameterName, data);
+        checkDivisibleDataLength(parameterName, data.length, expectedComponents);
     }
 
-    private void verticesCheck(int dataLength, int expectedComponents) {
+    private void checkVectorLengthWithVertices(String parameterName,
+            int dataLength, int expectedComponents) {
         int verticesNumber = getVertices().length / 3;
         int numberOfElements = dataLength / expectedComponents;
         if (dataLength / expectedComponents != verticesNumber) {
             throw Exceptions
                     .IllegalArgument(
-                            "The input array should be an array of %d-component elements and the number of elements should match the number of vertices. The current number of elements is %d, but the current number of vertices is %d.",
-                            expectedComponents, numberOfElements,
-                            verticesNumber);
+                            "The input array %s should be an array of %d-component elements and the number of elements should match the number of vertices. The current number of elements is %d, but the current number of vertices is %d.",
+                            parameterName, expectedComponents,
+                            numberOfElements, verticesNumber);
         }
     }
 }
--- a/GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
+++ b/GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
@@ -1,0 +1,137 @@
+/* Copyright 2015 Samsung Electronics Co., LTD
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gearvrf.utility;
+
+import android.text.TextUtils;
+
+public abstract class Preconditions {
+
+    /**
+     * The goal of this class is to improve the readability and to provide some
+     * common checking method to improve the robustness of GVRF. But in some
+     * cases, this may be a significant performance cost. Try to use and further
+     * extend this class carefully with concern on performance issue.
+     * 
+     * Actually, one purpose of this class may help us avoid from using
+     * GVRJniException which is not a proper way to throw/simulate an exception
+     * with JNI crash.
+     */
+
+    /**
+     * Check that the parameter is not {@code null}.
+     * 
+     * @param parameterName
+     *            The name of the user-supplied parameter that we are validating
+     *            so that the user can easily find the error in their code.
+     * @param object
+     *            the parameter object can be any Java reference type, including
+     *            arrays.
+     * @throws IllegalArgumentException
+     *             if the object is null pointer.
+     */
+    public static void checkNotNull(String parameterName, Object object) {
+        if (object == null) {
+            throw Exceptions.IllegalArgument("Input %s can't be null.",
+                    parameterName);
+        }
+    }
+
+    /**
+     * This function is to ensure that any input string value is not null or
+     * empty
+     * 
+     * @param value
+     *            String value to be checked.
+     * @param parameterName
+     *            The name of the user-supplied parameter that we are validating
+     *            so that the user can easily find the error in their code.
+     * @throws IllegalArgumentException
+     *             If the key is null or empty.
+     */
+    public static void checkStringNotNullOrEmpty(String parameterName,
+            String value) {
+        if (TextUtils.isEmpty(value)) {
+            throw Exceptions.IllegalArgument("Current input string %s is %s.",
+                    parameterName, value == null ? "null" : "empty");
+        }
+    }
+
+    /**
+     * This function is to ensure that the length of input array is legal
+     * according to the expected components' number.
+     * 
+     * @param parameterName
+     *            The name of the user-supplied parameter that we are validating
+     *            so that the user can easily find the error in their code.
+     * @param dataLength
+     *            Length of array data.
+     * @param expectedComponents
+     *            Expected number of components.
+     * @throws IllegalArgumentException
+     *             If the dataLength is zero or not divisible by the expected
+     *             number.
+     */
+    public static void checkDivisibleDataLength(String parameterName,
+            int dataLength, int expectedComponents) {
+        if (dataLength == 0 || dataLength % expectedComponents != 0) {
+            throw Exceptions
+                    .IllegalArgument(
+                            "The input array of %s should be an array of %d-component elements whose length is non-zero and divisible by %d. But current data length is %d.",
+                            parameterName, expectedComponents,
+                            expectedComponents, dataLength);
+        }
+    }
+
+    /**
+     * In common shader cases, NaN makes little sense. Correspondingly, GVRF is
+     * going to use Float.NaN as illegal flag in many cases. Therefore, we need
+     * a function to check if there is any setX that is using NaN as input.
+     * 
+     * @param parameterName
+     *            The name of the user-supplied parameter that we are validating
+     *            so that the user can easily find the error in their code.
+     * @param data
+     *            A series of input float data.
+     * @throws IllegalArgumentException
+     *             if the data includes NaN.
+     */
+    public static void checkFloatNotNaNOrInfinity(String parameterName,
+            float... data) {
+        for (float element : data) {
+            checkFloatNotNaNOrInfinity(parameterName, element);
+        }
+    }
+
+    /**
+     * This is checkNotNaN for only single float.
+     * 
+     * @param parameterName
+     *            The name of the user-supplied parameter that we are validating
+     *            so that the user can easily find the error in their code.
+     * @param data
+     *            A single float data.
+     * @throws IllegalArgumentException
+     *             if the data includes NaN.
+     */
+    public static void checkFloatNotNaNOrInfinity(String parameterName,
+            float data) {
+        if (Float.isNaN(data) || Float.isInfinite(data)) {
+            throw Exceptions.IllegalArgument(
+                    "%s should never be NaN or Infinite.", parameterName);
+        }
+    }
+
+}


JonShemitz:
Thanks, Chao. I'll review this tomorrow morning.

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
```diff
@@ -0,0 +1,103 @@
...
+    /**
+     * In common shader cases, NaN makes little sense. Correspondingly, GVRF is going to use Float.NaN as illegal 
+     * flag in many cases. Therefore, we need a function to check if there is any setX that is using NaN as input.
+     * @param data
+     *         A series of input float data.
+     * @throws IllegalArgumentException if the data includes NaN.
+     */
+    public static void checkNotNaN(float ...data){
+        for(float element : data){
+            if(element == Float.NaN){
```
jason2kim: I am not sure this if(element == Float.NaN) works for any true. Isn't there any method provided from java Float or Double for checking this, like isNon()?

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
```diff
@@ -0,0 +1,103 @@
...
+            throw Exceptions
+                    .IllegalArgument(
+                            "The input array should be an array of %d-component elements whose length is non-zero and divisible by %d. But current data length is %d.",
+                            expectedComponents, expectedComponents, dataLength);
+        }
+    }
+    
+    /**
+     * In common shader cases, NaN makes little sense. Correspondingly, GVRF is going to use Float.NaN as illegal 
+     * flag in many cases. Therefore, we need a function to check if there is any setX that is using NaN as input.
```
jason2kim: Do we really need to check NaN for every setty()?
Are you intending to to check NaN in any getty() as a mean for illegal flag?

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
```diff
@@ -0,0 +1,103 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gearvrf.utility;
+
+public final class Preconditions {
```
jason2kim: Name of the class seems not quite right form me.

GVRf/Framework/src/org/gearvrf/GVRMaterial.java
```diff
@@ -267,6 +269,7 @@ public GVRTexture getTexture(String key) {
     }
 
     public void setTexture(String key, GVRTexture texture) {
+        checkKeyNotNullOrEmpty(key);
         NativeMaterial.setTexture(getPtr(), key, texture.getPtr());
```
jason2kim: It will crash at this point if texture is null

GVRf/Framework/src/org/gearvrf/GVRMesh.java
```diff
@@ -66,8 +67,7 @@ protected final boolean registerWrapper() {
      *            Array containing the packed vertex data.
      */
     public void setVertices(float[] vertices) {
-        isNonNull(vertices);
-        isValidData(vertices.length, 3);
+        validOtherCheck(INPUT_TYPE_VERTICES, vertices, vertices.length, 3);
```
jason2kim: It will crash if parameter vertices is null at this point.

GVRf/Framework/src/org/gearvrf/GVRCameraRig.java
```diff
@@ -156,6 +159,8 @@ public float getFloat(String key) {
      *            The {@code float} value to map.
      */
     public void setFloat(String key, float value) {
+        checkKeyNotNullOrEmpty(key);
```
JonShemitz: This particular `String` is a key - but there might well be cases where you don't want a null-or-empty value. Suggest changing method to

    void checkStringNotNullOrEmpty(String name, String value) throws // ...

GVRf/Framework/src/org/gearvrf/GVRCameraRig.java
```diff
@@ -156,6 +159,8 @@ public float getFloat(String key) {
...
+        checkNotNaN(value);
```
JonShemitz: Is infinity a valid value, here? Suggest

    void checkNormalFloat(String name, float value) throws // ...

GVRf/Framework/src/org/gearvrf/GVRMaterial.java
JonShemitz: Jason's right.

    void checkNonNull(String name, Object value) throws // ...

GVRf/Framework/src/org/gearvrf/GVRMesh.java
JonShemitz: `validOtherCheck` is a sort of meaningless name.

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
chenchao1407: Actually I "steal" the name from Google Guava... :-)

GVRf/Framework/src/org/gearvrf/GVRMesh.java
```diff
@@ -180,9 +177,7 @@ public void setTriangles(char[] triangles) {
      *            Data to bind to the shader attribute.
      */
     public void setFloatVector(String key, float[] floatVector) {
-        isNonNull(floatVector);
-        isValidData(floatVector.length, 1);
-        verticesCheck(floatVector.length, 1);
+        validVectorCheck(key, floatVector, 1);
```
JonShemitz: I think you should be more consistent in your naming: `validVectorCheck` should probably be `checkValidVector` or (probably better) `checkVectorLength`

GVRf/Framework/src/org/gearvrf/GVRMesh.java
```diff
@@ -289,22 +278,18 @@ public void setVec4Vector(String key, float[] vec4Vector) {
      */
     public GVRMesh getBoundingBox() {
         return new GVRMesh(getGVRContext(), NativeMesh.getBoundingBox(getPtr()));
+    }    
+    
+    private void validVectorCheck(String key, float[] vector, int expectedComponents){
```
JonShemitz: Shift+Ctrl+F

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
JonShemitz: 1. `Preconditions` seems OK, if perhaps a bit 'academic'. I'm not sure what, if anything, would be any better: `Expectations`, `CheckParameters`, `ValidateParameters`?
2. Not clear why the class is `final` - I can certainly _imagine_ it having `protected` members at some point in the future.
3. Perhaps the class should be `abstract`, to keep anyone from ever `new` ing an instance ...

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
```diff
@@ -0,0 +1,103 @@
...
+     * The goal of this class is to improve the readability and to provide some common checking method to improve 
+     * the robustness of GVRF. But in some cases, this may be a significant performance cost. Try to use and further
+     * extend this class carefully with concern on performance issue.
+     * 
+     * Actually, one purpose of this class may help us avoid from using GVRJniException which is not a proper way to 
+     * throw/simulate an exception with JNI crash.
+     */
+    
+
+    public static final String INPUT_TYPE_ARRAY = "Input array";
```
JonShemitz: 1. `public` members should have JavaDoc.
2. These strings really don't belong here: if the same code is used in a new way, someone would have to come here and change the set of strings. It would be better if each `checkX()` method took a `String parameterName`, so the exception message could refer to the name in the calling code.

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
```diff
@@ -0,0 +1,103 @@
...
+    public static final String INPUT_TYPE_VECTOR = "Input vector";
+    public static final String INPUT_TYPE_NORMALS = "Input normals";
+    public static final String INPUT_TYPE_TRIANGLE = "Input triangle";
+    public static final String INPUT_TYPE_VERTICES = "Input vertices";
+    public static final String INPUT_TYPE_TEXCOORDS = "Input texture coordinates";
+
+    /**
+     * A simple check to see if the object is non-null or otherwise the function will throw an
```
JonShemitz: Maybe just `Check that the parameter is not {@code null}.`

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
```diff
@@ -0,0 +1,103 @@
...
+     *  IllegalArgumentException.
+     *  
+     * @param object
+     *          the parameter object can be any instance of any class even an array.
```
JonShemitz: Actually, what you're checking is whether there is any instance at all ... I'd suggest `any Java reference type, including arrays`

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
```diff
@@ -0,0 +1,103 @@
...
+     * @param dataMessage
```
JonShemitz: Probably should be `parameterName`

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
```diff
@@ -0,0 +1,103 @@
...
+     *          To point out the type of data, for the use when we try to throw an exception.
+     *          Type may not refer to the class name but a wider concept of class type such as vertices, normals.
+     * @throws IllegalArgumentException if the object is null pointer.
+     */
+    public static void checkNotNull(Object object, String dataMessage) {
```
JonShemitz: I'd swap these parameters.

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
```diff
@@ -0,0 +1,103 @@
...
+        if (object == null) {
+            throw Exceptions.IllegalArgument("%s can't be a null pointer.",
```
JonShemitz: Java doesn't have pointers. `"% may not be null"`

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
```diff
@@ -0,0 +1,103 @@
...
+                    dataMessage);
+        }
+    }
+    
+    /**
+     * We can see that there is a wide use of key-value pair for the parameters which are used in shader.
```
JonShemitz: Please remember that the first sentence - the text before the first unquoted '.' character - will be used as the method summary in the Methods table at the top of the JavaDoc page. This doesn't work, there. `Validate {@code String} used as a key name.` might be better ... except that, as above, I think what you really want here is a more generic 'String parameter is not null or empty' test.

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
```diff
@@ -0,0 +1,103 @@
...
+     * This function is to ensure that the key is meaningful.
+     * @param key
+     *          Key in the key-value pair.
+     *          
+     * @throws IllegalArgumentException if the key is null or empty.
+     */
+    public static void checkKeyNotNullOrEmpty(String key){
```
JonShemitz: Shift+Ctrl+F

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
```diff
@@ -0,0 +1,103 @@
...
+        if(key == null || key.trim().isEmpty()){
```
JonShemitz: 1. I suggest you use `TextUtils.isEmpty()`
2. `key.trim().isEmpty` is expensive - `trim()` is (potentially) returning a copy of the `String`. I'd suggest you introduce a private method that returns `true` if the `String` contains any non-whitespace (or, perhaps, if the first and last characters of a `String` are not whitespace).

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
```diff
@@ -0,0 +1,103 @@
...
+            throw Exceptions.IllegalArgument("Current key is %s which is not a proper key.", key == null ? "null" : "empty");
+        }
+    }
+
+    /**
+     * In GVRF, there are a lot of arrays uses with expected number of components inside.
```
JonShemitz: `Check the array length.`

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+    
+    /**
+     * In common shader cases, NaN makes little sense. Correspondingly, GVRF is going to use Float.NaN as illegal 
+     * flag in many cases. Therefore, we need a function to check if there is any setX that is using NaN as input.
+     * @param data
+     *         A series of input float data.
+     * @throws IllegalArgumentException if the data includes NaN.
+     */
+    public static void checkNotNaN(float ...data){
```
JonShemitz: A `float...` parameter with only one parameter generates code like

    float[] __temporary__ = new float[1];
    __temporary__[0] = parameter;

... and is thus comparatively expensive. (`calloc()` costs, plus making the next garbage collection happen that much sooner). It's fine to offer this method ... but please also offer an overload that takes a single float parameter.

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
JonShemitz: It doesn't - I seem to recall that there are actually more than one `NaN` value. `Float.isNaN()` is what you want.

JonShemitz:
-1 ...

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
chenchao1407: Both of you are right . I will fix this issue by using Float.isNaN

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
chenchao1407: The reason of this is for your fix on getFloat(). Currently your pull request is going to use the GVRJniException. I talked to Tom that he also thinks Float.NaN should be an illegal parameter for shader at any case. So that if we use Float.NaN as an illegal identifier in getFloat(). Then we can throw a Java Exception when getFloat() is doing something that is not correct.

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
chenchao1407: Actually in Google Guava, the Preconditions class makes its constructor private which serves same purpose as your point 3.

I am thinking that if we should keep this class simple and not used that frequently so that I made it final. I think abstract is also an option. I will update the patch based on your comments.

GVRf/Framework/src/org/gearvrf/GVRMaterial.java
chenchao1407: Yes. I think I made a mistake. I thought the crash and error will occurs only when getTexture() of null texture is called. I will update it.

GVRf/Framework/src/org/gearvrf/GVRMesh.java
chenchao1407: Actually I am also quite sure about this part. Because there are general two types of check operations. One is for all kinds of vector, the other one is for the data other than vector.

For INPUT_TYPE_VERTICES, imho, I am thinking that there are two types of design to manage those constant String, one is let Preconditions class hold all those Strings. The pros is that it is easy to manage and extend, the cons is that actually as you say, it is going to put something that doesn't belong to Preconditions and it may disobey the concept of OOD. 

The second way is to let the class itself to hold those constant String. The pros is that we are going to put something to where it should be. But the cons, in my opinion, is that there may cause some duplicates. For example, INPUT_TYPE_VECTOR actually applies to many classes not only limited to GVRMesh, if we put it at every class it may be used, it would be a little redundant.

Jon, How do you think?

GVRf/Framework/src/org/gearvrf/GVRCameraRig.java
chenchao1407: To GL, infinity is not a valid value. But the reason why I just check NaN is because NaN can be used as an illegal state identifier when we try to throw Java exception. I mean I am trying to balance the overhead of this class and how many conditions we should check with this class. Maybe we can discuss about it tomorrow.

GVRf/Framework/src/org/gearvrf/GVRMesh.java
chenchao1407: Ok. I will update it.

GVRf/Framework/src/org/gearvrf/GVRMesh.java
chenchao1407: :-( .....

chenchao1407:
And also, I am considering if this class should be public. Can you guys help think of if there is any use case that this class can benefit developers?

Thanks in advance.

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
JonShemitz: > the Preconditions class makes its constructor private 

Yes, explicitly coding a private constructor is the Java Way to do it. Declaring a class `abstract` sounds like it is meant to be inherited from.

But when I think of all the meanings of `final`, overloading `abstract` sure seems very Java like. And, it's *much* less  code.

>  simple and not used that frequently so that I made it final. 

`final` means you can't inherit from it, not that you can never change the code: I have gotten into more trouble by making things `final` than by not. We (or, maybe, Brazil) will probably grow this class slowly, as we need new tests. Who knows, maybe at some point there will be some sort of instance state and/or methods that people _will_ want to be polymorphic. At that point, we'd have to remove the `final` that we  really shouldn't be adding now.

GVRf/Framework/src/org/gearvrf/GVRMesh.java
JonShemitz: > Actually I am also quite sure about this part. Because there are general two types of check operations. One is for all kinds of vector, the other one is for the data other than vector.

Sorry, you lost me. I'm not sure what you're referring to ...


>  Preconditions class hold all those Strings ... is easy to manage and extend

Actually, it's not easy to manage or extend. If everyone puts their strings in a shared class, you can end up with merge conflicts and/or naming chaos. Every change invalidates (re-compiles) every module that uses it. Keeping the tag values in the client means that a new tag can be local to the client module, and adding a new tag only invalidates one module.

> let the class itself to hold those constant String ... may cause some duplicates. 

Java does a fine job of making each and every reference to a constant `"string"` refer to the same run-time `String` instance. The names would be local.

> Jon, How do you think?

I've told you twice, now ...

GVRf/Framework/src/org/gearvrf/GVRCameraRig.java
JonShemitz: We might want to use infinities to hold error codes - `{NaN, +INF, -INF}` is three times as many error codes as `{NaN}`.

>  I am trying to balance the overhead of this class and how many conditions we should check

Fair enough. But once we're calling static methods in a different package, I don't know that `Float.isNan(value) || Float.isInfininite(value)` is significantly more expensive than `Float.isNan(value)`.

JonShemitz:
> I am considering if this class should be public

If it's not, you won't be able to call it from outside of the `utility` package.

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
jason2kim: I agree that NaN is illegal for the shader. But, using it as return value checking when calling getFloat() is not correct. They should not coexist.
Error codes can be non-existent, invalid data value (including NaN), or any other errors when returning from the callee.

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
jason2kim: Even if you create a private method, you still need to scan the string for whitespaces, something similar to trim().
Considering code readability, using String.isEmpty() can be still a good way.

JonShemitz: > Considering code readability, using String.isEmpty() can be still a good way.

Disagree. We don't want to allocate a new `String` if we can help it. A single, well chosen method name is more readable than `trim().isEmpty()` and cheaper, too.

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
jason2kim: Precondition usually used for checking some condition before allowing to go into its main action. If the condition is not met, the whole log of task will abort. However, our new class introduced here seems just a collection of utility methods.
I got pointer from Chao about this Guava’s Preconditions, and taking a look, it does some state checking, I guess this is its major role, as well as general argument checking methods. Anyway, it is Chao’s decision for the class name, but I think the big name can invite unnecessary confusions.

chenchao1407:
Updated.
Thanks.

chenchao1407:
For the class name, I don't have a better choice. But I am open to change the class name if any one comes up with a better idea.

GVRf/Framework/src/org/gearvrf/GVRMesh.java
```diff
@@ -66,8 +74,8 @@ protected final boolean registerWrapper() {
      *            Array containing the packed vertex data.
      */
     public void setVertices(float[] vertices) {
-        isNonNull(vertices);
-        isValidData(vertices.length, 3);
+        checkValidDataNotVector(INPUT_TYPE_VERTICES, vertices, vertices.length,
```
JonShemitz: Do `vertices.length` inside the method, after you know its not `null`.

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
```diff
@@ -0,0 +1,108 @@
...
+     * This function is to ensure that any input string value is not null or empty
+     * @param value
+     *          String value to be checked.
+     * @param parameterName
+     *          The name/type of the string value.
+     *          
+     * @throws IllegalArgumentException if the key is null or empty.
+     */
+    public static void checkStringNotNullOrEmpty(String parameterName, String value){
+        if(value == null || TextUtils.isEmpty(value)){
```
JonShemitz: Shift+Ctrl+F

GVRf/Framework/src/org/gearvrf/GVRMesh.java
chenchao1407: Yes. This will cause a NullPointer problem. The reason I use this kind of implementation is because checkValidDataNotVector is because of the input parameters, there are both char[] and float[] so that I can't tell which is passed in since they are primitive type array. I will figure out another way to handle this.

Thank you.

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
chenchao1407: This has been auto-formatted, any issue here?

chenchao1407:
Updated.

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
JonShemitz: Are you sure? Auto-format should have put a space between the ) and the {

chenchao1407: I see. Just updated the format. Don't know why but "Shift+Ctrl+F" doesn't work this time. I can only format this file by doing "Source -> Format"

chenchao1407:
Any objection on this or  I will merge it so that Jason can continue working on his pull request based on this.

hTanks

chenchao1407:
Updated and no more uses constant String as type identifier.

GVRf/Framework/src/org/gearvrf/GVRCameraRig.java
```diff
@@ -156,6 +159,8 @@ public float getFloat(String key) {
      *            The {@code float} value to map.
      */
     public void setFloat(String key, float value) {
+        checkStringNotNullOrEmpty("key", key);
+        checkFloatNotNaNOrInfinity(value);
```
JonShemitz: We need a parameter name here, too.

GVRf/Framework/src/org/gearvrf/GVRMaterial.java
```diff
@@ -267,6 +268,8 @@ public GVRTexture getTexture(String key) {
     }
 
     public void setTexture(String key, GVRTexture texture) {
+        checkStringNotNullOrEmpty("key", key);
+        checkNotNull(GVRTexture.class.getSimpleName(), texture);
```
JonShemitz: Parameter name, not parameter type. That is, `"texture"` not `GVRTexture.class.getSimpleName()`.

GVRf/Framework/src/org/gearvrf/GVRMesh.java
```diff
@@ -120,8 +120,7 @@ public void setNormals(float[] normals) {
      *            Array containing the packed texture coordinate data.
      */
     public void setTexCoords(float[] texCoords) {
-        isNonNull(texCoords);
-        isValidData(texCoords.length, 2);
+        checkValidFloatArray("texture coordinates", texCoords, 2);
```
JonShemitz: `""texture coordinates"` s/b "`texCoords`"

GVRf/Framework/src/org/gearvrf/GVRMesh.java
```diff
@@ -153,8 +152,8 @@ public void setTexCoords(float[] texCoords) {
      *            Array containing the packed triangle index data.
      */
     public void setTriangles(char[] triangles) {
-        isNonNull(triangles);
-        isValidData(triangles.length, 3);
+        checkNotNull("triangle", triangles);
+        checkDivisibleDataLength(triangles.length, 3);
```
JonShemitz: Need a parameter name.

GVRf/Framework/src/org/gearvrf/GVRMesh.java
```diff
@@ -180,9 +179,7 @@ public void setTriangles(char[] triangles) {
      *            Data to bind to the shader attribute.
      */
     public void setFloatVector(String key, float[] floatVector) {
-        isNonNull(floatVector);
-        isValidData(floatVector.length, 1);
-        verticesCheck(floatVector.length, 1);
+        checkValidFloatVector(key, floatVector, 1);
```
JonShemitz: Parameter names, please.

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
```diff
@@ -0,0 +1,126 @@
...
+     * cases, this may be a significant performance cost. Try to use and further
+     * extend this class carefully with concern on performance issue.
+     * 
+     * Actually, one purpose of this class may help us avoid from using
+     * GVRJniException which is not a proper way to throw/simulate an exception
+     * with JNI crash.
+     */
+
+    /**
+     * Check that the if the parameter is {@code null}.
```
JonShemitz: `Check that the parameter is not {@code null}.`

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
```diff
@@ -0,0 +1,126 @@
...
+     * 
+     * @param object
+     *            the parameter object can be any Java reference type, including
+     *            arrays.
+     * @param parameterName
+     *            To point out the type of data, for the use when we try to
```
JonShemitz: 'The name of the user-supplied parameter that we are validating, so that the user can easily find the error in their code.`

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    /**
+     * This function is to ensure that any input string value is not null or
+     * empty
+     * 
+     * @param value
+     *            String value to be checked.
+     * @param parameterName
+     *            The name/type of the string value.
```
JonShemitz: `The name of the user-supplied parameter that we are validating, so that the user can easily find the error in their code.`

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
```diff
@@ -0,0 +1,126 @@
...
+
+    /**
+     * This is checkNotNaN for only single float.
+     * 
+     * @param data
+     *            A single float data.
+     * @throws IllegalArgumentException
+     *             if the data includes NaN.
+     */
+    public static void checkFloatNotNaNOrInfinity(float data) {
```
JonShemitz: Good, thanks.

JonShemitz:
Thanks - better, but still -1.

chenchao1407:
Fixed

GVRf/Framework/src/org/gearvrf/GVRCameraRig.java
```diff
@@ -156,6 +159,8 @@ public float getFloat(String key) {
      *            The {@code float} value to map.
      */
     public void setFloat(String key, float value) {
+        checkStringNotNullOrEmpty("key", key);
+        checkFloatNotNaNOrInfinity("Float value for shader", value);
```
JonShemitz: `"Float value for shader"` s/b just `"value"`: A literal parameter name, not any sort of description.

GVRf/Framework/src/org/gearvrf/GVRMaterial.java
```diff
@@ -289,6 +292,8 @@ public float getFloat(String key) {
     }
 
     public void setFloat(String key, float value) {
+        checkStringNotNullOrEmpty("key", key);
+        checkFloatNotNaNOrInfinity("Float value for shader", value);
```
JonShemitz: `"value"`

GVRf/Framework/src/org/gearvrf/GVRMesh.java
```diff
@@ -153,8 +153,8 @@ public void setTexCoords(float[] texCoords) {
      *            Array containing the packed triangle index data.
      */
     public void setTriangles(char[] triangles) {
-        isNonNull(triangles);
-        isValidData(triangles.length, 3);
+        checkNotNull("triangle", triangles);
```
JonShemitz: This is why it would be nice if Java had a macro to turn an identifier into a string: `"triangle"` s/b `"triangles"`.

GVRf/Framework/src/org/gearvrf/GVRMesh.java
```diff
@@ -291,31 +283,32 @@ public GVRMesh getBoundingBox() {
         return new GVRMesh(getGVRContext(), NativeMesh.getBoundingBox(getPtr()));
     }
 
-    private static void isNonNull(Object data) {
-        if (data == null) {
-            throw new IllegalArgumentException(
-                    "The input array should not be null.");
-        }
+    private void checkValidFloatVector(String parameterName, String key,
```
JonShemitz: Sorry - you are testing two different parameters here: you should have two different parameter names.

    void checkValidFloatVector(String keyName, String key, String vectorName, float[] vector, int expectedComponents)

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
```diff
@@ -0,0 +1,137 @@
...
+     *            String value to be checked.
+     * @param parameterName
+     *            The name of the user-supplied parameter that we are validating
+     *            so that the user can easily find the error in their code.
+     * @throws IllegalArgumentException
+     *             If the key is null or empty.
+     */
+    public static void checkStringNotNullOrEmpty(String parameterName,
+            String value) {
+        if (value == null || TextUtils.isEmpty(value)) {
```
JonShemitz: `TextUtils.isEmpty()` "Returns true if the string is null or 0-length.".

That is, you don't need the `value == null` test.

chenchao1407:
Updated

JonShemitz:
This looks fine, to me. We just need to validate it. @dknaan - could you perhaps do that? I'm in the middle of What's Wrong With This JNI Code ...

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
```diff
@@ -0,0 +1,137 @@
...
+     *            String value to be checked.
+     * @param parameterName
+     *            The name of the user-supplied parameter that we are validating
+     *            so that the user can easily find the error in their code.
+     * @throws IllegalArgumentException
+     *             If the key is null or empty.
+     */
+    public static void checkStringNotNullOrEmpty(String parameterName,
+            String value) {
+        if (TextUtils.isEmpty(value)) {
```
jason2kim: how about whitespaces, including "   " and "\t"?

JonShemitz: > how about whitespaces

What problem do you see with allowing whitespace? I don't think the `GVRShaderMaps` care much - they're basically just a collection of `Map<String, String>`.

jason2kim: This method checkStringNotNullOrEmpty() implies checking null and empty string. In general, white spaces are not regarded as data but empty. If we are restricting use of this method for some special cases only like GVRShaderMaps map, we can rename the method or document it at java doc.

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
JonShemitz: >  In general, white spaces are not regarded as data but empty.

Perhaps this is true in some wider context. In the Java API world, though, `String.isEmpty()` "Returns true if the length of this string is 0" while `TextUtils.isEmpty()` "Returns true if the string is null or 0-length."

dknaan:
I checked and it works fine, but I’m not sure if any of the features is used by the samples.

GVRf/Framework/src/org/gearvrf/utility/Preconditions.java
chenchao1407: It looks to me that trim().isEmpty() is pretty much more expensive than TextUtils.isEmpty(). However, if a string only contains spaces may be a little bit meaningless as a key. But right now the function is not only checking keys but also for more general purpose. I think it should be fine here.

Issue #19, opened 2015-04-17
GVRAccelerateDecelerateInterpolator::mapRatio(float) - Error message


nathan-almeida:
The message when passing an invalid mapRatio to **GVRAccelerateDecelerateInterpolator::mapRatio()** could be more user friendly.

```java
public void testMapRatioGreaterThanOne() {
        try {
            GVRAccelerateDecelerateInterpolator interpolator = GVRAccelerateDecelerateInterpolator.getInstance();
            interpolator.mapRatio(2f);
            fail();
        } catch (InvalidParameterException e) {
            assertNotNull(e.getMessage(), "ratio - the current time ratio, >= 0 and <= 1");
        }
    }
```

JonShemitz:
I'm not sure there's any point to testing this parameter: This is _not_ user-facing code; it is only called by the animation engine, and the inputs are always in the range [0, 1].

PR #34, opened 2015-04-29
Add 3D screenshot API.

rongguodong:
Also added a new GVRF sample to test the result of 3D screenshot (by using 6 2D textures).

gvr-sample shows how to use the new API to capture 3D screenshot and save it as six images.
gvr-test-screenshot3Dresult tests the screenshot result by mapping the six images to a cube.


GVRf/Framework/src/org/gearvrf/GVRContext.java
```diff
@@ -1382,4 +1382,6 @@ protected synchronized static void resetOnRestart() {
      *            none.
      */
     public abstract void captureScreenRight(GVRScreenshotCallback callback);
+
+    public abstract void captureScreen3D(GVRScreenshot3DCallback callback);
```
JonShemitz: Needs JavaDoc.

GVRf/Framework/src/org/gearvrf/GVRScreenshot3DCallback.java
```diff
@@ -0,0 +1,35 @@
...
+
+package org.gearvrf;
+
+import java.util.List;
+
+import android.graphics.Bitmap;
+
+/**
+ * Callback interface for asynchronous 3D screenshot capture. The screenshot
+ * results will be passed to the caller as the parameter of onScreenCaptures.
```
JonShemitz: `onScreenCaptures` s/b `{@link GVRScreenshot3DCallback#onScreenCaptured(List)}`

GVRf/Framework/src/org/gearvrf/GVRScreenshot3DCallback.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+public interface GVRScreenshot3DCallback {
+    /**
+     * Callback method dealing with the returned screenshots
+     * 
+     * @param bitmapList
+     *            List containing six screenshot results (one for each face of a
```
JonShemitz: Is there any order to the faces?

GVRf/Framework/src/org/gearvrf/GVRScreenshot3DCallback.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.util.List;
+
+import android.graphics.Bitmap;
+
+/**
+ * Callback interface for asynchronous 3D screenshot capture. The screenshot
+ * results will be passed to the caller as the parameter of onScreenCaptures.
+ */
+public interface GVRScreenshot3DCallback {
```
JonShemitz: Maybe - maybe! - this should be `GVR3DScreenshotCallback`. It's a 3D screenshot, right?

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -32,6 +32,7 @@
 import org.gearvrf.utility.Log;
 import org.gearvrf.utility.Threads;
 
+import android.R.integer;
```
JonShemitz: This does not _seem_ to be used.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -320,6 +327,98 @@ public void run() {
...
+        centerCamera.setRenderMask(GVRRenderMaskBit.Left
+                | GVRRenderMaskBit.Right);
+        GVRSceneObject centerCameraObject = new GVRSceneObject(this);
+        centerCameraObject.attachCamera(centerCamera);
+        mainCameraRig.getOwnerObject().addChildObject(centerCameraObject);
+        GVRTransform centerCameraTransform = centerCameraObject.getTransform();
+
+        // render +x face (the one user is currently looking at)
+        centerCameraTransform.rotateByAxis(-90, 0, 1, 0);
+        renderCamera(mActivity.appPtr, mMainScene, centerCamera,
```
JonShemitz: There seems to be a lot of repeated code here, six instances of

    // Rotate the camera
    // Same three statements

Probably the three duplicate statements should be moved into a private method.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -320,6 +327,98 @@ public void run() {
...
+
+    void returnScreenshot3DToCaller(final GVRScreenshot3DCallback callback,
+            final List<byte[]> byteArrayList, final int width, final int height) {
+        if (byteArrayList.size() != 6) {
+            throw new IllegalArgumentException("byteArrayList length is not 6.");
+        } else {
+            // run the callback function in a background thread
+            Threads.spawn(new Runnable() {
+                public void run() {
+                    List<Bitmap> bitmapList = new ArrayList<Bitmap>();
```
JonShemitz: We know `bitmapList` will have 6 and only 6 elements: `new ArrayList<Bitmap>()` s/b `new ArrayList<Bitmap>(6)`

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -330,8 +429,10 @@ void onDrawEyeView(int eye, float fovDegrees) {
             int width = 0, height = 0;
             if (mScreenshotCenterCallback != null
                     || mScreenshotLeftCallback != null
-                    || mScreenshotRightCallback != null) {
-                GVRRenderTexture rendernTexture = mRenderBundle.getRightRenderTexture();
+                    || mScreenshotRightCallback != null
+                    || mScreenshot3DCallback != null) {
+                GVRRenderTexture rendernTexture = mRenderBundle
```
JonShemitz: `rendernTexture` s/b `renderTexture`

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -388,6 +489,17 @@ void onDrawEyeView(int eye, float fovDegrees) {
                     mScreenshotCenterCallback = null;
                 }
 
+                // if mScreenshot3DCallback is not null, capture 3D screenshot
+                if (mScreenshot3DCallback != null) {
+                    List<byte[]> byteArrayList = new ArrayList<byte[]>();
```
JonShemitz: `new ArrayList<byte[]>(6)`

GVRf/Sample/gvr-sample/AndroidManifest.xml
```diff
@@ -33,7 +33,7 @@
         android:icon="@drawable/icon"
         android:label="@string/app_name"
         android:theme="@style/AppTheme" >
-        <meta-data android:name="com.samsung.android.vr.application.mode" android:value="vr_only"/>
+        <!-- meta-data android:name="com.samsung.android.vr.application.mode" android:value="vr_only"/-->
```
JonShemitz: This is for debugging only - should not have been checked in.

GVRf/Sample/gvr-sample/src/org/gearvrf/sample/SampleCubeScript.java
```diff
@@ -224,4 +231,41 @@ public void onScreenCaptured(Bitmap bitmap) {
...
         };
     }
+
+    private GVRScreenshot3DCallback newScreenshot3DCallback(
+            final String filename) {
+        return new GVRScreenshot3DCallback() {
+
+            @Override
+            public void onScreenCaptured(List<Bitmap> bitmapList) {
+                Log.d("SampleActivity", "Length of bitmapList TEST");
```
JonShemitz: ?

GVRf/Sample/gvr-test-screenshot3D/src/org/gearvrf/testscreenshot3D/TestScreenshot3DScript.java
```diff
@@ -0,0 +1,108 @@
...
+public class TestScreenshot3DScript extends GVRScript {
+
+    private static final float CUBE_WIDTH = 20.0f;
+    private GVRContext mGVRContext = null;
+
+    @Override
+    public void onInit(GVRContext gvrContext) {
+        mGVRContext = gvrContext;
+
+        GVRTexture front = mGVRContext.loadTexture(new GVRAndroidResource(
```
JonShemitz: I don't think that we should add any new samples that do synchronous loading - it would probably be better to use `loadFutureTexture()`

GVRf/Sample/gvr-test-screenshot3D/src/org/gearvrf/testscreenshot3D/TestScreenshot3DScript.java
```diff
@@ -0,0 +1,108 @@
...
+        GVRTexture right = mGVRContext.loadTexture(new GVRAndroidResource(
+                mGVRContext, R.drawable.screenshot3d_0));
+        GVRTexture top = mGVRContext.loadTexture(new GVRAndroidResource(
+                mGVRContext, R.drawable.screenshot3d_2));
+        GVRTexture bottom = mGVRContext.loadTexture(new GVRAndroidResource(
+                mGVRContext, R.drawable.screenshot3d_3));
+
+        GVRScene scene = mGVRContext.getNextMainScene();
+
+        GVRSceneObject mFrontBoardObject = new GVRSceneObject(gvrContext, CUBE_WIDTH,
```
JonShemitz: What does "Board" mean, here?

JonShemitz:
I can't seem to add 'file notes' - I don't think you should have included `oculussig_8be8658d`.

GVRf/Framework/src/org/gearvrf/GVRScreenshot3DCallback.java
rongguodong: Since I named the capture function "captureScreen3D" (which means "capture screenshot in 3D mode"), I also put "3D" later here. Also, in OpenGL, a 3D texture is called "GL_TEXTURE_3D". :)

If you do think putting 3D ahead is better, I can definitely change it accordingly.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
chenchao1407: I am thinking if we know there will be 6 and only 6 elements in an array. Why don't we just use Bitmap[] bitmapArray = new Bitmap[6];

This will cut off a little bit overhead which are from ArrayList compared to simple array.

GVRf/Sample/gvr-test-screenshot3D/src/org/gearvrf/testscreenshot3D/TestScreenshot3DScript.java
rongguodong: The cube is composed of six boards. These names are inherited from the original GVRSample.

GVRf/Framework/src/org/gearvrf/GVRScreenshot3DCallback.java
JonShemitz: > If you do think putting 3D ahead is better, I can definitely change it accordingly.

No, it's fine as is.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
JonShemitz: That would be even more efficient, though the `List` gives you enough extra functionality that it may be worth it. Both this and my original comment are pretty small issues, though.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -320,6 +327,98 @@ public void run() {
...
+    void returnScreenshot3DToCaller(final GVRScreenshot3DCallback callback,
+            final List<byte[]> byteArrayList, final int width, final int height) {
+        if (byteArrayList.size() != 6) {
+            throw new IllegalArgumentException("byteArrayList length is not 6.");
+        } else {
+            // run the callback function in a background thread
+            Threads.spawn(new Runnable() {
+                public void run() {
+                    List<Bitmap> bitmapList = new ArrayList<Bitmap>();
+                    for (byte[] bytearray : byteArrayList) {
```
JonShemitz: You know, some of our devices do have 8 cores, these days. Doing a `Threads.spawnLow()` for each face and waiting for all 6 to finish is going to be up to 6x faster than looping. I don't think call-to-callback time is incredibly important to the app - but there _is_ benefit to releasing these large byte arrays quickly. Minimizes heap fragmentation and all that.

GVRf/Sample/gvr-test-screenshot3D/src/org/gearvrf/testscreenshot3D/TestScreenshot3DScript.java
JonShemitz: I think "quad" is probably a more common term than "board" - I don't think we should be continuing the use of "board".

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
```diff
@@ -143,29 +143,76 @@ public GVRSceneObject(GVRContext gvrContext, GVRMesh mesh,
...
      * @param texture
      *            Basically, a stream containing a texture file. This can be
      *            either a compressed texture or a regular Android bitmap file.
-     *            
-     * @since 1.6.7            
+     * 
+     * @since 1.6.7
      */
-    public GVRSceneObject(GVRContext context, GVRAndroidResource mesh,
+    public GVRSceneObject(GVRContext gvrContext, GVRAndroidResource mesh,
```
rongguodong: renamed context to gvrContext to be consistent with other constructors.

rongguodong:
I have made changes according to Jon and Chao's comments. Please review it again. Thanks!

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -320,6 +326,127 @@ public void run() {
...
+                public void run() {
+                    final Bitmap[] bitmapArray = new Bitmap[6];
+                    Runnable[] threads = new Runnable[6];
+
+                    for (int i = 0; i < 6; i++) {
+                        final int index = i;
+                        threads[i] = new Runnable() {
+                            public void run() {
+                                byte[] bytearray = byteArrayArray[index];
+                                bitmapArray[index] = generateBitmap(bytearray,
```
JonShemitz: Part of the point of parallelizing is to reduce memory pressure as soon as possible - it would probably be a good idea to insert

    byteArrayArray[index] = null;

between lines 414 and 415.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -320,6 +326,127 @@ public void run() {
...
+                                        width, height);
+                                synchronized (this) {
+                                    notifyAll();
```
JonShemitz: This really should be `notify()` - there's only one thread waiting ...

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -320,6 +326,127 @@ public void run() {
...
+                        };
+                    }
+
+                    for (Runnable thread : threads) {
+                        Threads.spawnLow(thread);
+                    }
+
+                    for (int i = 0; i < 6; i++) {
+                        synchronized (threads[i]) {
+                            while (bitmapArray[i] == null) {
```
JonShemitz: This "busy wait" defeats the whole point of using notify and wait.

rongguodong:
Changed notifyAll to notify, and added a condition to prevent the wait is called too early before notify.

GVRf/Framework/src/org/gearvrf/GVRContext.java
```diff
@@ -995,7 +995,7 @@ public void loadTexture(TextureCallback callback,
      * {@link #loadTexture(GVRAndroidResource.TextureCallback, GVRAndroidResource, int)
      * : you don't have to implement a callback; you don't have to pay attention
      * to the low-level details of
-     * {@linkplain GVRSceneObject#attachRenderData(GVRRenderData) attaching} a
+     *{@linkplain GVRSceneObject#attachRenderData(GVRRenderData) attaching} a
```
JonShemitz: This change looks strange - did you Shift+Ctrl+F?

GVRf/Framework/src/org/gearvrf/GVRContext.java
```diff
@@ -1382,4 +1382,22 @@ protected synchronized static void resetOnRestart() {
...
+     * The screenshot capture is done asynchronously -- the function does not
+     * return the result immediately. Instead, it registers a callback function
+     * and pass the result (when it is available) to the callback function. The
+     * callback will happen on a background thread: It will probably not be the
+     * same thread that calls this method, and it will not be either the GUI or
+     * the GL thread.
+     * 
+     * @param callback
+     *            Callback function to process the capture result. It may not be
+     *            none.
```
JonShemitz: `none` s/b `{@code null}`

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
```diff
@@ -143,29 +143,76 @@ public GVRSceneObject(GVRContext gvrContext, GVRMesh mesh,
...
+     *            the scene object's width
+     * @param height
+     *            the scene object's height
+     * @param texture
+     *            Basically, a stream containing a texture file. This can be
+     *            either a compressed texture or a regular Android bitmap file.
+     * 
+     * @since 1.6.7
+     */
+    public GVRSceneObject(GVRContext gvrContext, float width, float height,
```
JonShemitz: I don't think that this is really the right approach: we have a lot of code that's _almost_ the same as the constructor above, and we're adding a `Future<>` version of just one of the many constructor permutations.

I think it would probably be better to refactor the above `(GVRContext, GVRAndroidResource, GVRAndroidResource)` constructor:

    // almost all the existing logic
    public GVRSceneObject(GVRContext, Future<GVRMesh>, Future<GVRTexture>);
    // calls above
    public GVRSceneObject(GVRContext, GVRAndroidResource, GVRAndroidResource);

... then we can add some trivial methods to GVRContext:

    Future<GVRMesh> futureMesh(GVRMesh);
    Future<GVRTexture> futureMesh(GVRTexture);

(I can write those two, if you like.)

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -320,6 +326,128 @@ public void run() {
...
+                    .removeChildObject(centerCameraObject);
+        }
+    }
+
+    Object testObject = new Object();
+
+    void returnScreenshot3DToCaller(final GVRScreenshot3DCallback callback,
+            final byte[][] byteArrayArray, final int width, final int height) {
+
+        if (byteArrayArray.length != 6) {
```
JonShemitz: I don't think this parameter check is 100% necessary, in an internal method, but it's not very expensive.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -320,6 +326,128 @@ public void run() {
...
+                    final Bitmap[] bitmapArray = new Bitmap[6];
+                    Runnable[] threads = new Runnable[6];
+
+                    for (int i = 0; i < 6; i++) {
+                        final int index = i;
+                        threads[i] = new Runnable() {
+                            public void run() {
+                                byte[] bytearray = byteArrayArray[index];
+                                byteArrayArray[index] = null;
+                                bitmapArray[index] = generateBitmap(bytearray,
```
JonShemitz: Almost!

Imagine line 431 sees `bitmapArray[i] == null` *just* before line 416 sets it - we might still do the `wait()` after the `notify()`. This stuff can be really tricky, but I'm 99% sure that setting a local on line 416 and setting bitmapArray[index] *within* the synchronized block will avoid any race conditions.

    Bitmap face = generateBitmap(bytearray, width, height);
    synchronized (this) {
      bitmapArray[index] - face;
      notify();
    }

JonShemitz:
Thanks, looking really close.

GVRf/Framework/src/org/gearvrf/GVRContext.java
rongguodong: This is generated by Ctrl+Shift+F. Even if I manually add the blank back, it will be removed by Ctrl+Shift+F. :(

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
rongguodong: I just pushed the modified code with FutureWrapper. Please review it again.

GVRf/Framework/src/org/gearvrf/FutureWrapper.java
```diff
@@ -0,0 +1,50 @@
...
+    @Override
+    public boolean cancel(boolean mayInterruptIfRunning) {
+        return false;
+    }
+
+    @Override
+    public boolean isCancelled() {
+        return false;
+    }
+
```
JonShemitz: This code is fine - for the current use cases. The only thing that gives me pause is that _some_ app will try to use it in a different context, and it will fail because `isCancelled()` and `isDone()` return the wrong values.

That is, if this checkin validates, I'll go ahead and merge it ... but I'd sure _prefer_ to see a Correct implementation of `Future` ...

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
```diff
@@ -143,29 +143,27 @@ public GVRSceneObject(GVRContext gvrContext, GVRMesh mesh,
...
-     *            Basically, a stream containing a texture file. This can be
-     *            either a compressed texture or a regular Android bitmap file.
-     *            
-     * @since 1.6.7            
+     * @param futureMesh
+     *            mesh of the object.
+     * @param futureTexture
+     *            texture of the object.
+     * 
+     * @since 1.6.7
```
JonShemitz: But this is not true ... I think you should probably add a `GVRVersion.V_1_6_8` ...

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
```diff
@@ -174,6 +172,48 @@ public GVRSceneObject(GVRContext context, GVRAndroidResource mesh,
...
+     */
+    public GVRSceneObject(GVRContext gvrContext, GVRAndroidResource mesh,
+            GVRAndroidResource texture) {
+        this(gvrContext, gvrContext.loadFutureMesh(mesh), gvrContext
+                .loadFutureTexture(texture));
+    }
+
+    /**
```
JonShemitz: Please remove this.

GVRf/Sample/gvr-test-screenshot3D/src/org/gearvrf/testscreenshot3D/TestScreenshot3DScript.java
```diff
@@ -0,0 +1,94 @@
...
+    private static final float CUBE_WIDTH = 20.0f;
+    private GVRContext mGVRContext = null;
+
+    @Override
+    public void onInit(GVRContext gvrContext) {
+        mGVRContext = gvrContext;
+
+        GVRScene scene = mGVRContext.getNextMainScene();
+
+        GVRSceneObject mFrontFace = new GVRSceneObject(gvrContext, CUBE_WIDTH,
```
JonShemitz: I think you should be using your new `GVRSceneObject(GVRContext, Future<GVRMesh>, Future<GVRTexture>)` constructor, here.

GVRf/Framework/src/org/gearvrf/FutureWrapper.java
chenchao1407: I think there should be a real implementation for isCancelled, cancel, and isDone.

rongguodong: Since this is a "faked" Future type, we cannot really cancel it. So "isCancelled" should always return "false" and "cancel" should do nothing but return "false".

For "isDone", since it is always "done" from the creating time, I have modified it to have it always return "true".

rongguodong:
I have modified FutureWrapper to make it behave correctly. I also modified the two sample files using the newly added constructor.

I changed the name of the sample which tests the screenshot result to make it more clearly.

GVRf/Framework/src/org/gearvrf/GVRVersion.java
```diff
@@ -55,5 +57,14 @@
...
     public static final String V_1_6_7 = "1.6.7";
 
-    public static final String CURRENT = V_1_6_7;
+    /**
+     * Add high-level {@link GVRSceneObject#GVRSceneObject(GVRContext,
+     * Future<GVRMesh>, Future<GVRTexture>)} constructor.
+     * 
+     * Add {@link FutureWrapper} to generate required Future<T> for the new
+     * constructor.
+     */
```
JonShemitz: The most important change: added the 3D screenshot API.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -320,6 +326,129 @@ public void run() {
...
+            ByteBuffer readbackBuffer, byte[][] byteArrayArray, int index) {
+        renderCamera(mActivity.appPtr, mMainScene, centerCamera,
+                mRenderBundle.getRightRenderTexture(), mRenderBundle);
+        readRenderResult(mRenderBundle.getPostEffectRenderTextureA().getPtr(),
+                readbackBuffer);
+        byteArrayArray[index] = Arrays.copyOf(readbackBuffer.array(),
+                readbackBuffer.array().length);
+    }
+
+    void renderSixCamerasAndReadback(final GVRCameraRig mainCameraRig,
```
JonShemitz: These three methods should probably be `private`, not package-visible.

GVRf/Sample/gvr-sample/src/org/gearvrf/sample/SampleCubeScript.java
```diff
@@ -41,116 +42,108 @@
...
-        scene.addSceneObject(mFrontBoardObject);
-        mFrontBoardObject.getTransform().setPosition(0.0f, 0.0f,
-                -CUBE_WIDTH * 0.5f);
-
-        mFrontBoardObject2 = new GVRSceneObject(gvrContext, CUBE_WIDTH,
-                CUBE_WIDTH, front);
-        mFrontBoardObject2.setName("front2");
-        scene.addSceneObject(mFrontBoardObject2);
-        mFrontBoardObject2.getTransform().setPosition(0.0f, 0.0f,
+        FutureWrapper<GVRMesh> future_mesh = new FutureWrapper<GVRMesh>(
```
JonShemitz: `future_mesh` is a C++-style name - please use `futureMesh', which follows Java conventions.

GVRf/Sample/gvr-sample/src/org/gearvrf/sample/SampleCubeScript.java
```diff
@@ -41,116 +42,108 @@
...
+                gvrContext.createQuad(CUBE_WIDTH, CUBE_WIDTH));
+
+        mFrontFace = new GVRSceneObject(gvrContext, future_mesh,
+                gvrContext.loadFutureTexture(new GVRAndroidResource(
+                        mGVRContext, R.drawable.front)));
+        mFrontFace.setName("front");
+        scene.addSceneObject(mFrontFace);
+        mFrontFace.getTransform().setPosition(0.0f, 0.0f, -CUBE_WIDTH * 0.5f);
+
+        mFrontFace2 = new GVRSceneObject(gvrContext, future_mesh,
```
JonShemitz: There's an awful lot of repeated code here ...

GVRf/Sample/gvr-sample/src/org/gearvrf/sample/SampleCubeScript.java
```diff
@@ -160,23 +153,20 @@ public void onInit(GVRContext gvrContext) {
     @Override
     public void onStep() {
         FPSCounter.tick();
-        mFrontBoardObject.getRenderData().getMaterial().setOpacity(1.0f);
-        mFrontBoardObject2.getRenderData().getMaterial().setOpacity(1.0f);
-        mFrontBoardObject3.getRenderData().getMaterial().setOpacity(1.0f);
+        mFrontFace.getRenderData().getMaterial().setOpacity(1.0f);
```
JonShemitz: Why are you setting opacity to constant values in `onStep()`? Shouldn't this happen once, in`onInit()`?

GVRf/Sample/gvr-test-screenshot3Dresult/src/org/gearvrf/testscreenshot3Dresult/TestScreenshot3DResultScript.java
```diff
@@ -0,0 +1,99 @@
...
+    private static final float CUBE_WIDTH = 20.0f;
+    private GVRContext mGVRContext = null;
+
+    @Override
+    public void onInit(GVRContext gvrContext) {
+        mGVRContext = gvrContext;
+
+        GVRScene scene = mGVRContext.getNextMainScene();
+
+        FutureWrapper<GVRMesh> future_mesh = new FutureWrapper<GVRMesh>(
```
JonShemitz: Same code comments as in `SampleCubeScript.java`.

These two look very similar - why are you adding two so-similar samples?

rongguodong: The modified version of "gvrample" shows how to use the new API to capture 3D screenshot, and save the result into six image files.

It is difficult for people (particularly reviewers) to tell whether the six images generated above are correct or not (i.e. whether they connect to each other perfectly) by simply view them by eyes. So I create the second sample "gvrtestscreenshot3Dresult". In this sample, I create a cube and map the six images to its faces. So the user can easily tell whether the result of 3D screenshot is correct or not. This sample is mainly for review reason. We can delete it after this commit is reviewed and merged.

GVRf/Sample/gvr-sample/src/org/gearvrf/sample/SampleCubeScript.java
rongguodong: This issue as well as the above code-repeat issue are both from the original gvrsample. I agree that there are lots of spaces to optimize it. But it has nothing to do with the 3D screenshot, and thus I would like to do the optimization in a separate commit and submit it later. This will keep each commit clearer and easy for maintain.

In fact, my only change to gvrsample is the screenshot processing methods. All the changes to onInit (as well as corresponding changes in GVRSceneObject) are not related to 3D screenshot, and should be done in a separate commit. But since they are already integrated, I think it is OK to leave them as is. But all further changes should be in a separate commit and should not block this 3D screenshot commit.

JonShemitz: OK.

rongguodong:
I have updated the version information, changed 5 screenshot-helper methods in GVRViewManager to private, and changed future_mesh to futureMesh in both samples.

I can change other parts in the sample code later in a separate commit.

JonShemitz:
Thanks. I’m in the middle of something: will re-review as soon as I can. Otoh, we’re down to the last nits – this would be a great time for someone else to read and verify this commit.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -320,6 +326,127 @@ public void run() {
...
+
+            centerCameraObject.detachCamera();
+            mainCameraRig.getOwnerObject()
+                    .removeChildObject(centerCameraObject);
+        }
+    }
+
+    private void returnScreenshot3DToCaller(final GVRScreenshot3DCallback callback,
+            final byte[][] byteArrayArray, final int width, final int height) {
+
```
chenchao1407: Just a minor issue. The naming of byteArrayArray seems a little bit weird to me...Can it be any other name which may be a little bit more readable or meaningful?

rongguodong: I also feel it is not good, but do not have a better idea. It is an array of type "byte[]". Do you have any suggestions for a better name?

rongguodong: Jon suggested byteArrays. I will modify and push it in soon.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -320,6 +326,127 @@ public void run() {
...
+        if (byteArrayArray.length != 6) {
+            throw new IllegalArgumentException(
+                    "byteArrayArray length is not 6.");
```
chenchao1407: Since it is a private call. I am thinking if this check is a little bit redundant...

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
rongguodong: Done.

GVRf/Sample/gvr-sample/src/org/gearvrf/sample/SampleCubeScript.java
chenchao1407: Actually gvrsample app and the following part of the codes can explain why it continues doing it in onStep().

It does want to change the texture into half transparent when your eye point to it. And when your eyes' focus leave the texture, it will return back to opaque.

chenchao1407:
Tried. Two problems found.

With 2D capture, there will be a chance to get native crash with following error message(I disable 3D capture while doing the test.):
05-01 19:35:40.205   360   360 I DEBUG   : Revision: '12'
05-01 19:35:40.205   360   360 I DEBUG   : ABI: 'arm'
05-01 19:35:40.205   360   360 I DEBUG   : pid: 15729, tid: 15749, name: Thread-6665  >>> org.gearvrf.sample                                                           <<<
05-01 19:35:40.205   360   360 I DEBUG   : signal 6 (SIGABRT), code -6 (SI_TKILL), fault addr --------
05-01 19:35:40.275   360   360 I DEBUG   : Abort message: 'sart/runtime/check_jni.cc:65] JNI DETECTED ERROR IN APPLICATION: field operation on invalid stack indirect reference table or invalid reference: 0x0'
05-01 19:35:40.275   360   360 I DEBUG   :     r0 00000000  r1 00003d85  r2 00000006  r3 00000000
05-01 19:35:40.275   360   360 I DEBUG   :     r4 a0fffdb8  r5 00000006  r6 00000002  r7 0000010c
05-01 19:35:40.275   360   360 I DEBUG   :     r8 00000000  r9 b4e4f520  sl a0d0b400  fp 000008e4
05-01 19:35:40.275   360   360 I DEBUG   :     ip 00003d85  sp a0ffdae0  lr b6ecefd5  pc b6ef2978  cpsr 60070010
05-01 19:35:40.275   360   360 I DEBUG   : 
05-01 19:35:40.275   360   360 I DEBUG   : backtrace:
05-01 19:35:40.275   360   360 I DEBUG   :     #00 pc 00037978  /system/lib/libc.so (tgkill+12)
05-01 19:35:40.275   360   360 I DEBUG   :     #01 pc 00013fd1  /system/lib/libc.so (pthread_kill+52)
05-01 19:35:40.275   360   360 I DEBUG   :     #02 pc 00014bef  /system/lib/libc.so (raise+10)
05-01 19:35:40.275   360   360 I DEBUG   :     #03 pc 00011531  /system/lib/libc.so (__libc_android_abort+36)
05-01 19:35:40.285   360   360 I DEBUG   :     #04 pc 0000fcbc  /system/lib/libc.so (abort+4)
05-01 19:35:40.285   360   360 I DEBUG   :     #05 pc 0021af5d  /system/lib/libart.so (art::Runtime::Abort()+160)
05-01 19:35:40.285   360   360 I DEBUG   :     #06 pc 000a76f3  /system/lib/libart.so (art::LogMessage::~LogMessage()+1322)
05-01 19:35:40.285   360   360 I DEBUG   :     #07 pc 000b0e19  /system/lib/libart.so (art::JniAbort(char const*, char const*)+1060)
05-01 19:35:40.285   360   360 I DEBUG   :     #08 pc 000b1375  /system/lib/libart.so (art::JniAbortF(char const*, char const*, ...)+60)
05-01 19:35:40.285   360   360 I DEBUG   :     #09 pc 000b3285  /system/lib/libart.so (art::ScopedCheck::CheckInstanceFieldID(_jobject*, _jfieldID*)+260)
05-01 19:35:40.285   360   360 I DEBUG   :     #10 pc 000b811b  /system/lib/libart.so (art::CheckJNI::GetLongField(_JNIEnv*, _jobject*, _jfieldID*)+50)
05-01 19:35:40.285   360   360 I DEBUG   :     #11 pc 001926d7  /system/lib/libart.so (art::JNI::GetDirectBufferAddress(_JNIEnv*, _jobject*)+18)
05-01 19:35:40.285   360   360 I DEBUG   :     #12 pc 000be83b  /system/lib/libart.so (art::CheckJNI::GetDirectBufferAddress(_JNIEnv*, _jobject*)+50)
05-01 19:35:40.285   360   360 I DEBUG   :     #13 pc 0016e974  /data/app/org.gearvrf.sample-1/lib/arm/libgvrf.so (Java_org_gearvrf_GVRViewManager_readRenderResult+24)
05-01 19:35:40.285   360   360 I DEBUG   :     #14 pc 00140f31  /data/dalvik-cache/arm/data@app@org.gearvrf.sample-1@base.apk@classes.dex
05-01 19:35:42.765   360   360 I DEBUG   : 
05-01 19:35:42.765   360   360 I DEBUG   : Tombstone written to: /data/tombstones/tombstone_04

In test sample, the SampleActivity.java, we have the following code block:
    @Override
    public boolean onTouchEvent(MotionEvent event) {
        if (event.getAction() == MotionEvent.ACTION_UP) {
            script.captureScreen(0, "screenshot_center");
            script.captureScreen(1, "screenshot_left");
            script.captureScreen(2, "screenshot_right");
            script.captureScreen3D("screenshot3d");
        }
        return super.onTouchEvent(event);
    }

I commented out "script.captureScreen3D("screenshot3d");" and do the test for 2D capture. Normally I can see that the drawing orders of capturing is center->left->right. But when crash happens, center is skipped. It means in GVRViewManager.onDrawEyeView(). , there is a chance that "if (mScreenshotCenterCallback != null) " will not get satisfied while doing capture screen and then crashed.

The second problem is about 3D capturescreen. After doing several times of 3D screen, you will get the following error and then native crash happens:
05-01 19:39:38.495   360   360 I DEBUG   : pid: 16232, tid: 16275, name: Thread-6715  >>> org.gearvrf.sample                                                           <<<
05-01 19:39:38.495   360   360 I DEBUG   : signal 11 (SIGSEGV), code 2 (SEGV_ACCERR), fault addr 0x4a494c85
05-01 19:39:38.565   360   360 I DEBUG   : Abort message: 'sart/runtime/check_jni.cc:65] JNI DETECTED ERROR IN APPLICATION: JNI CallVoidMethodV called with pending exception 'java.lang.OutOfMemoryError' thrown in unknown throw location'
05-01 19:39:38.565   360   360 I DEBUG   :     r0 73fdaf4f  r1 00000000  r2 beaa5530  r3 4a494c49
05-01 19:39:38.565   360   360 I DEBUG   :     r4 73fdaf4f  r5 b6f24e04  r6 a0ffdd38  r7 b4d2af75
05-01 19:39:38.565   360   360 I DEBUG   :     r8 a0556240  r9 00000000  sl 73fdaf87  fp 76a1aa40
05-01 19:39:38.565   360   360 I DEBUG   :     ip a0556244  sp a0ffdbd0  lr b4ceb013  pc b4cda80e  cpsr 00070030
05-01 19:39:38.575   360   360 I DEBUG   : 
05-01 19:39:38.575   360   360 I DEBUG   : backtrace:
05-01 19:39:38.575   360   360 I DEBUG   :     #00 pc 001d780e  /system/lib/libart.so (art::mirror::ArtMethod::IsProxyMethod()+1)
05-01 19:39:38.575   360   360 I DEBUG   :     #01 pc 001e800f  /system/lib/libart.so (art::Monitor::VisitLocks(art::StackVisitor*, void (*)(art::mirror::Object*, void*), void*, bool)+90)
05-01 19:39:38.575   360   360 I DEBUG   :     #02 pc 00227f15  /system/lib/libart.so (art::StackDumpVisitor::VisitFrame()+348)
05-01 19:39:38.575   360   360 I DEBUG   :     #03 pc 0022258f  /system/lib/libart.so (art::StackVisitor::WalkStack(bool)+250)
05-01 19:39:38.575   360   360 I DEBUG   :     #04 pc 0022a4ef  /system/lib/libart.so (art::Thread::Dump(std::__1::basic_ostream<char, std::__1::char_traits<char> >&) const+210)
05-01 19:39:38.575   360   360 I DEBUG   :     #05 pc 00232e2d  /system/lib/libart.so (art::ThreadList::DumpLocked(std::__1::basic_ostream<char, std::__1::char_traits<char> >&)+104)
05-01 19:39:38.575   360   360 I DEBUG   :     #06 pc 0021acb9  /system/lib/libart.so (art::AbortState::Dump(std::__1::basic_ostream<char, std::__1::char_traits<char> >&)+236)
05-01 19:39:38.575   360   360 I DEBUG   :     #07 pc 0021af05  /system/lib/libart.so (art::Runtime::Abort()+72)
05-01 19:39:38.575   360   360 I DEBUG   :     #08 pc 000a76f3  /system/lib/libart.so (art::LogMessage::~LogMessage()+1322)
05-01 19:39:38.575   360   360 I DEBUG   :     #09 pc 000b0e19  /system/lib/libart.so (art::JniAbort(char const*, char const*)+1060)
05-01 19:39:38.575   360   360 I DEBUG   :     #10 pc 000b1375  /system/lib/libart.so (art::JniAbortF(char const*, char const*, ...)+60)
05-01 19:39:38.575   360   360 I DEBUG   :     #11 pc 000b4485  /system/lib/libart.so (art::ScopedCheck::ScopedCheck(_JNIEnv*, int, char const*)+1284)
05-01 19:39:38.575   360   360 I DEBUG   :     #12 pc 000bb93f  /system/lib/libart.so (art::CheckJNI::CallVoidMethodV(_JNIEnv*, _jobject*, _jmethodID*, std::__va_list)+30)
05-01 19:39:38.575   360   360 I DEBUG   :     #13 pc 0016d774  /data/app/org.gearvrf.sample-2/lib/arm/libgvrf.so (_JNIEnv::CallVoidMethod(_jobject*, _jmethodID*, ...)+36)
05-01 19:39:38.575   360   360 I DEBUG   :     #14 pc 0016e578  /data/app/org.gearvrf.sample-2/lib/arm/libgvrf.so (gvr::GVRActivity::DrawEyeView(int, float)+524)
05-01 19:39:38.575   360   360 I DEBUG   :     #15 pc 0019342c  /data/app/org.gearvrf.sample-2/lib/arm/libgvrf.so (OVR::AppLocal::DrawEyeViewsPostDistorted(OVR::Matrix4<float> const&, int)+736)
05-01 19:39:38.575   360   360 I DEBUG   :     #16 pc 0016da80  /data/app/org.gearvrf.sample-2/lib/arm/libgvrf.so (gvr::GVRActivity::Frame(OVR::VrFrame)+688)
05-01 19:39:38.575   360   360 I DEBUG   :     #17 pc 0018eb1c  /data/app/org.gearvrf.sample-2/lib/arm/libgvrf.so (OVR::AppLocal::VrThreadFunction()+10212)
05-01 19:39:38.575   360   360 I DEBUG   :     #18 pc 0018f358  /data/app/org.gearvrf.sample-2/lib/arm/libgvrf.so (OVR::AppLocal::ThreadStarter(void*)+16)
05-01 19:39:38.575   360   360 I DEBUG   :     #19 pc 000137bb  /system/lib/libc.so (__pthread_start(void*)+30)
05-01 19:39:38.575   360   360 I DEBUG   :     #20 pc 0001189b  /system/lib/libc.so (__start_thread+6)


I guess that it is due to memory leak in current code logics. Thanks

chenchao1407:
Update: The crash for 3D capture screen may be because if  I triggered a new screencapture callback before the previous one is done. It will get accumulated and once the accumulated callbacks reach to some boundary and then the crash happens.

rongguodong:
I updated the memory allocation code for readbackBuffer, and played with the sample many times. I do not see crashes now. Please test it again.

chenchao1407:
Hi Guodong.

I tried. Unfortunately those two issues are still there...

rongguodong:
I pushed a new commit fixing the two crashing problems (none-pointer and out-of-memory). Please review and test it again. Thanks!

chenchao1407:
Hi Guodong. 

I reviewed the codes, I think it looks fine to me and those crashes are gone. Those are great.

However I tried to compare the cube I build from gvrtestscreen3Dresult and original cube in gvrsample, I am not sure if  I got what it should be. Maybe we can discuss about it tomorrow.

Moreover, there may be a discussion about initialization of readbackBuffer such as I talked to guodong. Once it gets initialized, we can either release it or reuse it in the future. It means, the buffer only exists when capturescreen happens. So that for each screen capture, we need to allocate memory -> render result -> deallocate memory. The pros is we can save memory from heap at the rest of time. The cons would be like each screen capture may cause a little bit more time. Or what we can do is like, allocate memory at the first screen capture, and then hold it. So in the future screen captures, it will benefit the performance from avoiding repeated memory allocation / deallocation. But the disadvantage will be like, the memory of heap will be occupied until the app ends. Or even there may be a third way like we provide a overload of captureScreenXX in GVRViewManager to let developer specify which way they want GVRF to go with.

In general, it will be like space-performance trade off issue,  I think we can have a open discussion about it.

chenchao1407:
One more issue, for those callbacks, do we need to consider the thread-safety? I mean using a synchronized lock to make GVRScript.captureScreenXX and onDrawEyeView not happen at the same time of different threads. It seems dangerous to me and the crash which Guodong fixed may be just one of cases where the multithreading issues happen.

JonShemitz:
> there may be a discussion about initialization of readbackBuffer such as I talked to guodong. 

I suspect that screen shots are going to be relatively rare in VR, just as on desktops or on phones. Many (most?) users will _never_ take a screen shot; I would be surprised if those who do take screen shots take one in every session. If I am right about this low usage, the chances of two or more screen shots in one session seem so low as to not be worth designing for.

JonShemitz:
> do we need to consider the thread-safety?

Well, of course we do. But what I recall from my first pass through the code is that

* The GL thread sees a callback handler: it allocates a `byte[]` or a `byte[][]` and (effectively) renders to that. (No thread safety issues.)
* It grabs a thread pool thread, to convert the byte array(s) to bitmap(s) (passing the byte array(s) and the callback handler to the new thread) then clears the member that triggered the screenshot. At this point, the GL thread should have no reference to the byte array(s).  (Again, no thread safety issues.) 
* The thread pool thread then converts the byte array(s) to bitmap(s), and passes them to the app, _via_ a callback on the thread pool thread. (The only thread safety issues here are on the app side, in the callback - no other GVRF code knows about the bitmap(s).)
* Once the thread pool thread returns from its `Runnable`, it no longer has any reference to the byte array(s), and they can be garbage collected.

I'll keep my eyes out as I re-read the code, but if this sketch is anywhere near right, there really shouldn't be any concurrency issues here.

GVRf/Framework/src/org/gearvrf/FutureWrapper.java
```diff
@@ -0,0 +1,50 @@
...
+ * versions of constructor for {@link GVRSceneObject} for different combinations
+ * of {@link GVRMesh} or Future<{@link GVRMesh}>, {@link GVRTexture} or Future<
+ * {@link GVRTexture}>.
+ * 
+ * @param <V>
+ *            Internal type to be wrapped. It can be either mesh or texture.
+ */
+public class FutureWrapper<V> implements Future<V> {
+
+    private V value_ = null;
```
JonShemitz: Fwiw, I would code this as a `final` field, but I don't think that you need to change this.

GVRf/Framework/src/org/gearvrf/GVRContext.java
```diff
@@ -1345,6 +1345,11 @@ protected synchronized static void resetOnRestart() {
      * same thread that calls this method, and it will not be either the GUI or
      * the GL thread.
      * 
+     * User should be responsible to wait until the last
```
JonShemitz: This is a good point, but it could perhaps be phrased a bit better.

> Users should not start a screenshot until any previous screen callbacks have returned. Overlapping screenshots will cause heavy load that could result in missed frames and/or out of memory errors.

GVRf/Framework/src/org/gearvrf/GVRContext.java
```diff
@@ -1345,6 +1345,11 @@ protected synchronized static void resetOnRestart() {
...
+     * {@code captureScreenCenter} finished to call a new
+     * {@code captureScreenCenter}. Failed to do so may result in out-of-memory
+     * error.
+     * 
      * @param callback
      *            Callback function to process the capture result. It may not be
      *            none.
```
JonShemitz: `none` s/b `{@ code null}`

GVRf/Framework/src/org/gearvrf/GVRScreenshot3DCallback.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.util.List;
+
+import android.graphics.Bitmap;
+
+/**
+ * Callback interface for asynchronous 3D screenshot capture. The screenshot
+ * result will be passed to the caller as the parameter of
+ * {@link GVRScreenshot3DCallback#onScreenCaptured(List)}.
+ */
```
JonShemitz: Probably should have an `@since`.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -104,13 +104,16 @@
     private GVRScreenshotCallback mScreenshotCenterCallback = null;
     private GVRScreenshotCallback mScreenshotLeftCallback = null;
     private GVRScreenshotCallback mScreenshotRightCallback = null;
+    private GVRScreenshot3DCallback mScreenshot3DCallback = null;
+    ByteBuffer readbackBuffer = null;
```
JonShemitz: We all know that I am no fan of the `'m'` prefix ... but consistency is important. If you add a field right next to other fields that have an `'m'`, it too should have an `'m'`.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -104,13 +104,16 @@
...
+    int readbackBufferWidth = 0, readbackBufferHeight = 0;
```
JonShemitz: These two fields appear to only be referenced  in `readRenderResult()` - they should probably be local variables.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -316,28 +338,131 @@ public void run() {
                 final Bitmap capturedBitmap = generateBitmap(byteArray, width,
                         height);
                 callback.onScreenCaptured(capturedBitmap);
+                capturedBitmap.recycle();
```
JonShemitz: Oh, please don't do this! It's harmless (though not necessary) if (and only if) the app does not use the `Bitmap` after the callback returns - but you have no way of knowing that this is the case! The app may do _anything_ besides save-to-disk-during-the-callback - imagine the app uploads the screenshot to a queue of images to be sent to a server somewhere,which is being consumed by Yet Another thread.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -316,28 +338,131 @@ public void run() {
...
+                                    e.printStackTrace();
+                                }
+                            }
+                        }
+                    }
+
+                    callback.onScreenCaptured(bitmapArray);
+
+                    for (Bitmap bitmap : bitmapArray) {
+                        bitmap.recycle();
```
JonShemitz: Again, this is very dangerous.

GVRf/Sample/gvr-sample/src/org/gearvrf/sample/SampleCubeScript.java
```diff
@@ -160,50 +154,77 @@ public void onInit(GVRContext gvrContext) {
...
-            if (eyePointeeHolder.getOwnerObject().equals(mFrontBoardObject3)) {
-                mFrontBoardObject3.getRenderData().getMaterial()
-                        .setOpacity(0.5f);
+            if (eyePointeeHolder.getOwnerObject().equals(mFrontFace3)) {
+                mFrontFace3.getRenderData().getMaterial().setOpacity(0.5f);
             }
         }
     }
 
+    private boolean lastScreenshotLeftFinished = true;
```
JonShemitz: I don't think you want four flags like this - **all** the screenshot APIs are memory and CPU intensive.

JonShemitz:
I'd have merged this with some JavaDoc issues and a small quibble with the flags in the sample  - but the `Bitmap.recycle()` codes are quite wrong.

rongguodong:
I have made changes according to Jon's comments, and pushed the modified code in. Please review again.

rongguodong:
For overlapping of screenshot calls, I think it is quite possible that a user call a captureScreenLeft and captureScreenRight together in the same frame. So I do not prevent all screenshot functions be executed in parallel, but just prevent the same screenshot function to be executed the second time before the first time finished. That is also the reason of using four booleans in the sample. 

Again, the sample code just demonstrate how to use the screenshot APIs, and I do not think the user will call all four APIs together in real applications. However, even the sample now call all four APIs together (which may cause four threads running in parallel), we do not have out of memory error as long as we do not call same API many times (although it is not guaranteed).

GVRf/Framework/src/org/gearvrf/GVRContext.java
```diff
@@ -1361,9 +1367,14 @@ protected synchronized static void resetOnRestart() {
      * same thread that calls this method, and it will not be either the GUI or
      * the GL thread.
      * 
+     * Users should start a {@code captureScreenLeft} until previous
```
JonShemitz: should _not_ start

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -104,13 +104,16 @@
     private GVRScreenshotCallback mScreenshotCenterCallback = null;
     private GVRScreenshotCallback mScreenshotLeftCallback = null;
     private GVRScreenshotCallback mScreenshotRightCallback = null;
+    private GVRScreenshot3DCallback mScreenshot3DCallback = null;
+    ByteBuffer mReadbackBuffer = null;
+    int readbackBufferWidth = 0, readbackBufferHeight = 0;
```
JonShemitz: `'m'` ...

JonShemitz:
Code looks good. JavaDoc is wrong in a few places, but this can be fixed later.

@chenchao1407 / @jason2kim  / @dknaan : Please re-test. If it runs, please merge.

rongguodong:
I have just corrected the JavaDoc, as well as renamed the two variables.

rongguodong:
Added one line at the end of onDrawEyeView to release memory if there is nor pending screenshot requests.

I keep mReadbackBuffer as a class variable so that it will not be automatically released after one call of onDrawEyeView. This is because we always call onDrawEyeView two times for left eye and right eye. If the user sets both mScreenshotLeftCallback and mScreenshotRightCallback (which is quite possible), the memory can be kept for the two calls of onDrawEyeView.

Since the user may not call screenshot APIs frequently, we release the memory if there is no pending screenshot requests.

rongguodong:
Hi Chao,

About the upside down result you got: The current code should be correct. I noticed that you captured screenshot with USB port pointing to the left (which is opposite to the direction when the device is in GearVR), so the upside down result should be expected. You can see that all 2D screenshots are also upside down. :)

chenchao1407:
I think the cache design may be necessary in some cases. I remember last time a Brazil engineer was talking about making game which ask players to take "pictures" in a VR jungle. The reuse of readbackBuffer in this case may help them improve the performance and player experience of the game. Actually I can see that there are several similar cases in current 3D gaming such as <Afrika> in PS3.

chenchao1407:
About synchronization issue, I think the relationship between GL thread and callback should be fine. The problem may occur since there may be another thread such as some sensor or listeners other than GL thread which can set the callback while running onDrawEyeView. The reason why previous crash happen is due to the onMotionEvent can set the callback while onDrawEyeView is running so that the GL was trying to render a buffer of null pointer. It got fixed by checking readbackBuffer each time before it tries to draw. I am still wondering if it is dangerous that currently we still can set callback(not run the callback) by another thread during onDrawEyeView.

rongguodong:
I think the current implementation should be fine (correct me if I am wrong).

The reason of previous crash is because we check the callback twice (once for memory allocation and once for actual readback), and the callback may be set/reset in between. In the current implementation, no matter when the callback is set and who set it, we only check it once. So there should not be problems any more.

chenchao1407:
Yeah. I know current thing should be fine so far. And actually I can't think of any real case that will cause another problem at this point. Just a little bit worried about if there is such kind of synchronization issue between set callback and onDrawEyeView while they are running in different threads. I mean, I will run this pull request and if no further issue occurs, I am going to merge it. And about my concern of synchronization, if there is really a case which may trigger another problem,  I think in the future we will see it.

rongguodong:
Sure, if we found other problems later, I can always modify it further.

One good thing is: we do not have methods to reset the callback. The user can only set it, but no way to reset it. :)

chenchao1407:
Verified.

PR #41, opened 2015-05-12
GVRF 2.0!

JonShemitz:
GVRF 1.x was basically a fairly thin Java wrapper on top of
a (primitive) C++ game engine. The native layer managed
object lifecycles using std::shared_ptr<> (and related
classes); there was lots of complex boiler-plate code
to map shared_ptr<>s to and from Java objects.

That complex code actually worked; the problem was that
doing something like creating an object and not immediately
adding it to the scene graph could leave you with a Java
object with no native implementation. Similarly, removing
objects and adding them back later could fail in the same
way. We added various hacks to 'manage' the reference counts
and maintain object-identity, but increasingly the whole
reference count scheme came to look like a classic
'leaky abstraction' - app developers had to know too much
about an implementation detail.

So, we finally decided to Do It Right, and let the Java
objects 'own' their native implementation objects. That
is: no more reference counting; Java finalize() methods
free their implementation objects. This eliminates a
surprisingly large amount of code: GVRF should even
run (a little) faster, now that there are no more
std::shared_ptr<>s.

Everything has been 'smoke tested' and seem to run just
fine; it's possible that extended use will reveal problems.
Similarly, while the 'magic of virtual destructors' means
that most 'hybrid objects' should be deleted cleanly, there
does remain the possibility that this version has added
some memory leaks. TEST CAREFULLY!

j.shemitz@samsung.com

j.shemitz@samsung.com


GVRf/Framework/jni/objects/components/camera_rig_jni.cpp
```diff
@@ -126,49 +120,24 @@ Java_org_gearvrf_NativeCameraRig_getLookAt(JNIEnv * env,
...
-Java_org_gearvrf_NativeCameraRig_getRightCamera(JNIEnv * env,
-        jobject obj, jlong jcamera_rig) {
-    std::shared_ptr<CameraRig> camera_rig = *reinterpret_cast<std::shared_ptr<
-            CameraRig>*>(jcamera_rig);
-    std::shared_ptr<Camera> right_camera(camera_rig->right_camera());
-    return right_camera.get() == NULL ?
-            0 :
-            reinterpret_cast<jlong>(new std::shared_ptr<Camera>(right_camera));
-}
-
```
chenchao1407: Why do we need to remove these two methods since we still keep right/left_camera() method to get the pointer of two cameras?

JonShemitz: Read The Source, Luke.

chenchao1407: I think I know the point here now. But right now I am thinking that actually there is nothing internal that is using camera_rig::left/right_camera() and also, the developer may never reach these two methods either. Please correct me if I am wrong. Thanks in advance.

JonShemitz: >  right now I am thinking that actually there is nothing internal that is using camera_rig::left/right_camera() and also, the developer may never reach these two methods either

Looks like you're right: No references to `CameraRig::left_camera()` or `CameraRig::right_camera()`. Otoh, cleaning out dead code was _not_ the point of this checkin - I only zapped jni methods that weren't being called any longer.

GVRf/Framework/jni/objects/mesh_eye_pointee_jni.cpp
```diff
@@ -39,28 +36,16 @@ Java_org_gearvrf_NativeMeshEyePointee_setMesh(JNIEnv * env,
...
-}
-
-JNIEXPORT jlong JNICALL
-Java_org_gearvrf_NativeMeshEyePointee_getMesh(JNIEnv * env,
-        jobject obj, jlong jmesh_eye_pointee) {
-    std::shared_ptr<MeshEyePointee> mesh_eye_pointee =
-            *reinterpret_cast<std::shared_ptr<MeshEyePointee>*>(jmesh_eye_pointee);
-    std::shared_ptr<Mesh> mesh = mesh_eye_pointee->mesh();
-    return reinterpret_cast<jlong>(new std::shared_ptr<Mesh>(mesh));
+    Mesh* mesh = reinterpret_cast<Mesh*>(jmesh);
```
chenchao1407: mesh() also becomes dead codes here which is similar to camera_rig.left/right_camera()

GVRf/Framework/jni/objects/post_effect_data.h
```diff
@@ -160,7 +160,7 @@ class PostEffectData: public HybridObject {
 
 private:
     ShaderType shader_type_;
-    std::map<std::string, std::shared_ptr<Texture>> textures_;
+    std::map<std::string, Texture*> textures_;
```
chenchao1407: similar here... Actually based on this patch, getTexture and the map textures_ will never be reached. Later after this patch get merged, some one else or  I will make a clean-up patch for these.

GVRf/Framework/jni/shaders/post_effect_shader_manager.h
```diff
@@ -53,20 +52,22 @@ class PostEffectShaderManager: public HybridObject {
     }
 
     ~PostEffectShaderManager() {
+        delete color_blend_post_effect_shader_;
+        delete horizontal_flip_post_effect_shader_;
+        // We don't delete the custom shaders, as their Java owner-objects will do that for us.
```
chenchao1407: Imho, I think in the destructor you have to do "delete" with all the pointers in std containers such as custom_post_effect_shaders_, otherwise only those pointers will be released but the content those pointers point to will become memory leak.

chenchao1407: Ok, please ignore my previous comment because I missed the comment here.

GVRf/Framework/src/org/gearvrf/GVRHybridObject.java
```diff
@@ -15,209 +15,97 @@
...
+     * exception is classes that maintain any sort of 'registry' of 'live'
+     * classes (like {@link GVREyePointeeHolder}). If you <em>do</em> override
+     * this method, be sure to call {@code super.destructor()}.
      * 
-     * See {@link #setKeepWrapper(boolean)} for details.
+     * @since 2.0.0
      */
-    public boolean getKeepWrapper() {
-        return (mFlags & KEEP_WRAPPER_MASK) != 0;
+    protected void destructor() {
```
ejsamsung: Deriving classes can neglect to call super from overriding implementations, which seems like it would punch a hole through the protection that is supposed to be provided by making finalize() and close() final.

JonShemitz: > Deriving classes can neglect to call super from overriding implementations, 

Very true - I was just thinking of that this morning, actually. I'll have think about that a bit. Clearly, I want to move native `delete` out of a method that might not get called, but I'm not sure that I can be 100% safe: `GVREyePointeeHolder` does override the current `destructor()` to remove a `WeakReference` from `sEyePointeeHolders`.

ejsamsung: Is there any reason not to move the call to `NativeHybridObject.delete()` up into `close()`, right after the call to `destructor()`?  Then `destructor()` is purely a hook method and the native object is valid when it's called.  We lose the ability to have `delete()` called before we do our own thing, but how often is it going to be useful or necessary to do something only *after* the native object has been destroyed?

JonShemitz: > Is there any reason not to move the call to NativeHybridObject.delete() up into close(), right after the call to destructor()?

Right - that's how I would "move native `delete` out of a method that might not get called" but that only guarantees that I call native `delete`. Obviously, not leaking memory is a good thing, but I would like to also be able to guarantee that (say) `GVREyePointeeHolder` will still deregister itself, even if a descendant class _doesn't_ call `super.destructor()`. I'm not sure I can do that, short of some sort of fragile and expensive system of `onFinalize` callbacks.

GVRf/Framework/jni/objects/components/eye_pointee_holder.h
```diff
@@ -69,7 +69,7 @@ class EyePointeeHolder: public Component {
 private:
     bool enable_;
     glm::vec3 hit_;
-    std::vector<std::shared_ptr<EyePointee>> pointees_;
+    std::vector<EyePointee*> pointees_;
```
chenchao1407: Just a quick question. I can see comments in other codes that the reason why you don't do delete with some std containers elements is because in Java, the finalize() will handle those pointers' content. So is pointees__?

JonShemitz: > Just a quick question. I can see comments in other codes that the reason why you don't do delete with some std containers elements is because in Java, the finalize() will handle those pointers' content. So is pointees__?

I am _not_ sure I know what you are asking! My best guess is 'Are these `EyePointee`s leaked?'

I don't think so. The `GVREyePointeeHolder` has a list of `GVREyePointee`s: when it is garbage, so are the  `GVREyePointee`s.

chenchao1407: Yeah. I just want to confirm if this is the same as others.

JonShemitz:
I've rebased this. The 'old' samples seem to work fine; the cubemap sample shows all-green, with no hints in the log. Tom says that Guodong saw similar behavior on another device (before 2.0) so it may be driver issue.

GVRf/Framework/src/org/gearvrf/GVRHybridObject.java
```diff
@@ -15,209 +15,97 @@
...
-     * opposed to another object that refers to the same native object). The
-     * most common scenario where this matters is one where you have declared
-     * classes that {@code extend GVRSceneObject} and you want to get your
-     * sub-classed object back from the {@link GVRPicker} because that instance
-     * contains app-specific values that a 'generic' {@link GVRSceneObject} does
-     * not.
+     * @since 2.0.0
+     */
+    @Override
+    protected final void finalize() throws Throwable {
```
JonShemitz: We've had some internal (off GitHub) discussion about the costs of using `finalize()`. While it looked like I _had_ to do this (because (at least on Lollipop) `Reference.get()` always returns `null` once the `Reference` has been placed on a `ReferenceQueue`, for all three `Reference` types) @ejsamsung suggested the clever work-around of using a class derived from `PhantomReference` that has a copy of the native pointer, so it can call `NativeHybridObject.delete()`.

The only problem with this optimization is that by the time a `Reference` has been placed on a `ReferenceQueue`, it has already been finalized: `~HybridObject()` _can't_ call an `onDestroy()` method that would (say) unregister `GVREyePointeeHolder`. I mention this now as a sort of early warning: the `destructor()` method is probably going to disappear in the very near future.

thomasflynn:
you need to rebase this change now.

JonShemitz:
OK, thanks.

JonShemitz:
Should be 'rebased' to current master; Solar System, Model Viewer, and OpacityAni all work: exercises unlit shader; async; custom shaders; custom post-effects.

thomasflynn:
I also tested pickandmove, gvr-video, simple-sample, and scene-objects.  I was unable to find any problems that were new.  I'm going to go ahead with this change.

PR #42, opened 2015-05-12
Get the Oculus Mobile SDK path from envvar

marctrem:
Signed-off-by: Marc-André Tremblay <m.tremblay@partner.samsung.com>


JonShemitz:
Looks right; does not break normal (in-Eclipse) build.

PR #44, opened 2015-05-13
More preconditions

JonShemitz:
Possible refactoring of parameter validation, per team meeting last week and a discussion with @guodongrong  today.


chenchao1407:
I would rather merge it at this time since some issues aren't concerns of this pull request. But I will make a follow up pull request which will refactor Assert a little bit and what is more important is to change current native crash mechanism for key-not-found case.

PR #47, opened 2015-05-15
Occlusion Query and Frustum Culling.

lokeshguddu:
We had some error related to the rebasing with the frustum-culling branch so we merged our changes with the upstream repo and created a new pull request. Please ignore the frustum-culling branch pull request.


thomasflynn:
1) rebase
2) how did you test this??  I just tested in the solar-system sample, and it crashes when frustum culling is enabled.
3) debugged it a bit, the bounding_box_info_ was being populated for the newly created bounding box for the new mesh inside getBoundingBox, but not populated for mesh object which you're calculating the bbox for.
4) did some hacking.  removed const from getBoundingBox so i could simply duplicate the bounding_box_info_.push_back()s for 'this' in addition to mesh->.  This probably isn't the right fix, but i could at least get things to compile.  I also had to add a check in getBoundingBoxInfo() to check if the size was 0, if so, call getBoundingBox() to populate the bounding_box_info_.  At this point, it at least doesn't crash.  Instead, it draws nothing.
5) if I also enable occlusion culling, then i finally at least see my scene.
6) it appears as if occlusion culling doesn't work as i expect.  I added a variable to mesh.h to store the number of triangles it has and added some logs to renderer.cpp to output the number of triangles drawn in the scene.  frustum culling appears to work.  when an object goes off-screen, the number of triangles decreases by the number of triangles of the object that left the screen.  good.  if that object is occluded by another planet (earth), the number of drawn triangles does not decrease.  I would expect it to.

Please fix this and test it.  I really do want this in ASAP.

lokeshguddu:
Hi,

We have fixed the issue relating to the bounding box info. The application was crashing because the value of the bounding_box_info_ was null unless the getBoundingBox() method was called. So we fixed it with a separate method for these specific purpose. Also, we were using vector<float> as a return type for this value, there was a significant drop in performance so we replaced the implementation with an array and its working as expected.

thomasflynn:
ok, it doesn't crash and frustum culling seems functional.  It appears as though it cannot be merged as is (the 'merge pull request' button is inactive).  I guess there are conflicts and it needs to be manually rebased.

chenchao1407:
Manual rebase is needed here.

Thanks.

phyuzion:
Need to rebase. can't click.

lokeshguddu:
Whenever I'm issuing a rebase on my office system it just hangs in there and never finishes it. I will  try that again on my home system and will update the status.

soumyajitdeb:
Hi Tom

Regarding the below:

>5) if I also enable occlusion culling, then i finally at least see my scene.
By default the visibility status of all objects is set to invisible. On the first frame occlusion queries are issued for all objects in the frustum. Once they pass, the objects become visible.

>6) it appears as if occlusion culling doesn't work as i expect. I added a variable to mesh.h to store >the number of triangles it has and added some logs to renderer.cpp to output the number of >triangles drawn in the scene. frustum culling appears to work. when an object goes off-screen, the >number of triangles decreases by the number of triangles of the object that left the screen. good. if >that object is occluded by another planet (earth), the number of drawn triangles does not decrease. >I would expect it to.

To test whether occlusion culling is working, we did the following: 
- We disabled frustum culling by always returning true from Renderer::is_cube_in_frustum(). This will make sure that if occlusion queries do not work correctly, we'll always see all objects rendered.
- We then put an explicit message whenever the visibility status of a scene object is set to invisible after the result of an occlusion query
- We measured the number of primitives rendered in the draw call. 
- We noticed that the number of draw calls seems to change based upon the visibility of objects on the screen indicating that the results of the occlusion queries is being used for determining visibility of objects.
- We also tested by checking if the draw calls reduce when the red planet goes behind earth - it appears to be the case.
- What we did observe is that occlusion queries with bounding boxes cause some amount of flickering at times. To avoid this, we only change visibility status once the object's return status from the occlusion query has stabilized and is the same for 12 frames (0.2s) or more.


Thanks
Soumyajit

GVRf/Framework/jni/engine/renderer/renderer.cpp
```diff
@@ -49,24 +49,147 @@ void Renderer::renderCamera(std::shared_ptr<Scene> scene,
...
+                        render_data_vector.push_back(render_data);
+                    }
                 }
             }
         }
         std::sort(render_data_vector.begin(), render_data_vector.end(),
                 compareRenderData);
 
-        glm::mat4 view_matrix = camera->getViewMatrix();
-        glm::mat4 projection_matrix = camera->getProjectionMatrix();
```
thomasflynn: you ended up renaming this to proj_matrix up above and didn't change it accordingly on lines 217 and 233.  the build broke.

lokeshguddu: My mistake! 
I tested it on my office system and when rebase didn't happen here I tried it on the c9 IDE. Must have missed these two lines during rebase conflict resolution. Thank you for pointing this out and fixing it.

thomasflynn:
there were other problems too.  it crashed when frustum culling was enabled and is_cube_in_frustum() had a hard-coded return true;.  It also needed to have the code formatter run on it (crtl-shift-F in eclipse).  I made a new pull-request with the necessary fixes.  let me know if you see any problems with it.

PR #49, opened 2015-05-17
Add a setDrawMode() method to GVRRenderData.

thomasflynn:
Add setDrawMode()/getDrawMode methods to GVRRenderData.  This will allow
users to draw GL_LINES and GL_POINTS in addition to the default of
GL_TRIANGLES.

tom.flynn@samsung.com

### DIFF
--- a/GVRf/Framework/jni/shaders/material/unlit_shader.cpp
+++ b/GVRf/Framework/jni/shaders/material/unlit_shader.cpp
@@ -97,7 +97,7 @@ void UnlitShader::render(const glm::mat4& mvp_matrix,
     glUseProgram(program_->id());
 
     glUniformMatrix4fv(u_mvp_, 1, GL_FALSE, glm::value_ptr(mvp_matrix));
-    glActiveTexture (GL_TEXTURE0);
+    glActiveTexture(GL_TEXTURE0);
     glBindTexture(texture->getTarget(), texture->getId());
     glUniform1i(u_texture_, 0);
     glUniform3f(u_color_, color.r, color.g, color.b);
--- a/GVRf/Framework/jni/shaders/material/unlit_shader.cpp
+++ b/GVRf/Framework/jni/shaders/material/unlit_shader.cpp
@@ -13,7 +13,6 @@
  * limitations under the License.
  */
 
-
 /***************************************************************************
  * Renders a texture without light.
  ***************************************************************************/
@@ -98,14 +97,15 @@ void UnlitShader::render(const glm::mat4& mvp_matrix,
     glUseProgram(program_->id());
 
     glUniformMatrix4fv(u_mvp_, 1, GL_FALSE, glm::value_ptr(mvp_matrix));
-    glActiveTexture(GL_TEXTURE0);
+    glActiveTexture (GL_TEXTURE0);
     glBindTexture(texture->getTarget(), texture->getId());
     glUniform1i(u_texture_, 0);
     glUniform3f(u_color_, color.r, color.g, color.b);
     glUniform1f(u_opacity_, opacity);
 
     glBindVertexArray(mesh->getVAOId());
-    glDrawElements(GL_TRIANGLES, mesh->triangles().size(), GL_UNSIGNED_SHORT, 0);
+    glDrawElements(render_data->draw_mode(), mesh->triangles().size(),
+            GL_UNSIGNED_SHORT, 0);
     glBindVertexArray(0);
 #else
     glUseProgram(program_->id());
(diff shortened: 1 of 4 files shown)


chenchao1407:
Looks good to me. Will test it later.

GVRf/Framework/src/org/gearvrf/GVRRenderData.java
```diff
@@ -323,6 +323,31 @@ public boolean getAlphaBlend() {
...
+     * Set the draw mode for this mesh. Default is GL_TRIANGLES.
+     * 
+     * @param drawMode
+     */
+    public void setDrawMode(int drawMode) {
+        if (drawMode != GL_POINTS || drawMode != GL_LINES
+                || drawMode != GL_LINE_STRIP || drawMode != GL_LINE_LOOP
+                || drawMode != GL_TRIANGLES || drawMode != GL_TRIANGLE_STRIP
+                || drawMode != GL_TRIANGLE_FAN) {
+            throw new IllegalArgumentException(
```
chenchao1407: OK. I tried so many times and finally found the logic problem here. I think you should use "&&" not "||" here...

thomasflynn:
you're right.  I added that in as a last-minute clean-up.  i'll fix it.  standby.

thomasflynn:
done.

chenchao1407:
Just tried several modes on solar systems and it works as I expected.

Looks good to me to do merge if there is no further objection.

GVRf/Framework/jni/shaders/material/unlit_shader.cpp
```diff
@@ -98,14 +97,15 @@ void UnlitShader::render(const glm::mat4& mvp_matrix,
     glUseProgram(program_->id());
 
     glUniformMatrix4fv(u_mvp_, 1, GL_FALSE, glm::value_ptr(mvp_matrix));
-    glActiveTexture(GL_TEXTURE0);
+    glActiveTexture (GL_TEXTURE0);
```
JonShemitz: Whitespace ...

GVRf/Framework/src/org/gearvrf/GVRRenderData.java
```diff
@@ -323,6 +323,31 @@ public boolean getAlphaBlend() {
...
+     * @param drawMode
+     */
+    public void setDrawMode(int drawMode) {
+        if (drawMode != GL_POINTS && drawMode != GL_LINES
+                && drawMode != GL_LINE_STRIP && drawMode != GL_LINE_LOOP
+                && drawMode != GL_TRIANGLES && drawMode != GL_TRIANGLE_STRIP
+                && drawMode != GL_TRIANGLE_FAN) {
+            throw new IllegalArgumentException(
+                    "drawMode must be one of GL_POINTS, GL_LINES, GL_LINE_STRIP, GL_LINE_LOOP, GL_TRIANGLES, GL_TRIANGLE_FAN, GL_TRIANGLE_STRIP.  Value passed was: "
+                            + drawMode);
```
JonShemitz: Minor point:  try to discourage string concatenation in `IllegalArgumentException` constructors. Using `Exceptions.IllegalArgument()` lets you use `String.format()` syntax, which gives you more control and is probably smaller than inline `String + int`.

Major point: I'm not sure the decimal value of `drawMode` helps, here. What matters most is that it's not GL_POINTS, GL_LINES, GL_LINE_STRIP, GL_LINE_LOOP, GL_TRIANGLES, GL_TRIANGLE_FAN, or GL_TRIANGLE_STRIP.

JonShemitz:
Please fix the whitespace - we really can't have one set of rules for contributors and another set of rules for the Owner.

GVRf/Framework/jni/shaders/material/unlit_shader.cpp
thomasflynn: that seems to have been introduced by the formatter.  perhaps my settings are off.  if i remove that space, and then hit crtl-shift-f, it re-introduces that space.  I can remove it though.

GVRf/Framework/src/org/gearvrf/GVRRenderData.java
thomasflynn: Good point on the decimal value of drawMode.  So it sounds like just truncating the message to say that its not one of the GL_* values would be sufficient.

JonShemitz: Yes, imho.

thomasflynn:
ok, done.

JonShemitz:
Look fine to me. @chenchao1407, I guess it's OK to go ahead and merge it the samples run normally.

PR #51, opened 2015-05-19
Update activity_jni.cpp

sisimma:
Since OpenGL specifies matrices as a one-dimensional array listed in column-major order while C does in row-major order, the gl matrix should be converted accordingly.


sisimma:
This fixed menu display such as volume control and fps display in debug mode. With this you can see volume bar at the center when you press volume keys up and down.

thomasflynn:
ah, thanks for the explanation.  at first check, it looked as though we're not using this matrix in our code.  but if it is needed for oculus' rendering of the volume bar (and probably other things) then this pull-request deserves a stronger look.

chenchao1407:
Hi Sean.

I can see this solve the problem of Volume bar. 

@thomasflynn  @JonShemitz  Any objection? Otherwise I am going to merge this one.

PR #54, opened 2015-05-19
Fix GVRConsole.

JonShemitz:
1. Fix some bit rot, that (partially) broke post-effect rendering.
2. Add an overloaded constructor, to be GVRContext.getNextMainScene()
   friendly.

j.shemitz@samsung.com


JonShemitz:
@thomasflynn: Your 'stats' branch works just fine, merged onto this. There _is_ a noticeable slowdown, in 'stats' but _not_ when the Solar sample uses this commit to display a frame count.

thomasflynn:
was there any change in render.cpp??  it looks like all spacing and formatting changes.  if there's a change, it's hard to see it.

tflynnt:
I see the change in renderer.cpp now.  Adding the glClearColor, correct?

JonShemitz:
> I see the change in renderer.cpp now. Adding the glClearColor, correct?

Yes.

thomasflynn:
you need to rebase this change now.

PR #56, opened 2015-05-20
Webvideocamera

thomasflynn:
Added GVRWebViewSceneObject, VideoSceneObject, and CameraSceneObjects to the scene-objects sample code.

tom.flynn@samsung.com

### DIFF
--- a/GVRf/Sample/scene-objects/src/org/gearvrf/sample/sceneobjects/SampleViewManager.java
+++ b/GVRf/Sample/scene-objects/src/org/gearvrf/sample/sceneobjects/SampleViewManager.java
@@ -37,6 +37,7 @@ import org.gearvrf.GVRTexture;
 import org.gearvrf.animation.GVRAnimation;
 import org.gearvrf.animation.GVRRepeatMode;
 import org.gearvrf.animation.GVRRotationByAxisAnimation;
+import org.gearvrf.scene_objects.GVRCameraSceneObject;
 import org.gearvrf.scene_objects.GVRConeSceneObject;
 import org.gearvrf.scene_objects.GVRCubeSceneObject;
 import org.gearvrf.scene_objects.GVRCylinderSceneObject;
@@ -48,7 +49,7 @@ public class SampleViewManager extends GVRScript {
 
     private int currentObject = 0;
     private GVRActivity mActivity;
-    
+
     SampleViewManager(GVRActivity activity) {
         mActivity = activity;
     }
@@ -72,6 +73,9 @@ public class SampleViewManager extends GVRScript {
                 gvrContext);
         GVRConeSceneObject coneObject = new GVRConeSceneObject(gvrContext);
         GVRWebViewSceneObject webViewObject = createWebViewObject(gvrContext);
+        GVRCameraSceneObject cameraObject = new GVRCameraSceneObject(
+                gvrContext, 8.0f, 4.0f,
+                ((SceneObjectActivity) mActivity).getCamera());
 
         objectList.add(quadObject);
         objectList.add(cubeObject);
@@ -79,9 +83,10 @@ public class SampleViewManager extends GVRScript {
         objectList.add(cylinderObject);
         objectList.add(coneObject);
         objectList.add(webViewObject);
-        
+        objectList.add(cameraObject);
+
         // turn all objects off, except the first one
-        for(int i=1; i<objectList.size(); i++) {
+        for (int i = 1; i < objectList.size(); i++) {
             objectList.get(i).getRenderData().setRenderMask(0);
         }
 
@@ -101,14 +106,7 @@ public class SampleViewManager extends GVRScript {
         cylinderObject.getTransform().setPosition(0.0f, 0.0f, -3.0f);
         coneObject.getTransform().setPosition(0.0f, 0.0f, -3.0f);
         sphereObject.getTransform().setPosition(0.0f, -1.0f, -3.0f);
-
-        // GVRAnimation animation = new
-        // GVRRotationByAxisAnimation(cylinderObject, 50, -3600, 0, 1, 0);
-        // animation.setRepeatCount(GVRRepeatMode.REPEATED).setRepeatCount(-1);
-        // animation.start(gvrContext.getAnimationEngine());
-        // cubeObject.getTransform().setRotationByAxis(45.0f, 1.0f, 0.0f, 0.0f);
-
-        // cylinderObject.getRenderData().setCullTest(false);
+        cameraObject.getTransform().setPosition(0.0f, 0.0f, -4.0f);
 
         // add the scene objects to the scene graph
         scene.addSceneObject(quadObject);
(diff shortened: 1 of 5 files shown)


phyuzion:
I think this is enough code for full sample. :)

tflynnt:
I was inspired by you :-).  Your pull-request the other week was a good idea.  I just wanted it in a separate sample.  There's one last scene object i want in: GVRTextSceneObject.  That needs some more debugging before it can go in.

GVRf/Sample/scene-objects/src/org/gearvrf/sample/sceneobjects/SceneObjectActivity.java
```diff
@@ -30,15 +37,87 @@
...
-        mViewManager = new SampleViewManager();
+
+        createWebView();
+        createCameraView();
+
+        mViewManager = new SampleViewManager(this);
         setScript(mViewManager, "gvr_note4.xml");
     }
 
+    private void createWebView() {
```
jason2kim: I can see it reflects my previous comments on Insu’s including the UI thread and non-commercial sample url. No further comment for me at this moment.

GVRf/Sample/scene-objects/src/org/gearvrf/sample/sceneobjects/SampleViewManager.java
```diff
@@ -64,15 +75,23 @@ public void onInit(GVRContext gvrContext) {
         GVRCylinderSceneObject cylinderObject = new GVRCylinderSceneObject(
                 gvrContext);
         GVRConeSceneObject coneObject = new GVRConeSceneObject(gvrContext);
+        GVRWebViewSceneObject webViewObject = createWebViewObject(gvrContext);
+        GVRCameraSceneObject cameraObject = new GVRCameraSceneObject(
+                gvrContext, 8.0f, 4.0f,
+                ((SceneObjectActivity) mActivity).getCamera());
```
JonShemitz: Why cast here? Why not just declare `mActivity` as `SceneObjectActivity` (on line 54) and pass it as a `SceneObjectActivity` (on line 56)?

GVRf/Sample/scene-objects/src/org/gearvrf/sample/sceneobjects/SampleViewManager.java
```diff
@@ -64,15 +75,23 @@ public void onInit(GVRContext gvrContext) {
...
         objectList.add(cylinderObject);
         objectList.add(coneObject);
-        
+        objectList.add(webViewObject);
+        objectList.add(cameraObject);
+        objectList.add(videoObject);
+
         // turn all objects off, except the first one
-        for(int i=1; i<objectList.size(); i++) {
+        for (int i = 1; i < objectList.size(); i++) {
```
JonShemitz: Calling `.size()` a lot ...

GVRf/Sample/scene-objects/src/org/gearvrf/sample/sceneobjects/SampleViewManager.java
```diff
@@ -123,7 +159,12 @@ public void setXAngle(float angle) {
 
     public void onTap() {
 
-        objectList.get(currentObject).getRenderData().setRenderMask(0);
+        GVRSceneObject object = objectList.get(currentObject);
+        object.getRenderData().setRenderMask(0);
```
JonShemitz: Should we add a `GVRRenderMaskBit` value of 0?

GVRf/Sample/scene-objects/src/org/gearvrf/sample/sceneobjects/SampleViewManager.java
```diff
@@ -131,12 +172,15 @@ public void onTap() {
...
-                        GVRRenderData.GVRRenderMaskBit.Left
-                                | GVRRenderData.GVRRenderMaskBit.Right);
+        object = objectList.get(currentObject);
+        if (object instanceof GVRVideoSceneObject) {
+            GVRVideoSceneObject video = (GVRVideoSceneObject) object;
+            video.getMediaPlayer().start();
+        }
+        
+        object.getRenderData().setRenderMask(
+                GVRRenderData.GVRRenderMaskBit.Left
```
JonShemitz: We probably should add a `Both` constant ...

GVRf/Sample/scene-objects/src/org/gearvrf/sample/sceneobjects/SceneObjectActivity.java
```diff
@@ -30,15 +37,87 @@
...
+            @Override
+            public boolean shouldOverrideUrlLoading(WebView view, String url) {
+                view.loadUrl(url);
+                return true;
+            }
+        });
+    }
+
+    private boolean checkCameraHardware(Context context) {
+        if (context.getPackageManager().hasSystemFeature(
```
JonShemitz: What's wrong with a simple 

    return context.getPackageManager().hasSystemFeature(PackageManager.FEATURE_CAMERA);

?

JonShemitz:
Looks OK - I don't see anything wrong, just a very few things that I'd do differently.

GVRf/Sample/scene-objects/src/org/gearvrf/sample/sceneobjects/SampleViewManager.java
thomasflynn: It was late at night.  What you suggest is certainly much cleaner.  Just made that change in my latest patchset.

GVRf/Sample/scene-objects/src/org/gearvrf/sample/sceneobjects/SampleViewManager.java
thomasflynn: well, like 8 times or so.  still, fixed it in my latest patchset.

GVRf/Sample/scene-objects/src/org/gearvrf/sample/sceneobjects/SampleViewManager.java
thomasflynn: Agreed.  we should do that in a separate check-in for GVRRenderData.  We should add a 'both' and 'none' or 'off' or something.

GVRf/Sample/scene-objects/src/org/gearvrf/sample/sceneobjects/SceneObjectActivity.java
thomasflynn: heh, it's what you get when you read the google documentation on the camera :)  http://developer.android.com/guide/topics/media/camera.html

I changed it to the simple return you suggest in my latest commit.

JonShemitz:
Am I missing something? I pulled the commit to test it, but can't import it into Eclipse because there doesn't seem to be a `.project` file.

thomasflynn:
i can add that .project file.  is that necessary?  i don't remember having to do that before.

JonShemitz:
> i can add that .project file. is that necessary? i don't remember having to do that before.

Yes, Eclipse won't import a project without the `.project` file. You probably don't remember doing so either because it's been a while since you've added a new project to a repository or just because it's the sort of fussy little detail that it's easy to forget.

thomasflynn:
ok, added the project file.

JonShemitz:
> ok, added the project file.

Woops - no `.classpath`, either.

thomasflynn:
ok done.

thomasflynn:
any further requests on this, guys?

JonShemitz:
Does it work, Chao?

chenchao1407:
Tried. Everything seems good. Only a small concern on video. Once it got stopped, it will never come back to play state. I took a look at the codes to see if it is based on design or a bug.  I think it is latter.

thomasflynn:
ok, paused the video rather than stopping it.

PR #64, opened 2015-05-21
Remove mipmap filter for compatiblity of cube map texture.

rongguodong:
Remove mipmap filter for compatiblity of cube map texture.
Changed 1.0f to 1.0 because Mali driver does not like it.

rongguodong@hotmail.com


rongguodong:
As Jon suggested, I move this change to a new pull request, so that we can merge it before merging the demo.

PR #66, opened 2015-05-22
GVRTextViewSceneObject

chenchao1407:
Just create a new GVRTextViewSceneObject for 2 purposes.

1. It is more Android way to handle text. Compared to drawing text with raw Paint class. It can better organize the input text such as automatic switch line of a long string. And it is more convenient to do many kinds of alignment stuffs.

2. I consider it as an attempt for importing more Android-style views into GVRf. I think somehow it will be beneficial for developers.

Also add a sample for this class, and add it into SceneObjectActivity sample.

I think there are still much space to improve this class. I think I will be add more API into it to better control the TextView if this is merged.

c.chen@samsung.com


GVRf/Sample/scene-objects/src/org/gearvrf/sample/sceneobjects/SampleViewManager.java
```diff
@@ -76,12 +70,17 @@ public void onInit(GVRContext gvrContext) {
         GVRCylinderSceneObject cylinderObject = new GVRCylinderSceneObject(
                 gvrContext);
         GVRConeSceneObject coneObject = new GVRConeSceneObject(gvrContext);
-        GVRTextSceneObject textObject = new GVRTextSceneObject(gvrContext, "Hello World!");
+        GVRTextSceneObject textObject = new GVRTextSceneObject(gvrContext,
```
tflynnt: I'll test tomorrow morning, but assuming TextViewSceneObject works as advertised, i'd be totally fine with removing TextSceneObject.

GVRf/Sample/gvrtextviewsample/AndroidManifest.xml
```diff
@@ -0,0 +1,43 @@
...
+ * limitations under the License.
+ */-->
+
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="org.gearvrf.gvrtextviewsample"
+    android:versionCode="1"
+    android:versionName="1.0" >
+
+    <uses-sdk
+        android:minSdkVersion="16"
```
tflynnt: Min version should be 19 i think.

tflynnt:
It's a bit more challenging to do a code review from my phone than i thought it would be :-).  Next time i'll lug my 20+lb laptop home.  I've only spotted minor things so far.

GVRf/Sample/gvrtextviewsample/AndroidManifest.xml
chenchao1407: I think I copied it from simple sample...:-(

mkeblx:
Great! this is exactly the kind of stuff that I (and I imagine many other devs) will find very useful. I'll probably use this for the SF VR hackathon this weekend. Thanks.

GVRf/Framework/src/org/gearvrf/scene_objects/GVRTextViewSceneObject.java
```diff
@@ -0,0 +1,310 @@
...
+import android.util.Log;
+import android.view.Surface;
+import android.view.View;
+import android.view.ViewGroup.LayoutParams;
+import android.webkit.WebView;
+import android.widget.LinearLayout;
+import android.widget.TextView;
+
+public class GVRTextViewSceneObject extends GVRSceneObject implements
+        GVRDrawFrameListener {
```
JonShemitz: Just fwiw, @chenchao1407 , it's OK for the component to implement `GVRDrawFrameListener`, but this _does_ make the callback `public`. It's often better to implement callback interfaces _via_ private, nested objects.

GVRf/Framework/src/org/gearvrf/scene_objects/GVRTextViewSceneObject.java
```diff
@@ -0,0 +1,310 @@
...
+
+    /**
+     * Shows a {@link TextView} on a {@linkplain GVRSceneObject scene object}.
+     * 
+     * @param gvrContext
+     *            current {@link GVRContext}
+     * @param gvrActivity
+     *            a {@link GVRActivity}
+     * @param width
+     *            width for createQuad to get a new {@link GVRMesh}
```
JonShemitz: This is true, strictly speaking, but probably not quite as helpful as it could be: There's potential for confusion between the __GVRf size__ and the __Android size__.

GVRf/Framework/src/org/gearvrf/scene_objects/GVRTextViewSceneObject.java
```diff
@@ -0,0 +1,310 @@
...
+     * with both view's default height and width and quad's default height and
+     * width. And the default text of this TextView will be "Hello World!"
+     * 
+     * @param gvrContext
+     *            current {@link GVRContext}
+     * @param gvrActivity
+     *            a {@link GVRActivity}
+     */
+    public GVRTextViewSceneObject(GVRContext gvrContext, GVRActivity gvrActivity) {
+        this(gvrContext, gvrActivity, "Hello World!");
```
JonShemitz: This seems wrong. Imho, the point of a constructor that doesn't take any initial text is that you will be setting it later. The default text, here, should probably be `""`.

GVRf/Framework/src/org/gearvrf/scene_objects/GVRTextViewSceneObject.java
```diff
@@ -0,0 +1,310 @@
...
+        mTextView.setTextColor(color);
+    }
+
+    /**
+     * Set the text string to be displayed.
+     * 
+     * @param string
+     *            The new text string to be displayed.
+     */
+    public void setText(String str) {
```
JonShemitz: Ooops, an abbreviation. :-(

GVRf/Framework/src/org/gearvrf/scene_objects/GVRTextViewSceneObject.java
```diff
@@ -0,0 +1,310 @@
...
+     * @param color
+     *            The view's background color, in Android {@link Color} format.
+     *            The {@linkplain Color#alpha(int) alpha component} is ignored.
+     */
+    public void setBackgroundColor(int color) {
+        mTextViewContainer.setBackgroundColor(color);
+    }
+
+    /**
+     * Set the view's background with resource drawable.
```
JonShemitz: It doesn't have to be a resource - it could be generated by the app.

GVRf/Framework/src/org/gearvrf/scene_objects/GVRTextViewSceneObject.java
```diff
@@ -0,0 +1,310 @@
...
+     * 
+     * @param drawable
+     *            The view's background.
```
JonShemitz: Should probably specify that `{@code null} will clear the background`.

GVRf/Framework/src/org/gearvrf/scene_objects/GVRTextViewSceneObject.java
```diff
@@ -0,0 +1,310 @@
...
+     */
+    public void setBackGround(Drawable drawable) {
+        mTextViewContainer.setBackground(drawable);
+    }
+
```
JonShemitz: Huh - `setBackground()` but not `getBackground()`?

GVRf/Framework/src/org/gearvrf/scene_objects/GVRTextViewSceneObject.java
```diff
@@ -0,0 +1,310 @@
...
+    /**
+     * Get the current text string.
+     * 
+     * @return The current text string that is being displayed.
+     */
+    public String getText() {
```
JonShemitz: Should just return a `CharSequence`. If the app needs a `String`, it can call `.toString()` itself.

(Yes, I know: I'm Mr Convenience Function, Mr Don't Bloat The App. But the underlying `TextView` has good reasons to return a `CharSequence` and not a `String`, and we should probably follow its lead. There's nothing wrong with adding

    public String getTextString() { return getText().asString(); }

GVRf/Sample/gvrtextviewsample/src/org/gearvrf/gvrtextviewsample/SampleViewManager.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Override
+    public void onInit(GVRContext gvrContext) {
+        sceneObject = new GVRTextViewSceneObject(gvrContext, mActivity);
+        textSize = sceneObject.getTextSize();
+
+        // set the scene object position
+        sceneObject.getTransform().setPosition(0.0f, 0.0f, -2.0f);
+
+        // add the scene object to the scene graph
```
thomasflynn: specifically, this adds it to the camera.  so, when you look around, the text stays in front of you.  i wasn't expecting that and went to look at the code to see if it was intentional :)

GVRf/Framework/src/org/gearvrf/scene_objects/GVRTextViewSceneObject.java
```diff
@@ -0,0 +1,310 @@
...
+     * 
+     * @param gravity
+     *            The gravity of this TextView;
+     */
+    public void setGravity(int gravity) {
+        mTextView.setGravity(gravity);
+    }
+
+    /**
+     * This is actually a callback for GVRDrawFrameListener. It shouldn't get
```
JonShemitz: Writing comments like this is a good cue to introduce a

    private final GVRDrawFrameListener mFrameListener = new GVRDrawFrameListener() {
        @Override
            if (++mCount > REFRESH_INTERVAL) {
                refresh();
                mCount = 0;
            }
        }
    };

GVRf/Sample/gvrtextviewsample/src/org/gearvrf/gvrtextviewsample/SampleViewManager.java
```diff
@@ -0,0 +1,72 @@
...
+        gvrContext.getNextMainScene().getMainCameraRig().getOwnerObject()
+                .addChildObject(sceneObject);
```
thomasflynn: getOwnerObject() is likely not necessary here.

GVRf/Sample/gvrtextviewsample/AndroidManifest.xml
JonShemitz: Dang - at one time I went through and updated all the samples to 19.

GVRf/Sample/gvrtextviewsample/src/org/gearvrf/gvrtextviewsample/SampleViewManager.java
```diff
@@ -0,0 +1,72 @@
...
+
+    }
+
+    @Override
+    public void onStep() {
+        counter++;
+        if (counter % 50 == 0) {
+            int curState = (counter / 50) % 5;
```
thomasflynn: i'm not sure of the intended logic here.  what i do see when running the app are 3 textviews:  one green, one white, one magenta.  from the code, it looks like i should be seeing a blue and a black (against a black background??).  if the blue is there, it's too hard to see.

PR #67, opened 2015-05-22
Custom post effect shaders: add the uniform to have the current eye.

KolorCompany:
It could be useful for the custom post effect shaders to have the
information of which eye it is currently rendering, so add the
information for that.


JonShemitz:
Thanks, @KolorCompany . This looks perfectly plausible. We have a bit of a crunch for a hackathon this weekend, so _may_ not be able to test and merge until next week.

chenchao1407:
Code looks good and the logic is also straight forward. Tested and it didn't break anything. I think it is good to merge.

JonShemitz:
Thanks, @chenchao1407 .

PR #73, opened 2015-05-22
Fixed the bug of resuing mesh with different shaders.

rongguodong:
Now, we generate VBO for different shaders and record them in a map.

rongguodong@hotmail.com

### DIFF
--- a/GVRf/Framework/jni/objects/mesh.cpp
+++ b/GVRf/Framework/jni/objects/mesh.cpp
@@ -167,7 +167,7 @@ void Mesh::generateVAO(Material::ShaderType key) {
     }
 
     GLuint vaoID_ = 0;
-	GLuint triangle_vboID_, vert_vboID_, norm_vboID_, tex_vboID_;
+    GLuint triangle_vboID_, vert_vboID_, norm_vboID_, tex_vboID_;
 
     glGenVertexArrays(1, &vaoID_);
     glBindVertexArray(vaoID_);
@@ -251,10 +251,10 @@ void Mesh::generateVAO(Material::ShaderType key) {
     }
 
     vaoID_map_[key] = vaoID_;
-	triangle_vboID_map_[key] = triangle_vboID_;
-	vert_vboID_map_[key] = vert_vboID_;
-	norm_vboID_map_[key] = norm_vboID_;
-	tex_vboID_map_[key] = tex_vboID_;
+    triangle_vboID_map_[key] = triangle_vboID_;
+    vert_vboID_map_[key] = vert_vboID_;
+    norm_vboID_map_[key] = norm_vboID_;
+    tex_vboID_map_[key] = tex_vboID_;
 
     // done generation
     glBindVertexArray(0);
--- a/GVRf/Framework/jni/objects/mesh.cpp
+++ b/GVRf/Framework/jni/objects/mesh.cpp
@@ -143,11 +143,11 @@ const float* Mesh::getBoundingBoxInfo() {
 }
 
 // generate vertex array object
-void Mesh::generateVAO() {
+void Mesh::generateVAO(Material::ShaderType key) {
 #if _GVRF_USE_GLES3_
     GLuint tmpID;
 
-    if (vaoID_) {
+    if (vaoID_map_.find(key) != vaoID_map_.end()) {
         // already initialized
         return;
     }
@@ -165,6 +165,9 @@ void Mesh::generateVAO() {
         throw error;
         return;
     }
+
+    GLuint vaoID_ = 0;
+	GLuint triangle_vboID_, vert_vboID_, norm_vboID_, tex_vboID_;
 
     glGenVertexArrays(1, &vaoID_);
     glBindVertexArray(vaoID_);
@@ -246,6 +249,12 @@ void Mesh::generateVAO() {
         glEnableVertexAttribArray(it->first);
         glVertexAttribPointer(it->first, 4, GL_FLOAT, 0, 0, 0);
     }
+
+    vaoID_map_[key] = vaoID_;
+	triangle_vboID_map_[key] = triangle_vboID_;
+	vert_vboID_map_[key] = vert_vboID_;
+	norm_vboID_map_[key] = norm_vboID_;
+	tex_vboID_map_[key] = tex_vboID_;
 
     // done generation
     glBindVertexArray(0);
--- a/GVRf/Framework/jni/objects/mesh.h
+++ b/GVRf/Framework/jni/objects/mesh.h
@@ -35,317 +35,317 @@
 namespace gvr {
 class Mesh: public HybridObject {
 public:
-	Mesh() :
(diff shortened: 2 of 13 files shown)


GVRf/Framework/jni/objects/mesh.h
```diff
@@ -30,344 +30,322 @@
...
+
+	// generate VAO
+	void generateVAO(Material::ShaderType key);
+
+	const GLuint getVAOId(Material::ShaderType key) const {
+		auto iterator = vaoID_map_.find(key);
+		return iterator != vaoID_map_.end() ? iterator->second : 0;
+	}
+
+	void cleanUp() {
```
JonShemitz: I think if you move this right after `~Mesh()`, the diff will be much easier to read.

rongguodong:
I have just moved cleanup() to right below ~mesh() following Jon's comment.

I noticed that this function seems to be introduced by the commit of frustum culling. They just simply duplicated all codes in the destructor in a new function cleanup(). I removed the duplication by calling this function in the destructor in this commit, so that it is easier to maintain.

JonShemitz:
@chenchao1407 ?

GVRf/Framework/jni/objects/mesh.cpp
```diff
@@ -166,6 +166,9 @@ void Mesh::generateVAO() {
         return;
     }
 
+    GLuint vaoID_ = 0;
+	GLuint triangle_vboID_, vert_vboID_, norm_vboID_, tex_vboID_;
```
JonShemitz: I don't know if you intend this indenting to signify anything ... if so, it won't survive a Shift+Ctrl+F

GVRf/Framework/jni/objects/mesh.cpp
```diff
@@ -247,6 +250,12 @@ void Mesh::generateVAO() {
         glVertexAttribPointer(it->first, 4, GL_FLOAT, 0, 0, 0);
     }
 
+    vaoID_map_[key] = vaoID_;
+	triangle_vboID_map_[key] = triangle_vboID_;
```
JonShemitz: I may be missing something, but `triangle_vboID_`, `vert_vboID_` and `tex_vboID_` never seem to be set in these diffs.

GVRf/Framework/jni/objects/mesh.h
```diff
@@ -30,344 +30,322 @@
...
+					-1), normalLoc_(-1), texCoordLoc_(-1), have_bounding_box_(
+					false) {
+	}
+
+	~Mesh() {
+		cleanUp();
+	}
+
+	void cleanUp() {
+		std::vector<glm::vec3> vertices;
```
JonShemitz: Most of the differences in this file appear to just be whitespace ... which makes it very hard to see what you really changed.

GVRf/Framework/jni/objects/mesh.cpp
rongguodong: I do not see any indenting on my PC. And it seems Shift+Ctrl+F does not work correctly for CPP files. Maybe my formatter setting has any problems?

GVRf/Framework/jni/objects/mesh.cpp
rongguodong: They are set in the original files, not shown in diff.

GVRf/Framework/jni/objects/mesh.cpp
JonShemitz: Maybe. I’ve had success formatting .h and .cpp files – it’s the *_jni.cpp files that don’t work well.

rongguodong:
I pushed a new commit after doing Ctrl-Shift-F for .cpp and .h files. It seems the original files have different format with mine. That's the reason of so many "blank" changes. :(

rongguodong:
At least, there is no strange indenting in mesh.cpp any more. :)

GVRf/Framework/jni/objects/mesh.h
```diff
@@ -30,344 +30,322 @@
...
-    }
-
-    // generate VAO
-    void generateVAO();
-
-    const GLuint getVAOId() const {
-        return vaoID_;
-    }
-
-    void cleanUp() {
```
thomasflynn: it looks like cleanUp() got deleted.  intentional?

GVRf/Framework/jni/shaders/material/bounding_box_shader.cpp
```diff
@@ -28,65 +28,65 @@
...
+	program_ = new GLProgram(VERTEX_SHADER, FRAGMENT_SHADER);
+	a_position_ = glGetAttribLocation(program_->id(), "a_position");
+	u_mvp_ = glGetUniformLocation(program_->id(), "u_mvp");
 }
 
 BoundingBoxShader::~BoundingBoxShader() {
-    if (program_ != 0) {
-        recycle();
-    }
+	if (program_ != 0) {
```
thomasflynn: not just in this file, but in general, it looks like your tabs are set to 8?

rongguodong:
It turns out that I did not set "use space only" for C++ formatter. Now it should be much better.

BTW: I tried a batch formatting for all Java code, and found out 840+ files are changed (most, if not all, changes are in Java Doc comments). So I have to abandoned that change, and only manually do Ctrl-Shift-F for all touched .cpp and .h files in this commit.

GVRf/Framework/jni/objects/mesh.h
rongguodong: No. It is moved up right below the destructor according to Jon's suggestion.

GVRf/Framework/jni/objects/mesh.h
```diff
@@ -354,8 +329,11 @@ class Mesh: public HybridObject {
     std::map<int, std::string> attribute_vec4_keys_;
 
     // add vertex array object and VBO
-    GLuint vaoID_;
-    GLuint triangle_vboID_, vert_vboID_, norm_vboID_, tex_vboID_;
+    std::map<Material::ShaderType, GLuint> vaoID_map_;
+    std::map<Material::ShaderType, GLuint> triangle_vboID_map_;
+    std::map<Material::ShaderType, GLuint> vert_vboID_map_;
+    std::map<Material::ShaderType, GLuint> norm_vboID_map_;
+    std::map<Material::ShaderType, GLuint> tex_vboID_map_;
```
chenchao1407: Just my minor concern here. As far as I can see, actually only we only need map for vaoIDs since only vaoID has getter which requires key. For all other vboID, we just need some kind of container to store them and finally call glDeleteBuffer with all of them.  In std, map is actually a tree(which may differentiate from hash_map), which may be a bit more expensive then something like a dequeue or vector.

chenchao1407:
Tried with sample cubemap and pickandmove, works as expected with no side effect. I think it is good to go.

GVRf/Framework/jni/objects/mesh.h
JonShemitz: > In std, map is actually a tree(which may differentiate from hash_map), which may be a bit more expensive then something like a dequeue or vector.

More memory consumption, yes. Slower, probably not.

JonShemitz:
>  I think it is good to go.

Thanks, @chenchao1407  - I'll take a closer look than I did yesterday.

PR #76, opened 2015-05-27
Possible Future<GVRMeshEyePointee> API

JonShemitz:
j.shemitz@samsung.com

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
@@ -431,6 +431,40 @@ public class GVRSceneObject extends GVRHybridObject {
     }
 
     /**
+     * Attach a default {@link GVREyePointeeHolder} to the object.
+     * 
+     * The default holder contains a single {@link GVRMeshEyePointee}, which
+     * refers to the {@linkplain GVRMesh mesh} in this scene object's
+     * {@linkplain GVRRenderData render data}. If you need anything more
+     * complicated (such as multiple meshes) use the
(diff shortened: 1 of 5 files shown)


rongguodong:
The code seems fine to me, and it works with Future<GVRMesh> correctly.

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
```diff
@@ -431,6 +431,40 @@ public void attachEyePointeeHolder(GVREyePointeeHolder eyePointeeHolder) {
...
+            return false;
+        }
+
+        GVREyePointeeHolder eyePointeeHolder = new GVREyePointeeHolder(
+                getGVRContext());
+        eyePointeeHolder.addPointee(eyePointee);
+        attachEyePointeeHolder(eyePointeeHolder);
+        return true;
+    }
+
```
chenchao1407: Imho, I am wondering if the method name s/b attachDefaultEyePointeeHolder() which may make more sense.

chenchao1407:
Tried using most of APIs you added here except getMeshEyePointee(Can't think out a use case right now). Works fine. Only a small concern on the method name, and also a little bit concern on those critical section( I may be over-thinking). But I think probably it should be good to go.

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
JonShemitz: Perhaps. I think, though, that overloading the existing `attachEyePointeeHolder()` is a good idea: emphasizes the similarity.

PR #77, opened 2015-05-27
Tune memory management code - eliminate finalize() method.

JonShemitz:
Finalizers are expensive; this update uses PhantomReference<>s
and a ReferenceQueue, which is cheaper.

j.shemitz@samsung.com

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/GVRHybridObject.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRHybridObject.java
@@ -17,78 +17,89 @@ package org.gearvrf;
 
 import java.io.Closeable;
 import java.io.IOException;
+import java.lang.ref.PhantomReference;
+import java.lang.ref.ReferenceQueue;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 import org.gearvrf.utility.Log;
 
-/** Base wrapper class for GVRF C++ classes */
+/**
+ * Root of the GVRF object hierarchy.
+ * 
+ * Descendant classes all have native (JNI) implementations; this base class
+ * manages the native lifecycles.
+ */
 public abstract class GVRHybridObject implements Closeable {
 
     private static final String TAG = Log.tag(GVRHybridObject.class);
 
-    private final GVRContext mGVRContext;
+    /*
+     * Instance fields
+     */
 
+    private final GVRContext mGVRContext;
+    /**
+     * This is not {@code final}: the first call to {@link #close()} sets
+     * {@link #mNativePointer} to 0, so that {@link #close()} can safely be
+     * called multiple times.
+     */
     private long mNativePointer;
 
-    GVRHybridObject(GVRContext gvrContext, long nativePointer) {
+    /*
+     * Constructors
+     */
+
+    /**
+     * Normal constructor
+     * 
+     * @param gvrContext
+     *            The current GVRF context
+     * @param nativePointer
+     *            The native pointer, returned by the native constructor
+     */
+    protected GVRHybridObject(GVRContext gvrContext, long nativePointer) {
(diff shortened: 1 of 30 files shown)


GVRf/Framework/src/org/gearvrf/GVRHybridObject.java
```diff
@@ -185,6 +196,189 @@ public int getUseCount() {
...
+
+    private static class GVRFinalizeThread extends Thread {
+
+        // private static final String TAG = Log.tag(GVRFinalizeThread.class);
+
+        private GVRFinalizeThread() {
+            setName("GVRF Finalize Thread");
+            setPriority(MAX_PRIORITY);
+            start();
+        }
```
chenchao1407: As a GC thread,  imho, I think it is better for you to call setDaemon(true) to make it a daemon thread.

And I am not quite understanding that why you set this thread with highest priority. Actually in my opinion, the GC thread should run with low priority since we don't want it to slow down or even block the other threads.

GVRf/Framework/src/org/gearvrf/GVRHybridObject.java
```diff
@@ -185,6 +196,189 @@ public int getUseCount() {
...
+                    GVRReference reference = (GVRReference) sReferenceQueue
+                            .remove();
+                    reference.close();
+                }
+            } catch (InterruptedException e) {
+                e.printStackTrace();
+            }
+        }
+    }
+
```
chenchao1407: Imho, I think the try catch block should be inside the while loop. 
Like:
while(true){
     try{
        .....
     }catch(InterruptedException e){
          ....
     }
}

Otherwise, if any exception happens, the app will be still running while this GC thread has already stopped working.

chenchao1407:
I think this replacement is very awesome. Just two concerns with Finalizer thread.

GVRf/Framework/src/org/gearvrf/GVRHybridObject.java
JonShemitz: > As a GC thread, imho, I think it is better for you to call setDaemon(true) to make it a daemon thread.

Huh?

    setDaemon (boolean isDaemon)

    Marks this thread as a daemon thread. A daemon thread only runs as long 
    as there are non-daemon threads running. When the last non-daemon 
    thread ends, the runtime will exit. This is not normally relevant to applications
    with a UI.

I don't see how this would be appropriate, here. A GVRf app _is_ an app with a UI, not a background service.

> And I am not quite understanding that why you set this thread with highest priority. Actually in my opinion, the GC thread should run with low priority since we don't want it to slow down or even block the other threads.

This is only going to wake up after GC has freed objects - this typically involves pausing all threads. Since this thread is _mostly_ invoking native `delete`, it seems like a good idea to make this happen as quickly as possible, before (say) low-priority threads allocate more memory.

GVRf/Framework/src/org/gearvrf/GVRHybridObject.java
JonShemitz: > Imho, I think the try catch block should be inside the while loop. 

But then we're entering and exiting a try/catch block every time GC adds a `PhantomReference` to the queue. This way, we only enter the try/catch block once.

> Otherwise, if any exception happens, the app will be still running while this GC thread has already stopped working.

But ... how is that going to happen? When we create the (self-starting) thread on line 216, we don't save any references to it. How are we going to interrupt a thread we have no reference to?

(We don't have to worry about the thread getting garbage collected, though: threads count as GC roots, just like static fields do.)

chenchao1407:
Talked to Jon and I think I made mistake on the thread's concerns. Actually I don't see difference between setDaemon/setPriority or not. Tested with very stressful case and still works fine.  I just found another issue probably not related to this pull request. Right now I think it is good to go.

chenchao1407:
Hi @JonShemitz . I think you may have to do the manual rebase on this pull request.

JonShemitz:
Yeah. Thanks.

PR #79, opened 2015-05-29
Fix a 'mandelbug' in async loading.

JonShemitz:
The root cause of the bug was using the same GVRAndroidResource in
multiple load calls. This is now illegal, as it is problematic in
another way, as well: If the later request(s) can't be consolidated
with the earlier request(s) (because the earlier request(s) have
already been completed) the later request(s) will reload the
resource ... using a closed stream.

The proximate cause of the bug was the way that the
private FutureResource class (in GVRAsynchronousResourceLoader)
was using the user-supplied GVRAndroidResource for
synchronization. This has now been corrected, so that each
FutureResource uses a unique, private Object for synchronization.

Also cuts a bit of log spam, and fixes one or two minor issues
I ran into while tracking down the mandelbug.

j.shemitz@samsung.com

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/GVRAndroidResource.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRAndroidResource.java
@@ -86,12 +86,7 @@ public class GVRAndroidResource {
      *             File doesn't exist, or can't be read.
      */
     public GVRAndroidResource(File file) throws FileNotFoundException {
-        stream = new MarkingFileInputStream(file);
-        debugState = DebugStates.OPEN;
-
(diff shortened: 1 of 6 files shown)


thomasflynn:
was finally able to sit down long enough without interruption to review the code.  +1.  haven't verified.

chenchao1407:
Can't find any real concern on the codes. And everything runs fine.

JonShemitz:
> Can't find any real concern on the codes. And everything runs fine.

Then, @chenchao1407 , why not merge it? @thomasflynn  already +1-ed it.

PR #101, opened 2015-05-30
Add a method to force monoscopic rendering.

thomasflynn:
Add a method to force monoscopic rendering per request in Issue 75.

tom.flynn@samsung.com

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/GVRActivity.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRActivity.java
@@ -45,7 +45,7 @@ public class GVRActivity extends VrActivity {
 
     private GVRViewManager mGVRViewManager = null;
     private GVRCamera mCamera;
-    private boolean forceMonoscopic = false;
+    private boolean mForceMonoscopic = false;
 
     static {
         System.loadLibrary("gvrf");
@@ -124,7 +124,7 @@ public class GVRActivity extends VrActivity {
      */
     public void setScript(GVRScript gvrScript, String distortionDataFileName) {
         if (getRequestedOrientation() == ActivityInfo.SCREEN_ORIENTATION_LANDSCAPE) {
-            if (isVrSupported() && !forceMonoscopic) {
+            if (isVrSupported() && !mForceMonoscopic) {
                 mGVRViewManager = new GVRViewManager(this, gvrScript,
                         distortionDataFileName);
             } else {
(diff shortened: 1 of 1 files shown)


GVRf/Framework/src/org/gearvrf/GVRActivity.java
```diff
@@ -136,6 +137,24 @@ public void setScript(GVRScript gvrScript, String distortionDataFileName) {
         }
     }
 
+    /**
+     * Sets whether to force rendering to be single-eye, monoscopic view.
+     *
```
chenchao1407: I think we may need to either add JavaDoc here to specify this setForceMonoscopic only takes effect before you call setScript or throws an exception like UnsupportedOperationException if someone calls this after setScript().

chenchao1407:
Codes looks almost fine and works as expected. Only a small concern on the JavaDoc comment.

thomasflynn:
added the javadoc.

GVRf/Framework/src/org/gearvrf/GVRActivity.java
```diff
@@ -136,6 +137,24 @@ public void setScript(GVRScript gvrScript, String distortionDataFileName) {
...
+    public void setForceMonoscopic(boolean force) {
+        forceMonoscopic = force;
+    }
+
+    /**
+     * Returns whether a monoscopic view will be created during {@linkplain setScript setScript()}.
+     */
+    public boolean getForceMonoscopic() {
+        return forceMonoscopic;
+    }
```
chenchao1407: Actually I think the logic here is also incorrect. What you should return is forceMonoscopic || !isVrSupported()

thomasflynn:
clarified the javadoc in getForceMonoscopic.

chenchao1407:
OK. I think it is good to me now.

JonShemitz:
I’ll review it as soon as I can.

GVRf/Framework/src/org/gearvrf/GVRActivity.java
```diff
@@ -45,6 +45,7 @@
 
     private GVRViewManager mGVRViewManager = null;
     private GVRCamera mCamera;
+    private boolean forceMonoscopic = false;
```
JonShemitz: Much as I dislike the `'m'` prefix, I think we should be internally consistent: when you add a new field to a block of fields, you probably should honor the local style.

GVRf/Framework/src/org/gearvrf/GVRActivity.java
```diff
@@ -136,6 +137,25 @@ public void setScript(GVRScript gvrScript, String distortionDataFileName) {
         }
     }
 
+    /**
+     * Sets whether to force rendering to be single-eye, monoscopic view.
+     *
+     * @param force
+     *            If true, will create a GVRMonoscopicViewManager when {@linkplain setScript setScript()} is called.  If false, will proceed to auto-detect whether the device supports VR rendering and choose the appropriate ViewManager.  This call will only an effect if it is called before {@linkplain setScript setScript()}.
```
JonShemitz: {@link syntax is off, and text is not auto-formatted.

     *            If true, will create a GVRMonoscopicViewManager when
     *            {@linkplain #setScript(GVRScript, String) setScript()} is
     *            called. If false, will proceed to auto-detect whether the
     *            device supports VR rendering and choose the appropriate
     *            ViewManager. This call will only have an effect if it is
     *            called before {@linkplain #setScript(GVRScript, String)
     *            setScript()}.

GVRf/Framework/src/org/gearvrf/GVRActivity.java
```diff
@@ -136,6 +137,25 @@ public void setScript(GVRScript gvrScript, String distortionDataFileName) {
...
+     *
+     */
+    public void setForceMonoscopic(boolean force) {
+        forceMonoscopic = force;
+    }
+
+    /**
+     * Returns whether a monoscopic view was asked to be forced during {@linkplain setScript setScript()}.
```
JonShemitz: Another invalid {@linkplain ...

thomasflynn:
ok, renamed variable and fixed formatting and javadoc.

GVRf/Framework/src/org/gearvrf/GVRActivity.java
```diff
@@ -136,6 +137,32 @@ public void setScript(GVRScript gvrScript, String distortionDataFileName) {
...
 
+    /**
+     * Sets whether to force rendering to be single-eye, monoscopic view.
+     * 
+     * @param force
+     *            If true, will create a GVRMonoscopicViewManager when
+     *            {@linkplain setScript setScript()} is called. If false, will
+     *            proceed to auto-detect whether the device supports VR
+     *            rendering and choose the appropriate ViewManager. This call
+     *            will only an effect if it is called before
```
JonShemitz: `will only an effect` s/b `will only have an effect`, but I think most people can deal with this sort of typo.

Issue #81, opened 2015-05-29
gvr-sample showing green


thomasflynn:
The gvr-sample was just showing green on a Nexus 5 in monoscopic mode.

PR #103, opened 2015-06-01
Fixed Issue #81 (gvr-sample showing green).

rongguodong:
We need to call reset() for the stream after setInSampleSize.

rongguodong@hotmail.com


chenchao1407:
Code logic looks good and straight forward. Verified with Nexus 5.

Issue #83, opened 2015-05-29
Add lighting.


thomasflynn:
Add initial support for lighting.

thomasflynn:
this is done as of Version 3.0

PR #104, opened 2015-06-01
Litshader

rongguodong:
Create a lit shader with texture. (issue #83)

rongguodong@hotmail.com


GVRf/Framework/jni/objects/components/render_data_jni.cpp
```diff
@@ -108,152 +112,158 @@ Java_org_gearvrf_NativeRenderData_setDrawMode(
...
 }
 
 JNIEXPORT jint JNICALL
 Java_org_gearvrf_NativeRenderData_getRenderMask(JNIEnv * env,
-        jobject obj, jlong jrender_data) {
-    RenderData* render_data = reinterpret_cast<RenderData*>(jrender_data);
-    return render_data->render_mask();
+    jobject obj, jlong jrender_data) {
+RenderData* render_data = reinterpret_cast<RenderData*>(jrender_data);
+return render_data->render_mask();
```
chenchao1407: The formatting here looks weird to me. It seems that it is because your eclipse can't parse codes well if there shows some syntax errors because of problematic CDT configuration.

GVRf/Sample/gvr-litshader/.project
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>gvrcubemap</name>
```
chenchao1407: I think you copied the project from gvrcubemap and then modified it. This property needs to be updated otherwise Eclipse will consider this one is the "same" as gvrcubemap.

chenchao1407:
Tried and worked as expected. There is only concern on the sample project property and the formatting in jni codes.

Also, there may be another concern like right now only lit shader supports GVRLight, but every renderdata with any type of material can do setLight. I think we may need to specify this in JavaDoc or do some kind of restriction like UnSupportedOperationException in GVRRenderData especially when our default shader currently is UnlitShader.

rongguodong:
Fixed the name in .project.

For format of JNI file, do you have any suggestions on how to fix it?

thomasflynn:
A couple design comments:
There should be an api call to turn the light on or off.
Position could perhaps be set via GVRTransform.  Thus allowing lights to be animated with our animation system.

thomasflynn:
One more:
if we were to add support for more than 1 light in the future (likely), would that require changes to the APIs?

tparisi:
Guys I hope you're looking at Three.js for inspiration on scene graph basics like this. +1 please have multiple lights, ability to turn on/off, and the canonical four light types. Spots are the hardest and most computationally expensive; also they are probably the lowest priority.

thomasflynn:
I'm drawing from my experience working on scene graphs over the years.  From working at SGI on the Performer team and open sourcing Open Inventor to building a scene graph at a startup with Don Burns (the original architect of Open Scene Graph) a few years ago.  For this project, there will be some inspiration all of the above.  It'll take some time to build a world-leading scene graph for VR, but we'll get there.  

And yes, we'll need the different types of lights and ability to turn them on/off.  Thus my above comments with the check-in :)  I do want to make sure that whatever is checked-in for short-term needs do not preclude longer-term goals.

tparisi:
Awesome... sounds like you're on it! thx

chenchao1407:
Hi @rongguodong 

Did you try to fix the CDT issue?

rongguodong:
No. I did not find a good way to fix it yet. If you have ideas, please let me know.

chenchao1407:
Maybe I can do the merge firstly and then I can do the reformatting for you since CDT at my side is good to do the ctrl+alt+f.

JonShemitz:
> Maybe I can do the merge firstly

I haven't reviewed this - I'm under the impression that @thomasflynn wants the API reworked.

thomasflynn:
yeah, don't merge this.  I do want Guodong to revise the api.  He's currently fixing some bugs elsewhere.   when he's done with those, he'll come back to this check-in.

rongguodong:
Sure. I will work on this right after I fixed the bug about CubeMap+pure GVRF.

rongguodong:
Pushed a new changes to enable/disable lighting with ubershader.

rongguodong:
Added query methods GVRLight.isEnabled() and GVRRenderData.isLightEnabled().

GVRf/Framework/jni/objects/components/render_data_jni.cpp
```diff
@@ -108,152 +120,172 @@ Java_org_gearvrf_NativeRenderData_setDrawMode(
 
 JNIEXPORT jlong JNICALL
 Java_org_gearvrf_NativeRenderData_ctor(JNIEnv * env,
-        jobject obj) {
-    return reinterpret_cast<jlong>(new RenderData());
+    jobject obj) {
+return reinterpret_cast<jlong>(new RenderData());
 }
```
chenchao1407: formatting here is not correct.

rongguodong:
Fixed the bug of switching from lighting-on to lighting-off, and rebased with the top of master.

chenchao1407:
Fix the problem I found, good to merge at my side. Is there any else objection?

jason2kim:
Small things but had to point out, 

Local variable sphere removed and replaced by rotateObject.

Unused:  import android.R.bool; 

Indentations in
Java_org_gearvrf_NativeShaderManager_addCustomShader()
Java_org_gearvrf_NativeShaderManager_getCustomShader()
Java_org_gearvrf_NativeShaderManager_ctor()

Unnecessary initialization for Java member variables, including mGVRContext = null;

jason2kim:
Indentations in the codes of light_jni.cpp, too.
Otherwise, I think the functionality and the Uber shader structure look good for me at this moment.

rongguodong:
I have just pushed a new commit addressing issues pointed out by Jason.

For the JNI code, since my Eclipse treats those code as grammatically wrong, Ctrl-Shift-F automatically make them no indent. I have manually added indent to them. However, we should figure out a way to fix Eclipse and thus let Ctrl-Shift-F works.

thomasflynn:
any further comments jason?

jason2kim:
No

PR #106, opened 2015-06-01
Stats2

thomasflynn:
Putting the infrastructure in place for displaying frame stats.  Currently just displaying number of draw calls and number of triangles.  In later check-ins, we will add application time, traversal time, and draw time.

tom.flynn@samsung.com

### DIFF
--- a/GVRf/Framework/jni/objects/scene.h
+++ b/GVRf/Framework/jni/objects/scene.h
@@ -27,6 +27,7 @@
 
 #include "objects/hybrid_object.h"
 #include "components/camera_rig.h"
+#include "engine/renderer/renderer.h"
 
 namespace gvr {
 class SceneObject;
@@ -57,6 +58,10 @@ public:
     void set_occlusion_culling( bool occlusion_flag){ occlusion_flag_ = occlusion_flag; }
     bool get_occlusion_culling(){ return occlusion_flag_; }
 
+    void resetStats() { if(!statsInitialized) { Renderer::initializeStats(); } Renderer::resetStats(); }
+    int getNumberDrawCalls() { return Renderer::getNumberDrawCalls(); }
+    int getNumberTriangles() { return Renderer::getNumberTriangles(); }
+
 private:
     Scene(const Scene& scene);
     Scene(Scene&& scene);
@@ -70,6 +75,7 @@ private:
     int dirtyFlag_;
     bool frustum_flag_;
     bool occlusion_flag_;
+    bool statsInitialized = false;
 
 };
 
--- a/GVRf/Framework/jni/objects/scene.h
+++ b/GVRf/Framework/jni/objects/scene.h
@@ -58,9 +58,19 @@ public:
     void set_occlusion_culling( bool occlusion_flag){ occlusion_flag_ = occlusion_flag; }
     bool get_occlusion_culling(){ return occlusion_flag_; }
 
-    void resetStats() { if(!statsInitialized) { Renderer::initializeStats(); statsInitialized = true; } Renderer::resetStats(); }
-    int getNumberDrawCalls() { return Renderer::getNumberDrawCalls(); }
-    int getNumberTriangles() { return Renderer::getNumberTriangles(); }
+    void resetStats() {
+        if (!statsInitialized) {
+            Renderer::initializeStats();
+            statsInitialized = true;
+        }
+        Renderer::resetStats();
+    }
+    int getNumberDrawCalls() {
+        return Renderer::getNumberDrawCalls();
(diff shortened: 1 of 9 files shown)


chenchao1407:
Hi @thomasflynn 
I think you may have to do the manual rebase before I can review this one.

thomasflynn:
ok, i think i rebased correctly.

GVRf/Framework/jni/objects/scene_jni.cpp
```diff
@@ -88,4 +100,29 @@ Java_org_gearvrf_NativeScene_setOcclusionQuery(JNIEnv * env,
...
+    Scene* scene = reinterpret_cast<Scene*>(jscene);
+    return scene->getNumberDrawCalls();
+}
+
+
+JNIEXPORT int JNICALL
+Java_org_gearvrf_NativeScene_getNumberTriangles(JNIEnv * env,
+        jobject obj, jlong jscene) {
+    Scene* scene = reinterpret_cast<Scene*>(jscene);
+    scene->getNumberTriangles();
```
chenchao1407: I think you missed the return statement here.(I know it will compile in some versions of ndkbuild..)

GVRf/Framework/jni/objects/scene_jni.cpp
```diff
@@ -88,4 +100,29 @@ Java_org_gearvrf_NativeScene_setOcclusionQuery(JNIEnv * env,
...
+    Scene* scene = reinterpret_cast<Scene*>(jscene);
+    return scene->getNumberDrawCalls();
+}
+
+
+JNIEXPORT int JNICALL
+Java_org_gearvrf_NativeScene_getNumberTriangles(JNIEnv * env,
+        jobject obj, jlong jscene) {
+    Scene* scene = reinterpret_cast<Scene*>(jscene);
+    return scene->getNumberTriangles();
```
thomasflynn: ok, added it.  interesting thing though, it actually worked without it.  ??  maybe it just happened to be at the right place on the stack?  weird.

GVRf/Framework/src/org/gearvrf/GVRScene.java
```diff
@@ -152,6 +153,57 @@ public void setFrustumCulling(boolean flag) {
...
+        } else if(!enabled && mStatsConsole != null) {
+            mStatsConsole.setEyeMode(GVRConsole.EyeMode.NEITHER_EYE);
+        }
+    }
+
+    void resetStats() {
+        if(mStatsEnabled) {
+            mStatsConsole.clear();
+            NativeScene.resetStats(getPtr());
+        }
```
chenchao1407: One more small concern is like you should use getNative() instead of getPtr() since getPtr() is an old version API.

GVRf/Framework/src/org/gearvrf/GVRScene.java
```diff
@@ -152,6 +153,66 @@ public void setFrustumCulling(boolean flag) {
...
+        return mStatsEnabled;
+    }
+
+    /**
+     * Set whether to enable display of stats for this scene.
+     *
+     * @param enabled
+     *     Flag to indicate whether to enable display of stats.
+     */
+    public void setStatsEnabled(boolean enabled) {
```
thomasflynn: ok, changed the way this works to match what you suggested.  will have to investigate the performance problem later today.

thomasflynn: uploaded new change that should improve performance.

GVRf/Framework/jni/objects/scene.h
```diff
@@ -57,6 +58,10 @@ class Scene: public HybridObject {
     void set_occlusion_culling( bool occlusion_flag){ occlusion_flag_ = occlusion_flag; }
     bool get_occlusion_culling(){ return occlusion_flag_; }
 
+    void resetStats() { if(!statsInitialized) { Renderer::initializeStats(); statsInitialized = true; } Renderer::resetStats(); }
```
JonShemitz: I can't imagine that this line would survive a Shift+Ctrl+F ...

JonShemitz:
Looks fine, outside of one long, single-line statement that will almost certainly wrap next time someone does a Shift+Ctrl+F.

@chenchao1407 , have you run this?

chenchao1407:
@JonShemitz Just ran it last night without recent two new commits. Will verify it again as soon as I finish my current stuff.

chenchao1407:
Verified. Performance actually gets improved by new commit though I still can see performance degradation while stats is on. Anyway, I think at this point this is acceptable.

Issue #46, opened 2015-05-14
GVRTransform::getRotationPitch - Result is NaN


santhyago:
We think that **pitch** value must be **between 0 and 360**, but the method is returning **NaN**.

```
    mSceneObject.getTransform().setRotation(1.0f,
    BoundsValues.getFloatList().get(0),
    BoundsValues.getFloatList().get(0),
    BoundsValues.getFloatList().get(0));

    float pitch = mSceneObject.getTransform().getRotationPitch();
    boolean result = (pitch >= 0) && (pitch <= 360);
    assertTrue("Pitch 1: " + pitch, result);
```
The first time, **pitch** gets a *"valid"* value (*1.63E-43*, I don't know if this is a valid value, but is in the range).
```
    mSceneObject.getTransform().setRotation(1.0f,
    BoundsValues.getFloatList().get(1),
    BoundsValues.getFloatList().get(1),
    BoundsValues.getFloatList().get(1));

    pitch = mSceneObject.getTransform().getRotationPitch();
    result = (pitch >= 0) && (pitch <= 360);
    assertTrue("Pitch 2: " + pitch, result);
```
The second time, with these values, **pitch** gets a **NaN** as value.

JonShemitz:
>     mSceneObject.getTransform().setRotation(1.0f,
    BoundsValues.getFloatList().get(0),
    BoundsValues.getFloatList().get(0),
    BoundsValues.getFloatList().get(0));
    // ...
    mSceneObject.getTransform().setRotation(1.0f,
    BoundsValues.getFloatList().get(1),
    BoundsValues.getFloatList().get(1),
    BoundsValues.getFloatList().get(1));

This is probably a genuine bug - but what are the values returned by `BoundsValues.getFloatList().get(0)` and `BoundsValues.getFloatList().get(1)`? Please try to give us self-contained bug reports that we can reproduce and triage.

santhyago:
```
private static final float UNPICKED_COLOR_R = 0.7f;
private static final float UNPICKED_COLOR_G = 0.7f;
private static final float UNPICKED_COLOR_B = 0.7f;
private static final float UNPICKED_COLOR_A = 1.0f;
```

```
BoundsValues.getFloatList().get(0) = Float.MIN_VALUE = 1.4E-45
BoundsValues.getFloatList().get(1) = Float.MAX_VALUE = 3.4028235E38
```

```
mSceneObject = getColorBoard(1.0f, 1.0f);
```

```
[...]

PR #108, opened 2015-06-02
Fix the bug for issue #46 (getRotationPitch returns NaN).

rongguodong:
rongguodong@hotmail.com

### DIFF
--- a/GVRf/Framework/jni/objects/components/transform.cpp
+++ b/GVRf/Framework/jni/objects/components/transform.cpp
@@ -13,7 +13,6 @@
  * limitations under the License.
  */
 
-
 /***************************************************************************
  * Containing data about how to position an object.
  ***************************************************************************/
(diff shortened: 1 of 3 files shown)


GVRf/Framework/jni/objects/components/transform.cpp
```diff
@@ -43,6 +42,16 @@ void Transform::invalidate() {
             (*it)->transform()->invalidate();
         }
     }
+    // scale rotation_ if needed to avoid overflow
```
JonShemitz: Looks like you do this test on every call to `invalidate()`, regardless of whether the `model_matrix_.isValid()` or not.

Seems like it would be better to do this in `Transform::getModelMatrix()`, if and only if the `Transform` has been invalidated.

chenchao1407: I am not quite sure but my concern is that if it is possible that when we call getModelMatrix(), actually the over/under flow has already happened by some calls like setRotation or even setModelMatrix and we don't have a way to know at getModelMatrix().

However I think to do the check on each invalidate is a little bit waste of performance. If it is possible to find a way that only when we change the rotation_ then we do such check.

rongguodong:
Now we only do scaling when rotation_ is updated.

chenchao1407:
Codes look good to me.
@santhyago Can you help check again with your test codes?

PR #109, opened 2015-06-03
Add resource caching.

JonShemitz:
Adds a simple Map<> from GVRAndroidResource to
WeakReference<GVRHybridObject>. Adds hooks to
all the resource loading methods to use the
cache. Now, loading a resource that's still
in memory should simply yield the existing
resource, even if that load was finished before
the new request.

Also closes all 'redundant' GVRAndroidResource-s,
to minimize the number of open file handles.

There is some ugliness here, because of the 
asynchronous package: the cache is often read
in the org.gearvrf package, but mostly populated
in the org.gearvrf.asynchronous package. I started
to move the asynchronous classes into the 'root'
package, but this made for a *lot* of changes: in 
the end, I put the new ResourceCache class in the
utility directory, and passed pointers to cache 
instances down into the async package.

j.shemitz@samsung.com

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/GVRContext.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRContext.java
@@ -32,9 +32,9 @@ import org.gearvrf.asynchronous.GVRCompressedTexture;
 import org.gearvrf.asynchronous.GVRCompressedTextureLoader;
 import org.gearvrf.periodic.GVRPeriodicEngine;
 import org.gearvrf.utility.Log;
+import org.gearvrf.utility.ResourceCache;
 
 import android.app.Activity;
-import android.app.Application;
 import android.content.Context;
 import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
@@ -174,10 +174,17 @@ public abstract class GVRContext {
      * @since 1.6.2
      */
     public GVRMesh loadMesh(GVRAndroidResource androidResource) {
-        GVRAssimpImporter assimpImporter = GVRImporter.readFileFromResources(
-                this, androidResource);
-        return assimpImporter.getMesh(0);
+        GVRMesh mesh = sMeshCache.get(androidResource);
+        if (mesh == null) {
+            GVRAssimpImporter assimpImporter = GVRImporter
+                    .readFileFromResources(this, androidResource);
+            mesh = assimpImporter.getMesh(0);
+            sMeshCache.put(androidResource, mesh);
+        }
+        return mesh;
     }
+
+    private final static ResourceCache<GVRMesh> sMeshCache = new ResourceCache<GVRMesh>();
 
     /**
      * Loads a mesh file, asynchronously, at a default priority.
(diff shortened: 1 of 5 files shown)


thomasflynn:
code looks fine as far as i can tell.  is there an existing app this was tested against?

JonShemitz:
> is there an existing app this was tested against?

I just added some `onStep()` code to Solar System to reload the sun texture or the sphere object: in both cases, it 'found' the existing resource.

chenchao1407:
Tested, worked as expected.  Just some concerns here:
1. For the asynchronous loading, right now it only supports texture by default without the support of mesh. It seems a little weird to me because right now in the current constructors of GVRSceneObject. It seems to me that if we want to make use of asynchronous calls when we load GVRScene, both mesh and texture need to be instances of Future. Currently there is no option that we just make one of them asynchronous and the other one is not. What I mean is that, if we want to make use of asynchronous loading, then it means cache only can work for texture. Or if we want the cache work for both texture and mesh, we have to give up the asynchronous loading.
2. I think it may be due to current package structure, but imho, I think cache system should be something which is behind the scene. What I mean is that we can expose some API calls like if developers wants to use cache in certain circumstance or what size they want to set. It seems a little weird to me that we expose the whole class to developer.
3. Actually in my test, it works. I mean I set different textures and meshes and the cache works as I expected. But I also can see that the Java heap size increase a lot without limitation. Do we need to set up some kind of APIs like how many items they want to put into cache and some mechanism(e.x. LRU cache) to help the system to expire something if current cache is out of size?

Sincerely
-Chao Chen

JonShemitz:
> 1. For the asynchronous loading, right now it only supports texture by default without the support of mesh. It seems a little weird to me because right now in the current constructors of GVRSceneObject. It seems to me that if we want to make use of asynchronous calls when we load GVRScene, both mesh and texture need to be instances of Future. Currently there is no option that we just make one of them asynchronous and the other one is not. What I mean is that, if we want to make use of asynchronous loading, then it means cache only can work for texture. Or if we want the cache work for both texture and mesh, we have to give up the asynchronous loading.

The `Future<>` APIs are meant to cover the most common cases; you can always use the underlying callback API if you want/need more control.

But your comment seems to show a confusion: caching is not connected to async loading. All the resource loading calls (except the original `GVRAndroidResource[]` cube map API) use the cache. The cache contains loaded resources: it has nothing to do with async request consolidation.

> 2. I think it may be due to current package structure, but imho, I think cache system should be something which is behind the scene. What I mean is that we can expose some API calls like if developers wants to use cache in certain circumstance or what size they want to set. It seems a little weird to me that we expose the whole class to developer.

Yes, the cache code should be private. Yes, it is not because I did not want to move the whole `asynchronous` package to the root. No, I don't think it matters much - the cache _instances_ are private.

> 3. Actually in my test, it works. I mean I set different textures and meshes and the cache works as I expected. But I also can see that the Java heap size increase a lot without limitation. Do we need to set up some kind of APIs like how many items they want to put into cache and some mechanism(e.x. LRU cache) to help the system to expire something if current cache is out of size?

No, we do not. The cache only contains otherwise-referenced resources. When the last hard-reference goes away, so does the cache entry.

chenchao1407:
Just did a further verification. The only issue is that in my test, there will be much more unreachable objects in the heap and which will make the cache mechanism looks consuming much more memory than normal. However after 2 or 3 GC cycle, the memory level(both native and Java side) will be even lower than the normal one. So I think it is good to go.

PR #112, opened 2015-06-04
Fix picking.

JonShemitz:
Frustrum culling broke eye picking, in at least
some scenarios. This fixs it.

j.shemitz@samsung.com


thomasflynn:
it's amazing things worked as well as they did without this.

Issue #98, opened 2015-05-29
Simplify picking


thomasflynn:
picking currently requires setting up a bounding Mesh for a object, creating an EyePointee, add the EyePointee to an EyePointeeHolder and then adding the EyePointeeHolder to the SceneObject.  While flexible, it is a lot for a developer to go through to have the ability to pick.  We saw this sort of trouble during the recent hackathon.  We can provide a simple enablePicking() api that will construct a default bounding box for the mesh, and add the appropriate EyePointee and EyePointeeHolders.

DanielNogueiraAndroid:
Excellent improvement !!!

PR #113, opened 2015-06-04
Add a picking-enabled property to scene objects.

JonShemitz:
Per https://github.com/Samsung/GearVRf/issues/98

j.shemitz@samsung.com

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
@@ -483,6 +483,42 @@ public class GVRSceneObject extends GVRHybridObject {
     }
 
     /**
+     * Simple, high-level API to enable or disable eye picking for this scene
+     * object.
+     * 
+     * The {@linkplain #attachEyePointeeHolder(GVREyePointeeHolder) low-level
+     * API} gives you a lot of control over eye picking, but it does involve an
+     * awful lot of details. Since most apps are just going to use the
+     * {@linkplain #attachEyePointeeHolder() simple API} anyhow, this method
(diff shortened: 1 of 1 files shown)


GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
```diff
@@ -483,6 +483,42 @@ public GVREyePointeeHolder getEyePointeeHolder() {
...
+     *            Should eye picking 'see' this scene object?
+     * 
+     * @since 2.0.2
+     */
+    public void setPickingEnabled(boolean enabled) {
+        if (enabled != getPickingEnabled()) {
+            if (enabled) {
+                detachEyePointeeHolder();
+            } else {
+                attachEyePointeeHolder();
```
thomasflynn: ah, cool, didn't know we had a method for attaching a default eyepointee.  though i'm disappointed that it uses the full mesh rather than the bounding box mesh.

thomasflynn:
looks good to me.

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
chenchao1407: Hi Jon.

I think the logic is reverse here. I mean if enabled we do attach otherwise we do detach.

Sincerely
-Chao Chen

JonShemitz: > I think the logic is reverse here. I mean if enabled we do attach otherwise we do detach.

You're right! When I added the 'if-any-change' test, I mixed up the 'current state' and the 'desired state'.

JonShemitz:
Perhaps this can be merged?

PR #115, opened 2015-06-05
Add stats for gl draw time.

thomasflynn:
Added stats for gl draw time using the disjoint_timer_query extension.

tom.flynn@samsung.com

### DIFF
--- a/GVRf/Framework/jni/engine/renderer/renderer.cpp
+++ b/GVRf/Framework/jni/engine/renderer/renderer.cpp
@@ -39,14 +39,57 @@ namespace gvr {
 
 static int numberDrawCalls;
 static int numberTriangles;
+static float drawTime;
+static float avgDrawTime;
+static bool monoscopic = false;
+static bool disjointTimerSupported = false;
+static int currentFrameQuery;
+static int lastFrameQuery;
+static GLuint queries[2][2];
+static GLuint queriesAvailable;
+static GLint queriesDisjointOccurred;
+static GLuint64 timeElapsed;
+static GLuint64 frameCount;
+
+#ifndef GL_TIME_ELAPSED_EXT
+#define GL_TIME_ELAPSED_EXT 0x88BF
+#define GL_GPU_DISJOINT_EXT 0x8FBB
+extern "C" {
+typedef void (*PFNGLGETQUERYOBJECTUI64VEXTPROC) (GLuint id, GLenum pname, GLuint64 *params);
+PFNGLGETQUERYOBJECTUI64VEXTPROC glGetQueryObjectui64vEXT__;
+}
+#endif
 
 void Renderer::initializeStats(){
-    // TODO: this function will be filled in once we add draw time stats
+    currentFrameQuery = 0;
+    lastFrameQuery = 0;
+    timeElapsed = 0;
+    frameCount = 0;
+    drawTime = 0.0f;
+    avgDrawTime = 0.0f;
+
+    const GLubyte *extensions = glGetString(GL_EXTENSIONS);
+    char *exist = strstr((char *)extensions, "GL_EXT_disjoint_timer_query");
+    if(!strncmp(exist, "GL_EXT_disjoint_timer_query", 26)) {
+        disjointTimerSupported = true;
+    }
+
+    if(disjointTimerSupported) {
+        glGenQueries(2, queries[0]);
+        glGenQueries(2, queries[1]);
+
+        glGetQueryObjectui64vEXT__ = (PFNGLGETQUERYOBJECTUI64VEXTPROC) eglGetProcAddress("glGetQueryObjectui64vEXT");
+    }
 }
 
 void Renderer::resetStats(){
     numberDrawCalls = 0;
     numberTriangles = 0;
+    queriesAvailable = 0;
+    queriesDisjointOccurred = 0;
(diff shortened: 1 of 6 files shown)


chenchao1407:
Hi @thomasflynn 

I think you have to do the manual rebase for this pull request.

Thanks in advance.
Sincerely
-Chao Chen

thomasflynn:
ok, i'll have to do it later tonight.

thomasflynn:
ok, should be rebased now.

GVRf/Framework/jni/engine/renderer/renderer.cpp
```diff
@@ -39,14 +39,57 @@ namespace gvr {
...
+    currentFrameQuery = 0;
+    lastFrameQuery = 0;
+    timeElapsed = 0;
+    frameCount = 0;
+    drawTime = 0.0f;
+    avgDrawTime = 0.0f;
+
+    const GLubyte *extensions = glGetString(GL_EXTENSIONS);
+    char *exist = strstr((char *)extensions, "GL_EXT_disjoint_timer_query");
+    if(!strncmp(exist, "GL_EXT_disjoint_timer_query", 26)) {
```
JonShemitz: I'm having trouble figuring out what this code means to do. First, it calls `strstr()` to get a pointer to the start of a substring. Then it passes that pointer - which may be `NULL` - to `strncmp()` with a `size_t n` argument of 26.

Why 26? I mean 1) I count 27 chars in `"GL_EXT_disjoint_timer_query"` and 2) why a magic number?

What is the point of the `strncmp()` call?

GVRf/Framework/jni/engine/renderer/renderer.cpp
```diff
@@ -57,6 +100,62 @@ int Renderer::getNumberTriangles(){
...
+    float gpuTime = 0.0f;
+    glGetQueryObjectuiv(queries[eye][lastFrameQuery], GL_QUERY_RESULT_AVAILABLE, &queriesAvailable);
+    if(queriesAvailable == 0) {
+        return gpuTime;
+    }
+
+    glGetIntegerv(GL_GPU_DISJOINT_EXT, &queriesDisjointOccurred);
+    if(!queriesDisjointOccurred) {
+        glGetQueryObjectui64vEXT__(queries[eye][lastFrameQuery], GL_QUERY_RESULT, &timeElapsed);
+        gpuTime = (float) ((double)timeElapsed / 1000000.0);
```
chenchao1407: It doesn't work on my Note4. Tried some further debug, found that the timeElapsed here always return 0.

GVRf/Framework/jni/engine/renderer/renderer.cpp
```diff
@@ -74,6 +173,11 @@ void Renderer::renderCamera(Scene* scene, Camera* camera, int framebufferId,
     numberDrawCalls = 0;
     numberTriangles = 0;
 
+    int currentEye = camera->render_mask() - 1;
```
JonShemitz: This makes assumptions about `render_mask()` - might be better to hide that in a `camera->get_eye()` call.

GVRf/Framework/src/org/gearvrf/asynchronous/AsyncBitmapTexture.java
```diff
@@ -217,9 +217,13 @@ private static Point getScreenSize(Context context, Point p) {
     private static void onGlInitialization() {
         if (glUninitialized) {
             int[] size = new int[] { -1 };
+            int errorCode = glGetError();
+            while(errorCode != GL_NO_ERROR) {
```
JonShemitz: * This loop isn't auto-formatted
* More importantly, what is the point of this loop? According to the OpenGL ES 3.0 Programming Guide, `glGetError()` "Returns the current error code and resets the current error coder to `GL_NO_ERROR`".

GVRf/Framework/jni/engine/renderer/renderer.cpp
thomasflynn: I've borrowed a Note4 and confirm that it's returning 0.  Must be a Qualcomm vs. Mali problem.  I'm out of time to debug this tonight :(.  I'll have to look into it sometime tomorrow.

thomasflynn: Correction.  This code also works fine on my Qualcomm based S5 (one of the devices i tested this against when i originally wrote this code on the weekend around a month ago).  So, it's something specific to the driver on Note4.

thomasflynn: update:  the note4 i was testing with last night was an engineering build with lollipop and various framework changes.  today i tested with a kitkat, user-build note4 and that works fine.  going to try upgrading to lollipop user-build.

thomasflynn: ok, lollipop user-build shows 0ms.  must be an issue on lollipop then.  that sucks.  i can try switching to gpu timestamps rather than using time elapsed.

chenchao1407:
@thomasflynn  You still want to go with this pull request? Or I will close this one and you can propose the stats change with another pull request.

thomasflynn:
it can be closed.  there is currently a bug with the Adreno driver on lollipop that just returns 0 all the time.  until that is fixed, this won't be practical.

PR #117, opened 2015-06-05
2.0 leaked GL objects.

JonShemitz:
My bad: It doesn't matter which thread calls the destructor ...
but it does matter which thread calls (say) glDeleteTextures().

Adds a set of std::vector<GLuint>, and purges them, once per
frame. (This *may* be a little too often.)

j.shemitz@samsung.com


chenchao1407:
Codes look straight forward. And also resolved the native leak issue I found last week. Looks great.

thomasflynn:
looks good to me as well.  let's merge.

PR #121, opened 2015-06-09
Previous collection of post-effect references was a bit inefficient.

JonShemitz:
j.shemitz@samsung.com


chenchao1407:
Good. I think it is good to merge.

PR #126, opened 2015-06-11
Push a bounding box parameter down to the GVRMeshEyePointee constructor.

JonShemitz:
Change GVRSceneObject.attachEyePointeeHolder() to use a bounding box.

j.shemitz@samsung.com


chenchao1407:
Just tried with guodong's pickandmove sample app. As far as I can see, if the object is small enough or enough far from camera(still small on screen), then it is fine. Otherwise, using bounding box will seems a little weird even though the object I test with is just a sphere.

Anyway, I think this is good to go because it just gives user another option to do eyepointee stuff.

thomasflynn:
+1, looks good to me.

PR #128, opened 2015-06-12
Collision detection.

lokeshguddu:
Detects the collision between two given Scene Objects. It takes into
account the bounding boxes of the meshes which are part of the Scene
Object.


soumyajitdeb:
This code uses AABBs for collision detection. Transformation of AABBs are taken care of correctly based on position. The main API call checks if two scene objects collide with each other.

There is a very basic sample with two spheres (created using the gvrsolarsystem sample) - when one sphere touches another sphere, a status message indicating that there is a collision is displayed.

-Soumyajit

GVRf/Framework/jni/objects/mesh.cpp
```diff
@@ -145,6 +146,64 @@ const float* Mesh::getBoundingBoxInfo() {
     return bounding_box_info_;
 }
 
+const float* Mesh::getTransformedBoundingBoxInfo(glm::mat4 *Mat) {
```
JonShemitz: This method returns a `float*` ... which always points to a member field. Far better to take/modify a `float*` parameter.

GVRf/Framework/jni/objects/mesh.cpp
```diff
@@ -145,6 +146,64 @@ const float* Mesh::getBoundingBoxInfo() {
...
+
+    //Inspired by Graphics Gems - TransBox.c
+    //Transform the AABB to the correct position in world space
+    //Generate a new AABB from the non axis aligned bounding box
+
+    transformed_bounding_box_info_[0] = M[3].x; transformed_bounding_box_info_[3] = M[3].x;
+    transformed_bounding_box_info_[1] = M[3].y; transformed_bounding_box_info_[4] = M[3].y;
+    transformed_bounding_box_info_[2] = M[3].z; transformed_bounding_box_info_[5] = M[3].z;
+
+    for(int i=0;i<3;i++)
```
JonShemitz: Whitespace ... again.

GVRf/Framework/jni/objects/mesh.h
```diff
@@ -348,6 +349,7 @@ class Mesh: public HybridObject {
     // bounding box info
     bool have_bounding_box_;
     float bounding_box_info_[6];
+    float transformed_bounding_box_info_[6];
```
JonShemitz: Kill this.

GVRf/Framework/jni/objects/scene_object.cpp
```diff
@@ -199,4 +200,27 @@ void SceneObject::set_visible(bool visibility = true) {
...
 
+bool SceneObject::isColliding(SceneObject *scene_object) {
+
+    //Get the transformed bounding boxes in world coordinates and check if they intersect
+    //Transformation is done by the getTransformedBoundingBoxInfo method in the Mesh class
+
+    float thisobj_bbox[6], checkobj_bbox[6];
+
+    glm::mat4 thisobj_model_matrix = this->render_data()->owner_object()->transform()->getModelMatrix();
+    memcpy(thisobj_bbox, (this->render_data()->mesh()->getTransformedBoundingBoxInfo(&thisobj_model_matrix)), sizeof(float)*6);
```
JonShemitz: These `memcpy)`s would not be necessary if `getTransformedBoundingBoxInfo()` populated a local variable, passed in as a parameter.

GVRf/Framework/jni/objects/scene_object.cpp
```diff
@@ -199,4 +200,27 @@ void SceneObject::set_visible(bool visibility = true) {
     }
 }
 
+bool SceneObject::isColliding(SceneObject *scene_object) {
+
+    //Get the transformed bounding boxes in world coordinates and check if they intersect
+    //Transformation is done by the getTransformedBoundingBoxInfo method in the Mesh class
+
+    float thisobj_bbox[6], checkobj_bbox[6];
```
JonShemitz: Please don't use abbreviations.

GVRf/Framework/src/org/gearvrf/GVRScene.java
```diff
@@ -30,6 +30,7 @@
 
     private final List<GVRSceneObject> mSceneObjects = new ArrayList<GVRSceneObject>();
     private GVRCameraRig mMainCameraRig;
+    private String mStatMessage = "";
```
JonShemitz: Seems rather hard to justify.

GVRf/Framework/src/org/gearvrf/GVRScene.java
```diff
@@ -212,8 +213,21 @@ void updateStats() {
 
             mStatsConsole.writeLine("Draw Calls: %d", numberDrawCalls);
             mStatsConsole.writeLine(" Triangles: %d", numberTriangles);
+            
+            if(mStatMessage!="")
```
JonShemitz: Whitespace. Please Shift+Ctrl+F before pushing your code.

GVRf/Framework/src/org/gearvrf/GVRScene.java
```diff
@@ -212,8 +213,21 @@ void updateStats() {
...
+            	mStatsConsole.writeLine("%s", mStatMessage);
```
JonShemitz: Have you tested this? It doesn't reset the field; you will get the same message, over and over.

GVRf/Framework/src/org/gearvrf/GVRScene.java
```diff
@@ -212,8 +213,21 @@ void updateStats() {
...
         }
     }
+
+    /**
+     * Add an additional string to stats for this scene.
+     *
+     * @param msg
+     *     String to add to stats.
+     */
+    public void addStatMessage(String msg) {
```
JonShemitz: Please don't use abbreviations, especially in a public API.

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
```diff
@@ -558,6 +558,18 @@ public void removeChildObject(GVRSceneObject child) {
     }
 
     /**
+     * Check if {@code otherObject} is colliding with this object.
+     * 
+     * @param otherObject
+     *            {@link GVRSceneObject Object} to check for collision with this
+     *            object.
+     * @return True if objects collide and False otherwise.
```
JonShemitz: In Java, it's `true` and `false`, or `{@code true}` and `{@code false}` if you don't want your JavaDoc to stick out like sore thumb.

GVRf/Framework/src/org/gearvrf/GVRScene.java
chenchao1407: Probably an TextUtils.isEmpty(mStatMessage) here is better and more safe than mStatMessage != "" check here.

GVRf/Framework/jni/objects/scene_object.cpp
```diff
@@ -199,4 +200,27 @@ void SceneObject::set_visible(bool visibility = true) {
...
+
+    glm::mat4 thisobj_model_matrix = this->render_data()->owner_object()->transform()->getModelMatrix();
+    memcpy(thisobj_bbox, (this->render_data()->mesh()->getTransformedBoundingBoxInfo(&thisobj_model_matrix)), sizeof(float)*6);
+
+    glm::mat4 checkobj_model_matrix = scene_object->render_data()->owner_object()->transform()->getModelMatrix();
+    memcpy(checkobj_bbox, (scene_object->render_data()->mesh()->getTransformedBoundingBoxInfo(&checkobj_model_matrix)), sizeof(float)*6);
+
```
chenchao1407: Same concern here. I don't see any reason why you need to do memcpy here since you are just doing comparison in the following.

GVRf/Framework/src/org/gearvrf/GVRScene.java
chenchao1407: Agreed. Probably you may need to reset mStatMessage within some time. It is not like the draw calls and triangles messages while they are updating their message every frame.

chenchao1407:
Just tried. I think you may need to add sphere.obj into the asset folder of the sample app otherwise it won't compile.

Moreover, I think the purpose of API addStatMessage is good. However it seems to me that right now it will make user a little bit hard to maintain since it won't clean up itself.

Overall, there are just some issues on formatting and implementation details, the feature of collision detection itself works fine.

lokeshguddu:
When I was pushing it to the remote repo, the file system showed that the sphere.obj is included. Don't know what happened while push operation. Will add it manually.

As for other issues, we are on it.

GVRf/Framework/jni/objects/mesh.cpp
```diff
@@ -145,6 +146,66 @@ const float* Mesh::getBoundingBoxInfo() {
...
 
+float* Mesh::getTransformedBoundingBoxInfo(glm::mat4 *Mat) {
+
+    if (have_bounding_box_ == false)
+        getBoundingBoxInfo();
+
+    glm::mat4 M = *Mat;
+    float a, b;
+    float *transformed_bounding_box = new float[6];
+
```
chenchao1407: Just minor concern here. Probably you can create a member float array of size 6 instead of doing new and delete the array each call.

GVRf/Framework/src/org/gearvrf/GVRScene.java
```diff
@@ -169,51 +170,73 @@ public boolean getStatsEnabled() {
...
+
+    /**
+     * Add an additional string to stats message for this scene.
+     * 
+     * @param message
+     *            String to add to stats message.
+     */
+    public void addStatMessage(String message) {
+        mStatMessage = new StringBuffer(message);
+    }
```
chenchao1407: Another minor concern here. Actually the use of StringBuffer confuses me a little bit. Normally we use StringBuilder/StringBuffer as a mutable String. But each time you just create a new StringBuffer if I want to put something on console. It seems to me that the use of StringBuffer is a little bit meaningless. And also, the only reason we may use StringBuffer instead of StringBuilder is for synchronization. But it seems to me that the way you use StringBuffer won't face any synchronization issue. I think probably String is fine in this case.

GVRf/Framework/jni/objects/scene_object_jni.cpp
```diff
@@ -176,7 +179,7 @@ Java_org_gearvrf_NativeSceneObject_attachEyePointeeHolder(
         jlong jeye_pointee_holder) {
     SceneObject* scene_object = reinterpret_cast<SceneObject*>(jscene_object);
     EyePointeeHolder* eye_pointee_holder =
-            reinterpret_cast<EyePointeeHolder*>(jeye_pointee_holder);
+    reinterpret_cast<EyePointeeHolder*>(jeye_pointee_holder);
```
chenchao1407: I think the format change here seems a little weird to me. Please make sure that you are using K&R built-in with modification on Tab policy(spaces only).

lokeshguddu: I did the Ctrl+Shift+F as per the guideline. This output might be the result of that operation.

GVRf/Framework/jni/objects/mesh.cpp
soumyajitdeb: Actually that is how it was earlier. John requested me to kill the member variable and allocate it on the heap. So it was changed to allocating here and then deallocating it in the function calling it.

GVRf/Framework/src/org/gearvrf/GVRScene.java
soumyajitdeb: This will be fixed in a few moments. The reason to use StringBuffer is exactly that - it is mutable and allows me to clear it when required. The new version allocates it once and then just modifies it according to need.

GVRf/Framework/jni/objects/mesh.cpp
JonShemitz: No, that's not what I suggested.

I suggested a local variable (on the stack) which is much cheaper.

soumyajitdeb: But allocating one on the stack and returning is not the most correct way to do it. Once the method returns, there is no guarantee that the contents would exist as the stack would get cleared out. Or do you mean allocating one on the stack in scene_object.cpp and passing it over?

chenchao1407: I guess Jon is asking you to modify the API to something like:
void Mesh::getTransformedBoundingBoxInfo(glm::mat4 *Mat, float* transformed_bounding_box)

@JonShemitz  Is that right?

GVRf/Framework/jni/objects/scene_object_jni.cpp
chenchao1407: Hi.
I just get this pull request and tried with the Ctrl+Shift+F. I find there is something different. 
Can you help go to Windows->Preference->C++->Code Style and check what your profile is?

Thanks in advance.

soumyajitdeb: It says K&R [built in] for me. Shall confirm in a minute with Lokesh.

GVRf/Framework/jni/objects/mesh.cpp
soumyajitdeb: Alright shall change the function signature to match the above. However it basically means a method call with side effects - something as a personal preference I do not like to have in my code.

GVRf/Framework/jni/objects/scene_object_jni.cpp
chenchao1407: I think that may be the problem.

Please go to the same place and click Edit, change Tab Policy from "tab only" to "space only". And change the profile name(otherwise you can't save the new profile.) That is currently GVRF's C code style.

Sincerely
-Chao Chen

GVRf/Framework/src/org/gearvrf/GVRScene.java
chenchao1407: Please think about if there is synchronization issue with your next update. If no, please use StringBuilder instead which can be 3x faster than StringBuffer.

Thanks in advance.

soumyajitdeb: We'll replace StringBuffer with StringBuilder. I do not see any reason for sync issues in this code.

Thanks!

GVRf/Framework/jni/objects/scene_object_jni.cpp
soumyajitdeb: Fixed it! Thanks. That was useful.

chenchao1407:
Modified the sample and took a closer look. The accuracy is not very good when the objects are close to the main camera. However since it is using bounding box to do the detection, I think it is acceptable to me. 
@jason2kim  Can you take a look at? It is almost fine to me, just a minor concern with the use of StringBuilder.

jason2kim:
Can these sdk settings of minSdkVersion and targetSdkVersion be set to 19, like other samples?
Otherwise, I have no additional comments at this pull request.

chenchao1407:
I will verify this tomorrow morning one more time and if no other issue then it will get merged.

Thanks in advance.

lokeshguddu:
The sdk versions have been updated. Regarding more accurate collisions, that would require per pixel detection which is kind of expensive. Maybe we can explore that avenue in the future.

chenchao1407:
Hi. 
It leaves a final small issue since Oculus 0.6.0 comes in. 

Can you help add the following four permission into the AndroidManifest.xml of the sample app here?(And also with the sample of JAssimp Importer) Otherwise it will trigger SecurityException.
    <uses-permission android:name="android.permission.MOUNT_UNMOUNT_FILESYSTEMS"/>
    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />

Thanks in advance.
Sincerely
-Chao Chen

chenchao1407:
Oops. I missed the permissions. Those four permissions are here:
     <uses-permission android:name="android.permission.MOUNT_UNMOUNT_FILESYSTEMS"/>
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />

lokeshguddu:
Done!

PR #129, opened 2015-06-12
bullet sample added

deepakrawat22:
A sample application using Bullet Physics along with GearVRf. The Bullet Physics is using Java wrapper from http://www.badlogicgames.com/wordpress/?p=248. Its compiled as a jar and native so file, sources can be found at https://github.com/deepakrawat22/BulletJniFramework.


GVRf/Sample/gvrbulletsample/AndroidManifest.xml
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */-->
+
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="org.gearvrf.bulletsample"
+    android:versionCode="1"
+    android:versionName="1.0" >
+
+    <uses-sdk
+        android:minSdkVersion="16"
```
thomasflynn: set minsdk to 19 as well.  we don't support anything below that.  we should go through the rest of the samples and make sure they're consistent as well.

GVRf/Sample/gvrbulletsample/project.properties
```diff
@@ -0,0 +1,15 @@
...
+# To customize properties used by the Ant build system edit
+# "ant.properties", and override values to adapt the script to your
+# project structure.
+#
+# To enable ProGuard to shrink and obfuscate your code, uncomment this (available properties: sdk.dir, user.home):
+#proguard.config=${sdk.dir}/tools/proguard/proguard-android.txt:proguard-project.txt
+
+# Project target.
+target=android-19
+android.library.reference.1=..\\..\\Framework
```
thomasflynn: i haven't tried this yet, but i think you'll need to change these to forward slashes for this to work on all the OSs.

GVRf/Sample/gvrbulletsample/.classpath
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="con" path="com.android.ide.eclipse.adt.ANDROID_FRAMEWORK"/>
+	<classpathentry exported="true" kind="con" path="com.android.ide.eclipse.adt.DEPENDENCIES"/>
+	<classpathentry kind="src" path="src"/>
+	<classpathentry kind="src" path="gen"/>
+	<classpathentry exported="true" kind="con" path="com.android.ide.eclipse.adt.LIBRARIES"/>
+	<classpathentry kind="lib" path="C:/Users/deepak.rawat/Desktop/Source/GearVRf/GVRf/Sample/gvrbulletsample/libs/bulletjniframework.jar"/>
+	<classpathentry kind="output" path="bin/classes"/>
```
chenchao1407: "C:/Users/deepak.rawat/Desktop/Source/GearVRf/GVRf/Sample/gvrbulletsample/libs/bulletjniframework.jar"
I am pretty sure that not every computer has this path.   :-)

GVRf/Sample/gvrbulletsample/src/org/gearvrf/bulletsample/BulletSampleViewManager.java
```diff
@@ -0,0 +1,225 @@
...
+	private GVRSceneObject quadWithTexture(float width, float height,
+			String texture) {
+		FutureWrapper<GVRMesh> futureMesh = new FutureWrapper<GVRMesh>(
+				mGVRContext.createQuad(width, height));
+		GVRSceneObject object = null;
+		try {
+			object = new GVRSceneObject(mGVRContext, futureMesh,
+					mGVRContext.loadFutureTexture(new GVRAndroidResource(
+							mGVRActivity, texture)));
+		} catch (IOException e) {
```
chenchao1407: Normally we uses GVRContext to do with GVRAndroidResource not GVRActivity. Just a minor issue here. Actually this is the only place that you uses mGVRActivity. Probably you actually don't need this in this sample.

chenchao1407:
I think you will have to change the way you refer to bullet library otherwise other people can't even compile this sample.

Moreover, please add header to those java files.

Thanks.

deepakrawat22:
Modified the sample app as per comments. In the last commit the libbullet.so file was missing as it was gitignored.

chenchao1407:
This one is awesome. I can see it grants GVRF a physical engine which we don't have. The bad thing is that it seems to me that the Bullet library itself is still not that mature. But it is really a good start for us. I think it is good to merge.

deepakrawat22:
This version of Bullet Physics JNI wrapper for Android seems to be not complete as they lack some features like there is no function to set linearVelocity to a Rigid body in the Java wrapper.

GVRf/Sample/gvrbulletsample/src/org/gearvrf/bulletsample/BulletSampleViewManager.java
```diff
@@ -0,0 +1,219 @@
...
+         * Throw a sphere from top
+         */
+        addSphere(scene, 1.0f, 1.5f, 100.0f, -10.0f, 20.0f);
+    }
+
+    @Override
+    public void onStep() {
+        mBullet.doSimulation(1.0f / 60.0f, 10);
+        for (RigidBody body : rigidBodiesSceneMap.keySet()) {
+            if (body.geometry.shape.getType() == ShapeType.SPHERE_SHAPE_PROXYTYPE) {
```
jason2kim: Minor code redundancy, the following change would look better for me: 
            if (body.geometry.shape.getType() == ShapeType.SPHERE_SHAPE_PROXYTYPE
                    || body.geometry.shape.getType() == ShapeType.BOX_SHAPE_PROXYTYPE) {

	// the second code block can be deleted.

GVRf/Sample/gvrbulletsample/src/org/gearvrf/bulletsample/BulletSampleViewManager.java
```diff
@@ -0,0 +1,219 @@
...
+    private void addCube(GVRScene scene, float x, float y, float z, float mass) {
+        BoxShape boxShape = new BoxShape(new Vector3(0.5f, 0.5f, 0.5f));
+        Geometry boxGeometry = mBullet.createGeometry(boxShape, mass,
+                new Vector3(0.0f, 0.0f, 0.0f));
+        MotionState boxState = new MotionState();
+        boxState.worldTransform = new Transform(new Point3(x, y, z));
+        RigidBody boxBody = mBullet
+                .createAndAddRigidBody(boxGeometry, boxState);
+
+        GVRSceneObject cubeObject = meshWithTexture("cube.obj", "cube.jpg");
```
jason2kim: For discussion purposes, 
I could see we can already use the Bullet library as a 3D app in GVRf, without any additional changes at the framework level. However, I was thinking of adding support of Physics in GVRf, say GVRBulletSceneObject(), which will make it simpler for our GVRf users to use, your opinions please.

thomasflynn: yeah, let's not tackle that in this check-in.  but we should take a closer look at physics integration in the future.

deepakrawat22: Removed the redundant code. Also, for adding support for Bullet Physics with GearVRf will be a bigger task and there will be some design decisions we need to consider. Like, do we need to allocate physics world all the time whenever create Scene.

jason2kim: Thanks for the opinions. Codes look good for me.

thomasflynn:
ok, any further comments on this?  @chenchao1407 ?

chenchao1407:
Yeah. I think it is good to me now. Please go ahead and merge it if you also think it is good.

PR #134, opened 2015-06-16
Fix #123 and allow user to set/get refresh frequency of GVRTextViewSceneObject

### LINKED ISSUE #123: Problem on GVRTextViewSceneObject
When we try to create more than 1 GVRTextViewSceneObject, the scene objects start flickering after moving head.

This even happen in the sample gvrtextviewsample, after creating another GVRTextViewSceneObject

chenchao1407:
Fix flickering while more than a certain number of GVRTextViewSceneObject within a Scene.
Actually the problem is caused by the limitation of SurfaceTexture's performance.
If we are trying to refresh many SurfaceTextures within a frame, then it will cause latency issue
which finally turns out to be the flickering.
The patch is something like a load balancer so that we distribute the refresh() calls into interval
frames instead of calling them within one frame. And only if the textview is changed, we trigger the
refresh().

The default interval is 30 and I changed it to 20, and allow user to set the frequency of 3 levels based
on their needs.

c.chen@samsung.com

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/scene_objects/GVRTextViewSceneObject.java
+++ b/GVRf/Framework/src/org/gearvrf/scene_objects/GVRTextViewSceneObject.java
@@ -48,15 +48,15 @@ public class GVRTextViewSceneObject extends GVRSceneObject {
      */
     public static enum IntervalFrequency {
         /*
-         * Frequency HIGH, means do refresh every 10 frames
+         * Frequency HIGH, means will do refresh every 10 frames
          */
         HIGH,
         /*
-         * Frequency MEDIUM, means do refresh every 20 frames
+         * Frequency MEDIUM, means will do refresh every 20 frames
          */
         MEDIUM,
         /*
-         * Frequency MEDIUM, means do refresh every 30 frames
+         * Frequency LOW, means will do refresh every 30 frames
          */
         LOW
     }
@@ -399,12 +399,12 @@ public class GVRTextViewSceneObject extends GVRSceneObject {
 
     public IntervalFrequency getRefreshFrequency() {
         switch (mRefreshInterval) {
-        case 10:
+        case HIGH_REFRESH_INTERVAL:
             return IntervalFrequency.HIGH;
-        case 20:
-            return IntervalFrequency.MEDIUM;
-        default:
+        case LOW_REFRESH_INTERVAL:
             return IntervalFrequency.LOW;
+        default:
+            return IntervalFrequency.MEDIUM;
         }
     }
 
--- a/GVRf/Framework/src/org/gearvrf/scene_objects/GVRTextViewSceneObject.java
+++ b/GVRf/Framework/src/org/gearvrf/scene_objects/GVRTextViewSceneObject.java
@@ -39,7 +39,32 @@ import android.widget.LinearLayout;
 import android.widget.TextView;
 
 public class GVRTextViewSceneObject extends GVRSceneObject {
-    private static final int REFRESH_INTERVAL = 30; // frames
+    private static final int HIGH_REFRESH_INTERVAL = 10; // frames
+    private static final int MEDIUM_REFRESH_INTERVAL = 20;
+    private static final int LOW_REFRESH_INTERVAL = 30;
+
+    /**
+     * The refresh frequency of this sceneobject.
+     */
+    public static enum IntervalFrequency {
+        /*
+         * Frequency HIGH, means do refresh every 10 frames
+         */
+        HIGH,
+        /*
+         * Frequency MEDIUM, means do refresh every 20 frames
+         */
+        MEDIUM,
+        /*
+         * Frequency MEDIUM, means do refresh every 20 frames
+         */
+        LOW
+    }
+
+    private static int sReferenceCounter = 0;
+    private boolean mFirstFrame;
+    private boolean mIsChanged;
+    private int mRefreshInterval = MEDIUM_REFRESH_INTERVAL;
 
     private static final int DEFAULT_WIDTH = 2000;
     private static final int DEFAULT_HEIGHT = 1000;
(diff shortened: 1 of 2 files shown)


GVRf/Framework/src/org/gearvrf/scene_objects/GVRTextViewSceneObject.java
```diff
@@ -39,7 +39,32 @@
...
+        /*
+         * Frequency HIGH, means do refresh every 10 frames
+         */
+        HIGH,
+        /*
+         * Frequency MEDIUM, means do refresh every 20 frames
+         */
+        MEDIUM,
+        /*
+         * Frequency MEDIUM, means do refresh every 20 frames
```
jason2kim: typo with this line.

GVRf/Framework/src/org/gearvrf/scene_objects/GVRTextViewSceneObject.java
```diff
@@ -103,6 +128,9 @@ public GVRTextViewSceneObject(GVRContext gvrContext, GVRMesh mesh,
         mSurface = new Surface(mSurfaceTexture);
         mSurfaceTexture.setDefaultBufferSize(mTextViewContainer.getWidth(),
                 mTextViewContainer.getHeight());
+        sReferenceCounter++;
```
jason2kim: Is this static variable intended to serialize drawing text frames one at a time for the purpose of performance load balancing? If so, it would be helpful to add some in source documentation for future reference.

GVRf/Framework/src/org/gearvrf/scene_objects/GVRTextViewSceneObject.java
```diff
@@ -335,12 +369,56 @@ public int getGravity() {
         return mTextView.getGravity();
     }
 
+    /**
+     * Set the refresh frequency of this scene object.
+     * 
+     * @param frequency
+     *            The refresh frequency of this TextViewSceneObject.
+     */
+    public void setRefreshFrequency(IntervalFrequency frequency) {
```
jason2kim: I believe you tested these setty()/getty() methods from a sample app. Your decision, but it would be helpful to submit the updated sample app, too.

GVRf/Framework/src/org/gearvrf/scene_objects/GVRTextViewSceneObject.java
```diff
@@ -39,7 +39,32 @@
 import android.widget.TextView;
 
 public class GVRTextViewSceneObject extends GVRSceneObject {
```
jason2kim: Just a quick note for the future,
Multiple text views under the default Linear Layout for some cases,
public GVRTextViewSceneObject(…,
            TextView[] textViews) { }
but it will require changes from the single text view based methods and its full view size.

chenchao1407:
Make some changes based on Jason's comments.

GVRf/Framework/src/org/gearvrf/scene_objects/GVRTextViewSceneObject.java
chenchao1407: I think finally we can make something like GVRViewGroupSceneObject to let developer to make their customized layout other than textviews' group.

GVRf/Framework/src/org/gearvrf/scene_objects/GVRTextViewSceneObject.java
```diff
@@ -39,7 +39,32 @@
...
+        /*
+         * Frequency HIGH, means do refresh every 10 frames
+         */
+        HIGH,
+        /*
+         * Frequency MEDIUM, means do refresh every 20 frames
+         */
+        MEDIUM,
+        /*
+         * Frequency MEDIUM, means do refresh every 30 frames
```
jason2kim: typo: MEDIUM -> LOW

GVRf/Framework/src/org/gearvrf/scene_objects/GVRTextViewSceneObject.java
```diff
@@ -335,12 +369,56 @@ public int getGravity() {
...
+
+    /**
+     * Get the refresh frequency of this scene object.
+     * 
+     * @return The refresh frequency of this TextViewSceneObject.
+     */
+
+    public IntervalFrequency getRefreshFrequency() {
+        switch (mRefreshInterval) {
+        case 10:
```
jason2kim: to minimize hardcodings, better to -> case HIGH_REFRESH_INTERVAL:

GVRf/Framework/src/org/gearvrf/scene_objects/GVRTextViewSceneObject.java
```diff
@@ -335,12 +369,56 @@ public int getGravity() {
...
+            return IntervalFrequency.HIGH;
+        case 20:
+            return IntervalFrequency.MEDIUM;
+        default:
+            return IntervalFrequency.LOW;
```
jason2kim: to be align with this definition of the default, mRefreshInterval = MEDIUM_REFRESH_INTERVAL;
        case MEDIUM_REFRESH_INTERVAL:
        default:
            return IntervalFrequency.MEDIUM;

GVRf/Sample/gvrtextviewsample/src/org/gearvrf/gvrtextviewsample/SampleViewManager.java
```diff
@@ -32,8 +36,14 @@
...
     private final int[] colors = new int[] { Color.RED, Color.YELLOW,
             Color.GREEN, Color.WHITE, Color.MAGENTA };
+    private final IntervalFrequency[] frequencies = new IntervalFrequency[] {
+            IntervalFrequency.HIGH, IntervalFrequency.MEDIUM,
+            IntervalFrequency.LOW };
     private float textSize;
     private int counter = 0;
+    Random random = new Random();
+
+    ArrayList<GVRTextViewSceneObject> textviews = new ArrayList<GVRTextViewSceneObject>();
```
jason2kim: LinearLayout mTextView; 
Can be deleted to avoid distraction. Seems there because of some historic reason…

jason2kim: boolean init = false;
This one also is not used^^

chenchao1407:
Fixed. Thanks

chenchao1407:
@thomasflynn @jason2kim 

Any else comment on this? Otherwise I am gonna merge it.

jason2kim:
it looks good for me.

PR #135, opened 2015-06-16
Initial check-in of port to OVR SDK 0.6.0

thomasflynn:
Not done yet. Still have compile errors.  Oculus changed DrawEyeViewsPostDistorted() to just DrawEyeViews().  Still need to update that.  There's several other errors as well.  I could use some help here.  I'm running low on time with this.  Anyone who can, please help out.  Thanks.

### DIFF
--- a/GVRf/Framework/jni/oculus/activity_jni.cpp
+++ b/GVRf/Framework/jni/oculus/activity_jni.cpp
@@ -116,8 +116,11 @@ jclass GVRActivity::GetGlobalClassReference( const char * className ) const
 void GVRActivity::Configure( OVR::ovrSettings & settings )
 {
     LOG( "GVRActivity::Configure");
-    // leave it as the oculus defaults for now.
+    settings.EyeBufferParms.multisamples = 2;
+    // leave the rest as default for now.
+    // TODO: take values specified in xml and set them here.
 }
+
 void GVRActivity::OneTimeInit( const char * fromPackage, const char * launchIntentJSON, const char * launchIntentURI )
 {
     LOG( "GVRActivity::OneTimeInit" );
--- a/GVRf/Framework/jni/oculus/activity_jni.cpp
+++ b/GVRf/Framework/jni/oculus/activity_jni.cpp
@@ -16,7 +16,7 @@
 #include "activity_jni.h"
 #include <jni.h>
 #include <glm/gtc/type_ptr.hpp>
-#include <VrApi/VrApi_Helpers.h>
+#include "VrApi_Helpers.h"
 
 static const char * activityClassName = "org/gearvrf/GVRActivity";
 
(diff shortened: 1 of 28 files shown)


thomasflynn:
ok, go ahead and review this now.

thomasflynn:
removed vrapi from project settings.  not sure how that was working for me.

jason2kim:
On Eclipse,
Framework Project -> Properties 
-> Android: add library for VrAppFramework
-> Project References: Check LibOVR and VrAppFramework

At least, our framework builds successfully and runs with ovr 0.6.0.1. We may need more tests with applications.

jason2kim:
I will also take a look soon in the code porting part changes including draw eye views in activity_jni.cpp

GVRf/Framework/jni/oculus/activity_jni.cpp
```diff
@@ -113,22 +113,11 @@ jclass GVRActivity::GetGlobalClassReference( const char * className ) const
...
-
-    // Always use 2x MSAA for now
-    app->GetVrParms().multisamples = 2;
-
-    // Always use bilinear texture filtering.
-    app->GetVrParms().textureFilter = TEXTURE_FILTER_BILINEAR;
-
-    app->GetSwapParms().WarpProgram = WP_SIMPLE;
+    LOG( "GVRActivity::Configure");
+    // leave it as the oculus defaults for now.
```
chenchao1407: I think you can remove modeParms.CpuLevel, modeParms.GpuLevel, textureFilter and WarpProgram settings here because by default those are the same as what you removed.

But for multisamples, I can see that it is loaded from system service and right now I can't know what it is.

To set it as what it was(2), you may need to add "settings.EyeBufferParms.multisamples = 2;" here to ensure it.

thomasflynn: what we _really_ should do is pay attention and pass down the value that is set in the xml file.  we should also expose some apis for setting the cpu/gpu levels, srgb framebuffers, etc. that would be part of ovrSettings in here.  That should be part of a separate check-in.  In the meantime, i'll take your advice and explicitly set multisample to 2 since that's what we already had hard-coded before.

deepakrawat22:
All the sample app are working correctly with the latest check-ins on Note 4

chenchao1407:
Tried. Apps run well with this build.

GVRf/Framework/jni/oculus/activity_jni.cpp
thomasflynn: ok, added back a multisample value of 2.

chenchao1407:
Here is what I get from my side:
1. To build with VrAppFramework. You need to import VrApi into workspace first(But no need to change any project properties).
2. But for Framework, you can't make VrApi as its reference since Framework will already put libvrapi.so into lib/armeabi-v7a. If you still put VrApi into Framework's reference list. Then it will lead to duplicate library issue.

thomasflynn:
i think you shouldn't be seeing this with my latest check-in.  are you?

chenchao1407:
Yes. I think current version is fine.

thomasflynn:
ok, merge it then?

jason2kim:
I took a look source codes briefly as well. As all the gvr samples are running well like Deepak said, I am okay with merging it at this point.

deepakrawat22:
I think now we need to update the project website regarding the build instruction especially the "Building the Oculus Mobile SDK" part.

PR #138, opened 2015-06-19
Jassimp Model Loader

lokeshguddu:
This integration takes help of Jassimp to load the Model with all its
meshes and textures.

The present implementation focuses on loading a model from a model file along with
it's textures. In future, we will add support for color and other
components.


GVRf/Framework/src/org/util/jassimp/AiConfigOptions.java
```diff
@@ -0,0 +1,745 @@
...
+    // * if you want triangulated data because #aiProcess_Triangulate is known
+    // to
+    // * have problems with the kind of polygons that the IFC loader spits out
+    // for
+    // * complicated meshes.
+    // * Property type: Bool. Default value: true.
+    // */
+    // #define AI_CONFIG_IMPORT_IFC_CUSTOM_TRIANGULATION
+    // "IMPORT_IFC_CUSTOM_TRIANGULATION"
+    //
```
thomasflynn: please don't check-in commented-out code.  especially a large volume of commented-out code.

GVRf/Framework/jni/engine/importer/assimp_importer_jni.cpp
```diff
@@ -47,4 +86,180 @@ Java_org_gearvrf_NativeAssimpImporter_getMesh(JNIEnv * env,
...
+
+error:
+    {
+        jclass exception = env->FindClass("java/io/IOException");
+        if (NULL == exception) {
+            /* thats really a problem because we cannot throw in this case */
+            env->FatalError("could not throw java.io.IOException");
+        }
+        env->ThrowNew(exception, aiGetErrorString());
+    }
```
chenchao1407: In current GVRF, since the overall environment is triggered and controlled by VrThread which is a pthread from native environment. It means, actually the life cycle is controlled by native side but not Java side. It also implies that all exception won't be caught when you throw it. It will only make an exception hanging there until next exception sensitive JNI call is called, and then throw a native uncaught exception crash. Therefore, if you really want to stop the app and throw some useful information, please use gvr_java_stack_trace which is in util package.

What's more, imho, please use goto statement only when there is no other choice. It seems to me that it is not the case here.

GVRf/Framework/jni/engine/importer/assimp_importer_jni.cpp
```diff
@@ -47,4 +86,180 @@ Java_org_gearvrf_NativeAssimpImporter_getMesh(JNIEnv * env,
...
+    env->ReleaseStringUTFChars(jnode_name, node_name);
+    return reinterpret_cast<jobject>(assimp_importer->mesh_material(env, current_node->mMeshes[index]));
+}
+
+JNIEXPORT jfloatArray JNICALL
+Java_org_gearvrf_NativeAssimpImporter_decomposeTransformationMatrix(JNIEnv * env,
+        jobject obj, jfloatArray jtransformation_matrix) {
+
+    jfloat* tm = env->GetFloatArrayElements(jtransformation_matrix,0); // Transformation Matrix
+
```
chenchao1407: I think you need to call ReleaseFloatArrayELments for tm before you jump out this function. Otherwise it seems to be a memory leak.

GVRf/Framework/src/org/gearvrf/GVRContext.java
```diff
@@ -417,6 +427,172 @@ public void loadMesh(MeshCallback callback, GVRAndroidResource resource,
...
+            }
+
+            @Override
+            public byte[] wrapVector3f(ByteBuffer arg0, int arg1, int arg2) {
+                byte[] warpedVector = new byte[arg2];
+                arg0.get(warpedVector, arg1, arg2);
+                return warpedVector;
+            }
+        };
+
```
chenchao1407: The argX seems to be a bad naming way to me. It is very hard to understand what it is doing. I think you'd better change them to better namings which can represent what they are.

GVRf/Framework/jni/engine/importer/assimp_importer_jni.cpp
```diff
@@ -30,6 +30,45 @@ Java_org_gearvrf_NativeAssimpImporter_getNumberOfMeshes(
...
+JNIEXPORT jint JNICALL
+Java_org_util_jassimp_Jassimp_getdoublesize(JNIEnv *env,
+        jclass cls);
+JNIEXPORT jint JNICALL
+Java_org_util_jassimp_Jassimp_getlongsize(JNIEnv *env,
+        jclass cls);
+JNIEXPORT jstring JNICALL
+Java_org_util_jassimp_Jassimp_getErrorString(
+        JNIEnv *env, jclass cls);
 }
```
chenchao1407: It seems to me that you are adding JNI methods of Jassimp class into this file which I don't think is good to well organize the project. I think you'd better create another file like jassimp_jni.cpp and put these JNI calls into that file.

GVRf/Framework/jni/engine/importer/assimp_importer.cpp
```diff
@@ -68,4 +68,428 @@ Mesh* AssimpImporter::getMesh(int index) {
...
+    jboolean return_value = env->CallBooleanMethod(object, method_id,
+            params[0].l);
+
+    return (bool) return_value;
+}
+
+bool AssimpImporter::callv(JNIEnv *env, jobject object, const char* type_name,
+        const char* method_name, const char* signature,/* const*/
+        jvalue* params) {
+    jclass java_class = env->FindClass(type_name);
```
chenchao1407: I think I personally know this is a wrapper method for CallVoidMethodA.  But just as we discussed before,  I think you'd better  not use the abbreviation name like this. Probably you can change it to something like callVoidMethod or callVoid.

Thanks in advance.

GVRf/Framework/src/org/util/jassimp/AiConfigOptions.java
```diff
@@ -0,0 +1,745 @@
+/*
+ * $Revision$
+ * $Date$
+ */
+package org.util.jassimp;
+
```
chenchao1407: Is this file from third-party or created by you? I don't see any valid file header here.

chenchao1407:
Beside the comments inside codes, I think I still have concerns on JAssimp related java files. You make those files into a package named org.util.jassimp which seems a little weird and somehow I think break the folder structure of GVRF project. Is there any concern that may prevent you from naming it like org.gearvrf.thirdparty.jassimp or anything similar?

Thanks in advance.
Sincerely
-Chao Chen

lokeshguddu:
Hi,

I did all the changes. If I missed something, please let me know.

chenchao1407:
Hi. Just tried. It seems that something fails with this pull request. I can only see the white background but without loaded model.

lokeshguddu:
Hi,

Try it on N4 for now. I'm looking into the S6 issue.

Thanks.

chenchao1407:
Actually my N4(Lollipop and Adreno) doesn't work well. I will try it one more time tmr.

lokeshguddu:
Hi,

It is working on the N4 I have (https://github.com/lokeshguddu/lokeshguddu.github.io/raw/master/img/gvrf/device-2015-06-18-151205.png). Will look into the reason for not working for other devices.

GVRf/Framework/src/org/gearvrf/vendor/jassimp/Jassimp.java
```diff
@@ -0,0 +1,275 @@
...
+OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT 
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY 
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT 
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE 
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+---------------------------------------------------------------------------
+ */
+package org.gearvrf.vendor.jassimp;
```
jason2kim: “vendor” in the class path name is quite confusing. Can you please rename ti to something more specific or simply remove it? Thank you.

jason2kim:
I have two Note 4 devices. One of them does not work. Android version is 5.0.1 and model is SM-N910F. 
There seems no notable error messages in the logcat. Adding more tight error / exception handlings and messages would be helpful in debugging this as well as for the future. Let us know if you need us help in addressing the issue.

lokeshguddu:
Hi,

In KK(Version 4.4.4), the model is showing up correctly because the draw calls are getting rendered. But in L(Version 5.0.1) even though the draw call are made, there is nothing being rendered. Any help in this regard will be appreciated.

Thanks

[Android KK]
jassimp_kk

[Android L]
jassimp_l

thomasflynn:
I don't have time to debug it tonight, but it does interest me.  I do see it on Note 4 KitKat.  All I see is white on S5, S6, and Note4 Lollipop (all are running lollipop actually).  The S6 is Mali while the others are Adreno, so it's not something specific to the GPU.  Interesting note: on the S5, it falls back to monoscopic mode (as it should), but rather than having a black screen where the eye is not drawn, the entire screen is white.  I've never seen that before.

lokeshguddu:
Hi,

I've fixed the issue of models not displaying on the android L devices. Please check and let me know if I missed something.

Thanks.

thomasflynn:
What was the key change that fixed it?

lokeshguddu:
Hi,

There was some problem with the method implementation where I was decomposing the Transformation matrix into Scale, Rotation and Position.

Everything seems fine now, but will still explore for any broken/missing components.

jason2kim:
Can you please fix the "vendor" issue as commented, before we can proceed?

GVRf/Framework/jni/engine/importer/assimp_importer.cpp
```diff
@@ -68,4 +68,428 @@ Mesh* AssimpImporter::getMesh(int index) {
 
     return mesh;
 }
+
+bool AssimpImporter::create_instance(JNIEnv *env, const char* class_name,
+        jobject& new_instance) {
```
deepakrawat22: IMHO the JNI logic should be in *_jni.cpp file. I can see in the framework all the JNI related code is restricted to *_jni.cpp files only.

chenchao1407: I think this is fine. These calls aren't used from Java native method. They are just using env for other purposes like calling back to Java. As you can see similar cases also happen with textures classes

lokeshguddu:
Hi,

@jason2kim Should we remove the name vendor or should we rename it to something like "thirdparty" as in org.gearvrf.thirdparty.jassimp . If you want something specific, please let us know, we will rename it accordingly.

GVRf/Framework/jni/engine/importer/assimp_importer.cpp
deepakrawat22: On Second thought, how about moving jassimp related code to different cpp files. IMHO to keep the GearVRf assimp code separate from jassimp, as we might be adding more functionality to it in future. I am not much familiar with jassimp, but I guess jassimp native code don't need GearVRf wrapper to assimp so easy to move to separate file. How about moving to contrib directory since its open source code.

lokeshguddu: Hi @deepakrawat22 ,

Thank you for going in such details to explain the inclusion of methods in assimp_importer.h/.cpp . 

In assimp_importer_jni.cpp we are parsing the object of GVRAssimpImpoter so that we can call those methods on the same object. Also, all these calls are somehow related to assimp and so our plan for the inclusion of jassimp methods inside assimp_importer. Jassimp is just a bridge for accessing similar functionality on the java side.

jason2kim:
Either way is okay. Isn’t it simpler and less deep to use org.gearvrf.jassimp, as long as it can tell? Thank you.

deepakrawat22:
Hi @lokeshguddu,

My concern is to avoid copying code from other open source project in GearVRf source files. For example the function create_instance() from createInstance(), etc. Otherwise its a code maintenance issue and may be licence issue (Not sure).

In my opinion if possible better to add jassimp.cpp and jassimp.h file to GearVRf project makefile and then use them.

lokeshguddu:
HI @deepakrawat22 .

It's not copying but making use of an open source project with proper mention of their licence. The method name was changed in similarity because that's the convention for most of the methods in jni codes for our GVR framework otherwise we might have kept the names and other occurrences exactly same as mentioned here(https://github.com/assimp/assimp/tree/master/port/jassimp).

Jassimp is an open source project which we have used with our framework. So technically it's not copying. It's just an extension for the Assimp library that we are using and for that use we need to follow our code convention for the framework as well as we need to modify it as per our needs. We have not done anything unique here, we have just ported it to suit our needs with the help of an open source library.

lokeshguddu:
Hi,

Is there any further comment this pull request?

lokeshguddu:
Hi,

I've rebased this branch with the latest update from the main branch.

jason2kim:
Thanks, and let me continue review soon.

chenchao1407:
Works with my lollipop device now. I think it is good to me.

jason2kim:
It looks good to me, too

chenchao1407:
Hi @lokeshguddu 

Since we just merge the null check for assimp importer, you need to do the manual rebase one more time.

Sorry for the inconvenience.

Sincerely
-Chao Chen

lokeshguddu:
Hi @chenchao1407 ,

We've rebased this branch with the latest commits from master. We've also rechecked the build and samples and it's working fine.

Please let us know if there is any issue.

Thanks.

lokeshguddu:
Hi @chenchao1407 ,

We've rebased again with the latest commits from the master branch.

Thanks.

chenchao1407:
Just tried, I think it is good to go.

PR #140, opened 2015-06-22
Fix changing of texture coordinates after a mesh has been created

liaxim:


thomasflynn:
nice.  looks good to me.  Chao?  Jason?

chenchao1407:
I think overall is fine.

The only concern is that it somehow conflicts with the implementation of guodong's litshader. In current guodong's litshader's implementation, with this pull request, the vaoID_Map doesn't work correctly.

However I think we can fix that if you guys don't have any other concern.

PR #141, opened 2015-06-24
Add bounding sphere info to the mesh.

thomasflynn:
Adds bounding sphere info for the mesh.  This bounding sphere is based
on the bounding box that we're already using for frustum culling.  As
such, the sphere will likely be a little larger than necessary, but will
be accurate enough for now.

The bounding sphere (at least the center of it) will be used in upcoming
check-ins dealing with transparency sorting, and LOD.

tom.flynn@samsung.com

### DIFF
--- a/GVRf/Framework/jni/objects/mesh.cpp
+++ b/GVRf/Framework/jni/objects/mesh.cpp
@@ -149,8 +149,9 @@ const float* Mesh::getBoundingBoxInfo() {
 void Mesh::getTransformedBoundingBoxInfo(glm::mat4 *Mat,
         float *transformed_bounding_box) {
 
-    if (have_bounding_box_ == false)
+    if (have_bounding_box_ == false) {
         getBoundingBoxInfo();
+    }
 
     glm::mat4 M = *Mat;
     float a, b;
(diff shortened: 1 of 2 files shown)


GVRf/Framework/jni/objects/mesh.cpp
```diff
@@ -204,6 +205,46 @@ void Mesh::getTransformedBoundingBoxInfo(glm::mat4 *Mat,
...
+// This gives us a really coarse bounding sphere given the already calcuated bounding box.  This won't be a tight-fitting sphere because it is based on the bounding box.  We can revisit this later if we decide we need a tighter sphere.
+const float *Mesh::getBoundingSphereInfo() {
+    if (have_bounding_box_ == false) {
+        getBoundingBoxInfo();
+    }
+
+    if (have_bounding_sphere_ == true) {
+        return bounding_sphere_info_;
+    }
+
```
chenchao1407: Very minor here. I think if(!have_bounding_box) and if(have_bounding_sphere_) would be a little better?

GVRf/Framework/jni/objects/mesh.cpp
```diff
@@ -204,6 +205,46 @@ void Mesh::getTransformedBoundingBoxInfo(glm::mat4 *Mat,
...
+    // assign the sphere
+    bounding_sphere_info_[0] = center_x;
+    bounding_sphere_info_[1] = center_y;
+    bounding_sphere_info_[2] = center_z;
+    bounding_sphere_info_[3] = radius;
+
+    have_bounding_sphere_ = true;
+
+    return bounding_sphere_info_;
+}
```
chenchao1407: It is not that related to this pull request but I am somehow thinking in deleteVaos, you may need to reset the have_bounding_box and have_bounding_sphere since all data are considered dirty with the dirty flag Mihail brought in in his pull request.

chenchao1407:
Looks good to me. Will test tmr. The only weird case I can think of is if you have a sphere, the bounding sphere's radius will be sqrt(2) times as original one. But should be ok in most of cases.

GVRf/Framework/jni/objects/mesh.cpp
thomasflynn: I was following the style that was already there in the file, but I honestly prefer what you suggest.  Done.

GVRf/Framework/jni/objects/mesh.cpp
thomasflynn: done.

thomasflynn:
Correct.  I really just need the center for the next few check-ins.  We can revisit making this a tighter sphere later.  We at least have a placeholder for now.

jason2kim:
Codes look okay for me. Just a question, are you planning to replace the bounding box with this sphere, or make it selective, including in the case of frustum culling?

thomasflynn:
That's much further down the line.  We'll need tighter bounding spheres for culling.  We'll also need to make sure the bounding volume (sphere or box) encompasses all its children for proper frustum culling.  That's all more work that I'm not trying to tackle at the moment :)

Issue #97, opened 2015-05-29
remove getOwnerObject from samples.


thomasflynn:
verify we can get rid of getOwnerObject from our samples.  and see we can possibly deprecate that api all together.

deepakrawat22:
After some analysis I found that get/setOwnerObject is used extensively in the Framework. Even in JNI code owner object is used by transform, render data. Owner object is used to main transformation hierarchies in scene objects.

So I think the objective of this enhancement will only be limited to removing getOwnerObject from Application usage?

In case of GVREyePointeeHolder, there is already a new API provided List<GVRPickedObject> findObjects(..), which do not return GVREyePointeeHolder so owner object can be avoided in this case.

jason2kim:
1) In Solar System demo app, … not able to get rid of this usage
=> Removing this is important for the app developers. Can you consider channg in JNI level if needed?

2) Once done, we need to make getOwnerObject(), if still alive, visible only to the framework package, not to the applications as public.

3) otherwise this change is only a second best solution^^

jason2kim:
If too much for changing the JNI side, a Java side trick would be overloading GVRSceneObject.addChildObject() with GVRComponent or GVRCameraRig. But, it would be not so consistent as you know.

PR #142, opened 2015-06-24
Removed the usage of getOwnerObject from sample applications

deepakrawat22:
Currently getOwnerObject is used in below scenarios in Sample applications (Also the changes made to get rid of the usage):

1 - GVRPicker.pickScene() returns an array of GVREyePointeeHolder and the GVRSceneObject which is currently in path of ray cast can be retrieved by doing getOwnerObject on eyePointeeHolder object.
=> A new API "GVRPicker.findObjects()" is provided since version 1.6 which will return List<GVRPickedObject> which is more cleaner to use. In this check-in this new API will replace all the usage of "GVRPicker.pickScene()"

2 - To transform the position of camera rig in the scene graph, currently the usage is 
mainScene.getMainCameraRig().getOwnerObject().getTransform().setPosition(0.0f, 6.0f, 1.0f);
=> A new API is provided by CameraRig which can set/get Transform

3 - Some of the sample attach scene objects to camera rig as child objects, so the object appear move with camera in scene graph. 
scene.getMainCameraRig().getOwnerObject().addChildObject(headTracker);
=> New API added so that scene object can be directly added to the camera rig instead of using getOwnerObject and then add scene object to it.

4 - At one place in Solar System demo app the owner object of camera rig is added as child object to some other scene object. I was not able to get rid of this usage of getOwnerObject().
moonRevolutionObject.addChildObject(mMainScene.getMainCameraRig().getOwnerObject());

Internally getOwnerObject is used though. This pull request is for fixing #97.

drawat@usc.edu

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/GVRCamera.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRCamera.java
@@ -252,7 +252,10 @@ public abstract class GVRCamera extends GVRComponent {
      *         {@code null}.
      */
     public GVRTransform getTransform() {
-        return getOwnerObject().getTransform();
+        if (getOwnerObject() != null) {
+            return getOwnerObject().getTransform();
+        }
+        return null;
     }
 }
 
--- a/GVRf/Framework/src/org/gearvrf/GVRCamera.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRCamera.java
@@ -257,6 +257,46 @@ public abstract class GVRCamera extends GVRComponent {
         }
         return null;
     }
+
+    /**
+     * Add {@code child} as a child of this camera owner object.
+     * 
+     * @param child
+     *            {@link GVRSceneObject Object} to add as a child of this camera
+     *            owner object.
+     */
+    public void addChildObject(GVRSceneObject child) {
+        if (getOwnerObject() != null) {
+            getOwnerObject().addChildObject(child);
+        }
+    }
+
+    /**
+     * Remove {@code child} as a child of this camera owner object.
+     * 
+     * @param child
+     *            {@link GVRSceneObject Object} to remove as a child of this
+     *            camera owner object.
+     */
+    public void removeChildObject(GVRSceneObject child) {
+        if (getOwnerObject() != null) {
+            getOwnerObject().removeChildObject(child);
+        }
+    }
+
+    /**
+     * Get the number of child objects that belongs to owner object of this
+     * camera.
+     * 
+     * @return Number of {@link GVRSceneObject objects} added as children of
+     *         this camera owner object.
+     */
+    public int getChildrenCount() {
+        if (getOwnerObject() != null) {
+            return getOwnerObject().getChildrenCount();
+        }
+        return 0;
+    }
 }
 
 class NativeCamera {
--- a/GVRf/Framework/src/org/gearvrf/GVRCamera.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRCamera.java
@@ -217,6 +217,43 @@ public abstract class GVRCamera extends GVRComponent {
         mPostEffects.remove(postEffectData);
         NativeCamera.removePostEffect(getNative(), postEffectData.getNative());
     }
+
+    /**
+     * Replace the current {@link GVRTransform transform} for owner object of
+     * the camera.
+     * 
+     * @param transform
+     *            New transform.
+     */
+    void attachTransform(GVRTransform transform) {
+        if (getOwnerObject() != null) {
+            getOwnerObject().attachTransform(transform);
+        }
+    }
+
+    /**
+     * Remove the object's (owner object of camera) {@link GVRTransform
+     * transform}.
+     * 
+     */
+    void detachTransform() {
+        if (getOwnerObject() != null) {
+            getOwnerObject().detachTransform();
+        }
+    }
+
+    /**
+     * Get the {@link GVRTransform}.
+     * 
+     * 
+     * @return The current {@link GVRTransform transform} of owner object of
+     *         camera. Applying transform to owner object of camera moves it. If
+     *         no transform is currently attached to the object, returns
+     *         {@code null}.
+     */
+    public GVRTransform getTransform() {
+        return getOwnerObject().getTransform();
+    }
 }
 
 class NativeCamera {
--- a/GVRf/Framework/src/org/gearvrf/GVRCameraRig.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRCameraRig.java
@@ -347,6 +347,75 @@ public class GVRCameraRig extends GVRComponent {
         return NativeCameraRig.getLookAt(getNative());
     }
 
+    /**
+     * Replace the current {@link GVRTransform transform} for owner object of
+     * the camera rig.
+     * 
+     * @param transform
+     *            New transform.
+     */
+    void attachTransform(GVRTransform transform) {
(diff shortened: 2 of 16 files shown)


GVRf/Framework/src/org/gearvrf/GVRCameraRig.java
```diff
@@ -347,6 +347,75 @@ void predict(float time) {
...
+    /**
+     * Get the number of child objects that belongs to owner object of this
+     * camera rig.
+     * 
+     * @return Number of {@link GVRSceneObject objects} added as children of
+     *         this camera rig owner object.
+     */
+    public int getChildrenCount() {
+        return getOwnerObject().getChildrenCount();
+    }
```
chenchao1407: Same here...

GVRf/Sample/gvr-sample/src/org/gearvrf/sample/SampleCubeScript.java
```diff
@@ -157,16 +158,15 @@ public void onStep() {
         mFrontFace.getRenderData().getMaterial().setOpacity(1.0f);
         mFrontFace2.getRenderData().getMaterial().setOpacity(1.0f);
         mFrontFace3.getRenderData().getMaterial().setOpacity(1.0f);
-        GVREyePointeeHolder[] eyePointeeHolders = GVRPicker
-                .pickScene(mGVRContext.getMainScene());
-        for (GVREyePointeeHolder eyePointeeHolder : eyePointeeHolders) {
-            if (eyePointeeHolder.getOwnerObject().equals(mFrontFace)) {
+        for (GVRPickedObject pickedObject : GVRPicker.findObjects(
+                mGVRContext.getMainScene(), 0, 0, 0, 0, 0, -1.0f)) {
+            if (pickedObject.getHitObject().equals(mFrontFace)) {
```
chenchao1407: Just a minor concern here. Imho, I think you are somehow giving up one use case of GVREyePointeeHolder. Probably I would rather add an API in GVREyePointeeHolder and do something like 
public boolean hitObject(GVRSceneObject sceneObject){
     return getOwnerObject().equals(sceneObject);
}

just my personal opinion.

chenchao1407:
Just did a quick review. Overall is fine, just a little bit different than I thought in GVREyePointeeHolder. It seems to me that you don't want to make wrapper functions for GVREyePointeeHolder's getOwnerObject(), do you?

GVRf/Framework/src/org/gearvrf/GVRCamera.java
```diff
@@ -217,6 +217,43 @@ public void removePostEffect(GVRPostEffect postEffectData) {
...
+     * 
+     * 
+     * @return The current {@link GVRTransform transform} of owner object of
+     *         camera. Applying transform to owner object of camera moves it. If
+     *         no transform is currently attached to the object, returns
+     *         {@code null}.
+     */
+    public GVRTransform getTransform() {
+        return getOwnerObject().getTransform();
+    }
```
chenchao1407: There is null pointer check for attach and detach but no such check for get? A little weird to me.

deepakrawat22:
Regarding the GVREyePointeeHolder case, Framework already provided a new high level API findObjects() which is better to use instead of pickScene(). Internally findObjects() uses GVREyePointeeHolder. But the use of getOwnerObject in application level is avoided with usage of findObjects().

For loop can be replaced as 

List<GVRPickedObject> pickedObjects = GVRPicker.findObjects(mGVRContext.getMainScene(), 0, 0, 0, 0, 0, -1.0f);
for (pickedObject : pickedObjects ) {
...
}
instead of (similar to what GVREyePointeeHolder one looks like)
for (GVRPickedObject pickedObject : GVRPicker.findObjects(
+                mGVRContext.getMainScene(), 0, 0, 0, 0, 0, -1.0f))

I will work on other review comments and I agree null check should be consistent.

chenchao1407:
My first concern is like: You are actually making GVREyePointeeHolder deprecated in some cases. That is fine if the new stuff you are going to use is actually better or at least equals.
But my second concern coming is like : I understand that actually there is little difference between pickScene and findObjects. But you are actually using an API hard to use(which take 7 parameters) to replace a simple call. At least, if it is really difficult to make a getOwnerObject wrapper for GVREyePointeeHolder. I think one thing maybe you can do is to add an overloaded function like

public static final List<GVRPickedObject>  findObjects(GVRScene scene){
       return findObjects(scene, 0, 0, 0, 0, 0, -1.0f);
}

Actually you can see that the simpler way of pickScene is also a overloaded function of a complicated one.
But I think you will finally need to make wrapper calls in some cases for GVREyePointeeHolder.

For example, if I want to do something like:
if(gvrEyePointeeHolder.getOwnerObjects().equals(someSceneObject)){
      gvrEyePointeeHolder.setEnable(false);
}

Imho, something like this is actually you can't do with GVRPickedObject.

deepakrawat22:
Consistent NULL check when calling getOwnerObject and easy to use findObjects.

Still need to work on Solar System sample and GVREyePointeeHolder.

deepakrawat22:
Regarding 

if(gvrEyePointeeHolder.getOwnerObjects().equals(someSceneObject)){
gvrEyePointeeHolder.setEnable(false);
}

This can be achieved with GVRPickedObject like

if (pickedObject.getHitObject().equals(someSceneObject)) {
    pickedObject.getHitObject().getEyePointeeHolder().setEnable(false);
}

Let me evaluate other use cases also.

deepakrawat22:
Hi, Added a new API so get rid of getOwnerObject from all sample application. As per Chao suggestion, still need to give some though to GVREyePointeeHolder. May be need to expose some API so that getOwnerObject should never be called in that case too (Even though app uses pickScene)

deepakrawat22:
Now can compile the framework successfully with getOwnerObject() as protected.

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
```diff
@@ -652,6 +652,36 @@ public void remove() {
...
+    /**
+     * Add {@code childComponent} as a child of this object (owner object of the
+     * component is added as child). Adding a component will increase the
+     * {@link getChildrenCount() getChildrenCount()} for this scene object.
+     * 
+     * @param childComponet
+     *            {@link GVRComponent Component} to add as a child of this
+     *            object.
+     */
+    public void addChildComponent(GVRComponent childComponent) {
```
jason2kim: It looks good to me now. I am also okay with overloading to addChildObject(GVRComponent com), instead of new term of “xxxComponent”.

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
```diff
@@ -652,6 +652,36 @@ public void remove() {
             throw new UnsupportedOperationException();
         }
     }
+
+    /**
+     * Add {@code childComponent} as a child of this object (owner object of the
+     * component is added as child). Adding a component will increase the
+     * {@link getChildrenCount() getChildrenCount()} for this scene object.
+     * 
+     * @param childComponet
```
thomasflynn: childComponent is misspelled here.

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
thomasflynn: agreed.  change to addChildObject(GVRComponent component)

chenchao1407:
Verified.

PR #143, opened 2015-06-26
Added Support to Multipass rendering and Created a Sample to showcase outline rendering.

phi-lira:
* Added support to cull front faces. (Used in outline and will be useful for some other effetcs)
* Added render_pass to native side which encapsulates for now material and cull state. More states like polygon offset can be added later. 
* Updated shaders to get material for the current pass.
* On Java side there are overrides to hide from normal user pass information. So, for most cases on a single pass, one will not need to worry about it. On Native side, there's no default parameters to query pass. It must be done explicitly, this was done to avoid errors assuming rendering only have one pass.
* A sample was created to showcase the multpass effect. For now, since fbx loading is not working, there's only a cube spinning with outline.

Future:
I plan to add GVRPass to java side. User can configure pass more easily and still maintain overrides to hide pass information for normal users.

f.lira@samsung.com


GVRf/Framework/src/org/gearvrf/GVRRenderData.java
```diff
@@ -321,22 +358,47 @@ public void setRenderingOrder(int renderingOrder) {
...
+    /**
+     * Set the face to be culled 
+     * 
+     * @param cullFace
+     *            {@code GVRCullFaceEnum.Back} Tells Graphics API to discard back faces,
+     *            {@code GVRCullFaceEnum.Front} Tells Graphics API to discard front faces,
+     *            {@code GVRCullFaceEnum.None} Tells Graphics API to not discard any face
+     * @param pass The rendering pass to set cull face state
      */
-    public void setCullTest(boolean cullTest) {
```
thomasflynn: it looks like you removed setCullTest() and replaced it with setCullFace().  While I like set setCullFace() better :), we should deprecate setCullTest()/getCullTest() first and then remove them in a later release.

GVRf/Framework/jni/engine/renderer/renderer.cpp
```diff
@@ -504,66 +502,82 @@ void Renderer::renderRenderData(RenderData* render_data,
...
-                LOGE(
-                        "Error detected in Renderer::renderRenderData; name : %s, error : %s",
-                        render_data->owner_object()->name().c_str(),
-                        error.c_str());
-                shader_manager->getErrorShader()->render(mvp_matrix,
-                        render_data);
+
+            for (int curr_pass = 0; curr_pass < render_data->pass_count(); ++curr_pass) {
+
+				numberTriangles += render_data->mesh()->getNumTriangles();
```
thomasflynn: could you check your formatting?  that looks like way too many spaces in indentation.  makes the diff harder to read.  check: http://gearvrf.org/bin/view/GearVRF/GearVRfWikiGetInvolved#CodeGuide

thomasflynn:
Just did a quick skim of the code.  overall, looks fine with some exceptions.  we should deprecate methods before removing them and formatting seems off in some places making it harder to read.

phi-lira:
Changes applied according to review.

GVRf/Framework/jni/engine/renderer/renderer.cpp
```diff
@@ -504,66 +502,80 @@ void Renderer::renderRenderData(RenderData* render_data,
...
-            } catch (std::string error) {
-                LOGE(
-                        "Error detected in Renderer::renderRenderData; name : %s, error : %s",
-                        render_data->owner_object()->name().c_str(),
-                        error.c_str());
-                shader_manager->getErrorShader()->render(mvp_matrix,
-                        render_data);
+            for (int curr_pass = 0; curr_pass < render_data->pass_count(); ++curr_pass) {
+            	numberTriangles += render_data->mesh()->getNumTriangles();
+				numberDrawCalls++;
```
thomasflynn: take a look at the diff here on github.  you'll see this line as well as all the rest below it are indented like 16 characters.  it looks like you fixed the indentation on the previous line, now just fix it for the rest :)

GVRf/Framework/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -0,0 +1,11 @@
+eclipse.preferences.version=1
+org.eclipse.jdt.core.compiler.codegen.inlineJsrBytecode=enabled
+org.eclipse.jdt.core.compiler.codegen.targetPlatform=1.7
+org.eclipse.jdt.core.compiler.codegen.unusedLocal=preserve
+org.eclipse.jdt.core.compiler.compliance=1.7
+org.eclipse.jdt.core.compiler.debug.lineNumber=generate
+org.eclipse.jdt.core.compiler.debug.localVariable=generate
+org.eclipse.jdt.core.compiler.debug.sourceFile=generate
+org.eclipse.jdt.core.compiler.problem.assertIdentifier=error
+org.eclipse.jdt.core.compiler.problem.enumIdentifier=error
```
chenchao1407: I think it is not recommended to push anything in .settings

GVRf/Framework/src/org/gearvrf/GVRRenderData.java
```diff
@@ -325,7 +362,34 @@ public void setRenderingOrder(int renderingOrder) {
      *         not.
```
thomasflynn: make sure to mark the methods as deprecated with the @deprecated tag.  See setKeepWrapper() in GVRHybridObject.java as an example.

GVRf/Framework/jni/objects/components/render_data.h
```diff
@@ -85,13 +120,19 @@ class RenderData: public Component {
...
+    	if (pass >= 0 && pass < render_pass_list_.size()) {
+    		return render_pass_list_[pass]->cull_face();
+    	}
 
-    void set_cull_test(bool cull_test) {
-        cull_test_ = cull_test;
-    }
+    	return nullptr;
+	}
+
```
chenchao1407: The format for this file looks weird to me.

Please make sure that in Preferences->C++->Code Style. Pick K&R[built-in] profile, change tab policy from "Tab only" to "Spaces only". And do Ctrl+Alt+F for each file you modified.

Thanks in advance.

GVRf/Framework/jni/engine/renderer/renderer.cpp
phi-lira: I'm sorry identation is correct here, even on the first time you commented I just removed the extra line.

This is probably due to a tab vs spaces configuration. Now I see you have a configuration for that. I'll apply the auto formatters and get it done. Sorry :(

GVRf/Framework/.settings/org.eclipse.jdt.core.prefs
phi-lira: Thank Chao. I'm unfamiliar with Eclipse and I'm sorry about that. Will remove from next commit. 
I'll also update .gitignore file so more people don't make the same mistake.

GVRf/Framework/src/org/gearvrf/GVRRenderData.java
phi-lira: Do you mean CullTest methods?

thomasflynn: oh, yes.  i mis-clicked.  I did mean the cullTest methods

GVRf/Framework/jni/objects/components/render_data.h
phi-lira: Will do.

GVRf/Framework/jni/objects/components/render_data.h
```diff
@@ -85,13 +120,19 @@ class RenderData: public Component {
...
+    bool cull_face(int pass = 0) const {
+    	if (pass >= 0 && pass < render_pass_list_.size()) {
+    		return render_pass_list_[pass]->cull_face();
+    	}
 
-    void set_cull_test(bool cull_test) {
-        cull_test_ = cull_test;
-    }
+    	return nullptr;
+	}
```
chenchao1407: Even though I think nullptr here equals with false, but I think "return false" is more accurate? Just my personal opinion.

phi-lira:
New commits according to reviews:

 * Removed .prefs files
 * Updated .gitignore file
 * Applied Auto Format on files that had tab changes.
 * Added Deprecated to CullMethods. (Removed CullMethods with pass since it didn't make sense anymore because of deprecation);

GVRf/Framework/src/org/gearvrf/GVRRenderData.java
phi-lira: Ok. Did changes but still have one concern. I did not had the care to keep and deprecate changes on native side. So, some interfaces changed. (such as culltest methods)

Do you want me to review interface changes on native?

thomasflynn:
native side is fine.  we can make whatever change we need there.  The public Java APIs are the ones we tell developers that we support.  So, the public Java APIs should be treated with care :).

GVRf/Framework/src/org/gearvrf/GVRRenderData.java
```diff
@@ -261,7 +279,12 @@ public boolean isDone() {
...
         }
     }
-
+    
+    public void addPass(GVRMaterial material, int cullFace) {
+    	NativeRenderData.addPass(getNative(), material.getNative(), cullFace);
+    }
+    
+    
     /**
```
chenchao1407: Imho, I think you may need a getter which return a list of RenderPass. And also, for better maintenance, you may need a java class for RenderPass. Currently it is a little bit difficult for developer to know exactly how many renderpasses are in a RenderData instance.

phi-lira: True. I commented this as Future Plans when I created the pull request. I planned to make this on next pull request. I implemented it this way because I won't be able to work on this for the next week and I wanted to push it as it is stable.

More rendering states that make sense to set per pass, like polygon offset, are going to be added on RenderPass too.

GVRf/Framework/src/org/gearvrf/GVRRenderData.java
```diff
@@ -261,7 +279,12 @@ public boolean isDone() {
             return mFutureMesh.isDone();
         }
     }
-
+    
+    public void addPass(GVRMaterial material, int cullFace) {
+    	NativeRenderData.addPass(getNative(), material.getNative(), cullFace);
+    }
+
```
chenchao1407: Since cullFace only has 3 values here. Probably an enum here as parameter is more accurate.
Just FYI, if I were you, I may write an enum like
public static enum GVRCullFaceEnum{
      Back(0), Front(1), None(2);
      private final int value;
      private GVRCullFaceEnum(int value){
             this.value = value;
      }
      int getValue(){
            return value;
      }
}

public void addPass(GVRMaterial material, GVRCullFaceEnum cullFace){
        NativeRenderData.addPass(getNative(), material.getNative(), cullFace.getValue());
}

By this, you can restrict the developers that they can only use 0,1,2 to set cullFace.
Just FYI, not a big deal here.

phi-lira:
New Commit with GVRCullFaceEnum changes suggested by chao. 

I had to make an interface apart from constructor to convert from native int to GVRCullFaceEnum.

GVRf/Framework/jni/engine/renderer/renderer.cpp
```diff
@@ -504,66 +500,83 @@ void Renderer::renderRenderData(RenderData* render_data,
...
+                            break;
+                        case Material::ShaderType::CUBEMAP_REFLECTION_SHADER:
+                            shader_manager->getCubemapReflectionShader()->render(
+                                    mv_matrix, glm::inverseTranspose(mv_matrix),
+                                    glm::inverse(view_matrix), mvp_matrix,
+                                    render_data, curr_material);
+                            break;
+                        default:
+                            shader_manager->getCustomShader(
+                                    curr_material->shader_type())->render(
```
jason2kim: the argument of getCustomShader(): isn't it supposed to an ID?

GVRf/Sample/gvr-outlinesample/src/org/gearvrf/gvroutlinesample/OutlineScript.java
```diff
@@ -0,0 +1,184 @@
...
+import org.gearvrf.GVRSceneObject;
+import org.gearvrf.GVRScript;
+import org.gearvrf.GVRTexture;
+import org.gearvrf.GVRMaterial;
+import org.gearvrf.GVRMesh;
+import org.gearvrf.GVRMaterial.GVRShaderType;
+
+public class OutlineScript extends GVRScript {
+
+    private GVRContext mGVRContext = null;
```
jason2kim: Unnecessary code to initialize as null in Java codes: = null;

GVRf/Sample/gvr-outlinesample/src/org/gearvrf/gvroutlinesample/OutlineScript.java
```diff
@@ -0,0 +1,184 @@
...
+            16, 17, 18, // top
+            18, 17, 19, //
+
+            20, 21, 22, // bottom
+            22, 21, 23 //
+    };
+
+    OutlineScript(GVRActivity activity) {
+        mActivity = activity;
+    }
```
jason2kim: Remove unused variable: mActivity, the same is this contructor with the arg

jason2kim:
I am not sure why these two files of framework projects were uploaded here? 
M GVRf/Framework/.cproject (60) 
M GVRf/Framework/.project (12)

GVRf/Framework/jni/engine/renderer/renderer.cpp
phi-lira: Yes. shader_type returns an ID despite the misleading name. My changes are related to curr_material taken from each pass. I did not made any changes to the getCustomShader signature. However I did changes to most of shader render signatures to add a material as parameter.

phi-lira:
Ok. I'll revert those.

GVRf/Framework/jni/engine/renderer/renderer.cpp
jason2kim: Yes you are right. The name gives some confusion for the case of custom shaders now.  
I agree that the signature was not changed with this pull request. But, I am seeing a crash when there is a call, other than the base pass, for example pass 1. The shader type or shader ID was not initialize correctly somehow. Do you see similar to this on your device?:
Fatal signal 6 (SIGABRT) at 0x000033af (code=-6), thread 13268 (OVR::VrThread)
pid: 13231, tid: 13268, name: OVR::VrThread  >>> org.gearvrf.gvroutlinesample <<<

thomasflynn:
Jason, have you re-checked after phil's latest check-ins addressing your concerns?

jason2kim:
Yes, I re-checked it today. I am waiting for Phil’s response on the Exception case above.

GVRf/Framework/jni/engine/renderer/renderer.cpp
phi-lira: Hi Jason, 

Did you get this crash on the orifinal gvroutlinesample or did you get it passing another shader other than unlit in pass 1? I'll check this tomorrow. I don't have access to gear at home.

phi-lira:
Ok. I'll take a look as sooner as I can on this.

@thomasflynn. Can you confirm GearVRf is running on S5 with latest commit? I tried it on my S5 this weekend and It's not launching. 

@chenchao1407 & @jason2kim I'm really appreciating the feedbacks. This is my first time working on Java and they are being really insightful for me. I'll get all of them fixed and if you happen to think about more improvement don't refrain to say. :D.

thomasflynn:
standby...

GVRf/Framework/jni/engine/renderer/renderer.cpp
jason2kim: Phil, to reproduce it, leave the app run for a while, say for more than 10 minutes. It can be problem of my test device, but is consistent on my Note 4.

thomasflynn:
had to fight with eclipse for a bit.  ran on the S5, S6, Note 4 KitKat, and Note 4 Lollipop.  I saw much more of a halo on the S6.  S6 has a Mali gpu, the others I tested have a QualComm Adreno gpu.  I'll take some screenshots.

thomasflynn:
screenshot_2015-06-29-19-04-37
this is from S6

thomasflynn:
screenshot_2015-06-29-22-07-43
This is from Note4

phi-lira:
Thanks Tom. It's something on my end then. 

Regarding the outline difference I'll keep a note to test thickness with highp to see if this is a mali precision error.

phi-lira:
Hi, I found the source of the crash. It was due to I'm forgetting to keep a reference on java side for additional pass material. Then when the GC kicked in it was removing this unreference material and deleting native pointer. (thus the explanation why it was consistently happening after some minutes)

The problem was aggraveted by the fact that NativeHybridObject_delete was not setting pointer to null after deleting it giving invalid memory access (instead of null pointer);

In order to fix this it will be better to implement GVRRenderPass on Java side, so hang on for a while and I'll tell you when I'm done.

jason2kim:
I am really happy to hear that you could reproduce it and locate the root cause. It is in line with my preliminary debugging as above. It showed random ID numbers when 
getCustomShader(int id) was called and crashed, e.g., id = 2043010648, 1075258008. Strongly suspected as initialization failure or gc memory reclaim.
I hope see your new patch soon. Thanks.

phi-lira:
Crash is fixed and I created a GVRRenderPass interface. 
Although demo is working I think still not stable to merge. 

I'm having a problem with this code on GVRRenderData:
public void setMaterial(GVRMaterial material, int passIndex) {
        if (passIndex < mRenderPassList.size()) {
            mRenderPassList.get(passIndex).setMaterial(material); <- this takes no effect 
            NativeRenderData.setMaterial(getNative(), material.getNative(), passIndex);  <- this was added as workaround.

I'll be looking into it as soon as I can.

phi-lira:
Regarding the framework itself I found two issues.

1) After a while the cube was distorting in a shear/scale like effect. This was caused by passing in rotateByAxis a non-normalized vector. I think rotateByAxis should normalize vector as most people would either expect this or not be aware of normalization problems. If performance is a concern I think it would be better to create a rotateByAxisNormalized or something similar that expect normalized vectors. 

I already have this normalization done in transform.cpp. I could commit it on a next pull request or along with this if you want.

2) I also noticed that even if normalized axis rotateByAxis is either translating object or rotating it around world (0, 0, 0). After 5 min runnning gvroutlinesample you'll see cube keeps going to the left. I don't know what's the cause for this, perhaps a small error on transform matrix.

Have you notices this somewhere else?

thomasflynn:
1) yes, i think it would make sense to normalize the vector.
2) I think the only thing we have that has something using that method would probably be the solar system app.  Since you're moving around in there anyway, i doubt anyone would notice if there was a drift.  Is the drift happening in the headset? or when holding up the phone.  If you're not docked into the headset, then the phone's sensors are being used and those sensors definitely experience drift.  It's inherent to those sensors.

thomasflynn:
With regards to your setMaterial() problem, is it a VAO thing?  Maybe it needs to be dirtied so the VAO gets rebuilt?

phi-lira:
Ok. It seems everything is fine now. Should I merge with upstream and commit or should I leave you to do it?

@thomasflynn The drift is happening in the headset. I didn't know about this sensor behaviourd. Good to know.

phi-lira:
I'll commit vector normalize on next PullRequest to not mix with this.

thomasflynn:
ok, it sounds like you're ready for this to be reviewed again?

phi-lira:
Yes. Let me know if you required any actions.

GVRf/Framework/jni/objects/hybrid_object_jni.cpp
```diff
@@ -28,7 +28,9 @@ Java_org_gearvrf_NativeHybridObject_delete(JNIEnv * env,
 JNIEXPORT void JNICALL
 Java_org_gearvrf_NativeHybridObject_delete(JNIEnv * env,
         jobject obj, jlong jhybrid_object) {
-    delete reinterpret_cast<HybridObject*>(jhybrid_object);
+    HybridObject* hybrid_object = reinterpret_cast<HybridObject*>(jhybrid_object);
+    delete hybrid_object;
+    hybrid_object = nullptr;
```
chenchao1407: I don't think you have to do this.

chenchao1407: Ok, I just saw the comment you made on this hybrid_object = nullptr.

Even though it still looks weird to me(which it means hybrid_object finally becomes a dangling pointer), but perhaps I made a mistake here.

chenchao1407:
It seems that it works fine at Note4. @thomasflynn  Can you help try on S6?

GVRf/Framework/jni/objects/hybrid_object_jni.cpp
chenchao1407: Ok. Finally I got something that proved my concern here.

The native crash on GC is still there, please check my log information:
07-06 14:38:20.875   358   358 I DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
07-06 14:38:20.875   358   358 I DEBUG   : Build fingerprint: 'samsung/trltexx/trlte:5.0.1/LRX22C/N910FXXE1BOD2:eng/test-keys'
07-06 14:38:20.875   358   358 I DEBUG   : Revision: '12'
07-06 14:38:20.875   358   358 I DEBUG   : ABI: 'arm'
07-06 14:38:20.875   358   358 I DEBUG   : pid: 21628, tid: 21686, name: GVRF Finalize T  >>> org.gearvrf.gvroutlinesample                                                 <<<
07-06 14:38:20.875   358   358 I DEBUG   : signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x0
07-06 14:38:20.915   358   358 I DEBUG   :     r0 af254dc0  r1 a048fadc  r2 af254dc0  r3 00000000
07-06 14:38:20.915   358   358 I DEBUG   :     r4 000002b6  r5 76efe7d0  r6 76efe740  r7 00000000
07-06 14:38:20.915   358   358 I DEBUG   :     r8 32c23e80  r9 b4e0e400  sl af254dc0  fp ffffffff
07-06 14:38:20.915   358   358 I DEBUG   :     ip b3bc9a3c  sp a048fac8  lr b3bc9a54  pc 00000000  cpsr a00d0010
07-06 14:38:20.915   358   358 I DEBUG   : 
07-06 14:38:20.915   358   358 I DEBUG   : backtrace:
07-06 14:38:20.915   358   358 I DEBUG   :     #00 pc 00000000  <unknown>
07-06 14:38:20.915   358   358 I DEBUG   :     #01 pc 000c4a50  /data/app/org.gearvrf.gvroutlinesample-2/lib/arm/libgvrf.so (Java_org_gearvrf_NativeHybridObject_delete+20)
07-06 14:38:20.915   358   358 I DEBUG   :     #02 pc 0008e339  /data/dalvik-cache/arm/data@app@org.gearvrf.gvroutlinesample-2@base.apk@classes.dex
07-06 14:38:22.765   358   358 I DEBUG   : 
07-06 14:38:22.765   358   358 I DEBUG   : Tombstone written to: /data/tombstones/tombstone_02


Actually as far as I can see "hybrid_object = nullptr" may not work as you expected since hybrid_object is just a copy of the native pointer from Java side(because primitive type parameter will always be passed by value not reference.) It means , probably, it may be the case that there are more than one java objects are holding the same native pointer and it finally leads to this kind of duplicate removal of dangling pointer. I will try to further evaluate this problem and see if I can reach to a conclusion later,

chenchao1407: OK. Problem found and here is a solution.

The reason why GC will crash with RenderPass class is not because of dangling pointer. The reason is that in Java side, you define GVRRenderPass as a subclass of GVRHybridObject. But in c++ side, RenderPass is not a subclass of HybridObject.

So when the codes reach to NativeHybridObject_delete, it tries to turn any pointer from GVRHybridObject into a HybridObject pointer in C++, but at the moment, RenderPass is actually not a HybridObject pointer. That is why delete will finally lead to a crash. Actually you don't have to do "hybrid_object = nullptr;" . Instead, you just need to set RenderPass in C++ as a subclass of HybridObject and the crash will go away(I tried).

phi-lira:
Ok. Fixed crash Chao pointed out by making RenderPass inherit from HybridObject. Also reverted changes to hybrid_object_jni.cpp

GVRf/Framework/jni/objects/hybrid_object_jni.cpp
phi-lira: Thanks Chao. I commited the updated changes to fix it plus reverted hybrid_object_jni.cpp.

I still have some questions. 

1) How should we set native pointer to null to avoid native code from mistakenly referencing deleted pointers? Cause I think right now, doing checks for nullptr in native code are non effective.

2) I'm a little confused with GVRHybridObject. There's a mNativePointer on GVRHybridObject and another copy of it in GVRReference. The GVRHybridObject initialized both references upon construction.

mNativePointer = nativePointer;
sReferenceSet.add(new GVRReference(this, nativePointer, cleanupHandlers));

When closing an object the GVRReference native pointer is deleted and set to null but shouldn't GVRHybridObject native pointer also be set to null?

jason2kim: So, you guys are still working on this, right? I am seeing a SIGSEGV crash:

07-07 12:46:10.770 27961 28013 F libc    : Fatal signal 11 (SIGSEGV) at 0x00000010 (code=1), thread 28013 (GVRF Finalize T)
07-07 12:46:10.800 27961 28041 W TimeWarp: V-sync 5756: Eye 1, CPU latency 0.007, GPU latency 0.002, Total latency = 0.009
07-07 12:46:10.830   317   317 I DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
07-07 12:46:10.830   317   317 I DEBUG   : Build fingerprint: 'samsung/trltetmo/trltetmo:4.4.4/KTU84P/N910TUVU1ANK4:user/release-keys'
07-07 12:46:10.830   317   317 I DEBUG   : Revision: '12'
07-07 12:46:10.830   317   317 I DEBUG   : pid: 27961, tid: 28013, name: GVRF Finalize T  >>> org.gearvrf.gvroutlinesample <<<
07-07 12:46:10.830   317   317 I DEBUG   : signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 00000010
…
07-07 12:46:10.980   317   317 I DEBUG   : backtrace:
07-07 12:46:10.980   317   317 I DEBUG   :     #00  pc 00000010  <unknown>
07-07 12:46:10.980   317   317 I DEBUG   :     #01  pc 000c4b40  /data/app-lib/org.gearvrf.gvroutlinesample-15/libgvrf.so (Java_org_gearvrf_NativeHybridObject_delete+20)
07-07 12:46:10.980   317   317 I DEBUG   :     #02  pc 000210cc  /system/lib/libdvm.so (dvmPlatformInvoke+112)
07-07 12:46:10.980   317   317 I DEBUG   :     #03  pc 00051daf  /system/lib/libdvm.so (dvmCallJNIMethod(unsigned int const*, JValue*, Method const*, Thread*)+398)
…

phi-lira: @jason2kim 

I tried in S6 and Note4 (Snapdragon) and could not reproduce this crash. 
I'm still using Mobile SDK 0.6.0 if that matters.

Do you have my latests commit? Can you share more details on how you got this crash?

chenchao1407:
OK. I think there is another problem which leads to crash according to jason's report.

The second problem which leads to the crash is because of dangling pointer. 

Let me shortly explain GVRReference.  You may need to know PhantomReference and ReferenceQueue in Java firstly. In each GC cycle, anything that is going to be recycled, before it gets released, you can find the PhantomReference(which is GVRReference in GVRF) of it in the ReferenceQueue(which is sReferenceQueue). That is how current GVRF works with its memory management. When any GVRHybridObject is going to be released, we check PhantomReference of it in the ReferenceQueue and try to call native delete on its mNativePointer to also release the according C++ memory.

Here is the root cause of the crash: in C++'s RenderData class, you delete all RenderPass elements in the destructor. So when a Java GVRRenderData is going to be release, the c++ RenderData will release all bound RenderPass memory. However, when the GVRRenderPass is going to be release, it is trying to release the RenderPass memory which was released before in RenderData's destructor.(Order is not fixed, it is possible that GVRRenderPass releases its own native memory firstly and then GVRRenderData will try to delete those deleted RenderPass pointers.) This is the root cause why we can see the crash when GC cycle is triggered.

So here are two potential solutions:
1, Not making RenderPass a subclass of HybridObject(Both Java and C++).
2. Just remove all codes in RenderData's destructor.

You can take either one.

phi-lira:
Thanks Chao. I did solution 2 as I think this will be more consistent to most of GVR code.

chenchao1407:
@phi-lira 

You may need to do some manual rebase since our code base is just changed by another pull request

phi-lira:
Merged.

GVRf/Framework/jni/objects/components/render_data.h
```diff
@@ -43,15 +44,20 @@ class RenderData: public Component {
...
+            Component(), mesh_(0), light_(0), use_light_(false), render_mask_(
                     DEFAULT_RENDER_MASK), rendering_order_(
-                    DEFAULT_RENDERING_ORDER), cull_test_(true), offset_(false), offset_factor_(
+                    DEFAULT_RENDERING_ORDER), offset_(false), offset_factor_(
                     0.0f), offset_units_(0.0f), depth_test_(true), alpha_blend_(
                     true), draw_mode_(GL_TRIANGLES) {
     }
 
     ~RenderData() {
+        render_pass_list_.clear();
```
jason2kim: The object management part looks okay now.

PR #144, opened 2015-06-29
Removing per-frame log spew.

thomasflynn:
Some log spew got checked in during the 0.6.0 port.  Removing it now.

tom.flynn@samsung.com


chenchao1407:
Looks good. I think it is ok to merge.

PR #145, opened 2015-06-29
Add a position animation.

thomasflynn:
We had animations for rotation and scale, but not position.  This
eliminates that oversight.  I'll check-in an example when I do my LOD
check-in later in the week.

tom.flynn@samsung.com


chenchao1407:
Code looks fine, will test later.

jason2kim:
codes look good to me too

Issue #89, opened 2015-05-29
Do culling per-scene, not per-eye.


thomasflynn:
Currently, frustum culling is done in renderCamera() in renderer.cpp.  renderCamera() is called per-eye.  It would be more ideal to do that culling once per frame, rather than twice.  That culling should be perhaps done against a combined left/right eye frustum.

thomasflynn:
![frustumcull](https://cloud.githubusercontent.com/assets/9667914/8237288/07753ffe-15a3-11e5-9717-044a32577df6.jpg)

thomasflynn:
I'm working on this now. 

The first step here will be to create a center camera for the camera rig.  That is done as part of https://github.com/Samsung/GearVRf/pull/202

Second step will be to move the culling to a method that happens before calling of renderCamera for either left or right eyes.

Third step will be to make the bounding volumes hierarchical so that whole subgraphs can be eliminated if their parent's bounding volume is not in the frustum.

Forth actually traverse the tree as a tree (rather than getWholeSceneObjects which returns the whole tree as a linked list) and do testing for frustum, visibility, and render mask.

thomasflynn:
this was done back in 2.3 or so.

PR #146, opened 2015-06-30
Lod

thomasflynn:
Added level of detail support.  SceneObject now has a setLODRange() method for setting the range in which that SceneObject should be visible.

tom.flynn@samsung.com


soumyajitdeb:
This code looks good overall. My main query is regarding whether using a pure distance from camera based metric might be sustainable in the long run. The ideal way to evaluate the LOD should be to use a coverage based metric based on pixels of screen space that the rendered object occupies. That way, objects that are close to the camera but tiny or at the peripheries of the screen would get the correct LOD and large sized semi distant objects would also get the correct LOD evaluated.

On desktop, the easy way to get this is to store the results of the previous frame's occlusion query (number of pixels rendered) and use that in the coverage function. However on android with binary occlusion queries, things get more difficult. While I do not have an exact solution at this very moment, we are exploring this issue in another context and it might be a good thing to revisit later.

-Soumyajit

thomasflynn:
As you mention, binary occlusion query limits our options.  Inventor did LOD selection based on screen space coverage, but did it by projecting the bounding box and measuring the resulting square area of pixels...all on the CPU.  As you can imagine, this wasn't necessarily a performance win.  Performer, OpenSceneGraph and others went the route of using distance from the camera which was fast and worked well enough to be a performance win.  OpenSceneGraph has the option of using screen pixels if you want to do so, but i believe they use occlusion query to accomplish that.

So, yeah, disappointingly, GLES still hasn't adopted full occlusion query and that limits our options :(

jason2kim:
Please remove those test oculussig_*

jason2kim:
Please remove those not used import lines in the sample Java files

thomasflynn:
d'oh!  will do.

GVRf/Framework/src/org/gearvrf/scene_objects/GVRSphereSceneObject.java
```diff
@@ -49,7 +49,35 @@
     public GVRSphereSceneObject(GVRContext gvrContext) {
         super(gvrContext);
 
-        generateSphere(STACK_NUMBER, SLICE_NUMBER);
+        generateSceneObject(gvrContext, STACK_NUMBER, SLICE_NUMBER);
+    }
+
+    /**
+     * Constructs a sphere scene object with a radius of 1 and 18 stacks, and 36
```
jason2kim: =>… of the given number of stacks and slices

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
```diff
@@ -571,6 +571,36 @@ public boolean isColliding(GVRSceneObject otherObject) {
     }
 
     /**
+     * Sets the range of distances from the camera where this object will be shown.
+     *
+     * @param minRange
+     *      The closest distance to the camera rig in which this object should be shown.  Default value is 0.
+     * @param maxRange
+     *      The farthest distance to the camera rig in which this object should be shown.  Default value is Float.MAX_VALUE.
```
jason2kim: Default value is 0. Default value is Float.MAX_VALUE.
=>Some dilemma…Are these meaningful as they will never be used until the app calls setLODRange()? Once called, the preset default values will be replaced. As you know, the current implementation does force LOD by default.

GVRf/Framework/jni/engine/renderer/renderer.cpp
```diff
@@ -259,12 +258,30 @@ void Renderer::frustum_cull(Scene* scene,
 
         // Only push those scene objects that are inside of the frustum
         if (!is_inside) {
-            (*it)->set_in_frustum(false);
+            scene_object->set_in_frustum(false);
+            continue;
+        }
+
```
jason2kim: How about adding another api here, e.g. scene_object->isLODUsed()? If false, we can skip calculating steps of camera distance (the default is in-range when lod is not in use)?
It can be a minor optimization considering setLODRange() is currently set only from app but frustum_cull() is in the framework.

thomasflynn: A reasonable request, but we will also need the camera distance for sorting transparent object back-to-front and opaque objects front-to-back.

jason2kim: Okay. The camera distance value calculated here can be used for multipurposes, like object ordering, in the future. May be some comments in the codes useful?

thomasflynn: ok, fair enough.  I'll add a TODO.

GVRf/Framework/jni/objects/scene_object_jni.cpp
```diff
@@ -213,6 +213,28 @@ Java_org_gearvrf_NativeSceneObject_isColliding(
...
+    return scene_object->getLODMinRange();
 }
+
+JNIEXPORT jfloat JNICALL
+Java_org_gearvrf_NativeSceneObject_getLODMaxRange(
+        JNIEnv * env, jobject obj, jlong jscene_object) {
+    SceneObject* scene_object = reinterpret_cast<SceneObject*>(jscene_object);
+    return scene_object->getLODMaxRange();
+}
+
```
chenchao1407: Normally, I would say, in a JNI file like this, you should declare a JNI function and then implement it. For setLODRange, getLODMinRange, and getLODMaxRange, you just provided the implementation here without declaration. However since it compiles and runs, then it should be fine even though it seems to me that in some environments it won't compile or at least it will show some warnings... Not quite sure about this.

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
thomasflynn: Yeah, good point.  Perhaps I don't need to point that out the defaults here, but instead in getLODMinRange() and getLODMaxRange().

GVRf/Framework/jni/objects/scene_object_jni.cpp
thomasflynn: No, that's fair enough.  I somehow missed the declarations up above.  Will be in my next check-in.

thomasflynn:
ok, new commit is up.  It addresses the feedback so far.

thomasflynn:
any further feedback?

jason2kim:
it looks good for me

thomasflynn:
tweaked the javadoc for setLODRange per suggestions by Chao.

chenchao1407:
Two small concerns:
1. Rotation of camera will make objects disappear or appear when the objects are close to the LOD edges. It makes sense to me since actually we have left camera and right camera so that it will somehow make rotation of camera change the distance. 

2. When you are rotating the camera, you will find that the appearance of the object in left and right cameras are not synchronized. I mean, you can see some object on the one camera but not on the other one. Actually it is still because of the reason I mentioned above.

Both side effects make sense to me but finally it is asking me to think about if there is any way to use centerview camera position to judge the LOD condition? And instead of doing the calculation twice for each frame, we can just do the filtering only once with centerview position.

thomasflynn:
This will be solved once issue #89 is implemented.  Culling in general, should be done once per-frame and  not per-eye.  The distance to camera calculation is done in the frustum_cull() routine.  Solving #89 is too much work to add into this check-in.

chenchao1407:
OK. Then good to me now.

Any other comment?

Issue #74, opened 2015-05-23
Input samples


mkeblx:
It would be useful to have some samples that demonstrate usage of external input like a gamepad and keyboard. A lot of applications beyond gaming could use good text entry support.

tflynnt:
I tried replying to the email notification of then when i got this earlier today.  Apparently it wasn't enough.  Here was my reply:

Fair enough.  We can create some samples that include various input.  Here are some references in the meantime:

for keyboard input, override onKeyDown in your activity:  http://developer.android.com/reference/android/app/Activity.html

Gamepad:  https://developer.android.com/training/game-controllers/controller-input.html

mkeblx:
Hmm, I couldn't get onKeyDown/onKeyUp to invoke. Had to resort to shimming with dispatchKeyEvent:

```java
    @Override
	public boolean dispatchKeyEvent(KeyEvent event) {
		int keyCode = event.getKeyCode();

		if (event.getAction() == KeyEvent.ACTION_DOWN ) {
			mScript.onKeyDown(keyCode, event);
		} else if (event.getAction() == KeyEvent.ACTION_UP) {
			mScript.onKeyUp(keyCode, event);
		}

		return true;
	}
```

thomasflynn:
interesting.  we'll have to take a closer look into that.  we had problems earlier with the oculus sdk intercepting touch events and not passing them to the activity.  we had to work-around it in a similar way.  maybe something similar is happening with the keyboard.

mkeblx:
That's likely it. The physical back button near the headset touchpad is using that code path.

thomasflynn:
hi jason, can you take care of the keyboard dispatching?  should probably be a quick change.

jason2kim:
sure, i am in the middle of something, let me get to this shortly.

chenchao1407:
Hi @mkeblx 

I think your issue with key event dispatching is resolved by pull request #116 

I think you can try it now. I think the sample app will be added later.

dknaan:
I created a new sample application (#147) to demonstrate how to use a gamepad to control the scene. Fire keys change background color. Left motion key translates the scene.

thomasflynn:
Input samples Have been merged

PR #149, opened 2015-07-05
[GearVRf] Sample application: how to handle input from a gamepad (related to #74)

dknaan:
A new sample application to demonstrate how to handle input from a gamepad. Fire keys change background colors. Left motion key translates the scene.


jason2kim:
Thanks Dotan, let me review your gamepad sample shortly. I’ve got a device now^^

chenchao1407:
One concern I found which is not related to input event(or maybe any kind of event not from GL Thread) is like, it can't handle change which directly calls GL commands. So you can't do something like Texture.update() or GVRConsole.writeLine() with these kinds of events.

Codes look good to me. However I may need to find a gamepad to do the verification.

dknaan:
OK, please let me know if you have any concern regarding this pull request or regarding the performance of this sample. Thanks

jason2kim:
Sdk settings in the manifest, need to be aligned with:
    <uses-sdk
        android:minSdkVersion="16"
        android:targetSdkVersion="21" 

Handlers never return false, either consumed or not: dispatchKeyEvent() and dispatchGenericMotionEvent()
   
Need to decide to catch on which key event, UP or DOWN, but should not be both: e.g.  public boolean dispatchKeyEvent(android.view.KeyEvent event) {
        if (event.getAction()== KeyEvent.ACTION_UP) {. . .

In code files,
Can remove unused lines of  “import . . .” in the .java files
Avoid unnecessary blank lines, especially at the beginning and end of each method. I assume this is your personal coding style, but...

dknaan:
Thanks for your comments. I will take care of it when I'm back from PTO.

dknaan:
Small corrections to the code:
- dispatchGenericMotionEvent and dispatchKeyEvent call super if the event is no$
- in dispatchKeyEvent only ACTION_DOWN is handled
- unused imports and empty lines were removed
- Android minSdkVersion and targetSdkVersion were changed

jason2kim:
It looks good to me.

thomasflynn:
looks good to me as well.

thomasflynn:
i'll probably add some additional code this weekend so people can see how to track the other axes and buttons.

dknaan:
Other buttons can be added, but I implemented only a very basic sample assuming we just want to demonstrate how to handle key events from the gamepad.

jason2kim:
I think adding additional features to this basic sample would be a good idea. GearVRf users also will do that based on this sample.
I saw Android documents specific to this topic, including code examples for various controller actions

PR #151, opened 2015-07-07
Expose setting

chenchao1407:
Some explanation of this patch:

1. It exposes all configuration parameters from Oculus.
2. VrSetAppings is actually the shadow Java class from OVR::ovrSettings.
3. Two ways to do the parameters change:
a. You can set the xml.
b. Override onInitAppSettings() method in GVRActivity.(onInitAppSettings will override what you set with xml.)
4.Add a sample which shows both ways to set the configuration.

c.chen@samsung.com


jason2kim:
Even though I couldn’t verify every configuration parameter,

I could see the VrSettings parameters, e.g. VrSettings.ShowLoadingIcon, were set in the sequence of,
Initial default set at AppLocal::AppLocal()
Set by XML at GVRActivity
Set by app codes at onInitAppSettings

jason2kim:
In the sample codes, there are screen captures on touch events. It would be helpful if you can give some description how it is related to the parameter tweaks, by giving a few parameter examples.

chenchao1407:
Yeah, you are right. Default settings' initialization is actually at different places in Oculus, not limited to AppLocal(), you can say it actually happens at AppLocal::CreateWindowSurface and then do Configure to ask developer to load their customized settings.

As I mentioned, there are two ways to load your own settings. The reason why I designed like onInitSettings can finally override XML ones is because XML is static setting, sometimes developer may want to do some dynamic settings based on different cases. So I'd like to let developer to do some common static settings with XML and then do dynamic programmatic settings with onInitSettings if they need.

chenchao1407:
And actually there is no relationship between capturescreen and the settings.. I just copied from gvrsample and modified it... So if you think this may be confusing, I can remove those functions.

jason2kim:
onInit() for cube drawing must be okay to show at least something, based on the current configuration. But, I think the touch handler and screen capturing need to be gone as people can be confused by it.

To make this sample meaningful itself and survive, we may need to add some more codes that can showcase the features. For example, setting changes on touch, then rendering according to the just changed setting parameter. Or, if it is not easy, simply dumping out the current settings for verification purpose (or this is still not significant, we may need ride on to an existing sample).

thomasflynn:
Changing settings based on user interaction may not be possible since a lot of these settings either configure the framebuffer or can only happen on startup.  Dumping out the current settings sounds useful.

chenchao1407:
As a said, VrAppSettings is a shadow class of OVR::ovrSetting. So for the parameters you don't do any change, I will copy the value from ovrSetting back to VrAppSettings. Therefore, for the dumping out, you can just simply call GVRActivity.getVrAppSettings()

jason2kim:
For the framework side changes, it is okay to me as commented earlier. The remaining part is how we can shape the sample app, if ever wanted to submit one.

GVRf/Framework/jni/oculus/activity_jni.cpp
```diff
@@ -75,150 +77,292 @@ GVRActivity::~GVRActivity() {
...
+                    "Z"));
+
+
+
+    //Settings for EyeBufferParms.
+    jobject eyeParmsSettings = env->GetObjectField(vrSettings,
+            env->GetFieldID(vrAppSettingsClass, "eyeBufferParms",
+                    "Lorg/gearvrf/utility/VrAppSettings$EyeBufferParms;"));
+    jclass eyeParmsClass = env->GetObjectClass(eyeParmsSettings);
+    settings.EyeBufferParms.multisamples = env->GetIntField(eyeParmsSettings, env->GetFieldID(eyeParmsClass, "multiSamples", "I"));
```
thomasflynn: so i put a log after this line to print out the value for multisamples.  no matter what i put in the xml file, this still ends up being 2.

phi-lira:
Just as a side note, one setting that's worth exposing to be set during runtime is cpu/gpu trottling. 

So, for instance, if one wants to set gpu levels to 0 while loading to save battery that's a good thing. Also, an app/game might have scenes that are gpu intensive while others (like menus) are not. It would save significant battery time to set this in runtime/loading.

We do this for our games, not all scenes/levels required the same trottling configuration.

thomasflynn:
that's good feedback.  so what you're saying is that when someone calls VrAppSettings.setCpuLevel(), it needs to call down to the native side and make the appropriate call to adjust CPU level and do it at any time while the app is running, not just during init.

phi-lira:
Exactly. So what we've done here on Unity side is to expose interface to set this in a friendly way.
We made a function that sets both cpu and gpu levels, something like SetPowerHint(PowerSettings settings)

It's something like:

void SetPowerHint(/...)
{
        switch (hint)
        {
                   case HIGH_FAVOR_GPU: 
                            setCpuLevel(2);
                            setGpuLevel(3);
                            break;
                     

                    .....
        }
}

We have many values like (MODERATE_BALANCED, MODERATE_FAVOR_GPU, MODERATE_FAVOR_CPU, HIGH_FAVOR_GPU, HIGH_FAVOR_CPU, LOW_FAVOR_CPU, POWER_SAVE, etc)

This way it's user friendly and we don't have to be concern about invalid values, (some combinations are not valid like setting cpu and gpu to MAX)

chenchao1407:
I am not sure if it is possible to do it with oculus. As far as I can see, in oculus, how they use gpulevel and cpulevel is not open sourced and it only happens for each time they create a new surface(in onResume). Without changing oculus source code, there is few way you can change cpulevel and gpulevel at runtime. And even if you can change them, they will only be used after a new surface is created in onResume. Of course, if you want to jump out of oculus framework to set cpu and gpu, then there may be another complicated issue. For this pull request, it is actually just a wrapper for VrAppInterface::Configure API.

thomasflynn:
@phi-lira can you open up a feature request for changing cpu/gpu at runtime.  we'll cover it in a subsequent check-in.  bonus points for showing the c++ calls you make to oculus to change the cpu/gpu levels :)

phi-lira:
@thomasflynn Will do. We get it through an OculusPlugin.dll available in UnityIntegration. I'll post more details on the feature request page. I''ll have to do some research to see if it's possible to do without Unity Plugin. (Perhaps in CApi classes). 

@chenchao1407 Regarding changing levels at runtime I can confirm we're achieving this through the Oculus Plugin for Unity. To be honest, we discovered this by a blog post from chris pruett from Oculus (https://developer.oculus.com/blog/squeezing-performance-out-of-your-unity-gear-vr-game-continued/) and ended up implementing a wrapper for this.

chenchao1407:
@phi-lira  As far as I know, the Oculus Plugin is a co-product between Unity and Oculus. It is possible that Oculus open that API for Unity. But as I can see, I haven't found such run time api from Oculus Mobile SDK. Are there any codes of those two APIs in Oculus plugin for Unity?

phi-lira:
@chenchao1407 No. Basically all access to Oculus plugin from Unity comes from this dll and as far as I know there's no source code for it, but I'll double check that. 

I know the source for PC SDK is available at: https://developer.oculus.com/downloads/pc/0.6.0.1-beta/Oculus_SDK_Source/ 

I'll check it tomorrow and see if I can find something about this.

chenchao1407:
New commit added for the following issues:
1. GVRXMLParser is refactored so that all setting in current GVRF will be read from VrAppSettings.
2. Add details about Monoscopic rendering in VrAppSettings. Also deprecated the use of forceMonoScopic
in GVRActivity.
3. Refactored codes in view manager according to modifications on GVRXMLParser and VrAppSettings.
4. Added sample xmls for gvrexposeapisample. So you can switch normal quality, high quality, low quality
, mono mode, mono full screen mode by using different xml files.
5. Replaced all xmls in current sample apps.
6. Default settings is set so that even though the xml file is empty you still can run the app with
default settings.
7. VrAppSettings.toString() overrided for debugging purpose.

GVRf/Sample/gvrtextviewsample/assets/gvr_note4.xml
```diff
@@ -14,9 +14,39 @@
  * limitations under the License.
  */-->
  
-<lens name="N4">
-    <distortion unit="0.02825" k0="1.0" k1="0.125" k2="0.125" k3="0.0" disparity="-0.008" distortion-scale="1.0"/>
```
jason2kim: Are these lines of information truncated in the new scheme?

jason2kim:
"5. Replaced all xmls in current sample apps."
1) we don't support backward compatibility?
2) or, can we change only the new sample that can showcase the new xml example entries?

GVRf/Sample/gvrexposeapisample/src/org/gearvrf/gvrexposeapisample/SampleActivity.java
```diff
@@ -0,0 +1,39 @@
...
+        super.onCreate(icicle);
+        setScript(script, "gvr_note4.xml");
+    }
+
+
+    @Override
+    protected void onInitAppSettings(VrAppSettings appSettings){
+        appSettings.getEyeBufferParms().setMultiSamples(1);
+    }
+}
```
jason2kim: I still don't feel this sample is enough to either showcase or verify the api. For verification purpose, how about adding some functions in the framework that displays the new/current settings, which can be called from?

GVRf/Sample/gvrtextviewsample/assets/gvr_note4.xml
jason2kim: If those were unused ones in the current context, I think it should be okay to clean them up.

GVRf/Sample/gvrexposeapisample/src/org/gearvrf/gvrexposeapisample/SampleActivity.java
jason2kim: I don’t have good suggestions at this moment but,
1)	dynamic showcasing of changed settings, on a touch event. Considering the current architecture of VR mode, I am not sure if  there are any parameters that can be changeable once the VR mode started.
2)	toString() can be used in dumping out settings from our Java object. I wonder if there is a way to print out the final ovr settings in effect from Native side.

chenchao1407:
Updated. Add log information for ovrSetting at the end of GVRActivity::Configure. For dynamic configuration in the runtime, we may need to discuss about it and put it into another pull request if possible.

chenchao1407:
Updated

jason2kim:
Thanks. I don’t have further feedback at this moment.

chenchao1407:
Rebased.

PR #153, opened 2015-07-07
Quad faces support to GearVRf

deepakrawat22:
Added support for quad faces in mesh. GearVRf used to ignore quad faces. Now add 2 triangles for quad faces. Tested with wavefront obj and fbx file, working alright.

Will create a separate pull request for fbx file because need to figure out a correct way to get extension from Android resources.

drawat@usc.edu


chenchao1407:
Codes look fine to me. Is there any test sample you can add?

I think it is part or prerequisite for fbx support, right?

deepakrawat22:
I will send you a wavefront obj file by email. Yes this is needed for FBX support.

jason2kim:
Code looks okay to me, and I believe it was verified by sample running an .OBJ file.

Issue #100, opened 2015-05-30
Crash with AssimpImporter.


chenchao1407:
The scenario is that I tried to create a bunny in performancetest sample about 2 or 3 times per second using the following codes(actually I just wanted to put stress on GC and see how our GC thread reacts):
        mMainScene = mGVRContext.getMainScene();
        GVRCameraRig mainCameraRig = mMainScene.getMainCameraRig();
        int i = (int)(counter / 15);

```
    GVRSceneObject bunny = null;
    try {
        // we assume that the mesh and the textures are valid
        bunny = new GVRSceneObject(mGVRContext,
                mGVRContext.loadFutureMesh(new GVRAndroidResource(mGVRContext,
                        "bunny.obj")),
                        mGVRContext.loadFutureTexture(new GVRAndroidResource(
                                mGVRContext, textureNames[i
                                % textureNames.length])));
    } catch (IOException e) {
        e.printStackTrace();
        mActivity.finish();
        Log.e(TAG,
                "Mesh or texture were not loaded. Stopping application!");
    }

    Random random = new Random();

    bunny.getTransform().setPosition(0.0f, 0.0f,
            random.nextFloat() * 3.0f + 2.0f);
    bunny.getTransform().rotateByAxisWithPivot(
            random.nextFloat() * 360.0f, 1.0f, 0.0f, 0.0f, 0.0f, 0.0f,
            0.0f);
    bunny.getTransform().rotateByAxisWithPivot(
            random.nextFloat() * 360.0f, 0.0f, 1.0f, 0.0f, 0.0f, 0.0f,
            0.0f);
    bunny.getTransform().rotateByAxisWithPivot(
            random.nextFloat() * 360.0f, 0.0f, 0.0f, 1.0f, 0.0f, 0.0f,
            0.0f);
    bunny.getTransform().translate(0.0f, 0.0f, -10.0f);
    //mainCameraRig.getOwnerObject().addChildObject(bunny);

    float x = random.nextFloat() - 0.5f;
    float y = random.nextFloat() - 0.5f;
    float z = random.nextFloat() - 0.5f;
    float length = (float) Math.sqrt(x * x + y * y + z * z);
    x /= length;
    y /= length;
    z /= length;

    new GVRRotationByAxisWithPivotAnimation(bunny, //
            5.0f + random.nextFloat() * 25.0f, //
            360.0f, //
            x, y, z, //
            0.0f, 0.0f, -10.0f) //
[...]

PR #154, opened 2015-07-09
FBX and other 3d mesh support

deepakrawat22:
Now passing hint parameter to Assimp::Importer.ReadFileFromMemory() function. Without hint Assimp fail to determine the format for some mesh type and result in no mesh loading. To verify this change can try to load any fbx file. Currently only FBX mesh is supported as GearVRf don't have support for embedded texture, UV from models.

PS - Pull request #153 is needed for this patch to work.

drawat@usc.edu

### DIFF
--- a/GVRf/Framework/jni/engine/importer/importer.cpp
+++ b/GVRf/Framework/jni/engine/importer/importer.cpp
@@ -13,7 +13,6 @@
  * limitations under the License.
  */
 
-
 /***************************************************************************
  * Imports a scene file using Assimp.
  ***************************************************************************/
@@ -21,11 +20,20 @@
 #include "importer.h"
 
 namespace gvr {
-AssimpImporter* Importer::readFileFromAssets(char* buffer,
-        long size, const char * hint_string) {
+AssimpImporter* Importer::readFileFromAssets(char* buffer, long size,
+        const char * filename) {
     Assimp::Importer* importer = new Assimp::Importer();
+    char* hint = 0;
+
+    if (filename != 0) {
+        hint = strrchr(filename, '.');
+        if (hint && hint != filename) {
+            hint = hint + 1;
+        }
+    }
     importer->ReadFileFromMemory(buffer, size,
-            aiProcess_JoinIdenticalVertices | aiProcess_FlipUVs, hint_string);
+            aiProcess_JoinIdenticalVertices | aiProcess_FlipUVs, hint);
+
     return new AssimpImporter(importer);
 }
 
--- a/GVRf/Framework/jni/engine/importer/importer.cpp
+++ b/GVRf/Framework/jni/engine/importer/importer.cpp
@@ -29,14 +29,10 @@ AssimpImporter* Importer::readFileFromAssets(char* buffer,
     return new AssimpImporter(importer);
 }
 
-AssimpImporter* Importer::readFileFromSDCard(std::string str) {
+AssimpImporter* Importer::readFileFromSDCard(const char * filename) {
     Assimp::Importer* importer = new Assimp::Importer();
-    const char* c_str = str.c_str();
-    importer->ReadFile(str,
+    importer->ReadFile(filename,
             aiProcess_JoinIdenticalVertices | aiProcess_FlipUVs);
-    /* No need to load call ReadFileFromMemory as ReadFile is already called */
-    /*importer->ReadFileFromMemory(c_str, str.size(),
-            aiProcess_JoinIdenticalVertices | aiProcess_FlipUVs, 0);*/
     return new AssimpImporter(importer);
 }
 }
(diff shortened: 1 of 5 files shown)


GVRf/Framework/jni/engine/importer/importer.cpp
```diff
@@ -34,8 +34,9 @@ AssimpImporter* Importer::readFileFromSDCard(std::string str) {
     const char* c_str = str.c_str();
     importer->ReadFile(str,
             aiProcess_JoinIdenticalVertices | aiProcess_FlipUVs);
-    importer->ReadFileFromMemory(c_str, str.size(),
-            aiProcess_JoinIdenticalVertices | aiProcess_FlipUVs, 0);
+    /* No need to load call ReadFileFromMemory as ReadFile is already called */
+    /*importer->ReadFileFromMemory(c_str, str.size(),
+            aiProcess_JoinIdenticalVertices | aiProcess_FlipUVs, 0);*/
     return new AssimpImporter(importer);
```
chenchao1407: Can you remove the commented codes since they are not needed in future?

And one more concern here, did you try with readFileFromSDCard? I haven't checked it before but currently I am doubting if you should leave ReadFileFromMemory here and comment out the ReadFile( because it is wrong). When I checked "Java_org_gearvrf_NativeImporter_readFileFromSDCard" in importer_jni.cpp, I was seeing that the std::string str for readFileFromSDCard is actually the content of the file. But as you may know, the ReadFile is accepting the file name/path but not the content of the file. 

For this reason, I think in this pull request you may want to either keep ReadFileFromMemory here or change the implementation of readFileFromSDCard in importer_jni.cpp even though I know it is not your fault.   :-)

chenchao1407:
Just leave a comment for importer.cpp.

Other looks good to me.  But I am still thinking it may be better if you can just write a sample app with this pull request.

And here is a small concern with assimp importer. Actually I think we may use the importer in a safer way since there is no any fail check in current codes. It was 1 month ago when I reported a null pointer crash #100 which is related to it.(Ok. I know that I assigned it to myself :-( ) I think it may be better to refactor codes or do anything else to handle it.

deepakrawat22:
Hey @chenchao1407,

Thanks for your comments. I will change the implementation of Java_org_gearvrf_NativeImporter_readFileFromSDCard in importer_jni.cpp file, to pass the filename instead of file content. Earlier I missed the code that is reading file content and I was under impression that filename is passed. I will prefer this method otherwise we will have the 2 functions in importer.cpp with same implementation (i.e. reading from memory).

I think a sample app is not required at this stage as no new functionality by framework is added and fbx file can be loaded just like other 3d mesh file. May be later on when we support embedded texture and lights we can add a sample. I remember there was an issue request that "Decrease the number of samples" #9 

I will add an API in GVRAndroidResource by name getResourceFilename() so that user can set a name for resource for case when loaded as android resource ID (raw.ResID). What do you think about this?

Also will add null check for importer in JNI.

chenchao1407:
When you mentioned getResourceFileName(), you are trying to get the resource file name for a resource ID? And then you can pass the extension name to readFileFromMemory?

deepakrawat22:
I am sorry for typo. I want to add an API setResourceFilename(). getResourceFilename() is already there and currently will fail for resources loaded using android resource ID.

deepakrawat22:
Modified the code as per review comments. One change is that no longer need to add API setResourceFilename() as I figured a way to get the filename for the resource which were added using Android resource ID.

GVRf/Framework/src/org/gearvrf/GVRAndroidResource.java
```diff
@@ -57,6 +58,8 @@
     private final String filePath;
     private final int resourceId;
     private final String assetPath;
+    // For hint to Assimp
+    private String resFilePath;
```
jason2kim: resourceFilePath

GVRf/Framework/jni/engine/importer/importer_jni.cpp
```diff
@@ -53,8 +53,14 @@ Java_org_gearvrf_NativeImporter_readFileFromAssets(JNIEnv * env,
...
 
-    AssimpImporter* assimp_scene = Importer::readFileFromAssets(
-            buffer, size);
+    AssimpImporter* assimp_scene = NULL;
+    /* Retrieve the file extension and pass it as hint */
+    char* dot = strrchr(native_string, '.');
+    if (!dot || dot == native_string) {
+        assimp_scene = Importer::readFileFromAssets(buffer, size, 0);
+    } else {
+        assimp_scene = Importer::readFileFromAssets(buffer, size, dot + 1);
```
jason2kim: Can we make this part reused, thinking of avoiding the similar codes written more than once? If yes,
1)	Make this part as a separate function, then make it called from NativeImporter_readFromByteArray(), too
2)	Change the arg jstring hint of NativeImporter_readFromByteArray() to jstring filename
3)	Change and simplify codes accordingly
4)	And, do the same for other callers in the future

deepakrawat22:
Removed duplicate code as @jason2kim suggested. Thanks.

jason2kim:
it looks good to me now

PR #158, opened 2015-07-13
Subdivide primitive objects for performance optimization.

rongguodong:
Subdivided GVRCubeSceneObject. Each face is subdivided into NxN quads, where N is given by user.

rongguodong@hotmail.com


GVRf/Framework/src/org/gearvrf/scene_objects/GVRCubeSceneObject.java
```diff
@@ -398,4 +436,415 @@ private void createSimpleCubeSixMeshes(GVRContext gvrContext,
...
+                subMeshes[index].setNormals(normals);
+                subMeshes[index].setTexCoords(texCoords);
+                subMeshes[index].setTriangles(indices);
+                grandchildren[index] = new GVRSceneObject(gvrContext,
+                        new FutureWrapper<GVRMesh>(subMeshes[index]),
+                        futureTextureList.get(5));
+                children[5].addChildObject(grandchildren[index]);
+            }
+        }
+
```
chenchao1407: Still the same as before, the level of code duplicate is a little bit high. We may find some way to have a better code structure that can reuse the codes well.

chenchao1407:
I think overall should be fine, may talk with you tomorrow to discuss about the code duplicate issue.

rongguodong:
Code for subdivided sphere is checked in. Please review it.

thomasflynn:
code looks good at first pass, just one change request:
'segmentNumber' sounds like you will be assigning a number to a segment.
I would prefer 'numberOfSegments' or 'numberOfSubdivisions' or something along those lines.

GVRf/Sample/gvr-cubemap/src/org/gearvrf/cubemap/CubemapScript.java
```diff
@@ -62,6 +65,26 @@ public void onInit(GVRContext gvrContext) {
         GVRMaterial cubemapMaterial = new GVRMaterial(gvrContext,
                 GVRMaterial.GVRShaderType.Cubemap.ID);
         cubemapMaterial.setMainTexture(futureCubemapTexture);
+        
+        ArrayList<Future<GVRTexture>> futureTextureList = new ArrayList<Future<GVRTexture>>(6);
+        futureTextureList.add(gvrContext
+                .loadFutureTexture(new GVRAndroidResource(gvrContext,
+                        R.drawable.back)));
```
thomasflynn: you appear to be missing 'back', 'right', 'front', 'left', 'top', 'bottom' from your drawables directory.  they appear to not be checked-in.

rongguodong:
Missing resources added.

thomasflynn:
ok, turned on stats, and toggled frustum culling on/off.  appears to work as expected.

rongguodong:
For variable names, I think both segmentNumber and numberOfSegment are quite clear in meaning. Since we also have sliceNumber, stackNumber, verticesNumber, etc. in other places, adding an extra "of" will only make all variable names longer.

thomasflynn:
ok, anyone else have any other issues with this pull-request?

chenchao1407:
I am currently busy with the following issue with #151. I think so far it is good to me now. May think of refactoring those duplicate codes later. But right now I think it is good to go.

PR #159, opened 2015-07-13
Null check in case Assimp is not able to load scene

deepakrawat22:
If due to bad file format Assimp is not able to load a file, assimp_importer_->GetScene() will give NULL but GearVRf was not checking for it. Added NULL check for that. Now if a bad file format is loaded GearVRf will not crash and show a error log instead.

drawat@usc.edu


chenchao1407:
Look good, I will check it later

PR #163, opened 2015-07-16
 Fixed a issues which was causing material native ptr to be null.

phi-lira:
On GVRRenderPass a material was being construct with wrong constructor call causing native ptr to be set to null by mistake.

f.lira@samsung.com


jason2kim:
@liaxim, can you please let us know if this cures your case?

liaxim:
@jason2kim, yes it does take care of the problem.

PR #165, opened 2015-07-17
Remove redundant uses of GVRRenderTexture in GVRRenderBundle.

chenchao1407:
Currently we are not using the left texture and right FBO created
on our own. So that those two GVRRenderTextures can be removed.
And related codes are also refactored.

### DIFF
--- a/GVRf/Framework/jni/engine/renderer/renderer.cpp
+++ b/GVRf/Framework/jni/engine/renderer/renderer.cpp
@@ -470,7 +470,7 @@ bool Renderer::is_cube_in_frustum(float frustum[6][4],
 }
 
 void Renderer::renderCamera(Scene* scene, Camera* camera,
-        RenderTexture* render_texture, ShaderManager* shader_manager,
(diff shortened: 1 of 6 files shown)


GVRf/Framework/jni/engine/renderer/renderer.cpp
```diff
@@ -470,7 +470,7 @@ bool Renderer::is_cube_in_frustum(float frustum[6][4],
 }
 
 void Renderer::renderCamera(Scene* scene, Camera* camera,
-        RenderTexture* render_texture, ShaderManager* shader_manager,
+        ShaderManager* shader_manager,
         PostEffectShaderManager* post_effect_shader_manager,
         RenderTexture* post_effect_render_texture_a,
         RenderTexture* post_effect_render_texture_b, glm::mat4 vp_matrix) {
```
rongguodong: This vp_matrix is also redundant.

chenchao1407:
Updated according to guodong's comment.

PR #171, opened 2015-07-17
Set material for parent object for subdivided sphere.

rongguodong:
rongguodong@hotmail.com


rongguodong:
This check-in is a quick way to enable setting material/texture after create the subdivided sphere. Since all segments in a subdivided sphere always share a same material/texture, this simple way should be enough. More general (and better) approach is explained in issue #172.

chenchao1407:
I think it is ok to go.. So simple... :-)

PR #174, opened 2015-07-18
Adding the VRBasketBall project

cesarramirez:
Adding the VRBasketball project from the Seattle VR Hackathon. This project contains three smaller components, the android client, a server and a x3dom page.


deepakrawat22:
Hi, I tried the VRBasketBall project and it is an awesome idea. Below are my findings while playing with VRBasketBall project.

1- Leap Left gesture is hard to detect or may be the leap I am using is broken. After countless tries will get one LEFT swipes gesture. For testing I rigged the server code to send LEFT command to Android application whenever onFrame is called.

2- Even though Android application is connected to the server it still show Disconnected, I think this is because server only sends CONNECTED commands once and if App miss it then it will always shows disconnected even though its connected. My solution for this was to set "connected = true" for all commands except DISCONNECT.

3- For Windows and Linux need to add Leap binaries for the server.

cesarramirez:
Thanks deepakrawat22  for your input, here is some context on the state of the VRBasketBalls project.

1- Left gesture by using the server or the HTML page? I did not work on the HTML site and the version that I have posted is all I could find about it, so currently this part if broken. I am in the process to get the correct file so the HTML page that accesses the LeapMotion can be fixed. Due to this issue I added a very basic code to detect left movement on the Leap Motion.  Since this was just a hack due to not having another way to read data from the LeapMotion, this solution is very unreliable. 

2- The CONNECTED command is fired by the Android client itself when it connects to the server (LeapClientThread:36). This difference may be due to some other condition when the DISCONNECTED command is fired even when the server is connected. I will look into this and update when I find something.

3- I will work on adding this too.

Timeframe: I will be very busy this week but I should have a new pull request by middle of next week.

PR #175, opened 2015-07-19
Outline sample

phi-lira:
* OutlineSample now loads a character to present outline shader in a real world scenario.
 
  * Added GVRImportSettings class to configure important assimp flags like triangulate meshes and calculate normals. 

  * Triangulate import settings is always performed as we should always render triangulated meshes for performance reasons.

I'm very interested in code review for the GVRImportSettings part. I was not sure how to pass GVRImportSettings class to jni so I ended up passing an int, which is bad because I end up checking for hardcoded values in native and that is trouble prone.

Also I used a class instead of enum because I couldn't find a way to use the bitwise operator (|) with enums in Java. So I codereview on that would also be great!


GVRf/Framework/src/org/gearvrf/GVRImportSettings.java
```diff
@@ -0,0 +1,62 @@
...
+     * Tell Importer to calculate smooth normals in case they are not present in imported model. 
+     */
+    public static final int CalculateTangents = 0x4;
+    
+    /**
+     * Tell Importer to optimize mesh to reduce drawcalls and improve cache locality. This might
+     * increase loading time.
+     */
+    public static final int OptimizeMeshData = 0x8;
+
```
chenchao1407: For static final member variable, the naming will be like OPTIMIZE_MESH_DATA, not camel naming.

phi-lira:
Thanks @liaxim That's exactly the kind of feedback I was looking for! I'll work on this today.

phi-lira:
Ok. If I understood correctly this new commit applies all suggestions. 
EnumSet is used to construct GVRImportSettings, settings map to assimp values and I added a function to convert from EnumSet to assimp flag.

So now importer.cpp takes in a int flag that maps directly to assimp values. This is not trivial to test, but I did some print and checked the flag to see if it was matching. 

I also added a "recommendedSettings" which I think is the minimal import options that should be passed in to assimp. I added some recommendedSettings for bumpmap and an easy way to add a flag to the already recommended settings.

PR #178, opened 2015-07-21
Adding a reset() to GVRTransform.

thomasflynn:
Adding a reset() to GVRTransform.  This will be the equivilent to
glLoadIdentity().

tom.flynn@samsung.com


chenchao1407:
Codes look fine. And work as expected(with my personal sample)

PR #179, opened 2015-07-21
Support head tracking based on sensor readings supplied by Oculus

liaxim:
oculus tracking disabled by default but whomever needs it can enable
it from GVRActivity::Configure

need to work on supporting all camera rig types but planning to do
so in subsequent commit as it is not immediately needed as far as I
know

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/GVRViewManager.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRViewManager.java
@@ -234,7 +234,7 @@ class GVRViewManager extends GVRContext implements RotationSensorListener {
          * GL Initializations.
          */
         mRenderBundle = new GVRRenderBundle(this, mLensInfo);
-        mMainScene = new GVRScene(this);
+        setMainScene(new GVRScene(this));
     }
 
     private void renderCamera(long activity_ptr, GVRScene scene,
@@ -536,7 +536,6 @@ class GVRViewManager extends GVRContext implements RotationSensorListener {
 
     /** Called once per frame, before {@link #onDrawEyeView(int, float)}. */
     void onDrawFrame() {
-        mActivity.setCameraRig(mMainScene.getMainCameraRig());
         if (mCurrentEye == 1) {
             mActivity.setCamera(mMainScene.getMainCameraRig().getLeftCamera());
         } else {
@@ -802,6 +801,9 @@ class GVRViewManager extends GVRContext implements RotationSensorListener {
                 mOnSwitchMainScene.run();
                 mOnSwitchMainScene = null;
             }
+        }
+        if (null != mMainScene) {
+            getActivity().setCameraRig(mMainScene.getMainCameraRig());
         }
     }
 
--- a/GVRf/Framework/src/org/gearvrf/GVRViewManager.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRViewManager.java
@@ -456,7 +456,6 @@ class GVRViewManager extends GVRContext implements RotationSensorListener {
             GVRCameraRig mainCameraRig = mMainScene.getMainCameraRig();
 
             if (eye == 1) {
-                mainCameraRig.predict(4.0f / 60.0f);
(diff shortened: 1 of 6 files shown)


GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -539,6 +536,7 @@ void onDrawEyeView(int eye, float fovDegrees) {
 
     /** Called once per frame, before {@link #onDrawEyeView(int, float)}. */
     void onDrawFrame() {
+        mActivity.setCameraRig(mMainScene.getMainCameraRig());
         if (mCurrentEye == 1) {
```
chenchao1407: Imho, I don't think you will have to setCameraRig for every new frame. I think you'd just need to set the camerarig in GVRViewManager.onSurfaceCreated() after new GVRScene(this) is called(for splashscreen). One more place you need to set is in the GVRViewManager.setMainScene() method(for every new main scene except splashscreen).

I tried above and it works for me.

chenchao1407:
Codes looks fine. There is only one small concern.

I tried the oculus sensor and it works awesome. It seems to me that it makes perfectly smoother compared to before. The only disadvantage is that if we turn on the oculus sensor then we can't debug without goggle. But overall this pull request is awesome.

chenchao1407:
Sorry the comments in the codes is somehow incorrect in the case that if any one sets a new camerarig into main scene by calling setMainCameraRig, it will lead to a problem. However I think we still can handle it since it is able for us to compare the scene which we call setMainCameraRig() with the main scene when setMainCameraRig() is being called.

liaxim:
Using setMainCameraRig in this manner would also mean that extra work needs to be done when setting a main scene. You could call setMainCameraRig on a GVRScene that is not the current main scene but you could switch to it later. I did look into avoiding setCameraRig on every frame but I had to turn this change into a refactoring also; which I'd rather have done separately. I will give it one more try.

I plan to add fallback to the internal sensor if not docked as a subsequent enhancement. Agreed that using Oculus for this has a surprisingly good effect - thank you Benoit for pointing it out.

chenchao1407:
In my opinion, there are three things you may need to do to avoid adding setCameraRig every frame. The first place is to call GVRActivity.setCameraRig in GVRViewManager.onSurfaceCreated. The second thing is to call GVRActivity.setCameraRig in GVRViewManager.setMainScene(). The third thing to do is in GVRScene.setMainCameraRig, you can check if gvrContext.getMainScene equals with current scene, if yes, you can do gvrContext.getActivity().setMainCameraRig(). As I can see, I think these three will cover all the cases. Please let me know if there is any other exceptional case.

Thanks in advance.
Sincerely
-Chao Chen

liaxim:
Three places to modify to keep track of the camera rig, feels too much.. Will have to add a reference to the Activity in GVRScene (or add event interface to it) just for this. I will do so if I can't come up with anything better. Not to mention that we are already setting the Camera every frame.. Would it be acceptable, as a stop-gap, if I merge the setCamera and setCameraRig calls we make on every frame into one?

chenchao1407:
I think in GVRScene you can directly get GVRActivity reference from its member reference gvrContext and its API getActivity()

liaxim:
This will do. Thanks.

liaxim:
Added one more commit that eliminates setting the camera rig on every frame.

chenchao1407:
Overall looks good to me except there is just a line with bad spacing. 

After that spacing get fixed, I think this one is good to go.

liaxim:
Tabs removed

chenchao1407:
Ok. I think it is good to go if there is no other objection.

liaxim:
Can this be merged? I have a subsequent commit that provides a fallback to the internal sensor if the device is not docked. I prefer to have it as a separate pull request as this one seems ready to go.

thomasflynn:
Looks good to me as well.  Tested it last night.  And it looks like it is opt-in for now so it won't immediately disrupt anyone.

PR #180, opened 2015-07-22
use CopyOnWriteArrayList instead of ArrayList

niusounds:
Removed `synchronized` block and lock object.


chenchao1407:
Since we probably read much more than write here. It makes sense to me to replace this.

PR #181, opened 2015-07-22
 Fixed Assimp warnings regarding gcc_struct being ignored. This chang…

phi-lira:
This fixes the many warnings regarding gcc_struct. This changed is self contained and it was taken from new version in assimp repo. 

f.lira@samsung.com


thomasflynn:
can someone test compiling with this on linux?

chenchao1407:
Tried. Compiled well. Also tried with some samples and worked fine. I can see that those warnings gone. I think it is good to merge.

PR #182, opened 2015-07-23
Texture enhancement

lokeshguddu:
Provides API to change/modify texture properties.


thomasflynn:
1 comment so far:  it looks like you are checking in libgvrf.so and other .so files.  That shouldn't be necessary.

There's another thing i'm having difficulty telling from the diffs, but I assume it compiles :).  When retrieving the support for aniso and storing the max aniso value, that info is gathered in GVRViewManager, but the variables are declared in GVRContext.  I think i'll have to do a check-out and look at the full source file.

lokeshguddu:
Hi @thomasflynn ,

We removed the .so files in the last commit. This happeened because of the modification in the .gitignore file which now allows for .so files to be tracked.

The aniso varible was declred in the GVRContext and the calculation was done in the GVRViewManager becuase we had to be run the aniso code on the gl thread and make the value available as public api. Since the GVRViewManager extends the GVRContext, it looked like a good place to put those things.

Please let us know if you find any more issue.

thomasflynn:
Thanks, that's what I was missing.  Makes sense now.

PR #184, opened 2015-07-25
Adding other axes and buttons.

thomasflynn:
Adding other axes and buttons for reference.  They will just be printed out.

tom.flynn@samsung.com

### DIFF
--- a/GVRf/Sample/gamepad/src/org/gearvrf/gamepad/SampleViewManager.java
+++ b/GVRf/Sample/gamepad/src/org/gearvrf/gamepad/SampleViewManager.java
@@ -194,10 +194,6 @@ public class SampleViewManager extends GVRScript {
             android.util.Log.d(TAG, "button SELECT");
             handled = true;
             break;
-        case 310:
-            android.util.Log.d(TAG, "Play");
-            handled = true;
-            break;
         }
 
         // change background color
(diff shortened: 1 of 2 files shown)


dknaan:
Everything works for me except for the "play" button (case 310 in the code) which started another application and didn't print anything to the log.

GVRf/Sample/gamepad/src/org/gearvrf/gamepad/SampleViewManager.java
```diff
@@ -72,6 +76,80 @@ public void onInit(GVRContext gvrContext) throws IOException {
...
+        // the left control stick, hat axis, or the right control stick.
+        float x = getCenteredAxis(event, mInputDevice, MotionEvent.AXIS_X,
+                historyPos);
+        float hatx = getCenteredAxis(event, mInputDevice,
+                MotionEvent.AXIS_HAT_X, historyPos);
+        
+        // Google refers to this axis as z, Samsung Gamepad uses RX.
+        float z = getCenteredAxis(event, mInputDevice, MotionEvent.AXIS_RX,
+                historyPos);
+
```
dknaan: When you write "Google refers to..." do you refer to the image in https://developer.android.com/training/game-controllers/controller-input.html?
I find it confusing to print to the log z, rz instead of axis_rx, axis_ry.

thomasflynn:
tried clarifying the whole second analog stick thing.  let me know if it makes better sense.

thomasflynn:
Removed reference to 'Play' button.  Seems it launches S-Console on Note 4 and there didn't seem to be a way to disable/override it.

PR #186, opened 2015-07-27
Make GVRSceneObject.detachEyePointeeHolder thread-safe in respect to GVRPicker.findObjects

liaxim:
GVRSceneObject.detachEyePointeeHolder() nulls out the owner object without locking (I have an app that does that quite frequently). If this happens while GVRPicker.findObjects() is running it may try to use an EyePointeeHolder without owner.


chenchao1407:
Overall looks good to me. A small concern is like, whether a global lock is a little bit wider than necessary. I know the point is similar to lock a whole scene when detachEyePointeeHolder and findObjects happen. But this lock will actually lock all scenes we currently have, even though it will be a rare case that an app will synchronously update more than one scene at the same time. But anyway, as I am saying the case is rare. It is good to go at my sight.

PR #187, opened 2015-07-27
Fix for "Enable dynamic switching between Oculus rotation sensor..."

liaxim:
previously never tried without the phone being connected to usb;
if it is not, there will be no undock event and
useOculusOrientationReading could be anything


liaxim:
Pull request #191 handles the whole dynamic switching more elegantly; this fix will be no longer necessary.

Issue #188, opened 2015-07-28
jAssimp: 3D Model loading fails if first level children has no mesh


deepakrawat22:
GearVRf jAssimp feature will fail to load a 3D model if the node hierarchy is such that first level children do not have any mesh.

![untitled](https://cloud.githubusercontent.com/assets/8952072/8920906/4f86b07a-3484-11e5-9876-975c104ab0a0.png)

Currently it just ignores rest of hierarchy in the function getAssimpModel().

The test file I used is f16a.fbx (https://github.com/deepakrawat22/GearVRApps/tree/master/gvrtest1/assets)

lokeshguddu:
Hi @deepakrawat22 ,

Thanks for bringing this to our notice. We will have a look into the implementation and revert back with a solution. Meanwhile if you find something else, please let us know.

Thanks.

lokeshguddu:
Hi @deepakrawat22 ,

We've fixed the issue. Please have a look [https://github.com/Samsung/GearVRf/pull/189].

In case we missed something, please let us know.

Thanks

PR #189, opened 2015-07-28
fix: Fixes the issue #188

lokeshguddu:
There was a problem of getAssimpModel(...) not loading models which has
no meshes in the initial node hierarchy. This commit fixes that issue by
recursive traversal of nodes.

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/GVRContext.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRContext.java
@@ -702,7 +702,9 @@ public abstract class GVRContext {
     }
 
     /**
-     * Helper method to create a new {@link GVRSceneObject} with either color or texture.
+     * Helper method to create a new {@link GVRSceneObject} with the mesh at the
+     * index {@link index} of the node mesh array with a color or texture
+     * material.
      * 
      * @param assetRelativeFilename
      *            A filename, relative to the {@code assets} directory. The file
--- a/GVRf/Framework/src/org/gearvrf/GVRContext.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRContext.java
@@ -672,9 +672,7 @@ public abstract class GVRContext {
             AiNode node,
             AiWrapperProvider<byte[], AiMatrix4f, AiColor, AiNode, byte[]> wrapperProvider) {
         try {
-            // for (AiNode childNode : childrenNodes) {
             for (int i = 0; i < node.getNumMeshes(); i++) {
-
                 FutureWrapper<GVRMesh> futureMesh = new FutureWrapper<GVRMesh>(
                         this.getNodeMesh(new GVRAndroidResource(this,
                                 assetRelativeFilename), node.getName(), i));
@@ -770,8 +768,6 @@ public abstract class GVRContext {
                         wholeSceneObject, node.getChildren().get(i),
                         wrapperProvider);
             }
-
-            // }
         } catch (Exception e) {
             // Error while recursing the Scene Graph
             e.printStackTrace();
--- a/GVRf/Framework/src/org/gearvrf/GVRContext.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRContext.java
@@ -199,7 +199,8 @@ public abstract class GVRContext {
      * @since 1.6.2
      */
     public GVRMesh loadMesh(GVRAndroidResource androidResource) {
-        return loadMesh(androidResource, GVRImportSettings.getRecommendedSettings());
+        return loadMesh(androidResource,
+                GVRImportSettings.getRecommendedSettings());
     }
 
     /**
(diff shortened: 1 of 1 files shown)


deepakrawat22:
Verified, now getAssimpModel(...) is able to load f16a.fbx file.

GVRf/Framework/src/org/gearvrf/GVRContext.java
```diff
@@ -628,118 +634,144 @@ public AiNode wrapSceneNode(Object parent, Object matrix,
...
+
+                    wholeSceneObject.addChildObject(sceneObject);
                 }
             }
+
+            for (int i = 0; i < node.getNumChildren(); i++) {
+                this.recurseAssimpNodes(assetRelativeFilename,
+                        wholeSceneObject, node.getChildren().get(i),
+                        wrapperProvider);
+            }
```
chenchao1407: Talked to deepak, we are thinking the logic here is not correct.

For example, if I have a hierarchy like man -> arms -> hands -> fingers. The hierarchy of scene objects should be like man is the direct parent of arms, arms are the parents of hands, etc.. But based on the algorithm here, man is direct parent of all other items. It may be wrong in the case that if we want to propagate a transformation from a certain parent to its successive children.

GVRf/Framework/src/org/gearvrf/GVRContext.java
```diff
@@ -628,118 +634,144 @@ public AiNode wrapSceneNode(Object parent, Object matrix,
...
+     *            A reference to the AiNode for which we want to recurse all its
+     *            children and meshes.
+     * 
+     * @param wrapperProvider
+     *            AiWrapperProvider for unwrapping Jassimp properties.
+     * 
+     */
 
+    public void recurseAssimpNodes(
+            String assetRelativeFilename,
```
chenchao1407: Do we need to set this API public?

Thanks in advance.

Sincerely
-Chao Chen

lokeshguddu: No it doesn't need to be a public API. Fixed this. 

Now for the other issue, we need to take a deeper look. Will revert back when we have a solution. If you have a solution please let us know, we will act accordingly.

Thanks

GVRf/Framework/src/org/gearvrf/GVRContext.java
lokeshguddu: Hi @chenchao1407 @deepakrawat22 ,

We have inspected the code and it looks fine.

"For example, if I have a hierarchy like man -> arms -> hands -> fingers. The hierarchy of scene objects should be like man is the direct parent of arms, arms are the parents of hands, etc.."

We think that the code that we have pushed is doing the same thing as you are saying. In fact, its the standard way of recursively loading an assimp model. Even the [assimp GitHub repo sample](https://github.com/assimp/assimp/blob/master/samples%2FSimpleTexturedOpenGL%2FSimpleTexturedOpenGL%2Fsrc%2Fmodel_loading.cpp) loads the models in a similar manner. And this [article](http://learnopengl.com/#!Model-Loading/Assimp) might clarify the reason behind this implementation.

Please have a look and if you still think that we have missed something please do let us know.

We will be happy to proceed in the suggested direction.

Thanks

soumyajitdeb: Just to add to what Lokesh mentioned. The function node.getNumChildren() only returns the immediate children of the node not all at any level children. 

In the above mentioned example, it would only return the 'arms'. The node 'arms' then must be recursed to further get the 'hands' and then one more pass gets the 'fingers'. At each level the correct transformation matrix will be computed and applied. Normally all model hierarchies are loaded in the same way. 

Please do let us know if we missed something else and we'll fix the same.

Thanks!

deepakrawat22:
Hi @lokeshguddu, I think the main problem here is a node can have more than one mesh, but in GearVRf a scene object can have only one mesh associated with it. How about if in case a node with more than one mesh, we make a gvrf scene object as parent to all the scene object associated with mesh. like

recurseAssimpNodes(GVRSceneObject rootscene, AiNode node) {
GVRSceneObject scene; 
if (node.getNumMeshes() == 1) {
  empty scene
}
else if (node.getNumMeshes() == 1) {
    scene = scene with mesh
} else {
    scene = empty scene
    // All other mesh as GVRSceneObject with parent as scene
}

for (int i = 0; i < node.getNumChildren(); i++) {
    recurseAssimpNodes(scene, node.getChildren.get(i))
}

I was following this approach to load model hierarchy to GearVRf in JNI code (which was not completed, and duplicate of what jAssimp feature does)

This article suggest http://assimp.sourceforge.net/lib_html/data.html to maintain hierarchy. About the  assimp sample code, as per my understanding the GitHub repo sample code is not doing the same as we currently have, it's just passing sc parameter to get material and its drawing immediately and not storing the hierarchy.

Also I think maintaining hierarchy might be necessary later on when bones and animation support would be added to GearVRf. I still don't have a very accurate knowledge of assimp, please correct me if I am wrong.

lokeshguddu:
Hi @deepakrawat22 ,

You are right about maintaining the hierarchy. It is just that we had just to render the models and not use it's animation aspect, so we didn't replicate the assimp hierarchy into the GVRScene with non-renderable GVSceneObject to reflect the node without meshes. This will only be useful if we need to implement skeletal animations using bones. 

For rendering purpose, the present implementation seemed fine.

We will push the changes with a new implementation which replicates the hierarchy which can be later modified to integrate animations.

lokeshguddu:
Hi,

We have done some modification. Please take a look and if you find something more to be changed, please let us know.

Thanks.

chenchao1407:
Looks good to me at this point. @deepakrawat22  Can you verify it again? Thanks in advance.

deepakrawat22:
@chenchao1407 Verified, the new code is able to load the model "f16a.fbx".

PR #191, opened 2015-07-28
Simplify switch between oculus and internal sensors

bfouet-gpsw:
Use the information that is already present in the Oculus VR frame, instead of listening to events.


liaxim:
Looks good and works as advertised.

PR #193, opened 2015-07-29
Support dynamic switching between Oculus and the internal sensor for head tracking

liaxim:
Using the Oculus docked flag is unreliable unfortunately (may have something to do with our lower-level changes but still..); have to resort to use our own dock observer to be certain switching would work.

### DIFF
--- a/GVRf/Framework/jni/oculus/activity_jni.cpp
+++ b/GVRf/Framework/jni/oculus/activity_jni.cpp
@@ -21,7 +21,6 @@
 #include "VrApi_Types.h"
 
 static const char * activityClassName = "org/gearvrf/GVRActivity";
-static const char * app_settings_name = "org/gearvrf/utility/VrAppSettings";
 
 namespace gvr {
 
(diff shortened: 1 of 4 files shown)


chenchao1407:
I think overall is fine. Just tried and the docking worked as expected. Since all my phones will quit the app when undock happens, so I can't check if undock works well. But as far as I can see in the codes, it should work as expected.

chenchao1407:
@liaxim  This one needs to be rebased also

liaxim:
Yep, needs rebasing too but I will do it after the refactoring is merged.

thomasflynn:
the refactoring is merged now.

liaxim:
Rebased and tested again

chenchao1407:
Function verified, looks good to me. Codes look almost fine. I am just thinking if we can reduce the code redundancy between KSensorPrediction and OculusSensorPrediction. Actually as we can see, in terms of codes, OculusSensorPrediction is actually an extension of KSensorPrediction. It is just a minor issue seems to me. How do you think? @liaxim

liaxim:
It seems so given the existing code but logically I think they are "peers" and Oculus does not "derive" from KSensor. Eventually, I'd like for Oculus to be completely separate from KSensor as they are substitutes for each other. What can and would be shared is the internal sensor but I currently have no plan how to do it exactly. I also have to do the whole thing over multiple iterations as my understanding of the codebase and the domain evolves.

Anyway, I am open to hearing what you suggest to do with the current code. How would you structure it so it requires noticeably less code?

thomasflynn:
Let's do that in a separate check-in.  this is holding up some other work.  i tested, looks good to me.

PR #195, opened 2015-07-30
Refactor the switching between Oculus and KSensor

liaxim:
Centralized things a bit and simplified the flow. I am aware there is small overlap with another change by Danke. The monoscopic renderer is kind of the odd one out for now.

### DIFF
--- a/GVRf/Framework/jni/objects/components/camera_rig.cpp
+++ b/GVRf/Framework/jni/objects/components/camera_rig.cpp
@@ -82,7 +82,12 @@ void CameraRig::setRotationSensorData(long long time_stamp, float w, float x,
             gyro_y, gyro_z);
 }
 
-void CameraRig::predict(float time) {
+void CameraRig::predictAndSetRotation(float time) {
+    glm::quat headRotation = predict(time);
+    setRotation(headRotation);
+}
+
+glm::quat CameraRig::predict(float time) {
     long long clock_time = getCurrentTime();
     float time_diff = (clock_time - rotation_sensor_data_.time_stamp())
             / 1000000000.0f;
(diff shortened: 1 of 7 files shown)


liaxim:
Uploaded another commit that enables Oculus prediction by default

chenchao1407:
Hi @liaxim 

Actually, with your new coming patch, we only can use Oculus sensor. I tried with two devices and only when I dock the goggle, the camera will work. It means, the device sensor will never be enabled even thought I undock the goggle. As far as I can see that vrFrame.DeviceStatus.DeviceIsDocked will always return true. At first sight I am thinking if this is a device's issue, until I find the following comments in VrApi_Android.h:

//-----------------------------------------------------------------
// Device Status
//
// Note that there is currently no way to tell what the initial dock/mount state
// is when the vrapi_Initialize() is first called. Only when the state changes after
// vrapi_Initialize() has been called we can tell what the real state is. To work
// around this problem, we assume the device is docked and the headset is mounted
// when we enter VR mode, and no dock/mount state changes have been recorded up
// to that point. Under normal circumstances this is usually correct. This is
// incorrect when in developer mode, but in that case you want the device to
// appear as docked and the headset as mounted anyway.
//
// These can be called any time from any thread.
//-----------------------------------------------------------------

Therefore, imho, unfortunately, even though the logic of codes seems no problem at all. But eventually this pull request won't work as expected in terms of auto mode switch until Oculus fix the issue at their site.

liaxim:
Yep, I am aware of it by now. That is why I have another patch where we observe for dock events ourselves. If this goes in first, I will merge/rebase the other one so the Oculus variable will be no longer used very soon.

chenchao1407:
Sure. I am thinking the same at this moment. You can just use broadcastreceiver and JNI to communicate with your OculusPrediction to decide which one you are going to use. I am ok to merge this first. Any other objection?

chenchao1407:
However, to merge it, I think you'd probably switch back to use KSensorPrediction by default. Otherwise the master branch will temporarily not work well until we merge the other one.

liaxim:
Ok, switched to KSensor

chenchao1407:
Hi @liaxim 

You may need to rebase this one.

liaxim:
Rebased and merged conflicts in activity_jni.cpp.

Issue #148, opened 2015-07-02
onKeyUp is called multiple times when back button is pressed


niusounds:
I want to handle GearVR back button pressing event to cancel showing default Oculus Home dialog.
I implement `public boolean onKeyUp(int keyCode, KeyEvent event)` in my Activity class.

```java
public boolean onKeyUp(int keyCode, KeyEvent event) {

  if (showingPlayerScene()) {
    backToContentBrowser();
    return true; // Cancel showing default dialog.
  }

  return false; // Show default Oculus Home dialog.
}
```

`return true;` cancels dialog. It's I expected. But `onKeyUp` method was actually called twice unexpectedly.

I suspect that would be wrong key event handling in `GVRActivity::OnKeyEvent`.

```cpp
bool GVRActivity::OnKeyEvent(const int keyCode, const int repeatCode,
        const OVR::KeyEventType eventType) {

    // 1: KeyState::KEY_EVENT_DOWN, 0: KeyState::KEY_EVENT_UP. Other information is lost from Oculus side.
    int isDown = (eventType == OVR::KEY_EVENT_DOWN) ? 1 : 0;

    return app->GetVrJni()->CallBooleanMethod(javaObject,
            onKeyEventNativeMethodId, keyCode, isDown);
}
```

In above code, `isDown` will be `0` when `eventType` is not only `KEY_EVENT_UP` but also `KEY_EVENT_SHORT_PRESS`, `KEY_EVENT_LONG_PRESS` or some other values.

These events (`KEY_EVENT_SHORT_PRESS` or `KEY_EVENT_LONG_PRESS`) should not be handled by `onKeyUp`.

jason2kim:
Hi, I remember the KEY UP event was tested working correctly with the previous Oculus version of 0.5.
Can you please let us know a more details, including Ovr version, which key code (BACK only or other keys), long press, and test log if you have? thanks.

niusounds:
I'm using Oculus Mobile SDK 0.6.0 because I found "Initial check-in of port to OVR SDK 0.6.0" commit message.
(...Is GVRf not supporting 0.6.0 yet?)

* When I push back button shortly -> onKeyUp is called twice.
* When I push back button long -> onKeyUp is called once.

My problem occurs with key code 4 (android.view.KeyEvent.KEYCODE_BACK) which is back button on GearVR near touch pad. I have not tried other keys (such as gamepad) because I have no gamepads.

dknaan:
[...]

PR #196, opened 2015-07-31
Improved key event handling

niusounds:
When I pressed back button `onKeyUp` is called twice. (see https://github.com/Samsung/GearVRf/issues/148)
Because actually 3 key event occurs when I press keys. These are DOWN->UP->SHORT PRESS.

And also, Oculus requires to implement to go to global menu when back button is long pressed. Framework should follow this requirement.

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/GVRActivity.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRActivity.java
@@ -43,6 +43,15 @@ public class GVRActivity extends VrActivity {
 
     private static final String TAG = Log.tag(GVRActivity.class);
 
+    // these values are copy of enum KeyEventType in VrAppFramework/Native_Source/Input.h
+    public static final int KEY_EVENT_NONE = 0;
+    public static final int KEY_EVENT_SHORT_PRESS = 1;
+    public static final int KEY_EVENT_DOUBLE_TAP = 2;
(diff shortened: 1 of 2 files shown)


chenchao1407:
I was thinking on improving this. I think the way you are working on matches my thinking. I will try on this and tell you my further thinking on this.

GVRf/Framework/src/org/gearvrf/GVRActivity.java
```diff
@@ -239,14 +248,34 @@ boolean onKeyEventNative(int keyCode, int eventType) {
          * native side. It can be more than two, DOWN and UP, if the native
          * supports in the future.
          */
-        final int cKeyDown = 1;
 
-        int action = (eventType == cKeyDown) ? KeyEvent.ACTION_DOWN
-                : KeyEvent.ACTION_UP;
-        KeyEvent event = new KeyEvent(action, keyCode);
+        switch (eventType) {
```
jason2kim: Thanks. It makes sense if the OVR library delivers all event types of these at least, not just the two of up and down.

thomasflynn:
jason, have you verified this?  is it good to merge?

jason2kim:
One thing yet to verify is to check if those event types are delivered from OVR library as intended. 
@chenchao1407 are you going to make some other comments here?

chenchao1407:
Overall is fine to me at this point.

PR #197, opened 2015-08-02
Upgraded shaders in model-viewer sample to GLSL3

thomasflynn:
Half of the shaders in the model-viewer sample didn't work on S6.  S6 is using Mali which is
apparently more strict when compiling shaders.  A lot of these shaders
required inverse(), which is only in GLSL3.  Upgraded the shaders to
GLSL3.  Things seem to look correct now.

tom.flynn@samsung.com


chenchao1407:
Looks good to me. Will try with a Mali Note 4 tmr

chenchao1407:
Confirmed with Mali note 4.

PR #199, opened 2015-08-04
Implement APIs for texture update in a non-GL thread.

chenchao1407:
Several changes included.

1. Introduced GVRFutureOnGlThread. This is a special implementation of RunnableFuture.
This should be used when you want a return value and run it with GVRContext.runOnGlThread(Runnable).

2. Add new APIs safeUpdate with GVRBitmapTexture. Previously GVRBitmapTexture.update
running on a non-GL thread won't return or report any error but you can't see your
update on screen. Using safeUpdate will make any update call from a non-GL thread
successful on screen.

3. Add new API GVRContext.isCurrentThreadGLThread. Previously, we only have API
assertGLThread inside GVRF to check if current thread is GL thread or throws an
Exception. I think it should be fine to open an API to check if current thread is
GL thread to judge if we should directly call anything that requires GL thread and
to see if we should use GVRFutureOnGlThread.

4. Fix GVRConsole. Previously, if you want to call GVRConsole.writeLine in a non-GL
thread, ex, through an input event, it won't update anything on screen. With
GVRBitmapTexture.safeUpdate, this is fixed.

5. Sample added for showing the fix on GVRConsole and safeUpdate on a non-GL thread.


liaxim:
The change itself is fine but:

1. I'd say completely hide the old update(); safeUpdate does the right thing depending on which thread it is getting called. My reason is the more apis are exposed, the harder it will be on people to know which one when to use and they have to work out the difference between safe and unsafe update..

2. As GVRF seems to focus a lot on convenience, then next comes the GVRBitmapTexture ctor.. which has to be run on the GL thread. The users have to be completely oblivious of the GL thread (or at least that should be the goal); it is very confusing if the threading "leaks" here and there.

chenchao1407:
Some update according to Mihail's comment. May further update after discussion with you tomorrow.

chenchao1407:
Hi @liaxim 
Just as you concerned, I just found that constructor of GVRBitmapTexture (actually of all kinds of GVRTexture) need to be called at gl thread. I will try to figure out this one.

chenchao1407:
Just talked to Mihail, I will make another pull request for the broader issues in Texture related classes. But as a pull request only for update() API, is there any other concern? If no, I am going to merge it.
@jason2kim @thomasflynn

GVRf/Framework/src/org/gearvrf/debug/GVRConsole.java
```diff
@@ -357,14 +370,20 @@ private void updateHUD() {
...
+        Future<Boolean> textureUpdated = null;
         if (texture != null) {
             textureUpdated = texture.update(HUD);
         }
 
-        if (textureUpdated != true) {
-            texture = new GVRBitmapTexture(getGVRContext(), HUD);
-            setMainTexture(texture);
+        try {
+            if (texture == null || textureUpdated.get() != null && !textureUpdated.get()) {
```
jason2kim: a confusing if statement, may need a bracket or redundant?  textureUpdated.get() != null && !textureUpdated.get()

GVRf/Framework/src/org/gearvrf/GVRFutureOnGlThread.java
```diff
@@ -0,0 +1,191 @@
...
+
+    /**
+     * The result we get after run() successfully executed.
+     * 
+     * @return The result from the Callable we put in.
+     * @throws CancellationException
+     *             If the task was successfully cancelled.
+     */
+    @Override
+    public T get() {
```
jason2kim: I think it would be nicer if we can reuse the get( , ) below, which is more inclusive.

GVRf/Sample/nonglthreadupdate/src/org/gearvrf/nonglthreadupdate/SampleActivity.java
```diff
@@ -0,0 +1,81 @@
...
+    public void onShowPress(MotionEvent e) {
+        // TODO Auto-generated method stub
+        
+    }
+
+    @Override
+    public boolean onSingleTapUp(MotionEvent e) {
+        // TODO Auto-generated method stub
+        mScript.onTapUp();        
+        return false;
```
jason2kim: minor, but still better to return true if we intended to mark as consumed

GVRf/Sample/nonglthreadupdate/src/org/gearvrf/nonglthreadupdate/SampleViewManager.java
```diff
@@ -0,0 +1,89 @@
...
+import org.gearvrf.debug.GVRConsole.EyeMode;
+import org.gearvrf.nontlthreadupdate.R;
+
+import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
+import android.graphics.Color;
+
+public class SampleViewManager extends GVRScript {
+
+    private GVRContext mGVRContext;
```
jason2kim: seems not being used

chenchao1407:
Updated. @jason2kim @thomasflynn

jason2kim:
it looks good to me

PR #203, opened 2015-08-05
GVRf: Add camera rotation support

danke-sra:
Two types of camera rotation are supported:
  1) Rotation based on the (game) character position and orientation
  2) Rotation based on head-rotation sensor

The former transform is accessed via CameraRig.getTransform(), the latter is accessed via
CameraRig.getHeadTransform().

They are cascaded in the object hierarchy inside CameraRig.

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/GVRCameraRig.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRCameraRig.java
@@ -70,6 +70,20 @@ public class GVRCameraRig extends GVRComponent {
 
     /** Constructor helper */
     private void init(GVRContext gvrContext) {
+        /*
+         * Create an object hierarchy
+         *
+         *             [camera rig object]
+         *                     |
+         *           [head transform object]
+         *            /        |          \
+         * [left cam obj] [right cam obj] [center cam obj]
+         *
+         * 1. camera rig object: used for camera rig moving and turning via
+         *    CameraRig.getTransform()
+         * 2. head transform object: used internally to do sensor-based rotation
+         */
+
         setOwnerObject(new GVRSceneObject(gvrContext));
         getOwnerObject().attachCameraRig(this);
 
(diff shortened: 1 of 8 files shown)


GVRf/Framework/src/org/gearvrf/GVRCameraRig.java
```diff
@@ -54,10 +57,27 @@
...
     }
 
     private GVRCameraRig(GVRContext gvrContext, long ptr) {
         super(gvrContext, ptr);
+        init(gvrContext);
+    }
+
+    /** Constructor helper */
+    private void init(GVRContext gvrContext) {
+        setOwnerObject(new GVRSceneObject(gvrContext));
```
chenchao1407: I am just thinking if this init() should be protected other than init() since in future if there is any  extension of GVRCameraRig, a protected init() may be more helpful.

chenchao1407:
Codes look fine. Just a little bit confused about the functionality of this pull request. Anyway, you will have to do the rebase since we merged Tom's center camera patch

GVRf/Framework/src/org/gearvrf/GVRCameraRig.java
```diff
@@ -240,9 +260,11 @@ public void setVec4(String key, float x, float y, float z, float w) {
      *            {@link GVRCamera Camera} to attach.
      */
     public void attachLeftCamera(GVRCamera camera) {
-        if (camera.getOwnerObject() == null) {
-            throw new IllegalArgumentException("Owner object not set correctly");
+        if (camera.hasOwnerObject()) {
+            throw new IllegalArgumentException("Owner object should not be set");
```
jason2kim: Is it better to replace if existing already?

danke-sra: done

danke-sra:
Rebased. Can we check it in before the master branch is changed?

chenchao1407:
Hi Danke.

Even though the problem of the marker in the middle of screen is fixed, but I can still see some other problems. The Picker of Eyepointee still not works well.

danke-sra:
@chenchao1407 

Fixed the issue with picker.cpp.

jason2kim:
it now looks okay to me

thomasflynn:
how's it look @chenchao1407 ?

chenchao1407:
Just verified with some samples, works fine. Should be good to merge.

danke-sra:
Sounds good. Please go ahead to merge it. 

BTW, if you prefer to collapse the 4 commits into 1, I can do it too.

PR #204, opened 2015-08-05
KSensor improvements

liaxim:
Can be reviewed.


liaxim:
Testing is very time consuming; will keep on testing it over the next couple of days. 

JNI crash with the eyepicking sample when dynamically switching from internal to ksensor/oculus. I have seen it before, without this change, but it is still suspicious. I will keep looking at it.

The most heavy-weight app that can possibly be used for testing needs updating as recent changes have broken it (it didn't seem like a straightforward removal of getOwnerObject references). So I haven't been able to use it today.

The sensored scene in the GVRViewManager is a pain and it is still half-baked. Once I move the internal sensor to the native code I will be probably able to improve that part.

liaxim:
Will test the mono renderer tomorrow; last time I broke it eventually as I forgot about it at the end.

thomasflynn:
tested.  seems to have a pretty noticeable difference.  anyone else have any comments?

dknaan:
I didn't see any difference on my S6 with the Eyepicking sample. 
@thomasflynn - which sample did you use to test?

dknaan:
I didn't see a difference also with the virtaul keyboard on my Note4

PR #208, opened 2015-08-08
Assimp shader, support for adding features to shader

deepakrawat22:
Added a new shader for 3D models which are imported using jAssimp. Now GearVRf can read all the available material property for 3D models using jAssimp, but lacks shader support. Assimp shader will be one shader for all assimp imported model and will be based on assimp viewer shader:
https://github.com/assimp/assimp/blob/master/tools/assimp_view/Shaders.cpp

This pull request is not complete yet and right now only support diffuse texture and material color constant. Adding new feature would be easy, just adding a new preprocessor and adding support for that feature in shader:
#define AS_DIFFUSE_TEXTURE                0x00000001

I feel the design I followed might look complicated, so any design feedback would be helpful.

EDIT: Basic assimp shader is completed. Right now the shader only support diffuse texture if present and if not present fall back to material color properties. Later on all the material properties supported by assimp can be easily added.

drawat@usc.edu

### DIFF
--- a/GVRf/Framework/jni/shaders/material/assimp_shader.h
+++ b/GVRf/Framework/jni/shaders/material/assimp_shader.h
@@ -32,8 +32,8 @@
 #define ISSET(num, i)                    ((num & (1 << i)) != 0)
 #define CLEARBIT(num, i)                 num = (num & ~(1 << i))
 
-#define AS_DIFFUSE_TEXTURE                0x00000001
-#define AS_SPECULAR_TEXTURE               0x00000002
+#define AS_DIFFUSE_TEXTURE                0x00000000
+#define AS_SPECULAR_TEXTURE               0x00000001
 
 /*
  * As the features are incremented, need to increase AS_TOTAL_FEATURE_COUNT
--- a/GVRf/Framework/jni/shaders/material/assimp_shader.h
+++ b/GVRf/Framework/jni/shaders/material/assimp_shader.h
@@ -1,0 +1,86 @@
+/* Copyright 2015 Samsung Electronics Co., LTD
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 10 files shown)


GVRf/Framework/jni/shaders/material/assimp_shader.h
```diff
@@ -0,0 +1,86 @@
...
+#include "GLES3/gl3.h"
+#include "glm/glm.hpp"
+#include "glm/gtc/type_ptr.hpp"
+
+#include "objects/recyclable_object.h"
+
+#define SETBIT(num, i)                   num = (num | (1 << i))
+#define ISSET(num, i)                    ((num & (1 << i)) != 0)
+#define CLEARBIT(num, i)                 num = (num & ~(1 << i))
+
```
deepakrawat22: Just realized this seems a bug. Used some old code, so didn't paid much attention.

thomasflynn:
i reviewed the code with Deepak.  I understand his code.  just needs testing now.

chenchao1407:
Is there any sample you can add for verification issue?

thomasflynn:
you can use the jassimp model viewer sample.  though, it currently crashes for me with this change.  Deepak is looking into it.

deepakrawat22:
I made some fail safe changes, like making shader_feature_set as unsigned int and a check for that it's value should not be greater that GL_PROGRAM_COUNT. But I am not able to reproduce the crash myself. I wonder why the value for shader_feature_set is not correct (Its passed from Java Material).

chenchao1407:
I get the same crash here also. I will help check this issue.

deepakrawat22:
Fixed the crash issue. Turns out the assimp_shader_ pointer was not set to null in constructor so it has garbage value.

thomasflynn:
yay!  it works now :)

PR #209, opened 2015-08-08
Moveculling

thomasflynn:
Moved culling outside of renderCamera().  Culling will now happen before the left eye is drawn.  Culling will only happen once per frame now.  Yay :).  Culling also now uses the center camera of the camera rig.

tom.flynn@samsung.com

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/GVRViewManager.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRViewManager.java
@@ -463,6 +463,11 @@ class GVRViewManager extends GVRContext implements RotationSensorListener {
         if (!(mSensoredScene == null || !mMainScene.equals(mSensoredScene))) {
             GVRCameraRig mainCameraRig = mMainScene.getMainCameraRig();
 
+            if(eye == 0) {
+                GVRPerspectiveCamera centerCamera = mainCameraRig.getCenterCamera();
+                cull(mMainScene.getNative(), centerCamera.getNative(), mRenderBundle.getMaterialShaderManager().getNative());
+            }
+
             if (eye == 1) {
                 GVRCamera rightCamera = mainCameraRig.getRightCamera();
                 renderCamera(mActivity.getAppPtr(), mMainScene, rightCamera,
@@ -639,7 +644,6 @@ class GVRViewManager extends GVRContext implements RotationSensorListener {
                     mSplashScreen = null;
                 }
             }
-            cull(mMainScene.getNative(), mMainScene.getMainCameraRig().getCenterCamera().getNative(), mRenderBundle.getMaterialShaderManager().getNative());
         }
 
         public void onDrawFrame() {
@@ -663,8 +667,6 @@ class GVRViewManager extends GVRContext implements RotationSensorListener {
             doMemoryManagementAndPerFrameCallbacks();
 
             mScript.onStep();
-
-            cull(mMainScene.getNative(), mMainScene.getMainCameraRig().getCenterCamera().getNative(), mRenderBundle.getMaterialShaderManager().getNative());
         }
 
         public void onDrawFrame() {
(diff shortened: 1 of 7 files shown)


thomasflynn:
any comments on this?

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -462,6 +463,11 @@ void onDrawEyeView(int eye, float fovDegrees) {
         if (!(mSensoredScene == null || !mMainScene.equals(mSensoredScene))) {
             GVRCameraRig mainCameraRig = mMainScene.getMainCameraRig();
 
+            if(eye == 0) {
+                GVRPerspectiveCamera centerCamera = mainCameraRig.getCenterCamera();
+                cull(mMainScene.getNative(), centerCamera.getNative(), mRenderBundle.getMaterialShaderManager().getNative());
+            }
+
```
dknaan: I think it would be better to use a flag to indicate if cull was already calculated. In the past I saw several places in the code where the left eye was not called before the right eye, so we can't assume it.

In general. I would prefer to use LEFT_EYE, RIGHT_EYE, BOTH_EYES enums instead of 0,1
There is already an enum definition in GVRConsole which we have to move to somewhere else and use everywhere.

chenchao1407: By general design, I am thinking onDrawEyeView, which will be called twice per frame, will handle rendering for each eye. And onDrawFrame is for pre-draw processing for a frame. I am just wondering if there is any reason that you put the culling here but not onDrawFrame?

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -462,6 +463,11 @@ void onDrawEyeView(int eye, float fovDegrees) {
         if (!(mSensoredScene == null || !mMainScene.equals(mSensoredScene))) {
             GVRCameraRig mainCameraRig = mMainScene.getMainCameraRig();
 
+            if(eye == 0) {
+                GVRPerspectiveCamera centerCamera = mainCameraRig.getCenterCamera();
+                cull(mMainScene.getNative(), centerCamera.getNative(), mRenderBundle.getMaterialShaderManager().getNative());
```
jason2kim: isn't this called again in the else { // if ... statement below?

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
thomasflynn: I actually did that initially, but was seeing weird things happening.  Objects would flicker in and out of existence right in front of me.  It was wierd.  Moved it to here and everything worked perfectly.  I agree this isn't the cleanest.  Let fight with it a little more and see if i can get it working from onDrawFrame.

chenchao1407: OK. Maybe by the end of this week to see if you can handle this. After that if there is still a problem, we can merge it firstly and then figure out if we can move it to onDrawFrame with another pull request.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -462,6 +463,11 @@ void onDrawEyeView(int eye, float fovDegrees) {
         if (!(mSensoredScene == null || !mMainScene.equals(mSensoredScene))) {
             GVRCameraRig mainCameraRig = mMainScene.getMainCameraRig();
 
+            if(eye == 0) {
```
jason2kim: Isn’t this if(eye == 0) statement is redundant, as it will go to the else statement below? Eye will be either 0 or 1. Also, it assumes it is called 0 first.

thomasflynn:
ok, moved it to onDrawFrame().  Seems to be working fine now.  Weird.  Please test before merging.

thomasflynn:
ok, @dknaan?

dknaan:
Code looks ok. I also tried to run some samples and everything looks fine.

PR #217, opened 2015-08-11
Avoid overdrawing except for transparent objects and overlays

salgubel:
To avoid over drawing, draw objects front to back based on camera distance (continue to draw transparent objects and overlays in reverse order back to front)

GearVRf-DCO-1.0-Signed-off-by: Srinivasa Algubelli
s.algubelli@samsung.com


chenchao1407:
Hi, please put the DCO signature into commit message. Thanks.

PR #223, opened 2015-08-15
Hierarchical Bounding Volumes

thomasflynn:
Changed the bounding box / bounding sphere code to be a unified bounding volume consisting of both.  In addition, add a bounding volume to SceneObject and that volume is encompasses all of the children of that SceneObject.

Next step (not covered by this check-in) will be to cull based on that SceneObject's bounding volume.

tom.flynn@samsung.com


thomasflynn:
any comments?

GVRf/Framework/jni/objects/bounding_volume.h
```diff
@@ -0,0 +1,112 @@
...
+
+        float s = (float) sqrt((radius_/3.0f));
+        min_corner_ = glm::vec3(center_[0] - s,
+                                center_[0] - s,
+                                center_[0] - s);
+
+        max_corner_ = glm::vec3(center_[0] + s,
+                                center_[0] + s,
+                                center_[0] + s);
+    }
```
dknaan: I probably miss-understand something here regarding the corners settings based on the center and radius: isn't it a bug that center_[0] is used for all the dimensions, and is it ok that 's' equals to sqrt((radius_/3.0f))  ?

GVRf/Framework/jni/objects/bounding_volume.h
```diff
@@ -0,0 +1,112 @@
...
+public:
+    BoundingVolume() {
+        center_ = glm::vec3(0.0f, 0.0f, 0.0f);
+        min_corner_ = glm::vec3(0.0f, 0.0f, 0.0f);
+        max_corner_ = glm::vec3(0.0f, 0.0f, 0.0f);
+    }
+
+    ~BoundingVolume() {
+    }
+
```
dknaan: Is there any special reason this code is not in a separate cpp file?

GVRf/Framework/jni/objects/bounding_volume.h
```diff
@@ -0,0 +1,112 @@
...
+            max_corner_[1] = point[1];
+        }
+        if(max_corner_[2] < point[2]) {
+            max_corner_[2] = point[2];
+        }
+
+        center_ = (min_corner_ + max_corner_)*0.5f;
+        radius_ = (max_corner_ - center_).length();
+    }
+
```
dknaan: Is there any special reason this code is not in a separate cpp file?

dknaan: I think that a documentation inside the code can make it clearer

chenchao1407:
Overall codes look good to me. The mechanism, in general, makes sense to me.

However, it will somehow reduce the efficiency of frustum culling. You can see in jassimp loader sample, the frustum culling performs worse than without this patch.

GVRf/Framework/jni/objects/bounding_volume.h
thomasflynn: sorry, there were some bugs in there.  center_[0] shouldn't be used for all the dimensions.  moved the implementation into a cpp file and tried to comment that method better.

thomasflynn:
added a new patch last night.  moved the implementation to a cpp file.  tried to comment and explain the one expand() method.

thomasflynn:
@chenchao1407, try testing again.  I did fix a bug in the bounding volume calculation.  I think i'm getting the expected results now.

dknaan:
Code looks fine. Thanks for the documentation.
Works well on my Mali Note4 and S6 Edge devices (I didn't compare performance with frustum culling)

thomasflynn:
ok, found the bug.  wasn't initializing my min and max corners correctly.  fixed.  should give the expected results now.  please test again @chenchao1407

chenchao1407:
Good to me now.

PR #226, opened 2015-08-19
stereo shaders: enable mono rendering.

bfouet-gpsw:
Add a management in the stereo shaders to be able to render stereo
sources as mono (right eye only).

Tested successfully with both OES stereo shaders. The unlit ones are untested.


thomasflynn:
Good to see another pull-request from you Ben :).  Hope things are going well.  The change looks innocent enough.  Will merge it after we test it locally.  Thanks!

PR #228, opened 2015-08-20
Add toString to GVRTransform

liaxim:


thomasflynn:
yeah, that looks handy.  looks good to me too.

PR #233, opened 2015-08-26
Prediction tweaks

liaxim:
This is not meant for merging; it shows the effect of the ::predict tweak I am planning to eliminate the shimmering. It doesn't improve the actual prediction but as far as I can tell doesn't degrade it while the shimmering is eliminated.

Still researching the motion blur - causes and effects are not clear to me. My impression is that it is not just a function of the prediction.

Exposed is a significant gyro drift problem with ksensor which causes at least some of the aliasing once the shimmering is removed.

By default the old KSensor is activated; long-pressing the back key switches modes:
0 - old KSensor
1 - the calculation leading to the shimmering is eliminated
2 - additional rotation (that doesn't do anything noticeable) eliminated also
3 - switches to Oculus

My plan is to use mode 2.

### DIFF
--- a/GVRf/Framework/jni/objects/components/camera_rig.cpp
+++ b/GVRf/Framework/jni/objects/components/camera_rig.cpp
@@ -136,19 +136,15 @@ glm::quat CameraRig::predict(float time, const RotationSensorData& rotationSenso
             / 1000000000.0f;
 
     glm::vec3 axis = rotationSensorData.gyro();
+    //the magnitude of the gyro vector should be the angular velocity, rad/sec
     float angle = glm::length(axis);
 
+    //normalize the axis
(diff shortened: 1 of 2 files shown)


thomasflynn:
ok, sat down and tested it.  2 looks fine.  the drift isn't any worse than 0 and is definitely more stable.  go ahead and clean this up.

liaxim:
Cleaned up the change. To switch to KSensor and actually test it modify activity_jni.h and replace "typedef GVRActivityT<OculusHeadRotation> GVRActivity;" with "typedef GVRActivityT<KSensorHeadRotation> GVRActivity;".

thomasflynn:
ok, built, tested, looks good to me.

PR #238, opened 2015-08-28
Fix some problems with loading models with getAssimpModel()

thomasflynn:
There was a problem with a jni table overflow because some references were not released.  Fixed that.
Also added the object's name to the SceneObject so that objects can be found and manipulated.
Finally, more gracefully handle the case where diffuse textures are not found.

tom.flynn@samsung.com

### DIFF
--- a/GVRf/Framework/jni/engine/importer/jassimp.cpp
+++ b/GVRf/Framework/jni/engine/importer/jassimp.cpp
@@ -57,7 +57,6 @@ public:
 bool create_instance(JNIEnv *env, const char* class_name,
         jobject& new_instance) {
     jclass java_class = env->FindClass(class_name);
-    DeleteLocalRef clazzRef(env, java_class);
 
     if (NULL == java_class) {
         return false;
@@ -73,6 +72,7 @@ bool create_instance(JNIEnv *env, const char* class_name,
     if (NULL == new_instance) {
         return false;
     }
+    DeleteLocalRef clazzRef(env, java_class);
 
     return true;
 }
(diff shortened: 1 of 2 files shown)


thomasflynn:
@lokeshguddu and @ejsamsung, could you both please review this?  thanks.

GVRf/Framework/jni/engine/importer/jassimp.cpp
```diff
@@ -36,6 +36,24 @@
...
+    DeleteLocalRef(JNIEnv* env, jfloatArray& farray) : mEnv(env), mObject((jobject&) farray) {};
+    DeleteLocalRef(JNIEnv* env, jintArray& iarray) : mEnv(env), mObject((jobject&) iarray) {};
+    DeleteLocalRef(JNIEnv* env, jstring& str) : mEnv(env), mObject((jobject&) str) {};
+    ~DeleteLocalRef() {
+        if (mObject != NULL) {
+            mEnv->DeleteLocalRef(mObject);
+        }
+    }
+};
+
```
chenchao1407: I think this class can be for wider use. My thinking is that creating a seperate class with its own h file may be a little bit better.

lokeshguddu:
Hi @thomasflynn ,

The modifications looks good to me. The addition of the name for the scene object was a good idea. It will help in tracing the right object when trying to get hold of the render data etc. for a specific node replicated as scene object.

ejsamsung:
Alright, the total stack of changes here looks OK.  Won't have a chance to build or test, but the code looks good.

GVRf/Framework/jni/engine/importer/jassimp.cpp
liaxim: If willing to go through the hassle to create a separate header you might consider making DeleteLocalRef a template class. It will yield fewer lines of code.

PR #242, opened 2015-09-02
Mono fix

danke-sra:


thomasflynn:
@chenchao1407 does the code look correct to you?  It involves getting the EyeBuffer parameters.  I can cherry-pick and test it, i'd just like your verification that he's getting the correct settings.

thomasflynn:
ok, verified the fix works as expected.

GVRf/Framework/src/org/gearvrf/GVRLensInfo.java
```diff
@@ -80,7 +100,20 @@ public int getFBOWidth() {
      * @return current distortion FBO height value
      */
     public int getFBOHeight() {
-        return mAppSettings.getEyeBufferParms().getResolution();
+        if (!mCustomFBOSize)
+            return mFBOWidth = mAppSettings.getEyeBufferParms().getResolution() *
+                    mAppSettings.getEyeBufferParms().getWidthScale();
```
chenchao1407: Here you are returning mFBOWidth for getFBOHeight?

chenchao1407:
There is a typo. And given the fact that this should be only for mono but the fix will also potentially affect VR mode which may cause unpredictable result. My advice is like you can add parameters into VrAppSettings.MonoModeParms to do kinds of restriction.

chenchao1407:
Just double checked the patch. Seems I made a mistake previously. Will verify that tmr morning.

chenchao1407:
@danke-sra Hi Danke, can you help rebase this pull request?

Thanks

danke-sra:
Rebased

PR #244, opened 2015-09-04
Cleanup of Mesh interface and ensure the VAO is generated when any of the underlying data defining the mesh is changed.

Vittalbabu:
holding vertices, normals, etc. via get methods except via a const
reference. Update the jni code to use the const variants of these
functions.
2. Set the dirty flag when any of the data structures are set. This will
force creation of the VAO at the next render call. Remove the
vaoInitialised flag I introduced in my last checkin and use vao_dirty_
to identify when VAOs need to be regenerated.

### DIFF
--- a/GVRf/Framework/jni/objects/mesh.cpp
+++ b/GVRf/Framework/jni/objects/mesh.cpp
@@ -180,8 +180,8 @@ void Mesh::getTransformedBoundingBoxInfo(glm::mat4 *Mat,
 void Mesh::generateVAO() {
 #if _GVRF_USE_GLES3_
 
-	if (vaoInitiliased_)
-		return;
+    if (vaoInitiliased_)
+        return;
     GLuint tmpID;
 
     if (vao_dirty_) {


Vittalbabu:
The changes pass the samples I tested like modelviewer, lishader. Will continue testing to validate that VAO is created/regenerated as appropriate.

GVRf/Framework/jni/objects/mesh.h
```diff
@@ -164,16 +144,7 @@ class Mesh: public HybridObject {
...
-
-    std::vector<glm::vec2>& getVec2Vector(std::string key) {
-        auto it = vec2_vectors_.find(key);
-        if (it != vec2_vectors_.end()) {
-            return it->second;
-        } else {
-            std::string error = "Mesh::getVec2Vector() : " + key + " not found";
-            throw error;
-        }
+        vao_dirty_ = true;
```
thomasflynn: I haven't looked at things closely enough here, so correct me if i'm wrong, but don't these set*Vector() methods map to setting uniforms?  if so, is it necessary to dirty the vao?

Vittalbabu:
I will check these calls and remove the dirty flag.

chenchao1407:
This one needs to be rebased.

GVRf/Framework/jni/objects/mesh.h
Vittalbabu: @thomasflynn These are used for attributes. The vao needs to be dirtied.

PR #245, opened 2015-09-04
Improve robustness of mesh.cpp

danke-sra:
Currently, some variables in mesh.cpp are not initialized, but
might be passed to glDelete* functions. This causes some glErrors
and might accidentally delete GL objects used by other meshes.

This patch initializes these variables to GL_INVALID (-1), and
added sanity checks in gl_delete.cpp. Stack trace is logged with
the tag "gvrf" to help debugging if invalid values are detected
by gl_delete.

GearVRf-DCO-1.0-Signed-off-by: Danke Xie
d.xie@samsung.com

### DIFF
--- a/GVRf/Framework/jni/engine/memory/gl_delete.cpp
+++ b/GVRf/Framework/jni/engine/memory/gl_delete.cpp
@@ -13,14 +13,25 @@
  * limitations under the License.
  */
 
-//#include "util/gvr_log.h"
+#include "util/gvr_log.h"
 #include "gl_delete.h"
+#include "util/gvr_cpp_stack_trace.h"
 
 namespace gvr {
 
 GlDelete gl_delete;
 
+void GlDelete::logInvalidParameter(const char *funcName) {
+    LOGW("GlDelete::%s is called with an invalid parameter", funcName);
+    printStackTrace();
+}
+
 void GlDelete::queueBuffer(GLuint buffer) {
+    if (buffer == GVR_INVALID) {
+        logInvalidParameter(__func__);
+        return;
+    }
+
     lock();
     buffers_.push_back(buffer);
 //    LOGD("queueBuffer(%d) buffers_.size() = %d", buffer, buffers_.size());
@@ -29,6 +40,11 @@ void GlDelete::queueBuffer(GLuint buffer) {
 }
 
 void GlDelete::queueFrameBuffer(GLuint buffer) {
+    if (buffer == GVR_INVALID) {
+        logInvalidParameter(__func__);
+        return;
+    }
+
     lock();
     frame_buffers_.push_back(buffer);
 //    LOGD("queueFrameBuffer(%d) frame_buffers_.size() = %d", buffer,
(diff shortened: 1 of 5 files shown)


GVRf/Framework/jni/engine/memory/gl_delete.cpp
```diff
@@ -13,14 +13,25 @@
...
+#include "util/gvr_log.h"
 #include "gl_delete.h"
+#include "util/gvr_cpp_stack_trace.h"
 
 namespace gvr {
 
 GlDelete gl_delete;
 
+void GlDelete::logInvalidParameter(const char *funcName) {
+    LOGD("GlDelete::%s is called with an invalid parameter", funcName);
```
thomasflynn: should this be a LOGE?

GVRf/Framework/jni/engine/memory/gl_delete.h
```diff
@@ -20,6 +20,8 @@
 #include <pthread.h>
 #include "GLES3/gl3.h"
 
+#define GL_INVALID ((GLuint)(-1))
```
thomasflynn: I'd prefer something like GVR_INVALID.  GL_INVALID looks like it could be a legit GL_ENUM and could confuse people.  Further 0 is likely a better value for invalid since it is usually reserved and you cannot delete vertex array/texture/program/buffer 0.

danke-sra:
There seems to be 2 ways of fixing this issue. 1) add checks for invalid values and warn users to correct them 2) set the initial values to 0, and let glDelete* silently ignore those values.

I think it is kinda a matter of style. In memory management, it is acceptable to allow deleting null pointers or 0 values (silently ignore them). So I think we can simply go for 2).

What do you think?

thomasflynn:
well, my point was that any number other than 0 is valid.  A driver could actually use the number you have set as GVR_INVALID.  While it seems most drivers seem to start ids at 1 and increment, that's not required and I've seen drivers return values that are large (I look for a pattern, just learned my assumption was wrong the hard way :).

While the GL spec says to ignore a delete of an id of 0, we could perhaps issue a warning.  Seems to me that if you're deleting 0, you may have a bug.

danke-sra:
Changed the invalid value GVR_INVALID to 0, and print a warning message if the invalid value is passed.

GVRf/Framework/jni/util/gvr_cpp_stack_trace.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+void printStackTrace(unsigned int max_frames)
+{
+    void* buffer[max_frames];
+    std::ostringstream oss;
+
+    dumpBacktrace(oss, buffer, captureBacktrace(buffer, max_frames));
+
+    __android_log_print(ANDROID_LOG_DEBUG, TAG, "%s", oss.str().c_str());
+}
```
chenchao1407: in util/gvr_log.h. There is already a macro wrapper for __android_log_print. Maybe you should use that one instead?

danke-sra:
Addressed @chenchao1407 's comment

chenchao1407:
Looks good to me. And it seems that everything works fine with this pull request.

PR #247, opened 2015-09-10
Fix JNI crash due to using a JNI environment attached to the UI thread ...

liaxim:
... from the Oculus native VR thread on shutdown. Plus small cleanup.

Unfortunately there is another crash, this time in Oculus so this
does't help. Not sure I will be able to do something about the Oculus
crash (in vrapi_Shutdown there seems to be use of an invalid JNI
object). The Oculus VrCubeWorld_Framework sample exhibits the same
behavior..

So this is it for now. Will try few more things but we might have to
hope that the new Oculus SDK remedies the problem.

### DIFF
--- a/GVRf/Framework/jni/oculus/activity_jni.cpp
+++ b/GVRf/Framework/jni/oculus/activity_jni.cpp
@@ -33,13 +33,17 @@ long Java_org_gearvrf_GVRActivity_nativeSetAppInterface(
         jstring fromPackageName, jstring commandString,
         jstring uriString)
 {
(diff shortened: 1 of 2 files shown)


thomasflynn:
perhaps you can try with 0.6.1, see if that still crashes?

thomasflynn:
checked the code, looks fine.  have confirmed the jni problem in oculus.  this does fix the jni problem in gvrf.  any objections to merging this?

liaxim:
No Oculus JNI crash in 0.6.1.

PR #253, opened 2015-09-15
Using GrowBeforeQueueThreadPoolExecutor as the threadPool

xcaostagit:
We are switching to a new thread pool since it is considered dangerous to have a thread pool with unlimited threads


liaxim:
Looks good to me. If there are no objections I will verify and merge tomorrow afternoon.

danke-sra:
Looks good.

PR #259, opened 2015-09-28
Cleanup of mesh interface -- allow access only via const references.

Vittalbabu:


thomasflynn:
looks good to me.  any other comments or objections to this?

GVRf/Framework/jni/objects/mesh.cpp
```diff
@@ -180,12 +180,11 @@ void Mesh::getTransformedBoundingBoxInfo(glm::mat4 *Mat,
...
 
-    if (vaoInitiliased_)
-        return;
+
     GLuint tmpID;
 
-    if (vao_dirty_) {
-         deleteVaos();
+    if (!vao_dirty_) {
+         return;
```
liaxim: I am curious why the deleteVaos() call is no longer necessary in case of a "dirty" mesh?

Vittalbabu: We reuse the old VAO for the mesh in the “common” case. This function will regenerate a new VAO only if the mesh is dirty – if we changed the attributes that define the VAO.

Liahim: I see. When a new VAO is generated in the dirty case, what happens with the old VAO? It doesn't need explicit cleanup? Thanks.

Vittalbabu: Good catch.

The old VAO should be deleted and a new VAO created. I missed the delete call. I will add it back and ensure it is properly called. I will call deleteVaos() but test.

PR #260, opened 2015-09-28
Controls sample final version

Maginador:
Controls sample update, adding worm shadow , animations duration change, play/replay buttons after changing any worm related animation.

This is the finished version of this sample.

GearVRf-DCO-1.0-Signed-off-by: Ricardo Maginador
ricardo.cm@samsung.com


thomasflynn:
looks good.

PR #263, opened 2015-09-30
Add setters/getters for near/far clipping planes to GVRCameraRig

thomasflynn:
We keep hitting this on one of our internal projects, so I'm creating a convenience function.  It'd be nice to set the near/far clipping planes for all the cameras attached to the camera rig from the camera rig itself.  So that's what this does.

tom.flynn@samsung.com


liaxim:
Looks good to me.

PR #265, opened 2015-10-07
FYI: Yaw drift correction - work-in-progress

liaxim:
Not meant for merging


liaxim:
Updated today; can keep the scene in-place if the phone is put on the table.

Having difficulty with very slow head rotation - the yaw drift correction can pull the scene along with the rotation. Looking into it more.

liaxim:
Ok, I thought I could get by without using different reference points but apparently I can't. The reference point needs to be close enough (<5 degrees) to avoid too large of a correction being generated and this "pull" that you get with my current change. The correction angle is a function of the angle between the current orientation and the orientation (both grow together). On to adding them reference points.

liaxim:
Just realized that we are drifting much much more than Oculus (turned off their yaw drift correction to see what happens). Say 5-6 times more. Unless the drift itself is minimized the correction is going to over-correct and be noticeable.

liaxim:
The extra drift (seems to be by an order of magnitude) is caused by the missing thermal calibration of the gyroscope readings. This means our readings are extra noisier and thus drift-ier.

liaxim:
I will get back to this at later time. For now I have another pull request that filters out some of the gyro noise and reduces the drift rate significantly.

Issue #239, opened 2015-08-29
Find SceneObject by name


thomasflynn:
I can't believe this doesn't already exist (that I could find anyway).  We should add a method to both GVRScene and GVRSceneObject to find a SceneObject by name.  So when people load in their models, they can find the articulated objects (by name) and move them.

chenchao1407:
Let me take care of it.

thomasflynn:
how's that coming along?-)  or should I take it up?

liaxim:
@thomasflynn, I could help with this one if you'd like.

thomasflynn:
Sure, go for it.  Should be quick enough.

liaxim:
The corresponding APIs have been added to GVRScene and GVRSceneObjects.

PR #266, opened 2015-10-07
Find SceneObject by name #239

liaxim:
Is this what @chenchao1407 had in mind?

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
@@ -563,6 +563,26 @@ public class GVRSceneObject extends GVRHybridObject {
     }
 
     /**
+     * Performs case-sensitive search
+     * 
+     * @param name
+     * @return null if nothing was found or name was null/empty
+     */
+    public GVRSceneObject[] getSceneObjectsByName(final String name) {
+        if (null == name || name.isEmpty()) {
+            return null;
+        }
+
+        final List<GVRSceneObject> matches = new ArrayList<GVRSceneObject>();
+        if (name.equals(getName())) {
+            matches.add(this);
+        }
+        GVRScene.getSceneObjectsByName(matches, mChildren, name);
+
+        return 0 != matches.size() ? matches.toArray(new GVRSceneObject[matches.size()]) : null;
+    }
+
+    /**
      * Check if {@code otherObject} is colliding with this object.
      * 
      * @param otherObject
--- a/GVRf/Framework/src/org/gearvrf/GVRScene.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRScene.java
@@ -144,6 +144,33 @@ public class GVRScene extends GVRHybridObject {
     }
 
     /**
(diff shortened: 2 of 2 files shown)


GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
```diff
@@ -563,6 +563,23 @@ public void removeChildObject(GVRSceneObject child) {
...
+     * Performs case-sensitive search
+     * 
+     * @param name
+     * @return null if nothing was found or name was null
+     */
+    public GVRSceneObject[] getSceneObjectsByName(final String name) {
+        if (null == name) {
+            return null;
+        }
+
```
thomasflynn: maybe i'm missing it, what if this object's name matches?

thomasflynn:
yeah, off-hand, this looks like it'll do it.  i only have the one question in the code.  @salgubel, i know you guys wrote your own version of this.  will this meet your needs so you don't need to use your own code for this?

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
liaxim: Good point, will upload a fix.

liaxim:
Checking the current GVRSceneObject's name and in addition to null queries, excluding empty string queries.

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
```diff
@@ -563,6 +563,26 @@ public void removeChildObject(GVRSceneObject child) {
...
 
     /**
+     * Performs case-sensitive search
+     * 
+     * @param name
+     * @return null if nothing was found or name was null/empty
+     */
+    public GVRSceneObject[] getSceneObjectsByName(final String name) {
+        if (null == name || name.isEmpty()) {
+            return null;
```
lalwani: Another small fail-safe check might be to do name.trim().isEmpty().

lalwani:
Looks good

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
Liahim: My goal was to filter out the name assigned by default (empty string). null could be considered an alias for this from user's perspective. Didn't want this API to be easily used as a replacement of getWholeSceneObjects.

If the user wants to use various number of spaces as names, I say let's not second guess the user.

lalwani: makes sense.

liaxim:
jMonkeyEngine's getChild(name) returns the first matching node.

liaxim:
Doesn't seem like OpenSceneGraph has direct match for this api; per http://trac.openscenegraph.org/projects/osg//wiki/Support/Tutorials/FindingNodes it is expected for the users to roll their own implementations.

thomasflynn:
perhaps we can also add a getSceneObjectByName (singular rather than plural) that returns the first matching node.  i'll merge this in the meantime.

PR #268, opened 2015-10-08
Use vrapi_GetPredictedDisplayTime instead of 0.0f for improved camera positioning

liaxim:


liaxim:
Seems to be helping but what I can notice is is in the fast moving case.

PR #273, opened 2015-10-12
Find SceneObject by name #239 - part 2

liaxim:
add the more convenient getSceneObjectByName api


thomasflynn:
Looks good.  Builds fine.

PR #275, opened 2015-10-12
Reduce default stack size

danke-sra:


thomasflynn:
off-hand, looks good.  @liaxim ?

liaxim:
My understanding is that ART uses unified java+native stack that unconditionally is 1MB+ (1MB + Java's stackSize + extras (per ./art/thread.cc)). The default should be comparable to Dalvik's - 1MB + 32KB + extras. So with this change are we sure that we are not actually increasing the stack size?

It doesn't seem like we can do much to meaningfully reduce the stack size.

danke-sra:
Can you please give reference for ART's stack size being unconditional?

liaxim:
204 static size_t FixStackSize(size_t stack_size) {
 205   // A stack size of zero means "use the default".
 206   if (stack_size == 0) {
 207     stack_size = Runtime::Current()->GetDefaultStackSize();
 208   }
 209 
 210   // Dalvik used the bionic pthread default stack size for native threads,
 211   // so include that here to support apps that expect large native stacks.
 212   stack_size += 1 * MB;
 213 
 214   // It's not possible to request a stack smaller than the system-defined PTHREAD_STACK_MIN.
 215   if (stack_size < PTHREAD_STACK_MIN) {
 216     stack_size = PTHREAD_STACK_MIN;
 217   }
 

Line 212

danke-sra:
Is Java default stack size still 32KB (as in Dalvik)? JRE seems to be 512KB and 1MB for 32bit and 64bit. If it is still around 32KB in ART, we can drop this change.

liaxim:
It is surprisingly tedious to find the actual default. I assume it is 32KB though as it is the one widely mentioned. Then again, it could be anything on our devices.. Will check with Kuldeep also.

liaxim:
The stack size of the gvrf worker threads is 1036KB. The stack size that can be controlled via java.lang.Thread is small enough already. Unfortunately there is no point in merging this change.

"adb shell kill -3 <pid>" would dump the threads (including the stackSize art chooses) into /data/anr/traces.txt.

danke-sra:
The stack size in Android seems to be small enough (Dalvik) or not configurable (ART). So close this commit and corresponding issue.

PR #276, opened 2015-10-13
Fix perf issue in loading models via assimp. The current code was

Vittalbabu:
re-reading the file twice for every single node with a mesh -- it was
loading the model file 1 + 2 * num_nodes.

For large model files (with a lot of nodes) this was very slow. One fbx
file with 146 nodes took 17 min to load.

The fix is to reuse the loaded data in later calls to load mesh and
material. The same fbx file now loads in 4 sec.


thomasflynn:
reviewed the code.  looks good.  wish we had caught it during the review when it was submitted.  the code builds and runs fine against the gvrjassimpmodelloader.

lokeshguddu:
@Vittalbabu great. Now we know what was the delay all about.

Thanks for the fix.

Issue #257, opened 2015-09-22
Bug in solar-system sample


elidelsonc:
Hello,

We have noted the following strange behaving with solar-system sample:

1 - the sample starts normally
2 - We press the back button and the "exit to oculus home ?" message appears
3 - if We press "no" to come back to the application, the application's sceneObjects don't appear anymore.
4 - Only the statistical display, switch to on/off if we touch in the pad.

what could be happening?

Best Regards,

Elidelson Carvalho
SIDIA BR - UX TEAM

thomasflynn:
i suspect we have a bug in onResume() or thereabouts with restoring the GL context and associated buffers.  will have to look more closely.  thanks for pointing this out.

liaxim:
Fix has been merged.

rahul27:
The texture loading issue seems to be fixed with this patch. But on relaunch the splash screen cuts off for me. Do you see the same?

liaxim:
Could you please open another issue then? I doubt it is related to the original reported here.

PR #277, opened 2015-10-13
Fix "Bug in solar-system sample #257"

liaxim:
it is actually a known issue; made texture and mesh caches non-static

plus fixed a potential npe in solar-view-manager


thomasflynn:
reviewed code, looks good.

PR #278, opened 2015-10-14
Minor fixes to gvr-vuforia sample

gaurav-mcl:
1. Avoid showing the teapot unless it is detected by the camera.
2. Change the teapot orientation so that its bottom is parallel to
the flat surface of the vuforia marker.

GearVRf-DCO-1.0-Signed-off-by: Gaurav Srivastava 
gaurav001.s@samsung.com


liaxim:
Verified

PR #280, opened 2015-10-14
fix: Astro boy model position and scale

lokeshguddu:
The sample app was not showing the astro boy model at properly so we changed some parameters. Now it's visible with proper orientation.


liaxim:
Verified

PR #281, opened 2015-10-14
Fix crash in loading complex models. The issue was we were not freeing

Vittalbabu:
one local refrence in load_scene_node() (for the jassimp_node). We need
to hold the local reference for the root node only.

Due to this we exhaust the limit of 512 local references and there is a
crash. Fix is to release local references except for the root node.


liaxim:
Looks good to me.

liaxim:
Verified using the gvrjassimpmodelloader sample.

In the future please make sure Eclipse uses spaces-only for indentation; also always using braces is highly desirable. Thanks.

Vittalbabu:
Will follow the guidelines in futre submissions.

Thanks,


-          Vittal

PR #282, opened 2015-10-14
Immersivepedia

nathan-almeida:
Immersivepedia, sample showing visual quality of GVRf in a interactive gallery.

GearVRf-DCO-1.0-Signed-off-by: Nathan Almeida
nathan.a@samsung.com


mteofilo:
Some instructions to better explore this sample:
1st scene (museum room) - Just to select dinosaur world. The others options are unavailable. Intends to illustrate an option menu.
2nd scene (dinosaur exposition) - each dinosaur has an associated totem, which has a respective action, like to show a gallery, a movie, a text or to rotate. These cited actions are triggered by looking the totem for a while. Moreover, the dinosaurs can be rotated by a swipe command.

GVRf/Sample/gvrf_immersivepedia/AndroidManifest.xml
```diff
@@ -0,0 +1,38 @@
...
+    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+
+    <application
+        android:allowBackup="true"
+        android:allowClearUserData="true"
+        android:hardwareAccelerated="true"
+        android:icon="@drawable/ic_launcher"
+        android:label="@string/app_name"
+        android:theme="@style/AppTheme"
+        android:vmSafeMode="true" >
```
liaxim: Is vmSafeMode still necessary?

GVRf/Sample/gvrf_immersivepedia/src/org/gearvrf/immersivepedia/scene/DinosaurScene.java
```diff
@@ -0,0 +1,183 @@
...
+                object.getRenderData().getMaterial().setOpacity(0f);
+            }
+        }
+    }
+
+    private GVRSceneObject createSkybox() {
+
+        GVRMesh mesh =  getGVRContext().loadMesh(new GVRAndroidResource(getGVRContext(), R.raw.environment_walls_mesh));
+        GVRTexture texture = getGVRContext().loadTexture(Loader.loadResourceFromFile(getGVRContext(),
+                "environment_walls_tex_diffuse.png", 0));
```
liaxim: This problem is present in few places. I don't have "environment_walls_tex_diffuse.png" in my externalStorageDirectory - should I? Otherwise this and few other textures do not load and the app bails out.

liaxim:
This is quite nice but there are few additional issues:

Note4: frame rate is pretty low; do you think you could improve it somewhat? For example, in other apps we've had similar scenes like this initial options menu that have much higher frame rate. I get some of the background elements on top of the dinosaur models, parts of the text ends up behind the model. Maybe you need to play less with the rendering order.

S6: frame rate is really bad but it seems everything looks much better and I didn't notice ordering issues.

If you could improve things for Note4 we can merge this in. We probably have to investigate what's goind on S6.

liaxim:
@nathan-almeida @mteofilo Does any of my feedback make sense? Thanks.

douglasalipio:
We've tested with version SM-N910C (Note 4) and SM-G925F (S6). 
The application ran with a high frame rate for these devices. 
We don't have any US device model available to check your opened issue. We've reviewed the source code a couple of times and didn't find any evidence about low frame rate.
The others issues was fixed in last two commits.
Thanks for review

liaxim:
I see. We will have someone other than me try it out just in case.

In the meantime, is there a chance you might cull some of the unused resources? For example dinos_video_wip.mp4 and tron.mp4 seem unused. Thanks.

liaxim:
This is by the way what I am seeing on my AT&T Note 4.

screenshot_2015-10-22-18-39-56

liaxim:
On my S6 the performance is very good but if I press the back key and choose to go back to the app, the models are no longer visible. Can you reproduce this? Do you think it is a GVRF issue or something in the app? Thanks.

mteofilo:
Hi @liaxim, the cited problem (back button) was registered in #257 issue. You've found this issue in our app and in other samples, like solar system sample. We'll remove unused resources.

mteofilo:
Hi @liaxim, I've tested right now the solar system sample and we can found the same error in this sample (the models are not been displayed after back button action). Then, I assume that is a GVRF issue.

douglasalipio:
@liaxim 
Remove unused videos dinos_video_wip.mp4 and tron.mp4. Anything else?

liaxim:
Good, we will look into the missing models. I don't have anything else.

liaxim:
Hi @douglasalipio @mteofilo , do you think you could update the sample to support API level 19? Specifically Totem.java:126 fails at runtime on phones at API 19. Thanks. I know that @thomasflynn wants that too.

Also if possible please use 2x or 4x msaa instead of 8x - there are phones that do not support 8x and the app crashes. It is a GVRF bug but until it is fixed this would help.

douglasalipio:
Hi Mihail

Ok, I will do.

Thanks

PR #287, opened 2015-10-22
Yaw drift reduction

liaxim:


liaxim:
The "Add filtering" commit is the one that will stay. The "Temporary" one is just so testing this change is easier - a long press of the back key toggles on and off the bias detection. I will remove it before merging.

The change to ktracker_sensor_filter.h is to use the proper c++ operators for memory management; the rest is just formatting.

With this change the drift is hardly noticeable. Without this change just leave the headset on the table and wait 10 mins - the drift will be like 90 degrees.

I have been using Note 4.

liaxim:
Dotan and Vittal, I would appreciate your help reviewing and verifying this change. Thanks.

dknaan:
I can confirm that on my Note4 the drift is about 5 times slower with the new version, but it still exists (which means that if you watch a 1.5 hour movie with the headset, you have to gradually rotate around yourself).

On S6 the drift is much slower for both old and new version, and is moving to the opposite direction. I'm not sure if the new version is better than the old one in this case. On Monday I can do additional experiments along longer time period to get more significant results.

liaxim:
Thanks for the feedback. I will look more into S6 - I guess the sensor in S6's GearVR is much less noisy.

liaxim:
I made some corrections and I see improvements on mine S6. First the filter could've used random values occasionally - I fixed the default vec3 ctor to zero-initialize members. Also I take into account changes in the temperature - as noise changes with temperature.

liaxim:
@dknaan , had any chance trying out the latest?

liaxim:
Dotan showed me that there is nothing noticeably better on his device. What I found out is that his headset produces occasionally peak readings that skew a lot the mean that is used as offset/bias. By increasing the size of the filter things get slightly better but way off the reference. The current situation is that on Note 4 the drift is reduced 5x but even with larger filter, the drift on S6 is the same with or without this change (still too large).

I discovered that there is a special ioctl call that can retrieve calibration information from the device (/dev/ovr0) itself. I need to find a way to put this information to use.

liaxim:
Ok, I have an updated version. Do not forget to change activity_jni.h from

```
typedef GVRActivityT<OculusHeadRotation> GVRActivity;
//typedef GVRActivityT<KSensorHeadRotation> GVRActivity;
```
to
```
//typedef GVRActivityT<OculusHeadRotation> GVRActivity;
typedef GVRActivityT<KSensorHeadRotation> GVRActivity;
```

before trying it out. I corrected few filtering bugs so the offset is much more stable now.

GVRf/Framework/jni/sensor/ksensor/k_sensor.h
```diff
@@ -77,11 +77,93 @@ class KSensor {
...
+
+    bool factoryCalibration = false;
+    vec3 factoryAccelOffset_;
+    vec3 factoryGyroOffset_;
+    mat4 factoryAccelMatrix_;
+    mat4 factoryGyroMatrix_;
+    float factoryTemperature_ = 0.0f;
+
+    static const int KGyroNoiseFilterCapacity = 6000;
+    static const int KRadiansToDegrees = 180/M_PI;
```
thomasflynn: did you intend this to be an int?

liaxim: No. Thanks for the catch.

liaxim:
Constant type fixed

thomasflynn:
tested on both note4 and s6.  what an amazing difference!  good work!

Issue #286, opened 2015-10-21
Textures not loading while loading a 3d model.


bonniemathew:
Hi, 

While working on loading a 3d model, we have found that some of the textures are not loading. It works on models with very few textures but the models with 8 - 12 texture references fails to load. This issue started happening from this commit c108b1af2220eb30b6df0062689c365e17b3caa0 (#253). 

-private static ExecutorService threadPool = Executors.newCachedThreadPool();
+private static ExecutorService threadPool = new GrowBeforeQueueThreadPoolExecutor("gvrf");

The usage of GrowBeforeQueueThreadPoolExecutor() seems to be the culprit. If I use Executors.newCachedThreadPool(), all textures are loading fine. 

Thanks, 
Bonnie M

liaxim:
Thank you for the detailed report. We will investigate. Do you happen to have a problematic model you could share?

bonniemathew:
I have sent the problematic model via email... (to m.marinov@samsung.com).

liaxim:
The pool gets exhausted. There are so many Future.get()'s waiting that they use up all the threads and prevent the actual texture loading from proceeding.

liaxim:
Please let me know whether pull request #289 works for you. Thanks.

bonniemathew:
It's loading all the textures. Verified with two other models. Thanks for the change..

PR #289, opened 2015-10-23
Fix for issue #286, "Textures not loading while loading a 3d model"

liaxim:
Reduces the number of threads needed while loading a model. setMainTexture
spawns a thread that waits for the texture to load. But if you do this
for enough textures at the right time, all the threads can be used up and
there will be no thread available to perform the actual loading..

The problem is still there and if someone who is overeager to use
Future<Texture>s can experience it. I will look at it again later and
might have to introduce one more pool.


liaxim:
This is an effective, though partial, fix. Any review comments?

PR #291, opened 2015-10-28
Fix to issue 227 - External Views framerate

ragner:
Fix to issue 227 - External Views framerate

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/scene_objects/GVRViewSceneObject.java
+++ b/GVRf/Framework/src/org/gearvrf/scene_objects/GVRViewSceneObject.java
@@ -1,0 +1,113 @@
+/* Copyright 2015 Samsung Electronics Co., LTD
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gearvrf.scene_objects;
+
+import org.gearvrf.GVRContext;
+import org.gearvrf.GVRDrawFrameListener;
+import org.gearvrf.GVRExternalTexture;
+import org.gearvrf.GVRMaterial;
+import org.gearvrf.GVRMaterial.GVRShaderType;
+import org.gearvrf.GVRMesh;
+import org.gearvrf.GVRSceneObject;
+import org.gearvrf.GVRTexture;
+import org.gearvrf.scene_objects.view.GVRView;
+
+import android.graphics.Canvas;
+import android.graphics.SurfaceTexture;
+import android.view.Surface;
+
+/**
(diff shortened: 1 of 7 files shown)


GVRf/Framework/src/org/gearvrf/scene_objects/GVRWebViewSceneObject.java
```diff
@@ -1,112 +0,0 @@
...
-import org.gearvrf.GVRTexture;
-import org.gearvrf.GVRMaterial.GVRShaderType;
-
-import android.graphics.Canvas;
-import android.graphics.SurfaceTexture;
-import android.util.Log;
-import android.view.Surface;
-import android.webkit.WebView;
-
-/**
```
thomasflynn: rather than removing the class outright, please deprecate it first.  also, note in the javadoc to reference the GVRViewSceneObject and the example in the scene-objects sample.  

it is generally more polite to deprecate an api/class before removing it.  we'll remove it in a subsequent release.

GVRf/Framework/src/org/gearvrf/scene_objects/GVRViewSceneObject.java
```diff
@@ -0,0 +1,113 @@
...
+
+        // TODO: Shader type maybe defined by some GVRView.getShaderType()
+        // according to view type
+        GVRMaterial material = new GVRMaterial(gvrContext, GVRShaderType.OES.ID);
+        material.setMainTexture(texture);
+        getRenderData().setMaterial(material);
+
+        mSurfaceTexture = new SurfaceTexture(texture.getId());
+        mSurface = new Surface(mSurfaceTexture);
+        mSurfaceTexture.setDefaultBufferSize(gvrView.getView().getWidth(),
```
thomasflynn: is this not tabbed right?  (4 spaces)

GVRf/Framework/src/org/gearvrf/scene_objects/view/GVRWebView.java
```diff
@@ -0,0 +1,65 @@
...
+        context.registerView(this);
+    }
+
+    @Override
+    public void draw(Canvas canvas) {
+        if (mSceneObject == null)
+            return;
+
+        // Canvas attached to GVRViewSceneObject to draw on
+        Canvas glAttachedCanvas = mSceneObject.lockCanvas();
```
thomasflynn: please don't prefix variables or functions with 'gl', it makes it look like it is some sort of official OpenGL variable / call.  in this case, simply 'attachedCanvas' or 'attachedGLCanvas' would be fine.

thomasflynn:
ok, on my first pass, i had a few nit-picky things plus i'd rather deprecate the WebViewSceneObject rather than remove it outright.  We'll remove it later.

GVRf/Framework/src/org/gearvrf/scene_objects/GVRViewSceneObject.java
ragner: Hi @thomasflynn,
Maybe I haven't understood what you was meaning ...
But it is just breaking the previous line, because of that it has 8 spaces.

ragner:
Thank you @thomasflynn,
I have applied  your comments.

GVRf/Framework/src/org/gearvrf/scene_objects/GVRViewSceneObject.java
thomasflynn: ok, maybe it's just the way it is showing up in github that makes it looks weird.

thomasflynn:
ok, looks good.  thanks.

PR #295, opened 2015-11-03
A simple example to use the new android camera2 api and renderscript with GearVRf

cydream:
Google introduced whole new camera api from API 21. (android.hardware.camera2.*)
This sample just create a quad and pass camera images through the surface using camera2 API.
By adding multiple surfaces, it is possible to apply some visual effects in real-time with  powerful renderscript features.


thomasflynn:
Cool!  Thanks for the pull-request!  Looks very interesting.  We'll be reviewing it over the next day or two and giving feedback.

GVRf/Sample/gvrcamera2renderscript/src/org/gearvrf/sample/gvrcamera2renderscript/Camera2RenderscriptManager.java
```diff
@@ -0,0 +1,102 @@
...
+		mGVRContext = gvrContext;
+		GVRScene mainScene = gvrContext.getNextMainScene(new Runnable() {
+			
+			@Override
+			public void run() 
+			{
+				try
+				{
+					mCameraHelper = new Camera2Helper(mActivity, 0);
+					Size previewSize = mCameraHelper.setPreferredSize(1920, 1080);
```
liaxim: The frame-rate gets noticeably better if 1/4 of width&height is used instead. What do you think?

liaxim:
Sample works nicely albeit somewhat choppy on my Note4. Let's see what @cydream thinks about my question. The code looks nice too.

GVRf/Sample/gvrcamera2renderscript/src/org/gearvrf/sample/gvrcamera2renderscript/Camera2RenderscriptManager.java
cydream: Hi liaxim, yes, the frame-rate depends on the preview size. Actually. I'm using 1280x720 in my real developments for the performance issue. And also that's why I made setPreferredSize method. Someone who would use this code might need to find the most proper sizes in their use cases.

liaxim:
Ok then. It looks good and I have no further comments. Thanks.

thomasflynn:
on the first pass, code looks fine.  tried running in on my S6 and didn't see anything.  charging up a Note4, will try it there.  did i perhaps miss something?

liaxim:
Seeing the same on my S6. This is probably relevant:

```
11-05 17:17:52.711  3111  3549 D CameraService: CameraService::connectDevice E (pid 12748 "org.gearvrf.sample.gvrcamera2renderscript", id 0)
11-05 17:17:52.721  3111  3549 I Camera2ClientBase: Camera 0: Opened. Client: org.gearvrf.sample.gvrcamera2renderscript (PID 12748, UID 10241)
11-05 17:17:53.021 12748 12901 W CameraDevice-JV-0: Stream configuration failed
11-05 17:17:53.021 12748 12901 E CameraCaptureSession: Session 0: Failed to create capture session; configuration failed
```

cydream:
I'm also not tested on S6, i'll check it as soon as possible and let you know the result.

cydream:
I have looked at the problem and knew that the camera preview image format is differ from Note4 and S6. In S6, the code should use YUV_420_888 format not RGBA_8888. I think I should use a if-else clause for the model name. Do you have any suggestions? It will be best if the gearvrf already has a such function.

liaxim:
If there is no way to dynamically discover these things you can always resort to Build.MODEL.contains("SM-G920") || (Build.MODEL.contains("SM-G925"). What GVRF function do you have in mind?

cydream:
I committed a patch for S6 devices (SM-G920, SM-G925, and SM-G928). Please check it again.

liaxim:
Ok, if you don't mind we will merge this and apply the minor fix ourselves.

cydream:
Thanks, liaxim. please do so.

PR #296, opened 2015-11-04
Fix bug in removeChildObject().

Vittalbabu:
Fix bug in removeChildObject(). If child is removed, the child's
transform has to be invalidated. Otherwise it will still hold the
complete transformation from root.

Invalidaing the transform forces a re-evaluation in the next
getModelMatrix() call.


thomasflynn:
looks fine to me.  @liaxim ?

liaxim:
Looks good to me too.

PR #297, opened 2015-11-04
Input Handling in GVRf

rahul27:
This patch lays out the groundwork for input handling
in GVRf.

The introduced GVRInputManager allows user defined InputEvents
to be dispatched to through the scene graph. This enables
the GVRSceneObjects to implement user interaction using the
SensorEvents generated from GVRBaseSensor implementations.

Also included is a sample that demonstrates using the
GVRInputManager to implement user interaction.

Note: Make sure you import both the input sample and the STL library project to run the sample application.

rahuldevan@gmail.com


rahul27:
Hi @thomasflynn, I've fixed most issues in this patch.

I've also addressed the comments you had in the last pull request.

GVRf/IO/STL/src/org/gearvrf/stl/STLConstants.java
```diff
@@ -0,0 +1,26 @@
+/* Copyright 2015 Samsung Electronics Co., LTD
```
thomasflynn: remind me again what STL stands for?  as a c++ programmer (prior to working at Samsung), I see STL and I think standard template library.

liaxim: "Standard Translation Library".. I find it confusing too.

GVRf/Framework/src/org/gearvrf/GVRContext.java
```diff
@@ -2133,6 +2133,33 @@ public GVRScene getNextMainScene() {
...
+     * 
+     * Alternatively the {@link GVRScene} can be changed using
+     * {@link GVRInputManager#setScene(GVRScene)}
+     * 
+     * @param scene
+     *            The {@link GVRScene} used to process the input.
+     * @return A {@link GVRInputManager} to use to dispatch {@link InputEvent}s
+     *         to the specified {@link GVRScene} object.
+     */
+    public abstract GVRInputManager setupInputManagerForScene(GVRScene scene);
```
liaxim: Having one way to set the scene for the GVRInputManager is more straightforward. Why have two?

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
```diff
@@ -698,6 +699,61 @@ public GVRSceneObject getChildByIndex(int index) {
...
+     * Note that the attached sensor affects this node and its descendants.
+     * 
+     * @param sensor
+     *            The {@link GVRBaseSensor} to be added to the object node.
+     *            
+     * @return <code>true</code> if the sensor was added successfully;
+     *         <code>false</code> otherwise, this could possibly be due to the
+     *         {@link GVRInputManager} not being set.
+     */
+    public boolean addSensor(GVRBaseSensor sensor) {
```
liaxim: Ok, these additions do not seem to me to belong "organically" to a GVRSceneObject. GVRSceneObject has numerous APIs the way it is. This "augmentation" of GVRSceneObject instances can easily be encapsulated by the GVRInputManager or SensorManager. Either one of these could associate a GVRSceneObject with a Sensor so later the SensorManager can lookup a Sensor for a SceneObject (referring to SensorManager.recurseSceneObject which uses GVRSceneObject.getSensor). Let's try to avoid making knowledge of this input/sensor concept too wide.

GVRf/IO/InputSample/src/org/gearvrf/io/Cube.java
```diff
@@ -0,0 +1,70 @@
...
+
+import org.gearvrf.GVRAndroidResource;
+import org.gearvrf.GVRContext;
+import org.gearvrf.GVRMaterial;
+import org.gearvrf.GVRMesh;
+import org.gearvrf.GVRRenderData;
+import org.gearvrf.GVRSceneObject;
+import org.gearvrf.GVRTexture;
+
+public class Cube extends GVRSceneObject {
```
liaxim: Fyi - there is GVRCubeSceneObject. Using it may help narrow down the sample to its core goal.

GVRf/IO/STL/src/org/gearvrf/stl/STLController.java
```diff
@@ -0,0 +1,401 @@
...
+import android.view.InputDevice;
+import android.view.KeyEvent;
+import android.view.MotionEvent;
+
+/**
+ * Use this class to translate MotionEvents and KeyEvents generated by the
+ * controller to values that can be used to create InputEvents for the GVRf
+ * GVRInputManager.
+ */
+public class STLController implements InputDeviceListener {
```
liaxim: Somewhat offtopic but it seems to me like this could be split into two parts - one, for lack of better wording, that manages the flow/processing and adapters for various devices (processJoystickInput and the device-specific if-s). Then it could be melded into the GVRInputManager.

liaxim:
After my first more thorough pass it feels like this could be a "gvrf extension" library. Call it gvrf-sensor-io or whatever and merge most of the code that resides in GVRF-proper and from "STL" library in there.

The benefit of packaging it like this is that it will self-contained and optional. The core library is applicable to pretty much all apps but this brings in somewhat more unique capabilities that not all apps will need. This will keep things simpler and more modular - always good in my book.

I will try to make another pass tomorrow after I get my hands on a gamepad and see it in action. Let me know what do you think. Thanks.

GVRf/Framework/src/org/gearvrf/GVRContext.java
rahul27: One is a public method to set up the Input Manager for a given scene and the other is a package protected method for the framework to get the Input Manager if one has been set.

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
rahul27: I've tried to stay in line with the User Interactivity concepts defined by X3D:

http://www.web3d.org/documents/specifications/19775-1/V3.2/Part01/components/pointingsensor.html

Since GVRf does not define an explicit grouping node, we make use of the Hybrid SceneObject node to define groups and hierarchies for User Interactivity.

GVRf/IO/InputSample/src/org/gearvrf/io/Cube.java
rahul27: Sure, I could have a look at it.

GVRf/IO/STL/src/org/gearvrf/stl/STLController.java
rahul27: I wanted to keep the device specific logic out of the GVRInputManager. The InputManager needs to be generic enough and not turn into an InputDeviceManager.

The plan is to grow the STL instead of the InputManager to support other IO devices, thereby minimizing the changes to the GVRf Framework.

rahul27:
@liaxim Thanks for your comments.

Let me wait for your final feedback before discussing your overall ideas about modifying this to be an extension to GVRf.

GVRf/IO/STL/src/org/gearvrf/stl/STLConstants.java
rahul27: As far as naming goes I am fine changing STL to something less ambiguous - along the lines of IOSupportLibrary.

GVRf/Framework/src/org/gearvrf/GVRContext.java
liaxim: I am referring to the fact that you set the scene via setupInputManagerForScene and also via GVRInputManager.setScene. If I have a new GVRScene I can either a) use GVRContext.setupInputManagerForScene (public) to set it, or b) if I already have an instance of GVRInputManager, use GVRInputManager.setScene (public too). You should settle on one and only one way to associate a GVRInputManager with a GVRScene.

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
liaxim: Until a grouping node is added to GVRF, I'd say let's not hijack the GVRSceneObject to be more than what it is supposed to be. To me the lack of a grouping node is one more reason why this should be an extension library.

If a grouping node is added then you can do it the proper way where each sensor is its own node. If grouping node is never added we are not polluting a core geometry class but still have the sensor capability.

GVRf/IO/STL/src/org/gearvrf/stl/STLController.java
liaxim: Looking at the sample's InputScript.onEvent() it is non-trivial and apparently every app that uses sensors needs to be supplying one. It feels something like this belongs to the framework.

So what else STL can be used for other than supporting the InputManager? If it is reusable, keep it out but otherwise if keeping it close to the InputManager means simpler (or ideally no need for) InputScript.onEvent() then why not?

liaxim:
Sample works nicely.

1. I feel stronger about having this as an extension library until after a grouping node is added to GVRF. Now the concept is just forced onto the existing GVRSceneObject and exposed very widely.

2. It would be nice if something could be done so end-user apps do not have to be aware of the STL library and the translation step (InputScript.onEvent).

You should wait for feedback from @thomasflynn before committing (or not) to changes.

rahul27:
This patch has been refactored quite a bit. Will push a new patch.

PR #298, opened 2015-11-05
Obey draw_mode and introduce set/getIndices (fix #293)

### LINKED ISSUE #293: GVRRenderData.setDrawMode is being ignored
If someone wants to render with, say,  GL_LINES, he or she can't currently do that. The draw_mode is not being used by the renderer.

thomasflynn:
Somehow, we weren't obeying the draw_mode.  That should be fixed now.
Also, deprecated get/setTriangles() in GVRMesh.  Replaced with
get/setIndices() which is a less confusing and more general name.

tom.flynn@samsung.com


liaxim:
Code looks good. Tried and it works as expected.

Issue #294, opened 2015-10-31
update to oculus v1.0.0 sdk


thomasflynn:
need to update to latest oculus sdk.

pdiaz:
+1

thomasflynn:
started.  compiles, but crashes at runtime.

https://github.com/thomasflynn/GearVRf/tree/ovr_100?files=1

liaxim:
@thomasflynn make sure you update VrApi.jar and VrAppFramework.jar under Framework/libs with the copies from 1.0.0.0. Once I did that the crash went away.

The next problem is obtaining updates to Oculus and Oculus Home. It didn't like my copies and I could not get it to update. It has been a hassle for some time and it is getting worse using Oculus with our custom builds..

The error was an exception that we have seen before and which I don't think has been resolved - bug 19350 - no INSTALL_PACKAGES permission.

thomasflynn:
ah, yes, I'm pretty sure I did indeed miss copying the new jars.  i'll try that if I get a chance in between meetings.

i'm doing my testing on my personal S6.  in other words, not on an engineering build which is where we often have problems with updating Oculus Home and other binaries.  yesterday, after installing simple-sample built with v1.0, and inserting into a GearVR, my phone showed the GearVR setup wizard and prompted me to download a bunch of updates.  This seems to match what Oculus describes in their 1.0 documentation.

so, hopefully, if I just update the jars, things will 'just work' :)  I'll try again today if I can squeeze some time in-between my meetings.

thomasflynn:
copied the two jars.  still crashing for me :(.  I can show you tomorrow.

liaxim:
Interesting. I can only get to the update screen on my device so I might have missed it. Send me the logs please.

liaxim:
Seeing it on my Note4. Seemed like SystemUtils.jar was missing. Added - still can't be found. Looks like the whole thing is meant to be build with Gradle (only). I can't even get the Oculus samples to work if built with Eclipse. Looking into it.

liaxim:
Ok, after few clean-rebuild cycles it is working. The only additional step is to copy ./VrAppSupport/SystemUtils/Libs/Android/SystemUtils.jar to Framework/libs. It is a good idea to manually remove all the ./bin/ directories and then clean-rebuild in Eclipse.

The Oculus samples had a case-sensitivity problem (VrGui vs VrGUI).

thomasflynn:
merged.

PR #299, opened 2015-11-05
Upgrade to Oculus SDK 1.0.0.0 (#294)

thomasflynn:
Need to make sure to copy VrApi.jar, VrAppFramework.jar and SystemUtils.jar into Framework/libs/ before building.  

Also, we'll follow Oculus' convention for the naming of their sdk directory.  Previously they were using 'ovr_mobile_sdk'.  now they are using 'ovr_sdk_mobile'.

We'll need to update the build instructions before merging this.

tom.flynn@samsung.com


liaxim:
Looks good to me.

PR #300, opened 2015-11-07
adding case for negative scale

roshanch:
Modifying code for extraction of scale matrix, adding case for negative scale

@thomasflynn  @liaxim Please review the changes.

### DIFF
--- a/GVRf/Framework/jni/objects/components/transform.cpp
+++ b/GVRf/Framework/jni/objects/components/transform.cpp
@@ -22,7 +22,7 @@
 #include "glm/gtc/type_ptr.hpp"
 
 #include "objects/scene_object.h"
-
+#include <math.h>
 namespace gvr {
 Transform::Transform() :
         Component(), position_(glm::vec3(0.0f, 0.0f, 0.0f)), rotation_(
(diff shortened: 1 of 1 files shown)


GVRf/Framework/jni/objects/components/transform.cpp
```diff
@@ -76,35 +76,44 @@ glm::mat4 Transform::getModelMatrix() {
...
-            matrix[0][0] * matrix[0][1] * matrix[0][2] * matrix[0][3] < 0 ?
-                    -1 : 1;
-    float ys =
-            matrix[1][0] * matrix[1][1] * matrix[1][2] * matrix[1][3] < 0 ?
-                    -1 : 1;
-    float zs =
-            matrix[2][0] * matrix[2][1] * matrix[2][2] * matrix[2][3] < 0 ?
-                    -1 : 1;
+
+	glm::vec3 new_position(matrix[3][0], matrix[3][1], matrix[3][2]);
```
liaxim: There is a tab in here. Please make sure for indentation you use spaces only.

Vittalbabu:
Looks good.

thomasflynn:
ok, can you fix the tab?  consistent spacing makes diffs easier.

Vittalbabu:
Math for this change is from 
http://stackoverflow.com/questions/33538035/extracting-scale-matrix-from-modelview-matrix

danke-sra:
Take the scaling matrix S = diag([-1, 1, 1, 1]) as an example, your algorithm will output S' = diag([-1, -1, -1, 1]). They are not equivalent. 

For example, v = [1, 2, 3, 1], S * v = [-1, 2, 3, 1] but S' * v = [-1, -2, -3, 1]. The y and z values are flipped.

The reason is stated in http://stackoverflow.com/questions/33538035/extracting-scale-matrix-from-modelview-matrix, the signs of xs, ys and zs are the same. When they are negative, either "one" or "three" axis are flipped. But you don't know which case is true. So this algorithm does not work either.

Please correct me if I am wrong...

roshanch:
@danke-sra : Yes, you are right. The fix is not complete and won't work if one of axis is inverted as we can't distinguish between rotation and scaling. However, this fix handles case for negative scale for all the axis, I am working on the algorithm for the missing case.

PR #307, opened 2015-11-13
Several modifications to clean-up / enhance the sample:

gaurav-mcl:
1. Fixed an initialization issue with GVRCylinderScneObject.
2. Removed the code for rotating scene objects as it seems superfluous
to what this sample intends to demonstrate.
3. Attached the GVRCameraSceneObject to camera rig instead of main scene
so that it can move with user's head motion.
4. Enabled VRMode for the camera so that 60 fps camera preview can be used.

GearVRf-DCO-1.0-Signed-off-by: Gaurav Srivastava 
gaurav001.s@samsung.com

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/scene_objects/GVRCameraSceneObject.java
+++ b/GVRf/Framework/src/org/gearvrf/scene_objects/GVRCameraSceneObject.java
@@ -26,7 +26,6 @@ import org.gearvrf.GVRSceneObject;
 import org.gearvrf.GVRTexture;
 import org.gearvrf.GVRMaterial.GVRShaderType;
 
-import android.app.Activity;
 import android.graphics.SurfaceTexture;
 import android.hardware.Camera;
 import android.hardware.Camera.Parameters;
--- a/GVRf/Framework/src/org/gearvrf/scene_objects/GVRCameraSceneObject.java
+++ b/GVRf/Framework/src/org/gearvrf/scene_objects/GVRCameraSceneObject.java
@@ -43,7 +43,7 @@ public class GVRCameraSceneObject extends GVRSceneObject implements
     private final SurfaceTexture mSurfaceTexture;
     private boolean mPaused = false;
     private Camera camera;
-    private Activity activity;
+    private GVRContext gvrContext;
     private boolean cameraSetUpStatus;
 
     /**
@@ -52,8 +52,6 @@ public class GVRCameraSceneObject extends GVRSceneObject implements
      * 
      * @param gvrContext
      *            current {@link GVRContext}
-     * @param activity
-     *            the parent {@link Activity} that created this scene object.
      * @param mesh
      *            an arbitrarily complex {@link GVRMesh} object - see
      *            {@link GVRContext#loadMesh(org.gearvrf.GVRAndroidResource)}
@@ -64,8 +62,8 @@ public class GVRCameraSceneObject extends GVRSceneObject implements
      *            should be sure to call it before you call
      *            {@link Camera#startPreview()}.
      */
-    public GVRCameraSceneObject(GVRContext gvrContext, Activity activity,
-            GVRMesh mesh, Camera camera) {
+    public GVRCameraSceneObject(GVRContext gvrContext, GVRMesh mesh,
+            Camera camera) {
         super(gvrContext, mesh);
         gvrContext.registerDrawFrameListener(this);
         GVRTexture texture = new GVRExternalTexture(gvrContext);
@@ -73,7 +71,7 @@ public class GVRCameraSceneObject extends GVRSceneObject implements
         material.setMainTexture(texture);
         getRenderData().setMaterial(material);
 
-        this.activity = activity;
+        this.gvrContext = gvrContext;
         this.camera = camera;
         mSurfaceTexture = new SurfaceTexture(texture.getId());
         try {
@@ -89,8 +87,6 @@ public class GVRCameraSceneObject extends GVRSceneObject implements
      * 
      * @param gvrContext
      *            current {@link GVRContext}
-     * @param activity
-     *            the parent {@link Activity} that created this scene object.
      * @param width
      *            the scene rectangle's width
      * @param height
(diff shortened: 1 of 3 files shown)


GVRf/Sample/scene-objects/src/org/gearvrf/sample/sceneobjects/SampleViewManager.java
```diff
@@ -124,16 +124,16 @@ public void onInit(GVRContext gvrContext) {
...
-        scene.addSceneObject(quadObject);
-        scene.addSceneObject(cubeObject);
-        scene.addSceneObject(sphereObject);
-        scene.addSceneObject(cylinderObject);
-        scene.addSceneObject(coneObject);
-        scene.addSceneObject(webViewObject);
-        scene.addSceneObject(cameraObject);
-        scene.addSceneObject(videoObject);
-        scene.addSceneObject(textViewSceneObject);
+		// add the scene objects to the scene graph.
```
thomasflynn: spacing looks off here.  do you have tabs rather than spaces?

GVRf/Sample/scene-objects/src/org/gearvrf/sample/sceneobjects/SceneObjectActivity.java
```diff
@@ -89,13 +114,47 @@ private void createCameraView() {
...
             if (camera != null) {
                 Parameters params = camera.getParameters();
-                params.setFocusMode(Parameters.FOCUS_MODE_CONTINUOUS_VIDEO);
-                camera.setParameters(params);
-                camera.startPreview();
-            }
-        } catch (Exception exception) {
-            android.util.Log.d(TAG, "Camera not available or is in use");
-        }
+
```
thomasflynn: can we add a method to GVRCameraSceneObject to set all these parameters to make it easier for developers to set this up?

chenchao1407:
Probably in this pull request, you are using tabs-only spacing policy instead of spaces only in formatter. I think you need to change it in Eclipse->Preference.(Sorry I don't have Eclipse in hands so that I can't provide detailed steps for you but I think you can check the settings in gearvrf.org.)

GVRf/Sample/scene-objects/src/org/gearvrf/sample/sceneobjects/SceneObjectActivity.java
```diff
@@ -76,6 +73,34 @@ private boolean checkCameraHardware(Context context) {
...
+        @Override
+        /**
+         * The byte data comes from the android camera in the yuv format. so we
+         * need to convert it to rgba format.
+         */
+        public void onPreviewFrame(byte[] data, Camera camera) {
+            numFrames++;
+            if (numFrames == 100) {
+                Parameters params = SceneObjectActivity.this.camera
```
thomasflynn: is setting the parameters at this point (frame 100) necessary?  if so, why?  seems odd.

gaurav-mcl: that was part of my attempt to prevent the camera from auto-focusing. but apparently it seems that it is not needed because even without this code, the camera doesn't auto-focus in 60 fps mode. So I will remove this code.

GVRf/Framework/src/org/gearvrf/scene_objects/GVRCameraSceneObject.java
```diff
@@ -54,8 +64,8 @@
      *            should be sure to call it before you call
      *            {@link Camera#startPreview()}.
      */
-    public GVRCameraSceneObject(GVRContext gvrContext, GVRMesh mesh,
-            Camera camera) {
+    public GVRCameraSceneObject(GVRContext gvrContext, Activity activity,
```
thomasflynn: sorry if i missed this before:  you can retrieve the activity from the GVRContext.  gvrContext.getActivity();  So there's no need to change the arguments.

GVRf/Framework/src/org/gearvrf/scene_objects/GVRCameraSceneObject.java
```diff
@@ -87,9 +101,9 @@ public GVRCameraSceneObject(GVRContext gvrContext, GVRMesh mesh,
      *            should be sure to call it before you call
      *            {@link Camera#startPreview()}.
      */
-    public GVRCameraSceneObject(GVRContext gvrContext, float width,
-            float height, Camera camera) {
-        this(gvrContext, gvrContext.createQuad(width, height), camera);
+    public GVRCameraSceneObject(GVRContext gvrContext, Activity activity,
```
thomasflynn: ditto here with the activity

liaxim:
Looks good to me. Haven't tried it though.

thomasflynn:
sample doesn't compile.  still has the mActivity argument in the constructor for GVRCameraSceneObject

gaurav-mcl:
@thomasflynn, not sure if you have the updated code. I don't see any Activity argument in the constructor and it compiles fine for me. The only redundant thing is 'import android.app.Activity' which I will remove in the next commit but it should not break compilation.

GVRf/Sample/scene-objects/src/org/gearvrf/sample/sceneobjects/SampleViewManager.java
```diff
@@ -84,12 +84,13 @@ public void onInit(GVRContext gvrContext) {
...
                 gvrContext, true, material);
         GVRCylinderSceneObject cylinderObject = new GVRCylinderSceneObject(
-                gvrContext, 0.5f, 0.5f, 1.0f, 10, 36, true, futureTextureList, 2, 4);
+                gvrContext, true, material);
         GVRConeSceneObject coneObject = new GVRConeSceneObject(gvrContext,
                 true, material);
         GVRViewSceneObject webViewObject = createWebViewObject(gvrContext);
         GVRCameraSceneObject cameraObject = new GVRCameraSceneObject(
-                gvrContext, 8.0f, 4.0f, mActivity.getCamera());
+                gvrContext, mActivity, 3.6f, 2.0f, mActivity.getCamera());
```
liaxim: Here is the extra mActivity..

gaurav-mcl: Yeah, somehow my eclipse project compiled with that bug. I did 'project clean' and rebuild and then I got the compile error too.

PR #308, opened 2015-11-14
Adding fix for transformation issue and support for importing texture parameters from assimp for FBX

roshanch:
Please review the changes.


GVRf/Framework/src/org/gearvrf/GVRContext.java
```diff
@@ -637,7 +668,22 @@ public AiNode wrapSceneNode(Object parent, Object matrix,
...
+		 if(wrapmodes!=null) 	
+         {
+        	 GVRContext context=assimpImporter.getGVRContext();
+        	 context.DEFAULT_TEXTURE_PARAMETERS.setWrapSType(WrapmodeMap.get(wrapmodes[GVRTextureWrapAxis.WrapS.getWrapAxis()]));
+        	 context.DEFAULT_TEXTURE_PARAMETERS.setWrapTType(WrapmodeMap.get(wrapmodes[GVRTextureWrapAxis.WrapT.getWrapAxis()]));
+         }
+        
+        
         return wholeSceneObject;
     }
```
chenchao1407: I think for the codes in this file you need to do Ctrl+Shift+F to do the formatting.

GVRf/Framework/jni/objects/textures/compressed_texture_jni.cpp
```diff
@@ -30,7 +30,7 @@ Java_org_gearvrf_asynchronous_NativeCompressedTexture_normalConstructor(JNIEnv *
 
 JNIEXPORT jlong JNICALL
 Java_org_gearvrf_asynchronous_NativeCompressedTexture_mipmappedConstructor(JNIEnv * env,
-        jobject obj, jint target);
+        jobject obj, jint target, jintArray jtexture_parameters);
```
danke-sra: By design, we didn't force to pass texture_parameters in the texture constructors (Texture, CompressedTexture, FloatTexture, etc). But the wrapping parameters can be passed using the Texture::updateTextureParameters function. They include wrap_s_type and wrap_t_type already. So please try to use the existing API.

GVRf/Framework/jni/contrib/assimp/include/assimp/scene.h
```diff
@@ -264,7 +267,13 @@ struct aiNode
  */
 #define AI_SCENE_FLAGS_TERRAIN 0x10
 
+enum aiTextureWrapAxis
```
danke-sra: Wrap parameters exist in include/assimp/material.h. Please check if those values are already parsed correctly from FBX file. If so, you can pass those parameters to GVRTexture. Generally, it is probably problematic to set wrapping parameters globally (for the whole scene).

danke-sra:
It's good to try to support fix texture wrapping in GVRf. However, please try to keep assimp header files unchanged if possible. I believe wrapping parameters are already supported by assimp. There is also a function GVRTexture.updateTextureParameters (or in native code, Texture::updateTextureParameters) to set wrapping. It would be very nice to get them to work. Thanks.

GVRf/Framework/jni/contrib/assimp/include/assimp/scene.h
```diff
@@ -118,6 +118,9 @@ struct aiNode
 	  * format. Importers that don't document any metadata don't write any. 
 	  */
 	C_STRUCT aiMetadata* mMetaData;
+	
+	/* Geometric Transformations of the node */
+	aiMatrix4x4 mGeomTransformation;
```
danke-sra: Is this referenced?

roshanch: I have updated assimp library for the transformation fix and it is a part of that. It is not referenced in GVRf but in assimp.

GVRf/Framework/src/org/gearvrf/GVRContext.java
roshanch: Thanks for pointing it out. Will do.

GVRf/Framework/src/org/gearvrf/GVRContext.java
```diff
@@ -130,6 +133,34 @@
     public final GVRTextureParameters DEFAULT_TEXTURE_PARAMETERS = new GVRTextureParameters(
             this);
 
+    /* *
+     * Store the mapping of integers with wrap modes   
+     */
+    public HashMap<Integer,TextureWrapType>WrapmodeMap=
```
liaxim: Deriving an anonymous class just for this? What's wrong with having a static array with 3 TextureWrapType elements and looking up by index? It will be more readable also.

GVRf/Framework/jni/objects/textures/compressed_texture_jni.cpp
liaxim: I for one don't see anything wrong with passing the parameters to the ctor. The GLTexture ctor is applying default values for these parameters so it seems natural to me to be able to pass values of your choice also.

PR #309, opened 2015-11-17
Integrate full Jassimp and upgrade assimp headers

danke-sra:
This port of Jassimp supports full assimp functionality. In addition to what we had, it can now access lights, cameras, bones and animations from Java.

This patch set also provides a way to convert AiScene to GVRf scene graph in pure Java. This makes GVRf easier to maintain for assimp/jassimp changes. User-defined hooks can be added to instantiate classes during the conversion process.

Optimized assimp loader performance from O(N) to O(1).


lokeshguddu:
Hi @danke-sra ,

All the changes looks good and was much needed improvement. I was working on something which required loading models from SD card and I see that you have taken care of those cases as well. The complexity improvement is really great. :+1:

thomasflynn:
@vittal and @roshanch  still waiting for your reviews on this.

Vittalbabu:
Changes look good. I haven’t run with this version of assimp locally – might get to that in a day.

Vittalbabu:
@danke Will we be updating the importer to use the new jassimp functions?

danke-sra:
@Vittalbabu Yes, we will be having one jassimp as you suggested. But maybe we can change it a bit later after the new one has been adequately tested.

PR #312, opened 2015-11-18
Trying to improve the head-tracking latency a little bit

liaxim:
Based on the Oculus samples; the head model is actually applied only if the position is not being tracked (there is check in vrapi_ApplyHeadModel).


thomasflynn:
looks good.

PR #314, opened 2015-11-20
Add Texture Capturer for reading back rendered textures

danke-sra:
- A GVRTextureCapturer can be added to the RenderData for capturing
  the rendered texture
- Currently, it is applied to the external renderer texture
- Added utilities to convert generate Bitmaps from captured data
  and to save it as PNG file

Performance of the capture is high, allowing 60 FPS capturing
without affecting rendering performance. It may be further improved
by using multiple pixel buffers. This feature can be used for
capturing screen at high FPS.

### DIFF
--- a/GVRf/Framework/jni/engine/renderer/renderer.cpp
+++ b/GVRf/Framework/jni/engine/renderer/renderer.cpp
@@ -581,6 +581,7 @@ void Renderer::renderRenderData(RenderData* render_data,
                             break;
                         case Material::ShaderType::EXTERNAL_RENDERER_SHADER:
                             shader_manager->getExternalRendererShader()->render(
+                                    mv_matrix, glm::inverseTranspose(mv_matrix),
                                     mvp_matrix, render_data);
                             break;
                         case Material::ShaderType::ASSIMP_SHADER:
--- a/GVRf/Framework/src/org/gearvrf/GVRRenderTexture.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRRenderTexture.java
@@ -15,6 +15,8 @@
 
 package org.gearvrf;
 
+import java.nio.IntBuffer;
+
 /** Frame Buffer object. */
 public class GVRRenderTexture extends GVRTexture {
     /**
@@ -37,7 +39,7 @@ public class GVRRenderTexture extends GVRTexture {
     /**
      * Constructs a GVRRenderTexture for a frame buffer of the specified size,
      * with MSAA enabled at the specified sample count.
-     * 
+     *
      * @param gvrContext
      *            Current gvrContext
      * @param width
@@ -51,7 +53,6 @@ public class GVRRenderTexture extends GVRTexture {
             int sampleCount) {
         super(gvrContext, NativeRenderTexture.ctorMSAA(width, height,
                 sampleCount));
-
         mWidth = width;
         mHeight = height;
     }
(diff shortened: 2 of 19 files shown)


GVRf/Framework/src/org/gearvrf/GVRRenderTexture.java
```diff
@@ -52,6 +52,31 @@ public GVRRenderTexture(GVRContext gvrContext, int width, int height,
         super(gvrContext, NativeRenderTexture.ctorMSAA(width, height,
                 sampleCount));
 
+    /**
```
rahul27: Is this here by mistake?

GVRf/Framework/jni/engine/renderer/renderer.cpp
```diff
@@ -581,6 +581,7 @@ void Renderer::renderRenderData(RenderData* render_data,
                             break;
                         case Material::ShaderType::EXTERNAL_RENDERER_SHADER:
                             shader_manager->getExternalRendererShader()->render(
+                                    mv_matrix, glm::inverseTranspose(mv_matrix),
```
rahul27: Any benefits of running an inverse as an argument. Can't this go inside the render function?

You could save one inverse operation if you place this call after the null check for the material in render().

GVRf/Framework/src/org/gearvrf/GVRRenderTexture.java
danke-sra: Not really. We are trying to add javadoc as much as possible. Patch formality is flexible in this project.

GVRf/Framework/jni/engine/renderer/renderer.cpp
danke-sra: The parameters just mirror what is used by the texture shader. It's not my intention and one inversion save is not significant given the GHz speed of modern processor. We are bearing with some redundancy in the code as you can see.

GVRf/Framework/src/org/gearvrf/GVRRenderTexture.java
liaxim: @danke-sra The Javadoc is *inside* the constructor. Something happened with the merging. This is how it looks:

```
 37     /**
 38      * Constructs a GVRRenderTexture for a frame buffer of the specified size,
 39      * with MSAA enabled at the specified sample count.
 40      *
 41      * @param gvrContext
 42      *            Current gvrContext
 43      * @param width
 44      *            Width of the frame buffer.
 45      * @param height
 46      *            Height of the frame buffer.
 47      * @param sampleCount
 48      *            MSAA sample count.
 49      */
 50     public GVRRenderTexture(GVRContext gvrContext, int width, int height,
 51             int sampleCount) {
 52         super(gvrContext, NativeRenderTexture.ctorMSAA(width, height,
 53                 sampleCount));
 54 
 55     /**
 56      * Constructs a GVRRenderTexture for a frame buffer of the specified size,
 57      * with MSAA enabled at the specified sample count, and with specified color
 58      * format, depth format, resolution depth and texture parameters.
 59      *
 60      * @param gvrContext
 61      *            Current gvrContext
 62      * @param width
 63      *            Width of the frame buffer.
 64      * @param height
 65      *            Height of the frame buffer.
 66      * @param sampleCount
 67      *            MSAA sample count.
 68      * @param colorFormat
 69      *            GVR color format. See {@linkplain org.gearvrf.utility.
 70      *            VrAppSettings.EyeBufferParams.ColorFormat ColorFormat}.
 71      * @param depthFormat
 72      *            Depth format. See {@linkplain org.gearvrf.utility.
 73      *            VrAppSettings.EyeBufferParams.DepthFormat DepthFormat}.
 74      * @param resolveDepth
 75      *            If true, resolves the depth buffer into a texture.
 76      * @param parameters
 77      *            Texture parameters. See {@link GVRTextureParameters}.
 78      *
 79      */
 80         mWidth = width;
 81         mHeight = height;
 82     }
```

danke-sra: Fixed. Thanks.

thomasflynn:
is this all the latest code for this?

dankex:
Yeah, this is the latest.

Issue #177, opened 2015-07-20
It would be desirable to have Quaternion and Vector classes on GearVRf Java API


phi-lira:
I think it would be a great addition to have a GVRVector class to handle things like cross/dot/forward and a GVRQuaternion class to handle quaternion creation.

mkeblx:
I believe GearVRf utilizes the Apache Commons Mathematics Library (v3.2): http://commons.apache.org/proper/commons-math which has such math objects and methods that work well.

dankex:
Good suggestion. I will try to make a patch to reflect this, and see how it goes.

liaxim:
We have JOML included nowadays. Planning to close this one.

PR #315, opened 2015-11-23
Add Java math library for OpenGL rendering calculations (JOML)

danke-sra:
Add new API GVRTransform.setModelMatrix(Matrix4f) and getModelMatrix4f
to use the JOML matrix.

Add OpenGL-compatible vectors and matrix classes.

Fix for https://github.com/Samsung/GearVRf/issues/177

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/GVRTransform.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRTransform.java
@@ -15,6 +15,8 @@
 
 package org.gearvrf;
 
+import org.joml.Matrix4f;
+
 /**
  * One of the key GVRF classes: Encapsulates a 4x4 matrix that controls how GL
  * draws a mesh.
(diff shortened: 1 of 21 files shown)


liaxim:
Not sure about this dependency just to wrap the matrix.

danke-sra:
This puts a convenient matrix library at GVRf developer's hands. As I tried to dogfood gvrf and write some app, I found it's still needed. I can't do anything with float[16]. This is probably why the original poster opened the issue. One thing to clarify is that it is *NOT* for the transform only or manipulate native-backed data structures. It's just standard tool for developers to avoid divergence.

liaxim:
You can choose any external library you want to use with a float[16]. No need for GVRF to make this choice.

thomasflynn:
there's also the apache commons library we include and Android has the android.opengl.Matrix classes as well.

danke-sra:
OK, if it's not needed, it can be canceled for now. This library is optimized for OpenGL's mat 4x4, 4x3, etc. Apache commons are for more general math purposes.

PR #318, opened 2015-11-25
Add support for skeletal animation

danke-sra:
- Add GVRBone and GVRBoneWeight and correpsonding native impl.
- Add GVRSkeleton to construct skeleton from a rigged model
- Modified assimp shader to animate mesh
- Add pretty print for GVRSceneObject for debugging scene graphs.
- Add a template GVRKeyFrameAnimation to hold animation

- Add a sample app (working in progress, pending animation import)
- Dependency on JOML (refactoring to use Android OpenGL Matrix).


GVRf/Framework/src/org/gearvrf/GVRHybridObject.java
```diff
@@ -310,7 +346,9 @@ private void close() {
                         handler.nativeCleanup(mNativePointer);
                     }
                 }
-                NativeHybridObject.delete(mNativePointer);
+                if (mOwnNative) {
```
liaxim: The whole purpose of the GVRHybridObject is to manage the native peer lifecycle automatically. Now it may or may not perform its very purpose. If a class doesn't need automatic native peer tracking then it shouldn't be deriving from GVRHybridObject.

GVRf/Framework/src/org/gearvrf/GVRMesh.java
```diff
@@ -17,20 +17,26 @@
...
 import org.gearvrf.utility.Exceptions;
+import org.gearvrf.utility.Log;
 
 /**
  * This is one of the key GVRF classes: It holds GL meshes.
  * 
  * A GL mesh is a net of triangles that define an object's surface geometry.
  */
-public class GVRMesh extends GVRHybridObject {
+public class GVRMesh extends GVRHybridObject implements PrettyPrint {
```
liaxim: Not sure why PrettyPrinting needs to be brought into core classes. There could be an external PrettyPrinting class that knows all it needs to know about certain GVRF classes and keep all the formatting in one place. Now we are polluting core classes with public methods and code that doesn't really belong to them.

GVRf/Framework/src/org/gearvrf/GVRMesh.java
```diff
@@ -17,20 +17,26 @@
...
     public GVRMesh(GVRContext gvrContext) {
-        super(gvrContext, NativeMesh.ctor());
+        this(gvrContext, NativeMesh.ctor());
     }
 
     GVRMesh(GVRContext gvrContext, long ptr) {
         super(gvrContext, ptr);
+        setBones(new ArrayList<GVRBone>());
```
liaxim: It seems to me like we need a special kind of a mesh - a mesh with bones, and keep this one generic.

GVRf/Framework/src/org/gearvrf/GVRTransform.java
```diff
@@ -15,6 +15,8 @@
 
 package org.gearvrf;
 
+import org.joml.Matrix4f;
```
liaxim: joml again? Is it used for something else other than these trivial wrappers?

GVRf/Framework/src/org/gearvrf/GVRVertexBoneData.java
```diff
@@ -0,0 +1,33 @@
+package org.gearvrf;
+
+public final class GVRVertexBoneData extends GVRComponent implements PrettyPrint {
+    /**
+     * Constructor.
+     */
+    public GVRVertexBoneData(GVRContext gvrContext, GVRMesh mesh) {
+        super(gvrContext, NativeVertexBoneData.get(mesh.getNative()), false /* not owned */);
```
liaxim: Who owns it then and how is it cleaned up?

GVRf/Framework/src/org/gearvrf/skinning/GVRSkeleton.java
```diff
@@ -0,0 +1,215 @@
...
+                // Debug: blank animation (z-axis rotation)
+                animationTransform = new Matrix4f();
+                animationTransform.rotation((float)(animationTime / 180 * Math.PI),
+                                            new Vector3f(1f, 0, 0));
+            }
+
+            Matrix4f boneTransform = new Matrix4f();
+
+            // Apply animation to bone
+            boneTransform.mul(boneOffsetInverse[nodeId]);
```
liaxim: Ok, I see joml being used for matrix multiplication. Still not good enough of a reason.

liaxim:
I will have to make more passes due to the size of complexity of this one. Main current issues:

- GVRHybridObject use and the introduction of ownNative
- Bones in the base mesh (java and c++ sides)
- Pretty-printing everywhere

danke-sra:
Thank you for pointing out the three additions/modifications/improvements:

- GVRHybridObject use and the introduction of ownNative
- Bones in the base mesh (java and c++ sides)
- Pretty-printing everywhere

Can you please be specific about what the issues are?

1) ownNative makes the 'delete native' logic optional. by design, what's the reason that all native objects should be owned by the hybrid class? if not, such a flag is needed and should've been there.

2) bones are part of mesh in assimp and it works well. i can see the reason behind it.

3) pretty-print is easy debugging. it's optional but quite crucial for developing bug-free application and framework.

GVRf/Framework/src/org/gearvrf/GVRMesh.java
danke-sra: it's intended to be a developer's helper to view internal data structures of so-called core classes. framework developers can use them to understand how importers create those structures and quickly detect issues.

GVRf/Framework/src/org/gearvrf/skinning/GVRSkeleton.java
danke-sra: i tried to get rid of math in java but it is hard. i will write a version using joml -- there is more convincing use-cases of joml when animation is added.

then, i will see if there is any way to replace it. if you know any good math library designed for openGL, please point out. I did good keywords like "java glm" and this is a top choice i can find.

GVRf/Framework/src/org/gearvrf/GVRVertexBoneData.java
danke-sra: this instance of VertexBoneData is a part of the native mesh class -- not a pointer. it does not need to be freed separately.

GVRf/Framework/src/org/gearvrf/skinning/GVRSkeleton.java
liaxim: If there is more than a trivial need for it - no problem. Then we can drop the apache math library as joml has a quaternion. There was a sample using apache but shouldn't be hard to substitute there too.

GVRf/Framework/src/org/gearvrf/GVRMesh.java
liaxim: Not questioning the usefulness at all - it is nice to have it. What if we had a separate pretty-printing class that does all the work, knows the nodes and performs the formatting in one place? Instead of having the interface and the code in all the nodes of interest.

GVRf/Framework/jni/objects/mesh.h
```diff
@@ -259,6 +296,21 @@ class Mesh: public HybridObject {
...
+    void setBones(std::vector<Bone*>&& bones) {
+        vertexBoneData_.setBones(std::move(bones));
+        shouldResetBoneBuffers = true;
+    }
+
+    VertexBoneData &getVertexBoneData() {
+        return vertexBoneData_;
+    }
+
+    void generateBoneArrayBuffers(Material::ShaderType key);
```
Vittalbabu: What is the use case for this? (ie) When will we have mesh (and bones) with different ShaderType keys?

An earlier checkin removed ShaderType key from meshes.

danke-sra: ShaderType key can be removed if it is not in the base code. I don't use it to distinguish meshes generated.

liaxim:
1) GVRHybridObject exists for the sole reason to manage the one-to-one association between Java and C++ objects and to control the deletion of the native objects automatically. This is so to reduce the complexity - you don't want to manually track the native objects from the Java side.. There is a single case in this change where there seems to be a different need and because of that suddenly the GVRHybridObject becomes something else.

2) because someone else does it doesn't mean we can't question it and look for a better way if possible; I see a mesh and a mesh with bones (specialization). Does every mesh have bones? If not, why not question the structure?

GVRf/Framework/jni/objects/mesh.h
```diff
@@ -259,6 +296,21 @@ class Mesh: public HybridObject {
...
+    }
+
+    void setBones(std::vector<Bone*>&& bones) {
+        vertexBoneData_.setBones(std::move(bones));
+        shouldResetBoneBuffers = true;
+    }
+
+    VertexBoneData &getVertexBoneData() {
+        return vertexBoneData_;
+    }
```
Vittalbabu: Should this be a const ref? Do we expect the callers to modify vertexBoneData_?

GVRf/Framework/jni/objects/mesh.h
Vittalbabu: Let's remove the map -- the code only sets the type as Material::ASSIMP_SHADER

danke-sra:
1) The purpose of GVRHybridObject is not changed. But it did not consider the case that the native class contains another native class which does not need to be deleted (yet it still needs to be accessed by Java -- therefore, another hybrid object). If I made that data member a pointer (rather than an object), then GVRHybridObject will work without change. However, I don't think this is a good reason to keep GVRHybridObject deleting all native objects (it is just a helper), when needed, it needs to be changed.

2) For creating a subclass of mesh, that might work, however, many usages of the code would need to check class type rather than just check "hasBones()". the current way does support mesh without bones.

Vittalbabu:
Is the bones data used in the Mesh class to generate VAOs? No. 

What is the main usage for Mesh? It holds data to create VAOs that are used to render the shaders.

Mesh is now a container which holds optional bones. Is there a linkage between bones with the other mesh data like geometry/normal/...? Why would we want to link VAOs and bones in one class? I am not saying this is wrong -- but is this the right way to hold data?

danke-sra:
@Vittalbabu Bone data are held in GVRVertexBoneData, GVRBone and GVRBoneWeight classes. Meshes only hold references to those bone data, which is correct -- We need to specify the association of meshes and bones. I think I agree with your point but I don't see what changes you suggest.

GVRf/Framework/jni/objects/mesh.h
danke-sra: Agreed. Will do if that code is merged.

GVRf/Framework/jni/objects/mesh.h
danke-sra: Will do. I don't think it is not meant to be const. Will just compile and see.

Vittalbabu:
@Danke -- placing Bones data in Mesh class is then not the answer (unless we decide to change the semantics of Mesh class). How about another class that holds Bones + other animation data (if present and needed).

danke-sra:
That was my initial thought but after reviewing assimp's implementation, I think it is actually a tighter design. There are details into that which made it a neat implementation. However, we can still do a refactoring if we really want to keep mesh leaner.

Vittalbabu:
We move up one level and look at RenderData. It holds the Mesh. Why can't it hold the Bones (or a general Animation) class instance?

Another question -- do we duplicate the same Bone information across multiple objects (Scene Objects) or are they likely to be unique? If there is significant duplication, then we can keep the Bones/Animation data at Scene level and RenderData holds an index to this information.

Vittalbabu:
Same comment regarding Material -- do want RenderPass to hold an index to the Material or the actual Material pointer?

danke-sra:
@Vittalbabu The reason assimp adds bones in mesh is each bone instance affects only that mesh -- render data can combine different meshes and materials, but you can't combine different bones with the mesh. bones are truly properties of that mesh. so there is nothing wrong to put it in the mesh in my opinion. A more subtle aspects of that is bones can split (you can see multiple bones with the same name, each associated with different mesh when they are loaded from assimp). This is to simplify skeletal animation and improve rendering performance. 

So that's the reason to treat bones as properties of the mesh (as done in assimp). I can definitely see other reasons to make them separate, and use other ways to associate them. There will be other consequences -- like more difficult error checking and performance cost, etc. So I am open to suggestions but need to get the code fully working before doing the refactoring.

liaxim:
Is it safe to assume that bones will come into play only in the case of meshes that are being loaded from external model files?

danke-sra:
@liaxim Typically, bones come from imported files. But there may be cases where app developers directly specify them. I haven't checked how the latter works. But I think they can use GVRMesh.setBones(List<GVRBone> bones) to do so. Do you mean they shouldn't be allowed to add bones manually?

liaxim:
Seems reasonable to not allow manual setting. My understanding is that bones definition in reality is doable only in modeling tools. Asking because that might simplify the separation of mesh and a mesh with bones, if we were to have that.

danke-sra:
If setBones is not public, even the current framework class cannot set it... any idea?

danke-sra:
Shouldn't it have same accessibility as setTriangles(), and setTexCoords()?

liaxim:
Would anyone want or even be able to come up with bones definitions at runtime?

danke-sra:
I am not sure if this has been done, but can google it. One use-case of bone generated at runtime is to animate a Kirigami (https://en.wikipedia.org/wiki/Kirigami). There are not so many vertices in a Karigami (say just a square with 4 vertices), by adding 1 or 2 bones, you can animate the folding of it into a triangle. It's pretty much like the tex-coords. Most of time they are loaded from a model but for simple geometries, all of the mesh and bones can be generated programmatically.

danke-sra:
(Don't merge this; This is being integrated with patches from SIDIA.)

PR #321, opened 2015-11-29
Minimal360photo

thomasflynn:
Minimal 360 photo viewer. 

Only handles fully spherical equirectangular photos.  landscapes and other panoramas end up getting stretched around the sphere.  would need to modify the texture coordinates based on the metadata of the panorama.  To handle those cases would require a sufficient amount of code that this would no longer be a minimal sample.  Will look to properly handle photospheres and other panoramas and share intents in a different sample.

tom.flynn@samsung.com


thomasflynn:
ready for review

liaxim:
Looks good

PR #325, opened 2015-12-01
With this line I can't import the project in Eclipse

liaxim:


liaxim:
This is about the minimal360photo sample.

PR #326, opened 2015-12-02
feat: Adds libgdx widget plugin

dattanand:
New implementation which provides a way to add 2D interactive widgets in GVRF as a plugin, widgets are libgdx (game engine) scene2d widgets. sample is also provided (widget-model-viewer)


thomasflynn:
Thanks for the pull-request!  It's a pretty large check-in and will take some time to review.  In fact, it's so large, github won't show the diffs for all the files.  it only shows diffs for the first 300 out of 452.  So, let's try to break this down into smaller pull-requests.  from what i can tell so far, there are at least these parts that can possibly be separate commits:

1) unlit fbo shader
2) shared texture
3) plugin interface
4) widget interface
... and other parts that seem to include libgdx and a sample.   i haven't looked at that closely enough to see how to break all that down.

smaller pull-requests will be easier to review (and fit within github's diff system :).

Issue #246, opened 2015-09-08
Can I set typeface font in GVRTextViewSceneObject class?


douglasalipio:
How do I set typeface font in GVRTextViewSceneObject?

```GVRTextViewSceneObject textView = new GVRTextViewSceneObject(gvrContext, activity);
textView.setTypeface();```

Android example:

```TextView textView = new TextView();
Typeface myTypeface = Typeface.createFromAsset(context.getAssets(), "fonts/futura-condensed-normal.ttf");
textView.setTypeFace(myTypeface);```

mkeblx:
Currently you cannot as `GVRTextViewSceneObject` uses a `TextView` internally that isn't fully accessible besides a few helper methods. (See:  https://github.com/Samsung/GearVRf/blob/master/GVRf/Framework/src/org/gearvrf/scene_objects/GVRTextViewSceneObject.java) 

`setTypeFace` would be a useful one to add, or a method that returns the `TextView` so one could call any of it's methods.

chenchao1407:
I think you can do that since at the beginning I created this type of scene object with just basic APIs. You can add whatever is available from TextView.

PR #329, opened 2015-12-03
Deprecate constructors which require GVRActivity

thomasflynn:
Deprecate constructors that require GVRActivity.  The activity can be gotten from the GVRContext.

tom.flynn@samsung.com


liaxim:
@ragner What is your concern?

ragner:
Hi @liaxim,
at current source code the change looks ok for me.

But, in my opinion, we should develop a org.gearvrf.scene_objects.view.GVRTextView 
to improve its rendering performance and fix some issues like #246 
so make GVRTextViewSceneObject deprecated.

liaxim:
I absolutely agree about the issues with GVRTextViewSceneObject. Just yesterday I had problems with it. We will deprecate it as soon as there is a better alternative.

PR #330, opened 2015-12-03
Gvrwidget

dattanand:
New implementation which provides a way to add 2D interactive widgets in GVRF as a plugin, widgets are libgdx (game engine) scene2d widgets. sample is also provided (widget-model-viewer)

As recommended we have broken down the earlier pull request #326 in distinguishable commits.

PR #332, opened 2015-12-04
feat: Adds unlit FBO shader support for libgdx widgets

dattanand:
Adding new unlit FBO shader type for use with libgdx widgets


GVRf/Framework/jni/shaders/material/unlit_fbo_shader.cpp
```diff
@@ -0,0 +1,125 @@
...
+static const char FRAGMENT_SHADER[] =
+        "precision highp float;\n"
+                "uniform sampler2D u_texture;\n"
+                "uniform vec3 u_color;\n"
+                "uniform float u_opacity;\n"
+                "varying vec2 v_tex_coord;\n"
+                "void main()\n"
+                "{\n"
+                "  vec4 color = texture2D(u_texture, v_tex_coord);"
+                "  gl_FragColor = vec4(color.r * u_color.r * u_opacity, color.g * u_color.g * u_opacity, color.b * u_color.b * u_opacity, color.a * u_opacity);\n"
```
Vittalbabu: Why don't you multiple each component of u_color by u_opacity in the C++ code and pass in the updated color as one vec3 uniform? Will save 4 multiplications for every fragment.

GVRf/Framework/jni/shaders/material/unlit_fbo_shader.cpp
```diff
@@ -0,0 +1,125 @@
...
+#include "util/gvr_gl.h"
+
+namespace gvr {
+static const char VERTEX_SHADER[] = "attribute vec4 a_position;\n"
+        "attribute vec4 a_tex_coord;\n"
+        "uniform mat4 u_mvp;\n"
+        "varying vec2 v_tex_coord;\n"
+        "void main() {\n"
+        "  v_tex_coord.x = a_tex_coord.x;\n"
+        "  v_tex_coord.y = 1.0f - a_tex_coord.y;\n"
```
Vittalbabu: Not as important as fragment shader opacity calculation (fewer vertex shader invocations compared for fragment shader invocations) -- but why not pass in the required y tex coord directly? Why do you need to subtract the y coord from 1?

Vittalbabu:
@dattanand Do you have a sample program that uses this shader?

dattanand:
@Vittalbabu 
This shader is same as existing unlit shader only being different in terms of y coordinate flipping. the sample code you can find in #326

liaxim:
@dattanand Then would't you be able to use the existing unlit shader but just specify different texture coordinates? @thomasflynn brought up this point.

dattanand:
probably yeah but this would be cleaner and texture coordinates can be specified in same way in application code for different kinds of scene objects meaning that much less to worry for developers

Vittalbabu:
@dattanand If we are adding a new "native"/buitlin shader, it would be nice to have a sample that tests it. We shouldn't be depending on a later checkin to create a sample -- part of which will test this shader. I am guessing that you would have tested this portion of the code before issuing the pull request -- ensured it built and ran. Could you attach the sample code?

Also, why would we do the y-flipping? If we anticipate app developers will use this shader, what is the use case for flipping the y-coord?

dattanand:
@Vittalbabu 
yes definitely it was written for purpose of widgets, Y flipping is required to flip the texture created by FBO rendering in libgdx. the shader is tested and working, sample code is present in #330 @ragner has already verified it as I know, this is part of same implementation and broken down as different commit, once it is is included, framework and sample code will also be requested for inclusion

Vittalbabu:
@dattanand We can wait for the samples in the later pull request. I also see that this shader is used for FBO rendering in libgdx. 

I am still confused a bit. From an end user persepctive, would it be common to do the flip in the y-coord? (eg) Do we anticipate only libgdx generated textures to behave in this manner and require a flip? (ie) If I were to create a FBO in a "regular" OpenGL app. would the texture y-coord need to be flipped? I am trying to see if the behavior of the FBO shader is intuitive to users who create FBOs and textures.

Thanks.

dattanand:
@Vittalbabu 
I think Y flipping would be required for textures created from FBO, its not perticular to libgdx.

dattanand:
tracked via #367

PR #333, opened 2015-12-04
Adds shared texture support for libgdx widgets

dattanand:
Adding shared texture support to be used with libgdx widgets


GVRf/Framework/src/org/gearvrf/GVRSharedTexture.java
```diff
@@ -0,0 +1,39 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gearvrf;
+
+/**
```
liaxim: You might want to update the javadoc to update the intended purpose of this class

liaxim:
The shared_texture part looks good to me.

liaxim:
I am not entirely sure about this but instead of having the SharedTexture class we could retrofit the base GVRTexture class to support the case SharedTexture exists for. It might be less confusing to have one type of texture less.

dattanand:
I can think of creating gvrtexture base type object and passing texture ID, Kindly suggest how can it be achieved if we want to explicitly remove the term 'shared'

dattanand:
tracked via #367

Issue #279, opened 2015-10-14
Support All Assimp Shaders


thomasflynn:
We should support the same built-in shaders that Assimp supports.  For reference we can look at the Shaders.cpp file in Assimp's github in tools/assimp_view/ directory.

danke-sra:
Skeletal animation for assimp shader is added in https://github.com/Samsung/GearVRf/pull/335

PR #335, opened 2015-12-07
Skeletal animation (skinning) development work for review

danke-sra:
SIDIA contribution for mesh animation
- Mesh animation support in native code
- Mesh animation sample

GVRf skinning infrastrcture
- Add GVRBone and GVRBoneWeight and correpsonding native impl.
- Add GVRSkinningController to construct bone tree from a rigged model
  and control bone transforms during skeletal animation
- Modified assimp shader to animate mesh
- Add pretty print for GVRSceneObject for debugging scene graphs
- Add getLocalModelMatrix to GVRTransform to get local transform
- Modify Jassimp wrapper provider to construct Quaternionf
- Add infrastructure for keyframe animation
  - Add Scale/Pos/Rotation keys
  - Add animation channels
  - Add interpolators to interpolate vectors and quaternions
  - Generate overall transform based on scale/pos/rotation keys
  - Fully integrated with GVRAnimationEngine
- Modified gvrjassimpmodelloader to demo skinning

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/GVRContext.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRContext.java
@@ -473,7 +473,7 @@ public abstract class GVRContext {
      * @throws IOException
      *             File does not exist or cannot be read
      *
-     * @deprecated Replaced by {@link #loadJassimpModel}
+     * @deprecated Replaced by {@link #loadModel}
      */
     public GVRSceneObject getAssimpModel(String assetRelativeFilename)
             throws IOException {
@@ -505,19 +505,112 @@ public abstract class GVRContext {
         return GVRImporter.getAssimpModel(this, assetRelativeFilename, settings);
     }
 
+    /**
+     * @deprecated Replaced by {@link #loadModelFromSD}
+     */
     public GVRModelSceneObject loadJassimpModelFromSD(String externalFile) throws IOException {
-        return loadJassimpModelFromSD(externalFile, GVRImportSettings.getRecommendedSettings());
+        return loadModelFromSD(externalFile);
     }
 
+    /**
+     * @deprecated Replaced by {@link #loadModelFromSD}
+     */
     public GVRModelSceneObject loadJassimpModelFromSD(String externalFile, EnumSet<GVRImportSettings> settings) throws IOException {
+        return loadModelFromSD(externalFile, settings);
+    }
+
+    /**
+     * @deprecated Replaced by {@link #loadModel}
+     */
+    public GVRModelSceneObject loadJassimpModel(String assetFile) throws IOException {
+        return loadModel(assetFile);
+    }
+
+    /**
+     * @deprecated Replaced by {@link #loadModel}
+     */
+    public GVRModelSceneObject loadJassimpModel(String assetFile, EnumSet<GVRImportSettings> settings) throws IOException {
+        return loadModel(assetFile, settings);
+    }
+
+    /**
+     * Simple, high-level method to load a scene object {@link GVRModelSceneObject} from
+     * a 3D model stored in the Android SD card.
+     *
+     * @param externalFile
+     *            A filename, relative to the SD card's root directory.
+     *
+     * @return A {@link GVRModelSceneObject} that contains the meshes with textures and bones
+     * and animations.
+     *
+     * @throws IOException
+     *             File does not exist or cannot be read
+     */
+    public GVRModelSceneObject loadModelFromSD(String externalFile) throws IOException {
+        return loadModelFromSD(externalFile, GVRImportSettings.getRecommendedSettings());
+    }
+
+    /**
+     * Simple, high-level method to load a scene object {@link GVRModelSceneObject} from
+     * a 3D model stored in the Android SD card.
+     *
+     * @param externalFile
+     *            A filename, relative to the SD card's root directory.
+     *
+     * @param settings
+     *            Additional import {@link GVRImportSettings settings}
+     *
+     * @return A {@link GVRModelSceneObject} that contains the meshes with textures and bones
+     * and animations.
+     *
+     * @throws IOException
+     *             File does not exist or cannot be read
+     */
+    public GVRModelSceneObject loadModelFromSD(String externalFile, EnumSet<GVRImportSettings> settings) throws IOException {
         return GVRImporter.loadJassimpModelFromSD(this, externalFile, settings);
     }
 
-    public GVRModelSceneObject loadJassimpModel(String assetFile) throws IOException {
-        return loadJassimpModel(assetFile, GVRImportSettings.getRecommendedSettings());
+    /**
+     * Simple, high-level method to load a scene object {@link GVRModelSceneObject} from
+     * a 3D model stored in Android application assets.
+     *
+     * @param assetFile
+     *            A filename, relative to the {@code assets} directory. The file
+     *            can be in a sub-directory of the {@code assets} directory:
+     *            {@code "foo/bar.png"} will open the file
+     *            {@code assets/foo/bar.png}
+     *
+     * @return A {@link GVRModelSceneObject} that contains the meshes with textures and bones
+     * and animations.
+     *
+     * @throws IOException
+     *             File does not exist or cannot be read
+     */
+    public GVRModelSceneObject loadModel(String assetFile) throws IOException {
+        return loadModel(assetFile, GVRImportSettings.getRecommendedSettings());
     }
 
-    public GVRModelSceneObject loadJassimpModel(String assetFile, EnumSet<GVRImportSettings> settings) throws IOException {
+    /**
+     * Simple, high-level method to load a scene object {@link GVRModelSceneObject} from
+     * a 3D model stored in the Android SD card.
+     *
+     * @param assetFile
+     *            A filename, relative to the {@code assets} directory. The file
+     *            can be in a sub-directory of the {@code assets} directory:
+     *            {@code "foo/bar.png"} will open the file
+     *            {@code assets/foo/bar.png}
+     *
+     * @param settings
+     *            Additional import {@link GVRImportSettings settings}
+     *
+     * @return A {@link GVRModelSceneObject} that contains the meshes with textures and bones
+     * and animations.
+     *
+     * @throws IOException
+     *             File does not exist or cannot be read
+     *
+     */
+    public GVRModelSceneObject loadModel(String assetFile, EnumSet<GVRImportSettings> settings) throws IOException {
         return GVRImporter.loadJassimpModel(this, assetFile, settings);
     }
 
--- a/GVRf/Framework/src/org/gearvrf/GVRContext.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRContext.java
@@ -32,6 +32,7 @@ import org.gearvrf.asynchronous.GVRAsynchronousResourceLoader;
 import org.gearvrf.asynchronous.GVRCompressedTexture;
 import org.gearvrf.asynchronous.GVRCompressedTextureLoader;
 import org.gearvrf.periodic.GVRPeriodicEngine;
+import org.gearvrf.scene_objects.GVRModelSceneObject;
 import org.gearvrf.utility.Log;
 import org.gearvrf.utility.ResourceCache;
 
@@ -504,19 +505,19 @@ public abstract class GVRContext {
         return GVRImporter.getAssimpModel(this, assetRelativeFilename, settings);
     }
 
-    public GVRSceneObject loadJassimpModelFromSD(String externalFile) throws IOException {
+    public GVRModelSceneObject loadJassimpModelFromSD(String externalFile) throws IOException {
         return loadJassimpModelFromSD(externalFile, GVRImportSettings.getRecommendedSettings());
     }
 
-    public GVRSceneObject loadJassimpModelFromSD(String externalFile, EnumSet<GVRImportSettings> settings) throws IOException {
+    public GVRModelSceneObject loadJassimpModelFromSD(String externalFile, EnumSet<GVRImportSettings> settings) throws IOException {
         return GVRImporter.loadJassimpModelFromSD(this, externalFile, settings);
     }
 
-    public GVRSceneObject loadJassimpModel(String assetFile) throws IOException {
+    public GVRModelSceneObject loadJassimpModel(String assetFile) throws IOException {
         return loadJassimpModel(assetFile, GVRImportSettings.getRecommendedSettings());
     }
 
-    public GVRSceneObject loadJassimpModel(String assetFile, EnumSet<GVRImportSettings> settings) throws IOException {
+    public GVRModelSceneObject loadJassimpModel(String assetFile, EnumSet<GVRImportSettings> settings) throws IOException {
         return GVRImporter.loadJassimpModel(this, assetFile, settings);
     }
 
--- a/GVRf/Framework/src/org/gearvrf/GVRHybridObject.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRHybridObject.java
@@ -154,7 +154,7 @@ public abstract class GVRHybridObject implements Closeable {
         return mNativePointer;
     }
 
-    public static long[] getNativePtrArray(Collection<? extends GVRHybridObject> objects) {
+    /*package*/ static long[] getNativePtrArray(Collection<? extends GVRHybridObject> objects) {
         long[] ptrs = new long[objects.size()];
 
         int i = 0;
--- a/GVRf/Framework/src/org/gearvrf/GVRHybridObject.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRHybridObject.java
@@ -20,6 +20,7 @@ import java.io.IOException;
 import java.lang.ref.PhantomReference;
 import java.lang.ref.ReferenceQueue;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -56,7 +57,7 @@ public abstract class GVRHybridObject implements Closeable {
 
     /**
      * Normal constructor
-     * 
+     *
      * @param gvrContext
      *            The current GVRF context
      * @param nativePointer
@@ -151,6 +152,17 @@ public abstract class GVRHybridObject implements Closeable {
      */
     public long getNative() {
         return mNativePointer;
+    }
+
+    public static long[] getNativePtrArray(Collection<? extends GVRHybridObject> objects) {
+        long[] ptrs = new long[objects.size()];
+
+        int i = 0;
+        for (GVRHybridObject obj : objects) {
+            ptrs[i++] = obj.getNative();
+        }
+
+        return ptrs;
     }
 
     @Override
--- a/GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
@@ -908,10 +908,6 @@ class NativeSceneObject {
 
     static native void detachEyePointeeHolder(long sceneObject);
 
-    static native void attachBone(long sceneObject, long bone);
-
-    static native void detachBone(long sceneObject);
-
     static native long setParent(long sceneObject, long parent);
 
     static native void addChildObject(long sceneObject, long child);
--- a/GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
@@ -17,11 +17,14 @@ package org.gearvrf;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Set;
 import java.util.concurrent.Future;
 
 import org.gearvrf.GVRMaterial.GVRShaderType;
+import org.gearvrf.utility.Log;
 
 /**
  * One of the key GVRF classes: a scene object.
@@ -38,7 +41,7 @@ import org.gearvrf.GVRMaterial.GVRShaderType;
  * {@link GVRRenderData} has a {@link GVRMesh GL mesh} that defines its
  * geometry, and a {@link GVRMaterial} that defines its surface.
  */
-public class GVRSceneObject extends GVRHybridObject {
+public class GVRSceneObject extends GVRHybridObject implements PrettyPrint {
 
     private GVRTransform mTransform;
     private GVRRenderData mRenderData;
@@ -784,10 +787,96 @@ public class GVRSceneObject extends GVRHybridObject {
         }
     }
 
+    private void findChildrenByName(String name, List<GVRSceneObject> objects,
+            boolean onlyFirst, Set<GVRSceneObject> visited) {
+        if (visited.contains(this))
+            return;
+
+        visited.add(this);
+
+        for (GVRSceneObject child : mChildren) {
+            String childName = child.getName();
+            if (name.equals(childName)) {
+                objects.add(child);
+                if (onlyFirst)
+                    return;
+            }
+
+            child.findChildrenByName(name, objects, onlyFirst, visited);
+            if (onlyFirst && !objects.isEmpty())
+                return;
+        }
+    }
+
+    /**
+     * Searches for the all children with the specified name.
+     *
+     * @param name the name of the children.
+     * @return the children found
+     */
+    public List<GVRSceneObject> findAllChildrenByName(String name) {
+        List<GVRSceneObject> objects = new ArrayList<GVRSceneObject>();
+        Set<GVRSceneObject> visited = new HashSet<GVRSceneObject>();
+        findChildrenByName(name, objects, false, visited);
+        return objects;
+    }
+
+    /**
+     * Searches for the first child with the specified name.
+     * 
+     * @param name the name of the child.
+     * @return the child found
+     */
+    public GVRSceneObject findChildByName(String name) {
+        List<GVRSceneObject> objects = new ArrayList<GVRSceneObject>();
+        Set<GVRSceneObject> visited = new HashSet<GVRSceneObject>();
+        findChildrenByName(name, objects, true, visited);
+        return objects.isEmpty() ? null : objects.get(0);
+    }
+
     /**
      * Called when the scene object has been loaded from a model.
      */
     public void onLoaded() {
+    }
+
+    /**
+     * Generate debug dump of the tree from the scene object.
+     * It should include a newline character at the end.
+     * 
+     * @param sb the {@code StringBuffer} to dump the object.
+     * @param indent indentation level as number of spaces.
+     */
+    public void prettyPrint(StringBuffer sb, int indent) {
+        sb.append(Log.getSpaces(indent));
+        sb.append(getClass().getSimpleName());
+        sb.append(" [name=");
+        sb.append(this.getName());
+        sb.append("]");
+        sb.append(System.lineSeparator());
+
+        if (mRenderData == null) {
+            sb.append(Log.getSpaces(indent + 2));
+            sb.append("RenderData: null");
+            sb.append(System.lineSeparator());
+        } else {
+            mRenderData.prettyPrint(sb, indent + 2);
+        }
+        sb.append(Log.getSpaces(indent + 2));
+        sb.append("Transform: "); sb.append(mTransform);
+        sb.append(System.lineSeparator());
+
+        // dump its children
+        for (GVRSceneObject child : getChildren()) {
+            child.prettyPrint(sb, indent + 2);
+        }
+    }
+
+    @Override
+    public String toString() {
+        StringBuffer sb = new StringBuffer();
+        prettyPrint(sb, 0);
+        return sb.toString();
     }
 }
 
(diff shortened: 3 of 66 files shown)


GVRf/Framework/jni/objects/animation/mesh_animation.cpp
```diff
@@ -0,0 +1,310 @@
...
+            return;
+        }
+    }
+}
+
+void MeshAnimation::setScene(const aiScene* scene) {
+    aiScene_ = scene;
+}
+
+void MeshAnimation::setGlobalInverTransform(const glm::mat4& globalTransform) {
```
thomasflynn: InverTransform?  typo?  or is the 'se' in Inverse deliberately left out?

GVRf/Framework/src/org/gearvrf/jassimp2/GVRJassimpSceneObject.java
```diff
@@ -14,15 +16,30 @@
...
+            recurseAssimpNodes(this, scene.getSceneRoot(GVRJassimpAdapter.sWrapperProvider));
+
+            // Animations
+            for (AiAnimation aiAnim : scene.getAnimations()) {
+                animations.add(GVRJassimpAdapter.get().createAnimation(aiAnim, this));
+            }
+        }
+    }
+
+    public List<GVRKeyFrameAnimation> getAnimations() {
```
liaxim: This method requires an explicit cast to GVRJassimpSceneObject. Isn't GVRJassimpSceneObject an implementation detail?

dankex: GVRJassimpSceneObject is not for implementation. It is an interface for loaded model. Before we have a generic model node, it mimics the functionality of aiScene.

dankex: How do you think about changing the return type of loadJassimpModel() return GVRJassimpSceneObject, or a new type GVRModelSceneObject?

GVRf/Framework/jni/objects/animation/mesh_animation.cpp
danke-sra: This is from fredduarte's code. Because we use full names, I will add a patch to fix it.

thomasflynn:
The included MeshAnimationSample does not compile.  The existing jassimpmodelloader sample does compile and show the animation on astro boy.

danke-sra:
Actually, the MeshAnimationSample does work on my side and shows the red dinosaur. Hopefully I didn't forget to check in any file. Can you please try to pull the whole branch: "git checkout -b localbranch; git pull https://github.com/Samsung/GearVRf.git pull/335/head"?

thomasflynn:
ok, did that.  still didn't build.  the sample refers to a MeshAnimation ShaderType on line 62.  That doesn't exist in GVRMaterial.java.  Also, the sample fails to compile line 58 with the characterMesh.animate(0.0f);  even though there is a public void animation(float time) in GVRMesh.h.  That one puzzles me.

danke-sra:
I checked those data are there. The ShaderType MeshAnimation is defined in GVRMaterial.java (https://github.com/danke-sra/GearVRf/blob/animation-dev/GVRf/Framework/src/org/gearvrf/GVRMaterial.java). Maybe the files got wrong timestamps. Can you please do a "git reset --hard FETCH_HEAD" following the above commands and force eclipse to clean the project? Thanks.

GVRf/Framework/src/org/gearvrf/animation/keyframe/GVRKeyFrameAnimation.java
```diff
@@ -0,0 +1,113 @@
...
+            if (nodeName != null && nodeName.equals(channel.getNodeName())) {
+                return i;
+            }
+            i++;
+        }
+        return -1;
+    }
+
+    @Override
+    protected void animate(GVRHybridObject target, float ratio) {
```
liaxim: This method in the case of the MeshAnimationSample takes from 50 to 350 ms, monotonically increasing; then starts over. It happens on the gl thread so the fps is horrendous. Is something wrong with my setup?

I was wondering for some time why do we have to execute all the animation logic on the gl thread. I haven't seen anything that makes it mandatory.

GVRf/Framework/jni/objects/animation/mesh_animation.cpp
```diff
@@ -0,0 +1,310 @@
...
+        if (weights[i] == 0.0) {
+            ids[i] = boneId;
+            weights[i] = boneWeight;
+            return;
+        }
+    }
+}
+
+void MeshAnimation::setScene(const aiScene* scene) {
+    aiScene_ = scene;
```
liaxim: I would imagine all the Assimp-stuff will be confined to an importer unit/package that converts to our internal representation and that's it. The MeshAnimation is unusable without an aiScene. Are we supposed to support just the case where this stuff comes from importing a model? That would be ok as long as it is spelled out and could simplify certain things. The decision should be applicable to bones too then.

We could refactor these things later in case we want them in soon. If that is the case I will just stop making needless noise.

GVRf/Framework/src/org/gearvrf/animation/GVRAssimpAnimation.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package org.gearvrf.animation;
+
+import org.gearvrf.GVRHybridObject;
+import org.gearvrf.GVRMesh;
+import org.gearvrf.GVRSceneObject;
+
+/** Animate Mesh from assimp **/
+public class GVRAssimpAnimation extends GVRAnimation {
```
liaxim: Assimp being "leaked". Assimp should be an implementation detail people are not aware of - we should strive for that at least.

GVRf/Framework/src/org/gearvrf/scene_objects/GVRModelSceneObject.java
```diff
@@ -0,0 +1,69 @@
...
+import org.gearvrf.GVRSceneObject;
+import org.gearvrf.PrettyPrint;
+import org.gearvrf.animation.GVRAnimation;
+import org.gearvrf.utility.Log;
+
+/**
+ * {@linkplain GVRSceneObject Scene object} that holds a loaded model.
+ */
+
+public class GVRModelSceneObject extends GVRSceneObject {
```
liaxim: Might be just me but I find this node type undesirable. Can't we have an animations node that is a child of the model's "root" or something? If we keep doing this we will end up with one super node that holds all possible state and does everything. Also is this node just for the case where we import a model?

GVRf/Framework/src/org/joml/AxisAngle4d.java
```diff
@@ -0,0 +1,597 @@
+/*
```
liaxim: The joml part should have been a separate pull request that we could have merged by now and made reviewing the rest a bit less tedious.

GVRf/Framework/src/org/gearvrf/animation/keyframe/GVRKeyFrameAnimation.java
danke-sra: I think the goal is to optimize it so that it doesn't matter where it runs. The I/O load for animation is tiny, and computation is O( number of animated nodes). Some places in the keyframe animation can be made faster, but I am still surprised to see lags happening like this.

GVRf/Framework/jni/objects/animation/mesh_animation.cpp
danke-sra: This code is transient. The code has been integrated in the java implementation using internal representations.

GVRf/Framework/src/org/gearvrf/animation/GVRAssimpAnimation.java
danke-sra: This is transient code too. We just keep it until all samples are verified to work with new implementation.

GVRf/Framework/src/org/gearvrf/GVRHybridObject.java
```diff
@@ -50,23 +51,41 @@
      */
     private long mNativePointer;
 
+    /**
+     * If true, the native object will be deleted by this class; otherwise,
+     * it won't be deleted.
+     */
+    private boolean mOwnNative;
```
liaxim: On first glance this is innocent enough but it is a symptom of the fact that the import process has not been abstracted enough. Assimp is now driving our design and we have to special-case the bone data by breaking a framework-wide assumption - that the Java objects own and manage the native peers, unconditionally.

This design is not set in stone but if it were to change, we need to see the general case and supply appropriate solution, not circumventing the current.

GVRf/Framework/src/org/gearvrf/scene_objects/GVRModelSceneObject.java
danke-sra: I know this is hard to understand. However, animations loaded from a model can only affect nodes in the model. We need something like this to keep the namespace of the model. Of course, we can created animations outside the model like we are doing right now.

GVRf/Framework/src/org/joml/AxisAngle4d.java
danke-sra: agreed.

GVRf/Framework/src/org/gearvrf/GVRHybridObject.java
danke-sra: There is some assimp issue in the native version (not my code). However, the native class (VertexBoneData) is a data member of Mesh (not a pointer). So it does not need to be deleted. The existing impl. assumes all native objects need to be deleted, which is untrue in C++ case.

GVRf/Framework/src/org/gearvrf/animation/keyframe/GVRKeyFrameAnimation.java
liaxim: Isn't the fact that the computation is on the gl thread and is O(animated nodes) somewhat problematic? What happens if someone has 10 dinosaurs running around? Or more? We could establish some limits, should we?

danke-sra: This patch is an initial one for skinning. Despite this, I doubt you would detect performance bottleneck with an order of just 10 dinosaurs. There are some traditional ways to optimize the rendering engine. 1) explicitly support instancing. this patch is prepared for that, but not really implemented it for clarify. 2) we can implement a pipeline in the rendering engine, spliting the GL thread into 3 stages (updating, culling and drawing). That is a bigger change than the scope of this.

liaxim: Sounds good in general. Have you tried the sample? How does it run for you?

danke-sra: It runs smoothly. I don't see any alarming issues. However, I can do an optimization pass after this first version. That will include a couple of things: 1) caching bone matrix locations for shader (no performance gain but just a peace of mind) 2) some pruning in the two animation controllers 3) speed-up keyframe look up (about .1ms / frame) and any other things.

danke-sra: Done 3). FPS with one dinosaur is 60. I will work on multiple later. Please see if this patch can be in with the condition that some clean-up will happen soon?

liaxim:
After undocking while running the mesh animation sample I got a crash because of this exception:

```12-10 09:46:27.981 15185 15237 F art     : sart/runtime/runtime.cc:292] Pending exception java.lang.RuntimeException thrown by 'unknown throw location'
12-10 09:46:27.981 15185 15237 F art     : sart/runtime/runtime.cc:292] java.lang.RuntimeException: Interpolation failed
12-10 09:46:27.981 15185 15237 F art     : sart/runtime/runtime.cc:292]   at java.lang.Object org.gearvrf.animation.keyframe.GVRAnimationChannel$KeyFrameInterplator.interpolate(float) (GVRAnimationChannel.java:83)
12-10 09:46:27.981 15185 15237 F art     : sart/runtime/runtime.cc:292]   at org.joml.Vector3f org.gearvrf.animation.keyframe.GVRAnimationChannel.getScale(float) (GVRAnimationChannel.java:356)
12-10 09:46:27.981 15185 15237 F art     : sart/runtime/runtime.cc:292]   at org.joml.Matrix4f org.gearvrf.animation.keyframe.GVRAnimationChannel.animate(float) (GVRAnimationChannel.java:312)
12-10 09:46:27.981 15185 15237 F art     : sart/runtime/runtime.cc:292]   at org.joml.Matrix4f[] org.gearvrf.animation.keyframe.GVRKeyFrameAnimation.getTransforms(float) (GVRKeyFrameAnimation.java:109)
12-10 09:46:27.981 15185 15237 F art     : sart/runtime/runtime.cc:292]   at void org.gearvrf.animation.keyframe.GVRNodeAnimationController.animate(float) (GVRNodeAnimationController.java:75)```

**Comment by liaxim** [comment]
Got the same exception some time during the animation - say a minute after I launched the app. The performance is better.

**Comment by fredduarte** [comment]
This issue was fixed the the top commit from my animation branch:
https://github.com/fredduarte/GearVRf/commit/1b29180e6dc6deab5210e20969c93a4d32bac29c
It is a temporary fix though. It works for single mesh animation only, for now.
@danke-sra can cherry-pick this commit from there.

**Comment by ragner** [comment]
At current implementation we need keep the GVRAssimpImporter's instance for each imported mesh
otherwise Android GC will delete it, which may be the cause of the Exception.

File: GVRf/Framework/jni/shaders/material/mesh_animation_shader.cpp [hunk 9]
```diff
@@ -0,0 +1,133 @@
+#include "mesh_animation_shader.h"
```
liaxim: I don't think the mesh_animation_shader is in use currently. What about it?

ragner: maybe @danke-sra  forgot to remove it . That shader is used at @fredduarte's implementation

danke-sra: Ok. I will clean up the previous implementation. Thanks @fredduarte for contributing it again. It really helps The crash is fixed -- I added one line to handle a corner case explicitly.

danke-sra:
These patches are probably ready for check in now. It has been tested on S6 and Note4 and works with no problems.

thomasflynn:
@fredduarte , @ragner and @liaxim please review again and give me the thumbs up/down.  Thanks.

GVRf/Framework/src/org/gearvrf/animation/keyframe/GVRNodeAnimationController.java
```diff
@@ -0,0 +1,86 @@
...
+        if (channelId != -1 && containsRenderable) {
+            animatedNodes.add(new AnimationItem(node, channelId));
+        }
+
+        return containsRenderable;
+    }
+
+    /**
+     * Update node transforms at each animation step.
+     */
```
fredduarte: The animate() method from GVRNodeAnimationController and from GVRSkinningController seem to have 
equal code for calculating the animation transforms. Is there a possibility for putting it on a common method
used by both classes?

liaxim:
If you are stationary looking at the animation it works fine; if you start rotating your head left and right 45 degrees as the animation is running it inevitably crashes within few seconds with this:

```
12-14 07:36:31.641 12616 12715 F art     : sart/runtime/runtime.cc:292] Pending exception java.lang.RuntimeException thrown by 'unknown throw location'
12-14 07:36:31.641 12616 12715 F art     : sart/runtime/runtime.cc:292] java.lang.RuntimeException: Interpolation failed
12-14 07:36:31.641 12616 12715 F art     : sart/runtime/runtime.cc:292]   at java.lang.Object org.gearvrf.animation.keyframe.GVRAnimationChannel$KeyFrameInterplator.interpolate(float) (GVRAnimationChannel.java:86)
12-14 07:36:31.641 12616 12715 F art     : sart/runtime/runtime.cc:292]   at org.joml.Vector3f org.gearvrf.animation.keyframe.GVRAnimationChannel.getScale(float) (GVRAnimationChannel.java:367)
12-14 07:36:31.641 12616 12715 F art     : sart/runtime/runtime.cc:292]   at org.joml.Matrix4f org.gearvrf.animation.keyframe.GVRAnimationChannel.animate(float) (GVRAnimationChannel.java:323)
12-14 07:36:31.641  6027 13211 V VRSVC   : at (HmtSensorEvent.java:149) handleMessageGetSerialForZero() [Already have same UUID uuidOnPhone=N00J50M2H2D2]
12-14 07:36:31.641 12616 12715 F art     : sart/runtime/runtime.cc:292]   at org.joml.Matrix4f[] org.gearvrf.animation.keyframe.GVRKeyFrameAnimation.getTransforms(float) (GVRKeyFrameAnimation.java:129)
12-14 07:36:31.641 12616 12715 F art     : sart/runtime/runtime.cc:292]   at void org.gearvrf.animation.keyframe.GVRNodeAnimationController.animate(float) (GVRNodeAnimationController.java:80)
12-14 07:36:31.641 12616 12715 F art     : sart/runtime/runtime.cc:292]   at void org.gearvrf.animation.keyframe.GVRKeyFrameAnimation.animate(org.gearvrf.GVRHybridObject, float) (GVRKeyFrameAnimation.java:121)
12-14 07:36:31.641 12616 12715 F art     : sart/runtime/runtime.cc:292]   at boolean org.gearvrf.animation.GVRAnimation.onDrawFrame(float) (GVRAnimation.java:365)
12-14 07:36:31.641 12616 12715 F art     : sart/runtime/runtime.cc:292]   at void org.gearvrf.animation.GVRAnimationEngine$DrawFrame.onDrawFrame(float) (GVRAnimationEngine.java:155)
12-14 07:36:31.641 12616 12715 F art     : sart/runtime/runtime.cc:292]   at long org.gearvrf.GVRViewManager.doMemoryManagementAndPerFrameCallbacks() (GVRViewManager.java:688)
12-14 07:36:31.641 12616 12715 F art     : sart/runtime/runtime.cc:292]   at long org.gearvrf.GVRViewManager.access$4(org.gearvrf.GVRViewManager) (GVRViewManager.java:672)
12-14 07:36:31.641 12616 12715 F art     : sart/runtime/runtime.cc:292]   at void org.gearvrf.GVRViewManager$3.beforeDrawEyes() (GVRViewManager.java:655)
```

It looks like some of the assumptions in KeyFrameInterplator.getKeyIndex need to be changed. Whenever it crashes the current "time" is always very close to the "lastFrameTime" (e.g. time: 231.41621, firstFrameTime: 85.0, lastFrameTime: 232.0); referring to KeyFrameInterplator.interpolate. The sample did crash once on launch too. Using S6.

liaxim:
And review-wise, I am tired of raising the same points again and again: special-casing in GVRHybridObject, special-casing is what I believe lead to the introduction of GVRModelSceneObject, pretty-printing code sprinkled everywhere and few more I raised. Maybe none of them make sense as none were backed by someone other than me. If that is the case, I guess this code is impossible to be improved further and should go in as-is after the crashes are fixed.

Most things are nothing that a future refactoring can't address but public api changes should be more scrutinized - see GVRModelSceneModel, GVRMesh.

liaxim:
And about the mesh sample - please move it to the new samples repo, thanks.

GVRf/Framework/src/org/gearvrf/animation/keyframe/GVRNodeAnimationController.java
danke-sra: It is possible to combine them but I chose to keep the skinning controller dedicated to only skinning. There are a couple of reasons:

1) The node animation code may be extended to support other types of animation: opacity / effects, etc. This can accidentally break the skinning controller or lower its performance

2) Both controllers have pruning to exclude nodes they don't control. So the total execution time is optimized, as they are one.

3) The animation transforms are in the animation class so there is no duplicate.

danke-sra:
About the crash, it's not that things are not fixed after reviewing. But there is something weird we should responsibly track down. Now, it only throws when in DEBUG mode, still we need to find the root cause of the exception.

During my development, it only throws from that location, when there is an JNI error! The JNI error crashes the application, getting some data GC'ed, but the animation engine still calls into the animation code, and leads to the exception.

I think it needs more testing so I kept the throw in the DEBUG mode and added the print of time to find any patterns, other than it requires a prior JNI crash. I will re-test this issue and see why the 45 tilting affects that.

danke-sra:
As for review comments, I have replied, and did not see any new points. Some of those were from unclear points about the design/implementation. They have been clarified and answered.

danke-sra:
Removed the throw in the animation code. It has been tested without issue.

GVRf/Framework/src/org/gearvrf/animation/keyframe/GVRAnimationChannel.java
```diff
@@ -0,0 +1,423 @@
...
+                float lastFrameTime = keys[keys.length - 1].getTime();
+                T firstFrameValue = keys[0].getValue();
+                T lastFrameValue = keys[keys.length - 1].getValue();
+
+                if (time <= firstFrameTime) {
+                    return firstFrameValue;
+                } else if (time >= lastFrameTime) {
+                    return lastFrameValue;
+                } else {
+                    // Shouldn't happen
```
liaxim: Ok, so why it does happen so frequently if it "shouldn't happen"?

danke-sra: I've seen this happening when the application is crashed by other reasons. My theory is the data is corrupted when other crash happens. I cannot hit this code in unit testing -- either the previous version or the new version.

liaxim: You've never managed to reproduce this? Once you upload the mesh sample I will try again and give you my device.

As what's happening, this is my take without delving into how the time calculations get us there: the "shouldn't happen" case does happen and when it does the RuntimeException is thrown. This is all called by a native thread and there is no Java-side try-catch(RuntimeException) block anywhere in the call-chain so the Java exception "escapes" to JNI and JNI does what it can with it - crash.

So I thought this really is about finding what is wrong with the calculations or maybe the throw is redundant; but since it was there, you have a certain invariant in mind. Invariant that might need tweaks.

If there is a corruption you would expect some randomness.. it can't always lead to this very callstack in logcat.

danke-sra: I can't just do a linear search to find randomness (it makes it slow). Random data is caught by this very line after search, if time is not found. The JNI crash is not the kind I referred to. What I have seen is the application has been crashed by other obvious reasons (null ptr in JNI code or even a Java Runtime exception thrown elsewhere), however, when any crash happens I see this exception in the stack trace. It confuses me. But after other crash is fixed, this is gone. I think this is sometimes due to animation engine calling the code frequently (60 times per second), and when the app crashes, it does not kill all threads within 16ms, so it is shortly exposed to corrupted data. I need to see how to repro this in normal cases.

liaxim: For whatever it is worth, it crashes for me because of the RuntimeException that used to be there. No other crashes whatsoever. Also the times are consistently "sane". At the time of the crash, the "time" is always something very close to "lastFrameTime" (232.0); I get "time" of 231.71883, 231.95538, 231.41621 when the crash occurs.

Anyway, it is not going to crash anymore so if you are happy with this resolution, it is ok with me.

danke-sra:
I reverted one change and added one back. I think the final version is best. Please re-test if you like. I've submitted the sample to the sample repo.

danke-sra:
Added tentative changes assuming GVRHybridObject always deletes the native object. (some debug is in progress)

danke-sra:
Simplified GVRVertexBoneData implementation. There is no need to inherit from GVRHybridObject. Code works as before.

GVRf/Framework/src/org/gearvrf/GVRHybridObject.java
```diff
@@ -153,6 +154,17 @@ public long getNative() {
         return mNativePointer;
     }
 
+    public static long[] getNativePtrArray(Collection<? extends GVRHybridObject> objects) {
```
liaxim: I believe this should be package-private. It is of no use to end users.

liaxim:
Looks good to me. Two things left; one major and one minor.

- GVRModelSceneObject; I know you are looking at what to do there.

- PrettyPrinting; I do not object to the functionality just the way it is structured; we expose a new method and a new interface to end users that we could avoid; dozens of lines of code that may or may not be relevant are added to multitude of classes; we could have a PrettyPrinter utility class that accepts a scene graph or individual nodes, does all the work in one place and keeps the pretty-printing out of the way.

danke-sra:
Any pending issues before this can be merged? I understand API rework will not be in this one.

liaxim:
I listed what I feel like issues, for the record. I do not mind having this merged "as-is".

GVRf/Framework/src/org/gearvrf/GVRHybridObject.java
danke-sra: I am a little bit dubious about treating public member as strictly end-user API. Java has no nested packages, package-private does not always provide adequate visibility to framework classes themselves -- unless all framework classes are in one package. This leads to disorganization in the framework. The gain here would lead to liabilities as the framework grows.

danke-sra: However, here I can change it to protected or package -- no problem.

danke-sra:
Can this be checked in now?

thomasflynn:
i'll do another review when i get a chance to breathe tonight/tomorrow

danke-sra:
I added cross-references for issues fixed by this. (#279 is partially fixed)

danke-sra:
For those who asked, test apps are here:

https://github.com/gearvrf/GearVRf-Demos/pull/1
https://github.com/gearvrf/GearVRf-Demos/pull/5

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
```diff
@@ -784,11 +787,97 @@ public void removeChildObject(GVRComponent childComponent) {
...
+        }
+    }
+
+    /**
+     * Searches for the all children with the specified name.
+     *
+     * @param name the name of the children.
+     * @return the children found
+     */
+    public List<GVRSceneObject> findAllChildrenByName(String name) {
```
thomasflynn: isn't this redundant with getSceneObjectsByName()?

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
```diff
@@ -784,11 +787,97 @@ public void removeChildObject(GVRComponent childComponent) {
...
+        return objects;
+    }
+
+    /**
+     * Searches for the first child with the specified name.
+     * 
+     * @param name the name of the child.
+     * @return the child found
+     */
+    public GVRSceneObject findChildByName(String name) {
```
thomasflynn: ditto with getSceneObjectByName

GVRf/Sample/gvrjassimpmodelloader/src/org/gearvrf/jassimpmodelloader/JassimpModelLoaderViewManager.java
```diff
@@ -27,7 +27,13 @@
 import org.gearvrf.animation.GVRAnimation;
 import org.gearvrf.animation.GVRAnimationEngine;
 import org.gearvrf.animation.GVRRepeatMode;
+<<<<<<< HEAD
```
thomasflynn: is this the result of a bad merge?  why are we seeing the <<<<<< HEAD, =======, >>>>> pattern added in this file?

GVRf/Framework/src/org/gearvrf/GVRContext.java
```diff
@@ -504,19 +505,19 @@ public GVRSceneObject getAssimpModel(String assetRelativeFilename,
         return GVRImporter.getAssimpModel(this, assetRelativeFilename, settings);
     }
 
-    public GVRSceneObject loadJassimpModelFromSD(String externalFile) throws IOException {
+    public GVRModelSceneObject loadJassimpModelFromSD(String externalFile) throws IOException {
```
thomasflynn: changing the return type to GVRModelSceneObject will break API compatibility.  maybe this is the time to introduce that loadModel() api discussed in another check-in.  That way we could introduce that new api and deprecate the loadJassimp* ones.  We will be able to remove the loadJassimp* apis in a future revision.

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
thomasflynn: are these functions still there in the latest patch?  it looks like they are.

danke-sra: done. uploaded a new patch for this.

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
danke-sra: done. uploaded a new patch for this.

thomasflynn:
tested with the associated gvrjassimpmodelviewer sample.  the sample requires a minor fix, but otherwise looks good at this point.

PR #343, opened 2015-12-11
Load model from a URL

lokeshguddu:
Also called Remote Model Loading. This uses the feature to load a model file from the External Storage of the device. We first download and parse a JSON file which contains the information about the model file and all its textures. We then download all the relevant files into the external storage and when the download is over, we lead the model with all its textures applied.


thomasflynn:
Thanks for the pull-request!  Looks like a very desirable feature.  I like how the number of framework changes are minimal and the bulk of the code is in the sample which is also somewhat small.  I'll try to take a closer look over the weekend.

danke-sra:
Thanks for the patch. I think for general model loading from network, we can't rely on the JSON file to map out required files to download. I did a similar patch to loadJassimpModel API. Internally, it unifies different file systems: android assets, SD card, linux filesystem, and URL. You may take a look and comment too: https://github.com/danke-sra/GearVRf/tree/jassimp_network

lokeshguddu:
Hi @ragner ,

Good and expected changes. Will update this branch with the recommended changes.

ragner:
:+1:

GVRf/Framework/src/org/gearvrf/GVRContext.java
```diff
@@ -501,7 +501,46 @@ public GVRSceneObject getAssimpModel(String assetRelativeFilename)
...
+     * @param fileName
+     *            A filename, present in the external storage of the device
+     * 
+     * @return A {@link GVRSceneObject} that contains the meshes with textures
+     * 
+     * @throws IOException
+     *             File does not exist or cannot be read
+     *
+     */
+    public GVRSceneObject getAssimpModelFromSDCard(String fileName)
```
liaxim: Did you consider using (and extending if necessary) the GVRAndroidResource class from GVRF? Seems like we need something like it to avoid the "fan out" GVRAndroidResource mentions. Also it is already used for textures and meshes..

GVRf/Sample/gvrurlmodelloader/src/org/gearvrf/urlmodelloader/URLModelLoaderViewManager.java
```diff
@@ -0,0 +1,203 @@
...
+        scene = gvrContext.getNextMainScene();
+        scene.setFrustumCulling(true);
+
+        // set background color
+        GVRCameraRig mainCameraRig = scene.getMainCameraRig();
+        mainCameraRig.getLeftCamera().setBackgroundColor(Color.BLACK);
+        mainCameraRig.getRightCamera().setBackgroundColor(Color.BLACK);
+        mainCameraRig.getTransform().setPosition(0.0f, 0.0f, 0.0f);
+
+        String jsonFileURL = "https://raw.githubusercontent.com/lokeshguddu/lokeshguddu.github.io/master/remote_model_urls.json";
```
liaxim: The url probably needs to change to the official GearVRf-Demos repo. Also please move the sample to that repo (we will remove all the samples from this repo soon). The samples repo is https://github.com/gearvrf/GearVRf-Demos.

liaxim:
Code looks good, sample works fine. The use of GVRAndroidResource is something we could do in the future if you are not up for it. We need to have the sample moved before this can be merged.

liaxim:
I personally like that all the mapping is left to the application using GVRF. I very much vote for keeping this kind of stuff out of GVRF.

PR #345, opened 2015-12-12
Network model loading (Fix #334 for the LoadJassimpModel API)

### LINKED ISSUE #334: File import fails on using loadJassimpModelFromSD(...)
Hi @danke-sra ,

I'm trying to load a model file which is present in the Device Storage on the Note4 test device. The models load fine if they are part of the assets folder of the application but fails when trying to access the same from external storage.

I'm copying the log that reflects the error.

```
...
...
...
11-12 08:23:16.674: W/System.err(19349): java.io.IOException: Unable to open file "astro_boy.dae".
11-12 08:23:16.674: W/System.err(19349): 	at org.gearvrf.jassimp2.Jassimp.aiImportFile(Native Method)
11-12 08:23:16.674: W/System.err(19349): 	at org.gearvrf.jassimp2.Jassimp.importFile(Jassimp.java:96)
11-12 08:23:16.674: W/System.err(19349): 	at org.gearvrf.GVRImporter.loadJassimpModelFromSD(GVRImporter.java:122)
11-12 08:23:16.674: W/System.err(19349): 	at org.gearvrf.GVRContext.loadJassimpModelFromSD(GVRContext.java:512)
...
...
...
```

You can also reproduce the same error if you keep the astro_boy.dae in the external storage and use 
```
loadJassimpModelFromSD("astro_boy.dae");
```
in the place of loadJassimpModel(...) at line 69 of [gvrjassimpmodelloader sample code](https://github.com/Samsung/GearVRf/blob/master/GVRf/Sample/gvrjassimpmodelloader/src/org/gearvrf/jassimpmodelloader/JassimpModelLoaderViewManager.java#L69).

danke-sra:
- Use a single entry-point to load models from all filesystems and network
- Add GVRResourceVolume to represent various filesystems from which resources
  can be loaded: android assets, linux filesystem, android sdcard and network.
- Resources referenced by a model file can be loaded from the same location of
  the model file using GVRReourceVolume.
- FileNameUtils is updated to help resolve relative paths in filesystems and URLs.
- Updated sample to show loading from network.


danke-sra:
This applies similar URL feature to the new API loadJassimpModel(...)

GVRf/Framework/src/org/gearvrf/GVRResourceVolume.java
```diff
@@ -0,0 +1,111 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.net.MalformedURLException;
+import java.net.URL;
+
+import android.os.Environment;
+
+/**
+ * Models a file system which supports stream I/O.
```
liaxim: There is some overlap with the purpose of GVRAndroidResource. We can't have them both with this being some kind of a dispatcher. Can't this stuff be rolled into GVRAndroidResource?

danke-sra: Good question. That's the same question I asked myself when started to add this feature. 

However, that turns out to be sub-optimal to me: 1) Resource and Resource Volume are different constructs like File vs. File System. Resource Volume is similar to the IOSystem interface that can be extended to be a clean module of virtual file system. Now it is simple but it is does become complex, mixing it up with GVRAndroid would make the code unreadable / unmaintainable. 

2) The only overlap is the function "openFileInSameLocation" in GVRAndroidResource. It it not implemented and can be removed so their functionality are separated.

liaxim: GVRAndroidResource is used for loading textures and meshes. It should be used (and extended if necessary) to load models. If GVRAndroidResource is not good enough - we should replace it with something better that is used for textures, meshes and models and prevents the overload fan out GVRAndroidResource does.

We can't have the implementation (or worse, future implementation predictions) drive the APIs we expose. Let's come up with best API we can and then we can worry about the implementation.

danke-sra: Let me point out something different from meshes and models to begin with (which is necessary):

1) Mesh and texture cannot include other files. -- so only a wrapper for a File would suffice -- this is what GVRAndroidResource is designed for. The newly added overload GVRAndroidResource(URL) is a compatible addition to original design and purpose. It can be used for texture / mesh as well.

2) Models require something that supports "openFile" (see IOSystem interface in assimp). FBX can include submodels in other FBX file. OBJ include MTL as a separate file. If we don't break original design of GVRAndroidResource, we would need an equivalent of IOSystem, for which I think the cheapest impl. will be something like the current GVRResourceVolume.

danke-sra:
Cleaned up the unused openFileInSameLocation method in GVRAndroidResource

danke-sra:
For the point about API design, I think this patch represents a step towards that. However, it is unfair to call this as a implementation-driven API. The goal is to refactor towards a unified and simple API for model loading. However, before the API is drastically changed, some refactoring towards this is shown to be practical in this patch. It works reliably.

GVRf/Framework/src/org/gearvrf/GVRResourceVolume.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    /**
+     * Opens a file from the volume. The filePath is relative to the defaultPath.
+     *
+     * @param filePath          File path of the resource to open.
+     *
+     * @throws IOException
+     */
+    public GVRAndroidResource openResource(String filePath) throws IOException {
```
liaxim: (1)
For the sake of the discussion, the various cases could have been different static factory methods in GVRAndroidResource:
```
GVRAndroidResource r = GVRAndroidResource.assetsResource(..)
GVRAndroidResource r = GVRAndroidResource.filesystemResource(..)
GVRAndroidResource r = GVRAndroidResource.externalStorageResource(..)
GVRAndroidResource r = GVRAndroidResource.networkResource(..)
```

People are already having to deal with loadTexture and loadMesh so they ideally should have to be familiar with one way of representing a resource (given the fact that we went through the trouble to come up with GVRAndroidResource).

GVRf/Framework/src/org/gearvrf/GVRContext.java
```diff
@@ -517,7 +518,16 @@ public GVRSceneObject loadJassimpModel(String assetFile) throws IOException {
     }
 
     public GVRSceneObject loadJassimpModel(String assetFile, EnumSet<GVRImportSettings> settings) throws IOException {
-        return GVRImporter.loadJassimpModel(this, assetFile, settings);
+        return GVRImporter.loadJassimpModel(this, assetFile, GVRResourceVolume.VolumeType.ANDROID_ASSETS, settings);
+    }
+
+    public GVRSceneObject loadJassimpModelFromURL(String urlString) throws IOException {
```
liaxim: (2)
And then, if you happen to have something like what I suggested in comment (1), users should be able to use one version of loadJassimpModel which accepts a GVRAndroidResource. And on a side note, I think we should rename the method to loadModel as the loader we use is irrelevant.

danke-sra:
Can we do the API redesign in another patch? I am a bit lost in the discussion. This is meant to be a fix to the filed bug -- just load textures from the location of the model. For API change, I think it can come next, and I am flexible with that.

thomasflynn:
ditto with the API change to loadModel().  that should be a different separate commit for a different time.

liaxim:
I have completely missed the point that this is a bug fix.

The gvrjassimpmodelloader changes have to go to the Demos repo.

liaxim:
Verified that loading the model from the SD card works - the bug is fixed.

Will add couple of comments you can choose to ignore. Other than the sample-to-Demos thing, this looks good to me.

GVRf/Framework/src/org/gearvrf/GVRResourceVolume.java
```diff
@@ -0,0 +1,111 @@
...
+        ANDROID_SDCARD,
+        LINUX_FILESYSTEM,
+        NETWORK,
+    }
+
+    protected GVRContext gvrContext;
+    protected VolumeType volumeType;
+    protected String defaultPath;
+
+    public GVRResourceVolume(GVRContext gvrContext, VolumeType volume) {
```
liaxim: We can't hide the introduction of a new publicly visible class but we can at least prevent the end users from instantiating it by making the ctors package-private.

GVRf/Sample/gvrjassimpmodelloader/src/org/gearvrf/jassimpmodelloader/JassimpModelLoaderViewManager.java
```diff
@@ -95,6 +96,14 @@ public void run() {
         mMainScene.addSceneObject(benchModel);
 
         rotateModel(astroBoyModel, 10f, astroBoyModelPosition);
+
+        // Model over network
+        String urlBase = "https://raw.githubusercontent.com/danke-sra/GearVRf/jassimp_network/GVRf/Sample/gvrjassimpmodelloader/assets/";
```
liaxim: I believe this needs to be updated to point to the (expected) corresponding location in the Demos repository.

danke-sra:
Removed sample from this patch, will submit to sample repo.

danke-sra:
Can this be checked in now?

thomasflynn:
i'll do another review when i get a chance to breathe tonight/tomorrow.

danke-sra:
Updated the patchset to make GVRResourceVolume ctors non-public

thomasflynn:
needs rebase

danke-sra:
Rebased and fixed one minor issue (missing @ in URL in case it contains user info). The test patch is here: https://github.com/gearvrf/GearVRf-Demos/pull/2

PR #348, opened 2015-12-15
Remove all the samples since they have been moved to a dedicated repo

liaxim:
https://github.com/gearvrf/GearVRf-Demos

Next step is to actually remove them from the history and
reduce the repo size.


liaxim:
@thomasflynn can I merge this?

PR #351, opened 2015-12-16
Fix occasional KSensor crash on docking/undocking

liaxim:
std::thread's dtor terminates if the thread is joinable
thus ensuring it is not

plus some precautions on the java side

Change-Id: Ia7821688ad7bacabb198827fc47d8a6f8bb5fe59

Conflicts:
	GVRf/Framework/src/org/gearvrf/GVRActivity.java

### DIFF
--- a/GVRf/Framework/jni/sensor/ksensor/k_sensor.cpp
+++ b/GVRf/Framework/jni/sensor/ksensor/k_sensor.cpp
@@ -95,7 +95,10 @@ KSensor::KSensor() :
 }
 
 KSensor::~KSensor() {
-    stop();
+    try {
+        stop();
+    } catch (const std::exception&) {
+    }
 }
 
 void KSensor::start() {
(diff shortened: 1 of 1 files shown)


danke-sra:
Code looks good. A little bit unusual this has to be fixed in the destructor.

Just elaborate a bit (not a huge concern): In the destructor, there is a potential lag (due to thread joining) and a possibility an exception can be thrown. So it is a bit dangerous and kinda rare to do. (any good example of this?)

liaxim:
Good point about the exception - catching exceptions now in the dtor.

I could have the stop method called before the dtor but the lag would still be there. Why would a dtor call that lags somewhat be different than a regular method call that lags somewhat?

danke-sra:
It is just desirable for destructors to complete fast. It can be achieved by interrupting the thread before joining. Destructors may be called in batch so the faster / leaner the better. However, it is not an issue here. So I am perfectly OK with the patch.

PR #352, opened 2015-12-16
Input Handling in GVRf

rahul27:
This patch lays out the groundwork for input handling
in GVRf.

The introduced GVRInputManager allows input events
to be dispatched to through the scene graph. This enables
the GVRSceneObjects to implement user interaction using the
SensorEvents generated from GVRBaseSensors.

The framework for now supports Mice and Gamepads.

External Input devices can be added using the
GVRCursorController class.

rahuldevan@gmail.com


rahul27:
@liaxim @thomasflynn this pull request is ready for review. I will push the sample to the demos repo once this is through.

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
```diff
@@ -704,6 +705,53 @@ public GVRSceneObject getChildByIndex(int index) {
...
+        return mSensor;
+    }
+
+    /**
+     * Remove the {@link GVRBaseSensor} from the object.
+     * 
+     * @return <code>true</code> if the sensor is successfully removed from the
+     *         object; <code>false</code> otherwise.
+     */
+    public boolean resetSensor() {
```
liaxim: We already have setSensor; this very thing could be achieved by setSensor(null) and people will have to deal with one api less. And it is pretty common "pattern".

liaxim:
On a first glance this looks ready. I think the sample pull request needs to be created now so people can try it out.

GVRf/Framework/src/org/gearvrf/GVRSceneObject.java
rahul27: Fixed this.

rahul27:
@liaxim Added the input sample here: https://github.com/gearvrf/GearVRf-Demos/pull/3.

The demo makes use of the GVRTextViewSceneObject which has been deprecated. But I can't find a sample that makes use of the new GVRTextView class. 

Would it be okay if I move to the new GVRTextView in a separate patch once I figure out how to use it?

liaxim:
Sure, no problem.

liaxim:
Fyi, GVRTextView usage:

- create a GVRTextView
```
        mTextView = new GVRTextView(this);
        mTextView.layout(0, 0, 500, 500);
        mTextView.measure(500, 500);
        mTextView.setBackgroundColor(Color.WHITE);
        mTextView.setTextColor(Color.BLACK);
        mTextView.setText("UI TextView");
```

- create a GVRViewSceneObject
```
        mViewSceneObject = new GVRViewSceneObject(gvrContext, mTextView,
                 <a_mesh>);
```

- add the scene object to the scene

liaxim:
Looks good to me.

In a (not so) future commit it would be desirable to destroy the controller's thread when a device is removed.

liaxim:
For other reviewers: the sample is intentionally more complicated; it uses depth in the event resolution thus the use of isColliding(). This is not the expected general use-case so onSensorEvent should be more straightforward usually.

rahul27:
@liaxim Sure I can stop the thread when the device is removed. Will add a patch to do the same.

I also plan to make another simple sample that demonstrates a simple use case, without isColliding().

thomasflynn:
@rahul27 are you planning on another patch (for stopping the thread)?

rahul27:
@thomasflynn will do it in a separate pull request.

PR #353, opened 2015-12-16
Simple scnene which can we make complex easily

dknaan:
Simple scnene for which we can add as many bauunies as we want to make it complex


liaxim:
@dknaan please move the sample to https://github.com/gearvrf/GearVRf-Demos

dknaan:
I close this pull request and create another one in https://github.com/gearvrf/GearVRf-Demos

PR #354, opened 2015-12-18
Unbind buffers before calling into the external renderer

liaxim:
Philippe pointed out the problem. VRTop finally is working pretty well.

### DIFF
--- a/GVRf/Framework/jni/shaders/material/external_renderer_shader.cpp
+++ b/GVRf/Framework/jni/shaders/material/external_renderer_shader.cpp
@@ -64,6 +64,12 @@ void ExternalRendererShader::render(
         scratchBuffer[5] = max_corner[2];
     }
 
+    //Oculus leaves buffers bound before calling us; SurfaceFlinger is ES 2.0
+    //so the following two lines ensure that SF doesn't end up using incorrect
+    //buffers
+    glBindBuffer(GL_ARRAY_BUFFER, 0);
+    glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, 0);
+
     TextureCapturer *capturer(render_data->get_texture_capturer());
     if (!capturer || !capturer->getAndClearPendingCapture()) {
         // Original rendering


liaxim:
The way it is, it is as close as possible to the place that has this special requirement. No other code in the framework cares about this thing. This is the reason I put it there.

danke-sra:
Ok. It's ok to keep it here. However, I've heard from another developer @rahul27 that this two lines are code are also missing for what he is doing. We then need to keep an eye for anything that may break by this...

liaxim:
@rahul27 should let us know what the problem is. What I do know is that without these two lines, the external renderer does not work on hybrid.

rahul27:
@danke-sra @liaxim For some of the stuff I am working on I wanted to make straight OpenGL calls at the application level i.e. inside onInit() and onStep(). In my case the draw calls failed and the glReadPixels() call always returned 0s, adding the two lines from this patch fixed that issue. To me it seems like this call should be made once the Oculus renderer is done with its drawing so that the buffers are reset for the application or any other external renderer to start drawing outside the GVRf context. 

You will find very simplified example of my OpenGL application here:

https://github.com/rahul27/GearVRf/tree/opengl

liaxim:
@rahul27 regardless of the issue, you shouldn't be making direct GL calls from the app. GVRF is supposed to suffice so let us know what GVRF doesn't do for you.

You can't assume anything about what is going on behind the scenes in GVRF so your GL calls may or may not work, the app might crash, etc.

thomasflynn:
more importantly, an application trying to make its own OpenGL calls cannot make any assumption about OpenGL state.  i haven't gotten to testing the libgdx plugin, but the same would apply there.  (if i understand correctly, libgdx makes its own gl calls).

thomasflynn:
ok, i've gotten a little lost in the discussion here.  merge or don't merge?

danke-sra:
I am neutral -- would prefer it to limit the scope of the leaked buffer binding (from Oculus). let others speak.

Vittalbabu:
It is not be good to allow an app. to dictate how the internal state should be modified. How do we arbitrate conflicting requests to set state?

I second Mihail's comment -- let's us know what is missing in GVRF and we will see if that can be added.

thomasflynn:
this pull-request is dealing with the external renderer and external renderer only.  

unfortunately, no matter we say, there will always be app writers who will make their own OpenGL calls whenever they can get their hands on something executing in a GL thread.  I've been through this before in other scenegraph projects.  We just need to make sure it is clear that the app writers cannot assume any OpenGL state is in the state they expect when they make their calls.

rahul27:
@thomasflynn @liaxim if developers aren't supposed to be making OpenGL calls at the app level then this fix works.

PR #359, opened 2015-12-23
GVRAnimation fix

mingsong:
--Add better timing for checking if an animation finishes.
--Add a api for resetting timing related fields for an animation if the animation is reused.

GearVRf-DCO-1.0-Signed-off-by: Mingsong Bi
mingsong.bi@samsung.com


thomasflynn:
looks good

liaxim:
Theoretically isFinished should be volatile. Possibly mElapsedTime and mIterations too.

PR #362, opened 2015-12-29
Make GVRRenderData.getMesh synchronous

liaxim:
Resolve a future mesh is present.


liaxim:
Avoiding a NPE in VRTop.

danke-sra:
Looks good.

PR #363, opened 2016-01-04
Fixes to the IO framework

rahul27:
1) Stop the event thread when no devices are found
2) Ability to set near and far depth
3) Cursor can now be attached to the camera rig
4) Fixed a bug when multple cursor controllers are
   returned for a given device
5) Fix NPE on SensorEvent.getCursorController()
6) Setting the initial position using setPosition() is fixed

rahuldevan@gmail.com

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/GVRBaseSensor.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRBaseSensor.java
@@ -35,6 +35,8 @@ import android.util.SparseArray;
  * 
  */
 public class GVRBaseSensor {
+    private static final String TAG = GVRBaseSensor.class.getSimpleName();
+    private static final float[] EMPTY_HIT_POINT = new float[3];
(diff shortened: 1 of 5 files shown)


thomasflynn:
@liaxim, could you look this over?  Thanks.

liaxim:
How does "3) Cursor can now be attached to the camera rig" work? How do I do that?

rahul27:
@liaxim in the input-sample add the scene object to the camera rig instead of the main scene. 

[Here](https://github.com/gearvrf/GearVRf-Demos/blob/master/input-sample/src/org/gearvrf/io/InputScript.java#L282) is a link to the line in the input-sample.

Before, the cursor used to only use its position and not the hierarchical position which caused issues when it was attached to the camera rig.

liaxim:
Ok, got it.

Code looks good and the sample works fine. From my perspective it can be merged.

PR #366, opened 2016-01-06
Hierarchical frustum culling

mingsong:
Implement hierarchical frustum culling

### DIFF
--- a/GVRf/Framework/jni/objects/scene_object.cpp
+++ b/GVRf/Framework/jni/objects/scene_object.cpp
@@ -28,9 +28,10 @@
 
 namespace gvr {
 SceneObject::SceneObject() :
-        HybridObject(), name_(""), children_(), visible_(true), in_frustum_(false),
-        query_currently_issued_(false), vis_count_(0), lod_min_range_(0), lod_max_range_(MAXFLOAT),
-        using_lod_(false), bounding_volume_dirty_(true) {
+        HybridObject(), name_(""), children_(), visible_(true), in_frustum_(
+                false), query_currently_issued_(false), vis_count_(0), lod_min_range_(
+                0), lod_max_range_(MAXFLOAT), using_lod_(false), bounding_volume_dirty_(
+                true) {
 
     // Occlusion query setup
 #if _GVRF_USE_GLES3_
@@ -55,7 +56,7 @@ void SceneObject::attachTransform(SceneObject* self, Transform* transform) {
     }
     transform_ = transform;
     transform_->set_owner_object(self);
-    dirtyBoundingVolume();
+    dirtyHierarchicalBoundingVolume();
 }
 
 void SceneObject::detachTransform() {
@@ -63,7 +64,7 @@ void SceneObject::detachTransform() {
         transform_->removeOwnerObject();
         transform_ = NULL;
     }
-    dirtyBoundingVolume();
+    dirtyHierarchicalBoundingVolume();
 }
 
 void SceneObject::attachRenderData(SceneObject* self, RenderData* render_data) {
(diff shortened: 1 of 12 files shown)


GVRf/Framework/jni/objects/scene_object.cpp
```diff
@@ -250,31 +250,217 @@ bool SceneObject::isColliding(SceneObject *scene_object) {
...
+        } else {
+            LOGD("FRUSTUM: mesh in frustum, render %s and all its children\n",
+                    name_.c_str());
+        }
+    }
+
+    return is_inside ? 2 : 1;
+}
+
+bool SceneObject::is_cube_in_frustum(float frustum[6][4],
```
rahul27: I get the following error here:

error: invalid conversion from 'const float (*)[4]' to 'float (*)[4]' [-fpermissive]

Does the frustum array need a 'const'?

GVRf/Framework/jni/objects/scene_object.cpp
```diff
@@ -250,31 +250,217 @@ bool SceneObject::isColliding(SceneObject *scene_object) {
...
+    }
+
+    return true; // fully inside
+}
+
+// frustumCull() return 3 possible values:
+// 0 when the bounding volume of the object is completely outside the frustum: cull it out and do not continue with its children
+// 1 when the bounding volume of the object is intersecting(or inside) the frustum but the object itself is not: cull it out and continue culling with its children
+// 2 when both the bounding volume and the mesh of the object is intersecting (or inside) the frustum: need to render it further and continue culling with its children
+int SceneObject::frustumCull(Camera *camera, const glm::mat4 &vp_matrix,
```
rahul27: This function does not use the VP matrix.

GVRf/Framework/jni/objects/scene_object.cpp
mingsong: It doesn't need to be const. It may be due to the compilation setting that complaining about the conversion, you can make a change to the code to make it compile or change the compilation setting.

GVRf/Framework/jni/objects/scene_object.cpp
```diff
@@ -250,31 +250,217 @@ bool SceneObject::isColliding(SceneObject *scene_object) {
...
-    if(render_data_ && render_data_->mesh()) {
-        bounding_volume_.expand(render_data_->mesh()->getBoundingVolume());
+    // Calculate the new bounding volume from itself and all its children
+    // 1. Start from its own mesh's bounding volume if there is any
+    if (render_data_ != NULL && render_data_->mesh() != NULL) {
+        // Future optimization:
+        // If the mesh and transform are still valid, don't need to recompute the mesh_bounding_volume
+        // if (!render_data_->mesh()->hasBoundingVolume()
+        // || !transform_->isModelMatrixValid()) {
+        mesh_bounding_volume.reset();
```
rahul27: The transform() call does a reset internally.

GVRf/Framework/jni/objects/scene_object.cpp
```diff
@@ -250,31 +250,217 @@ bool SceneObject::isColliding(SceneObject *scene_object) {
 }
 
 void SceneObject::dirtyBoundingVolume() {
```
rahul27: What do you think about renaming this function to reflect that the value could also be a HBV?

It might not be straightforward for developers to differentiate between this call and renderdata->mesh->boundingvolume() without looking at the source code.

GVRf/Framework/jni/objects/scene_object.cpp
mingsong: good call, thanks.

GVRf/Framework/jni/objects/scene_object.cpp
mingsong: right, it  should be cleaned up

GVRf/Framework/jni/objects/scene_object.cpp
mingsong: I agree, and will add more comments about mesh's BV and HBV.

mingsong:
All comments have been addressed, thanks.

PR #341, opened 2015-12-10
Adds GVRF widget Framework and sample

dattanand:
Adding GVRF widget (libgdx) framework implementation and sample code.

PR #367, opened 2016-01-07
Adding GVRF widget framework

dattanand:
this has all the changes for GVRF widget framework including rework done for review comments given in #341. only exception being the unlit shader is not removed since the widget scene object can be created using .obj meshes which have standard texture coordinates, also input dispatch for widget too behaves in the same way, to avoid using this shader would involve additional changes which dont seem right to me and would require swapping at every stage and involve more complication since input dispatch is little complicated


GVRf/Framework/src/org/gearvrf/plugins/widget/GVRWidgetPluginActivity.java
```diff
@@ -0,0 +1,757 @@
...
+
+    public GVRScript getScript() {
+        return mScript;
+    }
+
+    public int getTextureId() {
+        return mWidget.getTexId();
+    }
+
+    public void setCurrentScript(GVRScript script) {
```
liaxim: What is the relationship between this method and GVRActivity's setScript?

GVRf/Framework/src/org/gearvrf/plugins/widget/GVRWidgetPluginActivity.java
```diff
@@ -0,0 +1,757 @@
...
+import com.badlogic.gdx.utils.Array;
+import com.badlogic.gdx.utils.Clipboard;
+import com.badlogic.gdx.utils.GdxNativesLoader;
+
+/**
+ * This provides GVR (libGDX) widget lifecycle and context management and brings
+ * it together with GVR activity context. Base activity for GVR apps which use
+ * GVRWidgets
+ */
+public class GVRWidgetPluginActivity extends GVRActivity implements
```
liaxim: Please help me understand - is there a technical reason this separate activity exists or is it desirable for some other reason?

GVRf/Framework/src/org/gearvrf/plugins/widget/GVRWidgetInputDispatcher.java
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+    }
+
+    /**
+     * gets the normalised hit extent from top left relative to current scene
+     * object e.g. if pointer is at centre of the scene object, x and y
+     * coordinates would be 0.5f each.
+     */
+    private void gethit(GVRWidgetSceneObjectMeshInfo info) {
```
liaxim: Camel-case please

GVRf/Framework/src/org/gearvrf/scene_objects/GVRWidgetSceneObject.java
```diff
@@ -0,0 +1,157 @@
...
+        mesh.setTexCoords(textcoords);
+        mesh.setIndices(SIMPLE_INDICES);
+
+        GVRRenderData renderData = new GVRRenderData(gvrContext);
+        renderData.setMaterial(mat);
+        attachRenderData(renderData);
+        renderData.setMesh(mesh);
+    }
+
+    public GVRWidgetSceneObjectMeshInfo getmeshInfo() {
```
liaxim: Camel-case please

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -863,4 +869,15 @@ GVRRenderBundle getRenderBundle() {
     public GVRInputManager getInputManager() {
         return mInputManager;
     }
+
```
liaxim: Please use spaces instead of tabs (lines 872-873 start with a tab).

GVRf/Framework/src/org/gearvrf/plugins/widget/GVRWidgetPluginActivity.java
```diff
@@ -0,0 +1,757 @@
...
+
+    public void setPickedObject(GVRSceneObject obj) {
+        mInputDispatcher.setPickedObject(obj);
+    }
+
+    public GVRSceneObject getPickedObject() {
+        return mInputDispatcher.getPickedObject();
+    }
+
+    public void initializeWidget(GVRWidget widget) {
```
liaxim: What is the role of this method? Every new widget instance needs to go through it? Would you consider a thread pool (gvrf has one already) instead of spinning off a new thread on every call?

liaxim:
Is there a sample we can try with this pull request or subsequent pull requests are needed before something can be run? Thanks.

GVRf/Framework/src/org/gearvrf/plugins/widget/GVRWidgetPluginActivity.java
dattanand: it sets the script in base plugin activity from app, similar to setscript which does it for GVRActivity.

GVRf/Framework/src/org/gearvrf/plugins/widget/GVRWidgetPluginActivity.java
dattanand: yes, since Libgdx and its context management happens through its own activity, so this activity serves purpose for creating the libgdx widget inside GVRF.

GVRf/Framework/src/org/gearvrf/plugins/widget/GVRWidgetPluginActivity.java
dattanand: widget will be initialized only once at the launch of activity. i dont think i can use GVRF thread, the libgdx widget rendering happens in different thread and GL context

dattanand:
sample is also available, but i dont see samples anymore in master branch, where have they disappeared?

liaxim:
Please use https://github.com/gearvrf/GearVRf-Demos for samples. We moved them all there. Thanks.

dattanand:
sample updated in 
https://github.com/gearvrf/GearVRf-Demos/pull/14

GVRf/Framework/src/org/gearvrf/plugins/widget/GVRWidgetInputDispatcher.java
dattanand: rectified

GVRf/Framework/src/org/gearvrf/scene_objects/GVRWidgetSceneObject.java
dattanand: rectified

thomasflynn:
Needs rebase

GVRf/Framework/jni/shaders/material/unlit_fbo_shader.cpp
```diff
@@ -0,0 +1,125 @@
...
+#include "util/gvr_gl.h"
+
+namespace gvr {
+static const char VERTEX_SHADER[] = "attribute vec4 a_position;\n"
+        "attribute vec4 a_tex_coord;\n"
+        "uniform mat4 u_mvp;\n"
+        "varying vec2 v_tex_coord;\n"
+        "void main() {\n"
+        "  v_tex_coord.x = a_tex_coord.x;\n"
+        "  v_tex_coord.y = 1.0f - a_tex_coord.y;\n"
```
liaxim: I am getting this error because of the "1.0f":

```
01-18 08:30:20.041 15696 15745 E gvrf    : Could not compile shader 35633:
01-18 08:30:20.041 15696 15745 E gvrf    : 0:7: L0002: Undeclared variable 'f'
```

After removing the trailing 'f' things work.

dattanand:
rebase was done, no further review yet

danke-sra:
@dattanand I have one question. The jar files are very large (total about 20Mb). This can make all apps very large and slow to install. We probably should find a way to avoid linking it when an application doesn't need the feature.

Can you please figure out a way to do that?

One thing I can think of is
1) Write a wrapper to load GDX jar files dynamically when the application uses it
2) Check in those JAR files not in the "libs" directory but in an "optional/libgdx" or "extra/libgdx" directory, so users can copy needed JAR files into their apps when they need them.

dattanand:
@danke-sra 
I added them based on review comments and made changes in make file to add them as part of framework build, if they are not required by default, I will revert those changes and add them in extra, but how do I go about loading them dynamically in code?

danke-sra:
To load gdx jar dynamically into framework, you can refer to this method (or similar): http://stackoverflow.com/questions/6857807/is-it-possible-to-dynamically-load-a-library-at-runtime-from-an-android-applicat

I haven't got too much time to look into this. Please let me know if you see other issues.

liaxim:
This discussion brings me back to my past suggestion to have this done as a gvrf extension library. Something that is not part of the core fw but if you need it, you just add the additional extension jar to your apk. If the gdx support has too large of a footprint I believe it is pretty much mandatory.

danke-sra:
@liaxim Can you provide a simple example for doing an extension to him?

liaxim:
What I have in mind is rather simple - create a separate jar that depends on Framework.jar and provides all you need for the libgdx integration. Regular GVRF apps use Framework.jar. Those which need libgdx include one more jar and get access to an expanded API that supports libgdx (and become larger in size).

liaxim:
@dattanand Any thoughts about making this something like an extension? Needs rebase also.

dattanand:
I have been thinking over it since GVRF as such is not designed to plug in your extensions, So I had written little sort of plugin, yes but the jar has to be part of it, Conditional compilation sort of think would have helped I guess, I will work something out, meanwhile If you have any suggestion, please let me know.

dattanand:
@liaxim I am thinking of implementing some kind of class loader which will load the classes from jar file present at specified location since you dont want it to be part of apk and I dont know how to dynamically build apk with or without jar included, I am not sure how to go about it but is this the way I can proceed?

liaxim:
@dattanand Class loaders and loading jars dynamically from somewhere is something we should ideally avoid.

What I have in mind is really simple:
1. Make the changes in such a way that doesn't require modifying the core fw. 
2. Package your classes and .so as an "extension" library.
3. Whomever needs the extended functionality drops this "extension" library in their libs folder.

Now, I over-simplified 1. but that should be the goal. There might be some API changes required in the fw but these should be as minimal as possible and generalized.

I will try to go over your change and give your more concrete suggestion on how to repackage this.

danke-sra:
For the GVRPlugin concept, I've got some suggestions based on recent framework changes.

1) The plugin is a good idea, but it's hard to design the interface for a general-purpose plugin. Say, some plugins need to be initialized upon onInit() callback, some provide shaders, and some provide input support for some devices, and some handle transforms (physics engine, animation), etc. So, we can consider the event system as an interface for plugins, and it supports loose coupling, and on-demand loading.

2) The widget/libgdx plugin modifications to GVRScript and GVRViewManager can be refactored into a IScriptEvent listener, and added to the GVREventReceiver of your GVRScript. This can be done in your GVRWidgetPluginActivity.

I think in this way this change won't require any framework changes any more. Your new activity class GVRWidgetPluginActivity would have wrapped all needed changes, and applications can derive from it instead of GVRActivity get all new features. Similar plugins can use the same pattern (providing a subclass of GVRActivity). Other plugins may be added as an event listener in similar ways.

Then, the widget package can be compiled as a JAR file, and the source can be structured as a library. When needed, users can add this library or JAR file, and change their GVRActivity to GVRWidgetPluginActivity to start using the new features.

Please let us know how that works. Thanks for your thoughts and contribution.

thomasflynn:
Actually, i would prefer if a user didn't have to subclass from a different activity if it wasn't absolutely necessary.

dattanand:
Thanks for your inputs, Let me go through the new architecture and get back to you if I need some help.

dattanand:
I have updates the framework according to suggestions, the widget framework including activity and scene object go into seperate plugin  now, libgdx jars are part of that plugin. So framework project and this plugin project have to be used by app. this implements the plugin concept and has minimal effect on framework, of course setting the plugin in the framework and initializing it only remains in framework with required shader related changes, rest all make their way. kindly check and let me know your comments.

GVRf/Framework/jni/shaders/material/unlit_fbo_shader.cpp
```diff
@@ -0,0 +1,125 @@
...
+    glUniform1i(u_texture_, 0);
+    glUniform3f(u_color_, color.r, color.g, color.b);
+    glUniform1f(u_opacity_, opacity);
+
+    glBindVertexArray(mesh->getVAOId(Material::UNLIT_FBO_SHADER));
+    glDrawElements(render_data->draw_mode(), mesh->indices().size(),
+            GL_UNSIGNED_SHORT, 0);
+    glBindVertexArray(0);
+#else
+
```
thomasflynn: no need for ES2 code anymore.  you can remove the #if 1 #else statements and remove the ES2 code.

NolaDonato: Does this mean I can use uniform buffer objects in my shaders?

Nola

thomasflynn: ES3 is our required minimum.

dattanand:
I have changed the plugin interface which is now loosely coupled with framework, the only thing being is initialization of plugins before drawing, though I have one question, The app does not seem to behave as it used to with older framework, I know there are so many changes went in framework, So I just wanted to know any significant change if any especially VrApiActivityHandler which I believe was not part of framework earlier, to sum it up, let me clarify what I am doing,
1.  I draw widget content on GLSurfaceView which is drawn on FBO in the shared context and I create texture from the FBO and place it on scene object, Now all this was working well, but with latest framework, things are little haywire.
2. I am attaching screenshot which tells me my GLSurfaceview content is being rendered on top of original content, I am little confused as to what might be happening, I also saw this log which I find is related to the problem

12-01 20:49:45.387: E/libEGL(23608): call to OpenGL ES API with no current context (logged once per thread)
12-01 20:49:45.427: W/VRSurfaceManager(23608): VRSurfaceManager : setFrontBuffer
12-01 20:49:45.437: W/TimeWarp(23608): eglGetCurrentSurface( EGL_DRAW ) != eglPbufferSurface

please help me figure this out.

liaxim:
It looks much better. There seems to be one last sticking point; it will be hard to get this in with the requirement to derive from GVRWidgetPluginActivity. There is absolutely no way to avoid that?

Please update the sample in the demo repo. I made trivial changes to the imports and it seems to work. Is that all? Thanks.

Looks like it doesn't work on S7; I get black-ish quads without content.

I am seeing the problem from the screenshot. Will work on figuring it out.

liaxim:
One time crash while starting and stopping the demo app:

```
03-29 10:39:29.480 21308 21308 E AndroidRuntime: FATAL EXCEPTION: main
03-29 10:39:29.480 21308 21308 E AndroidRuntime: Process: org.gearvrf.GVRWidgetViewer, PID: 21308
03-29 10:39:29.480 21308 21308 E AndroidRuntime: java.lang.RuntimeException: Unable to pause activity {org.gearvrf.GVRWidgetViewer/org.gearvrf.widgetViewer.GVRWidgetViewer}: java.lang.NullPointerException: Attempt to invoke virtual method 'boolean com.badlogic.gdx.backends.android.AndroidGraphics.isContinuousRendering()' on a null object reference
03-29 10:39:29.480 21308 21308 E AndroidRuntime: 	at android.app.ActivityThread.performPauseActivity(ActivityThread.java:4626)
03-29 10:39:29.480 21308 21308 E AndroidRuntime: 	at android.app.ActivityThread.performPauseActivity(ActivityThread.java:4585)
03-29 10:39:29.480 21308 21308 E AndroidRuntime: 	at android.app.ActivityThread.handleRelaunchActivity(ActivityThread.java:5389)
03-29 10:39:29.480 21308 21308 E AndroidRuntime: 	at android.app.ActivityThread.access$1200(ActivityThread.java:222)
03-29 10:39:29.480 21308 21308 E AndroidRuntime: 	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1837)
03-29 10:39:29.480 21308 21308 E AndroidRuntime: 	at android.os.Handler.dispatchMessage(Handler.java:102)
03-29 10:39:29.480 21308 21308 E AndroidRuntime: 	at android.os.Looper.loop(Looper.java:158)
03-29 10:39:29.480 21308 21308 E AndroidRuntime: 	at android.app.ActivityThread.main(ActivityThread.java:7294)
03-29 10:39:29.480 21308 21308 E AndroidRuntime: 	at java.lang.reflect.Method.invoke(Native Method)
03-29 10:39:29.480 21308 21308 E AndroidRuntime: 	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1230)
03-29 10:39:29.480 21308 21308 E AndroidRuntime: 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1120)
03-29 10:39:29.480 21308 21308 E AndroidRuntime: Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'boolean com.badlogic.gdx.backends.android.AndroidGraphics.isContinuousRendering()' on a null object reference
03-29 10:39:29.480 21308 21308 E AndroidRuntime: 	at org.gearvrf.widgetplugin.GVRWidgetPluginActivity.onPause(GVRWidgetPluginActivity.java:339)
03-29 10:39:29.480 21308 21308 E AndroidRuntime: 	at org.gearvrf.widgetViewer.GVRWidgetViewer.onPause(GVRWidgetViewer.java:310)
03-29 10:39:29.480 21308 21308 E AndroidRuntime: 	at android.app.Activity.performPause(Activity.java:7033)
03-29 10:39:29.480 21308 21308 E AndroidRuntime: 	at android.app.Instrumentation.callActivityOnPause(Instrumentation.java:1339)
03-29 10:39:29.480 21308 21308 E AndroidRuntime: 	at android.app.ActivityThread.performPauseActivity(ActivityThread.java:4612)
03-29 10:39:29.480 21308 21308 E AndroidRuntime: 	... 10 more
```

liaxim:
Taking back my statement on S7 - it actually works and works better than S6.

dattanand:
@liaxim Initially I had started it without having to derive from GVRWidgetPluginActivity, but since it required the libgdx runtime and associated contex which is tightly coupled with android activity lifecycle so it was diffcult to bypass. I will check try to fix this crash and update the demo. Thanks!!

liaxim:
About the viewport thing, this is what I found. Add the following code to the start of GVRWidgetPluginActivity.init:

```java
    private void init(ApplicationListener listener,
            AndroidApplicationConfiguration config, boolean isForView,
            EGLContext sharedcontext) {
        config.resolutionStrategy = new FillResolutionStrategy() {
            public MeasuredDimension calcMeasures(int widthMeasureSpec, int heightMeasureSpec) {
                return new MeasuredDimension(2560, 1440);
            }
        };
```

The viewport is correct afterwards; not suggesting this is *the* fix but I hope it helps identifying the root cause. I think the gdx widgets have gone missing after this though.. Let me know if you still think there is a problem on the framework side.

GVRf/Framework/src/org/gearvrf/GVRScript.java
```diff
@@ -303,4 +313,25 @@ SplashScreen createSplashScreen(GVRViewManager viewManager) {
         onSplashScreenCreated(splashScreen);
         return splashScreen;
     }
+
+    public void addPlugin(GVRPlugin plugin) {
```
danke-sra: @dattanand Do you think there is another way of adding in the plugin? Do you think we can only override onInit() to perform needed initialization?

dattanand:
@liaxim I have updated the new demo and fixed the crash, the app works fine when I commented this in VrApiActivityHandler 

holder.setFormat(PixelFormat.TRANSLUCENT);

I dont know the root cause but please check what might be causing this.

dattanand:
@danke-sra In my case the plugin is added at runtime and the initialization happens in the plugin class, the plugin has its own init() function which does initialization before oninit() is called.

danke-sra:
@dattanand I see your concern. After we add IActivityEvents, these issues can be solved easily. The patch is pending https://github.com/Samsung/GearVRf/pull/479.

When it is merged, it helps solve two issues:

1) Instead of overriding GVRActivity, you can handle activity lifecycle by implementing your IActivityEvents, and add it using gvrActivity.getEventReceiver().addListener(...);

2) You may add an onInit() handler by implementing IScriptEvents, and add it to the GVREventReceiver of the GVRScript.

Of course, GVRScript belongs to the application. Framework might need to override IActivityEvents.onSetScript() to add a listener to it. I've tried the above scheme, and it seems to work. Please go ahead to resolve the resolution issue, and kindly consider the architectural scheme here. It will make the patch pretty ready for merging.

dattanand:
@danke-sra I am not sure how can I avoid overriding GVRActivity, there is a reason for it and I have explained it as well. Also the plugins core initialisation should not be part of oninit() I believe.

danke-sra:
What's the reason you have to override GVRActivity, except that you need the event callbacks?

dattanand:
@danke-sra currently I need it to use script in my application, otherwise in plugin I can derive from Activity and use IActivityCallback but my app will always derive from GVRWidgetPluginActivity, see I cannot avoid it since I need libgdx context in my application, this is what I can modify, actually to see the apps dervice from GVRActivity, in my case this happens in plugin, thats the only difference. please let me know if you any other thoughts.

dattanand:
@liaxim please let me know if you found anything regarding the issue or do I need to make any change in my code to make it work

danke-sra:
@dattanand Thanks for the information. I have update the events so that it can support the requirements of the plugin. Actually, I understand the application needs libgdx context. The reason we don't want a framework class to derive from GVRActivity is that we cannot use two subclasses of GVRActivity together, so we prefer composition here. Here is some results on my side:

1) GVRWidgetPluginActivity could be refactored into GVRWidgetPlugin which can be added to application Activity which derives from GVRActivity. 

2) Now I understand the widget needs to be initialized before application's onInit. I have added an event onEarlyInit to perform tasks like widget initialization. Do you think we could use it instead of the GVRPlugin interface? From current code, I see only GVRPlugin.init is used, other methods in GVRPlugin are unused. So I think we might need to wait before we can define a complete and general interface for plugins.

My temporary patches are here for your reference: https://github.com/danke-sra/GearVRf/commit/945ca4ccb915fb85f8171b1abbe24c89a5464b9d

The demo app patch is here: https://github.com/danke-sra/GearVRf-Demos/commit/053395beb3f0ddec63c7464b3bad45375da042a8

Please let me know if you are OK for using the new events?

dattanand:
@danke-sra I will have a look at your changes but I need to fix the issue I mentioned earlier in screenshot, the framework change fixes this issue but I am not sure whether that is the correct one and would it cause any side effect.

dattanand:
holder.setFormat(PixelFormat.TRANSLUCENT); in VrApiActivityHandler(), seems it is added in recent changes and was not there earlier when VrApiActivityHandler was not present.removing this makes my app work correctly, can it be removed, if not how can I fix the issue in my code? I need some help with this.

dattanand:
I have resolved the weird display issue in my app, using suggestion by @liaxim , I made changes in widgets to accommodate it, I will push it in demo branch, currently due to pathetic firewall problems I am not able to push it, after this I would try to refactor it according to patches provided by @danke-sra and update. thanks!

liaxim:
@dattanand Is holder.setFormat(PixelFormat.TRANSLUCENT) still a problem? I think it is necessary but I can study further.

dattanand:
@liaxim it is the issue but as you suggested, I changed the resolution of my view, and it seems to work properly without removing holder.setFormat(PixelFormat.TRANSLUCENT) ; I have updated the app in demo branch with the changes, thanks!

GVRf/Framework/jni/shaders/material/unlit_fbo_shader.cpp
```diff
@@ -0,0 +1,100 @@
...
+    glUseProgram(program_->id());
+
+    glUniformMatrix4fv(u_mvp_, 1, GL_FALSE, glm::value_ptr(mvp_matrix));
+    glActiveTexture (GL_TEXTURE0);
+    glBindTexture(texture->getTarget(), texture->getId());
+    glUniform1i(u_texture_, 0);
+    glUniform3f(u_color_, color.r, color.g, color.b);
+    glUniform1f(u_opacity_, opacity);
+
+    glBindVertexArray(mesh->getVAOId(Material::UNLIT_FBO_SHADER));
```
liaxim: Github shows no merge problems but mesh->getVAOId no longer needs arguments.

dattanand:
I have now updated the widget plugin and the demo app with the activity events according to patches by @danke-sra, seems to work fine, I have updated the demo as well, now the code is as per the expectations. it does not intrude the framework, no intermediate activity, no additional overhead of jar file, please check the changes and let me know if anything more is required to get it pulled. thanks!!

GVRf/Framework/project.properties
```diff
@@ -11,6 +11,6 @@
 #proguard.config=${sdk.dir}/tools/proguard/proguard-android.txt:proguard-project.txt
 
 # Project target.
-target=android-19
+target=android-20
```
liaxim: Why the change from android-19 to android-20?

GVRf/Framework/src/org/gearvrf/GVRActivity.java
```diff
@@ -349,17 +349,16 @@ public boolean dispatchGenericMotionEvent(MotionEvent event) {
 
     @Override
     public boolean dispatchTouchEvent(MotionEvent event) {
+
```
liaxim: Please cleanup these unnecessary changes to GVRActivity.

GVRf/Framework/src/org/gearvrf/GVRScript.java
```diff
@@ -15,9 +15,15 @@
 
 package org.gearvrf;
 
+import java.util.ArrayList;
```
liaxim: Please cleanup these unnecessary changes to GVRScript.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -697,13 +697,19 @@ public void beforeDrawEyes() {
             }
 
             try {
+
```
liaxim: Please cleanup these unnecessary changes to GVRViewManager.

GVRf/Framework/src/org/gearvrf/plugin/GVRPlugin.java
```diff
@@ -0,0 +1,45 @@
...
+package org.gearvrf.plugin;
+
+
+
+import android.os.Bundle;
+
+/**
+ * Plugin interface to be implemented by GVRF plugins
+ */
+public interface GVRPlugin {
```
liaxim: Please remove the unused GVRPlugin class.

liaxim:
Ok, this is something else altogether. There is a very minor cleanup to do. We have to wait for the multilight pull request to be merged first though. Shouldn't have much of an impact on yours but you could evaluate it by trying to rebase yours on top of the multilight pull request.

dattanand:
@liaxim oops, some merge issues I overlooked, I will correct it.

liaxim:
Please squash the commits.

GVRf/Framework/jni/engine/renderer/renderer.cpp
```diff
@@ -817,6 +817,10 @@ void Renderer::renderMaterialShader(RenderData* render_data,
                 shader_manager->getLightMapShader()->render(mvp_matrix,
                         render_data, curr_material);
                 break;
+			case Material::ShaderType::UNLIT_FBO_SHADER:
+                shader_manager->getUnlitFboShader()->render(
+                        mvp_matrix, render_data, curr_material);
```
liaxim: "mvp_matrix" should be "uniforms.u_mvp" post-multilight. Needs to be updated otherwise we will merge something that does not compile.

liaxim:
Please move the Widget_plugin under Extensions folder (GVRf/Extensions/Widget_plugin). There is a plan for other extension libraries and they will reside there. Thanks.

dattanand:
well due to being git novice and moreover intermittent git connection problems, I am facing issues creating clean commit from my branch, I have correctly updated the demo though. here I ended up creating new pull request #503 for the same from new branch (accidently) which is clean commit but I wanted to remove libgdx jars from framework directory in that commit which I am not able to do, I will create fresh commit in this pull request if required. thanks!!

liaxim:
@dattanand Can we close this obsolete pull request?

dattanand:
yes, this one can be closed

PR #370, opened 2016-01-08
Fixed compilation error on unused sphereInFrustum() function

rahul27:


rahul27:
@mingsong I got a compilation error after pulling in your changes. Could you verify this fix is correct?.

I had to change the function signature to match the header.

mingsong:
Right, it is fixed.

PR #371, opened 2016-01-08
Script framework

danke-sra:
- Modify GVRViewManager.java to invoke events via GVREventManager
- Add GVREventManager for delivering events (to both attached scripts
  and Java event handling methods)
- Add GVRScriptManager to attach/detach scripts
- Add GVRScriptFile to hold script text

Mainly added a way to define what events an object can handle by using the IEvents tag and sub-interfaces). Then, the event manager and script manager are added to deliver events to scripts, as well as existing handling methods.

The source code for JnLua is also added as part of this patchset. Please let me know if there is a better way to handle that.

The changes can be tested using the patch https://github.com/gearvrf/GearVRf-Demos/pull/11


danke-sra:
Added missing .so due to .gitignore

The basic scripting framework is kinda complete. Please review and comment if anything needs to be done before merging. I will look at other events to which scripts can be attached in separate patches.

rahul27:
Did you mean to add the sample "GVRf/Sample/gvr-lua" in this pull request?

danke-sra:
Removed sample from this pull request

GVRf/Framework/src/org/gearvrf/GVRActivity.java
```diff
@@ -391,6 +391,10 @@ public void unregisterView(View view) {
         mRenderableViewGroup.removeView(view);
     }
 
+    public GVRContext getGVRContext() {
```
rahul27: Is this method really necessary as it is the GVRActivity that instantiates the View Manager. The context in essence should always be available to the activity from an applications point of view after onInit().

danke-sra: I couldn't find a way to get context from sub-class of GVRActivity. Any idea?

rahul27: I don't think I fully understand this function. The context should be available irrespective of the hierarchy. It is the GVRActivity that calls setScript. The script in turn could be used to get the context. Do  you have an example of where this is used in a subclass? maybe that would give a better idea.

danke-sra: You can see why getGVRContext() is needed in this file: https://github.com/danke-sra/GearVRf-Demos/blob/script-dev/gvr-lua/src/org/gearvrf/sample/gvrlua/GearVRLua.java

I agree that GVRContext should be made available regardless of hierarchy. However, it was assumed that it is not needed in sub-class of GVRActivity (which owns GVRContext). This just corrects this issue.

rahul27: Oh I see now, wouldn't this method break the intended life cycle for the Context? i.e. the context would only be wholly complete and correct after onInit(). It is fine in your case since you only use it to access the script manager which has been initialized by the constructor.

danke-sra: I see the idea. As we discussed, it is a good idea to protect against scene initialization before onInit, but because context is needed for many other things, such a protection cannot be easily enforced. In the worst case, hiding context from subclass of GVRActivity could only make the subclass redundant -- since it cannot do anything without the context. We are currently only use it to specifying an XML -- which could be done in the manifest file as well.

In Android, context is not hidden from subclass of Activity either. So let's bear with it. I would be great rethink the design and resolve all these issues simultaneous but we could do it later.

thomasflynn:
quick question:  do you expect we'll be modifying the source of jnlua as we often end up doing with jassimp?  could we package it up as a jar?  it seems as if most of the files in this check-in are the jnlua source files.

danke-sra:
We probably don't need to modify the Java files of JNLua very often. But there is some stability issue in the JNI code of JNLua. We might need to make small fixes like what we did to jassimp.cpp.

thomasflynn:
ok, thanks.  that's what i was looking to understand.

danke-sra:
Removed JNLua source and checked in libraries. Those libraries can be maintained here: https://github.com/danke-sra/jnlua-android

Please see if there is other changes before this patch can be merged.

GVRf/Framework/src/org/gearvrf/GVREventManager.java
```diff
@@ -0,0 +1,96 @@
+package org.gearvrf;
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
+import org.gearvrf.script.GVRScriptFile;
+import org.gearvrf.script.IScriptable;
+
```
thomasflynn: need to add javadoc for the class and all public methods.

GVRf/Framework/src/org/gearvrf/script/GVRLuaScriptFile.java
```diff
@@ -0,0 +1,33 @@
+package org.gearvrf.script;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+import org.gearvrf.GVRContext;
+
+public class GVRLuaScriptFile extends GVRScriptFile {
```
thomasflynn: javadoc

GVRf/Framework/src/org/gearvrf/script/GVRScriptFile.java
```diff
@@ -0,0 +1,84 @@
...
+import java.io.InputStream;
+import java.io.InputStreamReader;
+
+import javax.script.Bindings;
+import javax.script.ScriptContext;
+import javax.script.ScriptEngine;
+import javax.script.ScriptException;
+
+import org.gearvrf.GVRContext;
+
```
thomasflynn: javadoc

GVRf/Framework/src/org/gearvrf/script/GVRScriptFile.java
```diff
@@ -0,0 +1,84 @@
...
+public abstract class GVRScriptFile {
+    private static final String TAG = GVRScriptFile.class.getSimpleName();
+    GVRContext mGvrContext;
```
thomasflynn: is this deliberately package visible?  (and not protected?)

GVRf/Framework/src/org/gearvrf/script/GVRScriptManager.java
```diff
@@ -0,0 +1,151 @@
...
+import org.gearvrf.GVRAndroidResource;
+import org.gearvrf.GVRContext;
+import org.gearvrf.utility.Log;
+
+import com.naef.jnlua.script.LuaScriptEngineFactory;
+
+
+/**
+ * Script manager
+ */
```
thomasflynn: more descriptive javadoc please :)

GVRf/Framework/src/org/gearvrf/GVREventManager.java
```diff
@@ -0,0 +1,96 @@
+package org.gearvrf;
```
thomasflynn: need copyright header

GVRf/Framework/src/org/gearvrf/IEvents.java
```diff
@@ -0,0 +1,8 @@
+package org.gearvrf;
```
thomasflynn: copyright header

GVRf/Framework/src/org/gearvrf/IScriptEvents.java
```diff
@@ -0,0 +1,10 @@
+package org.gearvrf;
```
thomasflynn: copyright header

GVRf/Framework/src/org/gearvrf/script/GVRLuaScriptFile.java
```diff
@@ -0,0 +1,33 @@
+package org.gearvrf.script;
```
thomasflynn: copyright header

GVRf/Framework/src/org/gearvrf/script/GVRScriptFile.java
```diff
@@ -0,0 +1,84 @@
+package org.gearvrf.script;
```
thomasflynn: copyright header

GVRf/Framework/src/org/gearvrf/script/GVRScriptManager.java
```diff
@@ -0,0 +1,151 @@
+package org.gearvrf.script;
```
thomasflynn: copyright header

GVRf/Framework/src/org/gearvrf/script/IScriptable.java
```diff
@@ -0,0 +1,4 @@
+package org.gearvrf.script;
```
thomasflynn: copyright header

thomasflynn:
also, libjnlua.so is missing from the check-in.

danke-sra:
Added the missing so file and added copyright headers and javadoc.

GVRf/Framework/src/org/gearvrf/GVRActivity.java
liaxim: getGVRContext() needs javadoc describing when it makes sense to be called; even better, in addition to that it should throw an IllegalStateException if it is called before setScript().

GVRf/Framework/src/org/gearvrf/GVREventManager.java
```diff
@@ -0,0 +1,155 @@
...
+        if (script == null)
+            return;
+
+        script.invokeFunction(eventName, params);
+    }
+
+    private void invokeMethod(Object target, Method method, Object[] params) {
+        try {
+            method.invoke(target, params);
+        } catch (IllegalAccessException e) {
```
liaxim: I believe this try-catch block should be catching all Exceptions as a precaution against RuntimeExceptions escaping from the invoked method.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -667,7 +675,9 @@ public void beforeDrawEyes() {
 
             doMemoryManagementAndPerFrameCallbacks();
 
-            mScript.onStep();
+            GVRViewManager.this.getEventManager().sendEvent(
```
liaxim: This is a pretty slow operation - 12-14ms most of times on my S6 (even for a script without onStep). I can see some things could be cached but I doubt it can ever be very fast considering what is involved.

I suggest moving the script execution off the GL thread. From the script perspective the threading shouldn't be apparent anyway. This may mean a lot of work on the GVRF side - we could leave it for later as long as it is ensured that no overhead is incurred for the use-case where no scripts are involved.

liaxim:
I would have preferred the scripting to be well-hidden from end-users (why would a normal dev need getEventManager/getScriptManager from GVRContext?). The EventManager could have been more conservative and be part of the ScriptManager (generalized later as needed; maybe never, who knows). Assuming this is just the first iteration though it is ok.

The only thing I do believe needs immediate attention is the overhead on every frame.

GVRf/Framework/src/org/gearvrf/GVREventManager.java
danke-sra: invokeMethod is not designed to do anything more than just a method call. If the method throws, the caller should catch it.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
danke-sra: Good to know the profiling result. We need to use this indirection for event calls. In typical event-based systems, the latency may event be longer. I doubt moving script off GL thread works, because threads need to manipulate scene / scene objects. Caching would be the right way to go.

danke-sra:
I can kinda agree on GVREventManager but GVRScriptManager needs to be visible, because users need to attach scripts.

danke-sra:
Just for clarification, I do think GVREventManager can be public, because when we have a nice event delivery and scripting infrastructure, app developers may choose to add their custom event sets, and attach their own scripts to handle their custom events. Why should we prevent extensibility?

danke-sra:
I ran the MeshAnimationSample with this patch, and it still got 60 FPS. If the onStep() call is delayed 12-14ms, we shouldn't get 60 FPS. Maybe the profiler interacted with classloader/reflection. There is no immediate performance issue we need to worry about. I will see if optimization is needed for that in a later patch.

GVRf/Framework/src/org/gearvrf/GVRActivity.java
thomasflynn: this was a good catch.  somehow i missed the lack of javadoc earlier.  also the suggestion about adding an exception if called before setScript() is a good one.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
thomasflynn: so, what happens inside this function when there are no scripts?  (i tried skimming it, but am too tired at the moment).  what line number inside sendEvent() does it exit out?

danke-sra: It exits from line 88 in GVREventManager.java. If the function is not defined in the script, GVRScriptFile.invokeFunction(...) internally receives a ScriptException and then it returns false. Since nothing is wrong in this case, the function sendEvent(..) exits normally.

GVRf/Framework/src/org/gearvrf/script/GVRScriptFile.java
```diff
@@ -0,0 +1,151 @@
...
+
+    /**
+     * Loads a script into the {@GVRScriptFile} object.
+     *
+     * @param inputStream
+     *     The input stream from which to load the script.
+     * @throws IOException
+     */
+    public void load(InputStream inputStream) throws IOException {
+        BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream));
```
rahul27: close the reader once done?

rahul27:
@danke-sra , I've gone through the pull request and here is what I think:

I look at the GVREventManager with respect to the scripting support. I do not see the benefit of using the event manager when my app is purely java (since the reflection would be an overhead).

If we do plan to use the event manager only with scripting support then will there be can instance where IEvent is not of type IScriptable?

GVRf/Framework/src/org/gearvrf/GVREventManager.java
liaxim: We are invoking something user-supplied; being defensive in this case is generally a good practice. Your call.

GVRf/Framework/src/org/gearvrf/script/GVRScriptFile.java
danke-sra: Good question. I checked into this a bit, and can confirm that "when the reader is closed, it will close the underlying InputStream". This is a side-effect I didn't want. The creator of the InputStream should close it instead?

danke-sra:
I understand that there is no benefit of using reflection when there is no script -- but there is benefit when there is script. We want the two cases to share API, because the module which sends out events doesn't know whether the listener is a script or a Java method, and we don't to distinguish the two cases for the event-producing side. This of cause is not saying this is the only mechanism, but it may be the one with least extra code when more events are added. Suppose reflection is not used, with each new event, we need to add a function to call it, or add an event class to represent it. It is not easy to scale.

Please let me know if this makes sense, and also would like to consider alternatives, as long as scalability and complexity is at the same level.

danke-sra:
For the interfaces IEvents and IScriptable, I can confirm that some classes can only implement IEvents (or subclass) but not IScriptable. This is just a developer's choice. There is no cost of also implementing IScriptable because it is just a tag interface.

With this said, hopefully we don't jump to the conclusion that we don't need event manager for classes like that, we want an event delivery mechanism that is adaptive and uniform for both java handlers and script handlers, and want framework / event developers to have no worries about that. Further, we recommend all external events to be scriptable, so script users can access full functionality as Java code can do.

liaxim:
**Why** do we need "an event delivery mechanism that is adaptive and uniform for both java handlers and script handlers"? Can't we get the scripting in and in decent shape first before generalizations come into play?

Also **why** we would want "script users can access full functionality as Java code can do"? I thought the scripting would present an even more abstracted (though somewhat limited) view of what gvrf does (eventually). If things are the same as what we already do on the Java side but the language syntax is slightly different then what is the benefit of scripting? I mean why would I use the scripting instead of Java if it is no easier?

thomasflynn:
It looks like in this case, code doesn't speak louder than words.  Let's meet up and discuss this in person.

danke-sra:
Some profiling results for your information:

Javascript (no compilation)
    onStep present                           .6 - .7 ms
    onStep absent                             1 ms

Lua (no compilation)
    onStep present                            5 ms
    onStep absent                             5 ms

When script is not attached, onStep time is .05 ms (only Java reflection).

Optimization ideas:

1) We can optimize the time to 0 ms when the handler function is undefined (for both JS and Lua).
2) Jaavscript runs pretty fast. If we need to make it faster, we may try V8 engine.
3) Lua can be made faster by compiling the script. Since it doesn't generate Java bytecode, Android doesn't prevent it.

danke-sra:
Because none of existing code binds a script and the patch adds no significant overhead, is this patch ready for checking in now?

liaxim:
I don't object having it merged.


---
For the record:

- This should have been kept strictly to enabling scripting and in my opinion it is too early to come up with an EventManager that encompasses both regular Java and scripts.

- Without any script the overhead is negligible but unnecessary. Assuming Java and scripts should share the same events and that Java even needs "events" is premature.

- With the lua script from the lua demo app this prints out 10-13ms most of the times.
```
        long start = SystemClock.uptimeMillis();
        try {
            mLocalEngine.eval(statement);
        } catch (ScriptException e) {
            return false;
        } finally {
            start = SystemClock.uptimeMillis() - start;
            Log.i("tag", "eval : " + start);
            removeBindings(localBindings, params);
        }
```
An exception gets thrown always (all that it says is "deallocated").

- In my opinion no scripts should be executed on the GL thread; for performance, safety, simplicity.

danke-sra:
Just for record as well, in a healthy system, time critical threads don't callback to unpredictable user code, unless the code is equally high priority / critical. onInit() and onStep() are things that define application behavior, and shouldn't be called from GL thread. This issue is unrelated to scripting. Scripting may add a bit overhead, but other things that fit in onInit() and onStep() can be equally unpredictable, even implemented in Java.

I think adding more features presents a stress test to the design of the system. Instead of worrying about overheads added by user code, we should try to make the system more robust so user code is safe. Framework doesn't run by itself. It provides a service to applications. Some de-coupling of system concerns and application concerns will benefit adoption of the framework.

PR #373, opened 2016-01-11
Test HBV before mesh for IO

rahul27:
To improve the performance of ray casting we test the input ray
versus the hierarchical bounding volume before running a ray cast
on the objects mesh.


mingsong:
Code looks good. 
Have you tried some measuring to see the how much time improvement we can have on average with this HBV test filter?

rahul27:
@mingsong Here are the average times (in nano seconds):
Before :
Hit : 88584
Miss: 71791.75

After:
Hit: 75177
Miss: 13739.75

With the filter in place you see that the time is ~20% the original for a miss. The times for a hit is comparable to the before case.

PR #377, opened 2016-01-13
Add support to export scene to add light map

ragner:
Just to review


GVRf/Framework/jni/engine/exporter/exporter.cpp
```diff
@@ -0,0 +1,193 @@
+/* Copyright 2015 Samsung Electronics Co., LTD
```
liaxim: Spaces instead of tabs for indentation, please.

ragner: done.

liaxim:
The demo is nice. How did you go about specifying the light source in the scene before exporting it?

We will need input from Vittal on this one as he is working on the lights support.

---
I think you could separate completely out the export part as we could merge that sooner. Maybe put the export in the GVRContext as it is the usual dumping ground for stuff like that.

ragner:
Hi @liaxim, it is exporting just the Mesh information already.  After that you can add the lights' source at the Beast application that I have sent by email with the steps to test.

liaxim:
I was referring to the lightMap flag in addition to the export functionality. Not too big of a deal.

Vittalbabu:
Looks good.

Regarding lights -- I have an initial version of multiple lights using the shaders present in the Texture Shader class under https://github.com/Vittalbabu/GearVRf/commits/light. This is work in progress (not a pull request yet) and has incomplete implementation (plus lots of debug code) of some of the multiple lights functionality. When this is complete, we should be able to query from the Scene all the lights that are defined. Currently the implementation takes the same parameters as the existing GVRLight class but we are now able to attach multiple lights. All shader types  get access to all the lights in the Scene and only LightShader uses these lights. Goal is to fully flesh out the interfaces in Scene that handle Light, look at updating the shader in LightShader class (create multiple shaders if required), complete further definition of Light types.

Finally, we have to address one question. Do we force all shaders to use the available lights in the Scene or do we allow some of the shader types to ignore lights in their rendering. If we allow shaders to ignore lights, we will end up with some objects in the scene lit up while other are not.

danke-sra:
@Vittalbabu As far as I can see, this patch does not directly handle light types corresponding shaders. It mainly adds an exporter to export the scene for offline lightmap backing (with Beast). The question may be a bit off topic in this thread. In my opinion, in our current architecture, we allow users to select shaders including custom shaders for each scene object, so there is no systematic guarantee that all lights are rendered. The developer needs to take care of consistency.

danke-sra:
@ragner Thanks for the patch for review. It can be used now for static lights and static scene objects. Since it is offline baking, it doesn't need to support moveable lights and scene objects. Moveable things need to be handled in another way.

So the difficult issue is to design some way of coordinating Beast and dynamic lighting. Do you have any plan for that?

I think on GVRf side, maybe we can add a flag "moveable" to Scene Objects and Lights, and only export the non-moveable scene objects and lights for Beast. 

Further versions of GVRf will ideally export lights (types, position, etc.), so offline baking of these lights, and online rendering can be coordinated.

What do others think?

Vittalbabu:
@danke-sra Movable lights will be implemented by the new Light class holding a SceneObject* (to get the position,direction from the SceneObject). If this pointer is NULL, light is static. We can provide Scene::getLights() and Scene::getStaticLights() methods or Scene::getLights(bool staticLightsOnly = false)

danke-sra:
@ragner A minor issue with the exporter: when I export to "A.obj", it creates two files A.obj and A.obj.mtl. It would be better to use the file name "A.mtl" for the material file. Is it possible?

ragner:
Hi all,
Thank you for your review.
How there is no visual editor to the scene of GVRf framework, the only way I have found to integrate it with Beast was exporting the scene.

I have made this patch just for you to see something working and to prove that it is possible.

If you agree with this I will implement some "final version".

I am not sure if .obj is the best format to export the scene, I am using it just because there a sample of beast that imports .obj file.

Maybe the best solution could be export at collada format.

Thank you all.

danke-sra:
@ragner Thanks for keeping us in the loop. Please go ahead to implement the final version. Please keep in mind a couple of things while you design / implement it:

1) We need to make sure both static scene objects / lights, and moveable scene objects / lights work. Since Beast only handles static things, we can make sure we can easily exclude moveable scene objects / lights from the exporter.

2) The Visual Studio project for generating the lightmap would also be part of GVRf. So please make that tool generic as well. For instance, it can process the exported lights. Since dependency on Visual Studio may be some extra burden for Android developers, if possible, could we implement it in Lua, since Beast SDK also provides a Lua binding? If not, Visual Studio is fine. We just need to support multiple platforms, such as Windows, Mac and ideally Ubuntu as well.

I understand GVRf doesn't currently have complete light information, we can use standard Collada format to import those lights.

These are just some suggestions. Look forward to seeing the next version. Thanks for this contribution. It is amazing work!

ragner:
Hi @danke-sra,
* I haven't found any docs about Lua binding to Beast. What I have found is that Beast offers just a Lua render pass,  which allows users to customise the baking.
*  I am not sure if Beast is multi platform, its evaluation SDK is just for Windows platforms. I may need help about it. What I know is that Unity3d is multi platform and has support to Beast.

danke-sra:
Hi @ragner,

I checked again and found Lua support quite difficult. Yeah, you are right there is a Lua interface is for render pass scripting but non-C programming API is difficult -- it only supports C# and Python. So let's stick with Visual Studio and C++. Thanks.

-

PR #380, opened 2016-01-19
Optimization for Hiearchical View Frustum Culling

mingsong:
1. Augment the frustum testing function, to be able to tell if the given object's AABB
is completely inside the frustum that can eliminate further tests for all its children;
2. Add plane masking when doing frustum testing to skip testing against masked planes for all
the children when the parent's AABB is completely inside those planes.


mingsong:
Port from merged branch in pure.

Issue #86, opened 2015-05-29
Support for shadows.


thomasflynn:
Need support for rendering shadows.

Maginador:
We are planning on shadows support implementation.
The main idea is to start with hard shadows based on something like shadow mapping technique.

Our initial definitions for parameters to expose for the user :

Light Parameters (GVRLight)

UseShadows (on/off) - bool useShadow
	if the light is going to use shadow or not, in the case of more than one light, error/warning.

Shadow Type (hard/soft) - enum shadowType FUTURE USE
	Defines the type of shadow that will be used, it interferes in what passes sequence to use.

Softness (0-1)  - float softness FUTURE USE
	How soft is the shadow going to be. The amount of feather applied to the borders.

Strength (0-1) - float strength
	Opacity of the shadow, 1 for full shadow and 0 for no shadow.

Resolution (low/medium/high) - enum shadowResolution
	Quality of the mapping (render resolution): 
low = worse quality and higher performance.
medium = not bad quality and not bad performance.
high = higher quality and worse performance.

Bias(0-1) - float shadowBias
	bias to prevent artifact (the higher the bias, higher shadow offset (contact shadows may disappear if bias is too high).

Other light important Parameters

Range(1 - x) - float lightRange
	If using directional light its range is infinity, otherwise range will define how far the light interfere in shadow projection.

Angle/Direction (vec3(0 - 360)) - Vector3D lightDirection
	Direction from where the light comes.

Object Parameters (GVRRenderData)

Cast Shadow (on/off) - bool castShadows
	if this object is going to cast shadows towards other objects.

Receive Shadow (on/off) - bool receiveShadows
	if this object is going to receive shadows cast by other objects.

fredduarte:
@thomasflynn 

Here is a Work In Progress implementation of Shadow Mapping on the GVRf:
https://github.com/fredduarte/GearVRf/tree/shadowmap (3 commits)

Again, this is NOT a final implementation, but a display of the shadowing capability on the framework.

There is an attached sample project with a couple of 3D models rendering shadows,
but, on the old directory structure (Samples dir along with the Framework).

This way, to use the shadow map, all is needed is the code that follows:
```Java
...
gvrDirectionalLight = new GVRDirectionalLight(gvrContext);

scene.setDirectionalLight(gvrDirectionalLight);

gvrDirectionalLight.setLightPosition(0, 7, 10);

gvrDirectionalLight.setLightDirection(0, 0, 0);
...
```

[...]

PR #381, opened 2016-01-19
Shadow mapping implementation

fredduarte:
- Implemented shadow mapping technique using 3-step rendering process:
* First, it renders the scene from the perspective of the light;
* Second, it renders the scene from the perspective of the camera;
* Finally, it combines both to produce the shadows;

- Implemented a directional light object which is used for calculating the shadowmap Z distance.

- Created a sample project for displaying and testing the shadow mapping (moved to GVRf-Demos repo).

### DIFF
--- a/GVRf/Framework/jni/engine/renderer/renderer.cpp
+++ b/GVRf/Framework/jni/engine/renderer/renderer.cpp
@@ -35,6 +35,8 @@
 #include "util/gvr_gl.h"
 #include "util/gvr_log.h"
 
+#include "objects/components/directional_light.h"
+
 namespace gvr {
 
 static int numberDrawCalls;
@@ -160,6 +162,49 @@ void Renderer::renderCamera(Scene* scene, Camera* camera, int framebufferId,
     // bone/weight/joint and other assimp data, we will put general model conversion
     // on hold and do this kind of conversion fist
 
+    //  LOGI(" Render Camera Framebuffer %d ", framebufferId);
+
+    DirectionalLight *cameraLight = scene->getDirectionalLight();
+
+    bool renderShadow = cameraLight; // TODO reader from scene;
+
+    if ((framebufferId != 0) && !renderShadow) {
+        renderCamera(scene, camera, framebufferId, viewportX, viewportY,
+                viewportWidth, viewportHeight, shader_manager,
+                post_effect_shader_manager, post_effect_render_texture_a,
+                post_effect_render_texture_b, ShadowShader::RENDER_DEFAULT);
+
+    } else {
+        shader_manager->getShadowShader()->setCameraLight(cameraLight);
+
+        renderCamera(scene, camera,
+                shader_manager->getShadowShader()->getFBOFromLight(), viewportX,
+                viewportY, viewportWidth, viewportHeight, shader_manager,
+                post_effect_shader_manager, post_effect_render_texture_a,
+                post_effect_render_texture_b, ShadowShader::RENDER_FROM_LIGHT);
+
+        renderCamera(scene, camera,
+                shader_manager->getShadowShader()->getFBOFromCamera(),
+                viewportX, viewportY, viewportWidth, viewportHeight,
+                shader_manager, post_effect_shader_manager,
+                post_effect_render_texture_a, post_effect_render_texture_b,
+                ShadowShader::RENDER_FROM_CAMERA);
+
+        renderCamera(scene, camera, framebufferId, viewportX, viewportY,
+                viewportWidth, viewportHeight, shader_manager,
+                post_effect_shader_manager, post_effect_render_texture_a,
+                post_effect_render_texture_b, ShadowShader::RENDER_WITH_SHADOW);
+    }
+
+}
+
+void Renderer::renderCamera(Scene* scene, Camera* camera, int framebufferId,
+        int viewportX, int viewportY, int viewportWidth, int viewportHeight,
+        ShaderManager* shader_manager,
+        PostEffectShaderManager* post_effect_shader_manager,
+        RenderTexture* post_effect_render_texture_a,
+        RenderTexture* post_effect_render_texture_b, int modeShadow) {
+
     numberDrawCalls = 0;
     numberTriangles = 0;
 
@@ -193,7 +238,7 @@ void Renderer::renderCamera(Scene* scene, Camera* camera, int framebufferId,
         for (auto it = render_data_vector.begin();
                 it != render_data_vector.end(); ++it) {
             renderRenderData(*it, view_matrix, projection_matrix,
-                    camera->render_mask(), shader_manager);
+                    camera->render_mask(), shader_manager, modeShadow);
         }
     } else {
         RenderTexture* texture_render_texture = post_effect_render_texture_a;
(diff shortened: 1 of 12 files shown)


thomasflynn:
Vittal, please review this.
@fredduarte, please separate the sample into the GearVRf-Demos repo.  Thanks.

Vittalbabu:
@fredduarte What changes in the frame rate do you see when shadows are enabled?

Vittalbabu:
@fredduarte How will this handle multiple directional lights? Also, is this approach extensible to other light types?

GVRf/Framework/jni/shaders/posteffect/shadow_shader.cpp
```diff
@@ -0,0 +1,499 @@
...
+        "varying vec4 v_position;\n"
+        "varying vec4 v_position_world;\n"
+        "varying vec4 v_light_biased_clip_position;\n"
+
+        "void main() { \n"
+
+        //////////////////////////////////////////////////////////////////////////////////////////
+        ///  Ignore render: get depth light mode
+        //  FBO created with: getFBO => glFramebufferTexture2D(GL_FRAMEBUFFER, GL_DEPTH_ATTACHMENT, GL_TEXTURE_2D, depth->id(), 0);
+        "	if (u_mode < 0) return;\n"
```
Vittalbabu: Should we discard the fragment in this case?

GVRf/Framework/jni/shaders/posteffect/shadow_shader.cpp
```diff
@@ -0,0 +1,499 @@
...
+
+        //////////////////////////////////////////////////////////////////////////////////////////
+        /// Default texture // Optional mode: replace TextureShader
+        "   if (u_mode == 1) { gl_FragColor = texture2D(u_texture, v_tex_coord); return; } //\n"
+
+        //////////////////////////////////////////////////////////////////////////////////////////
+        /// Test Mode
+        "   if (u_mode == 2) { gl_FragColor = vec4( pow( texture2D(u_texture_light_color, (v_position_world.xy / v_position_world.w + 1.0) / 2.0).z, 10.0) ); return; } \n" //
+        "   if (u_mode == 3) { gl_FragColor = vec4( pow( texture2D(u_texture_camera_depth, (v_position_world.xy / v_position_world.w + 1.0) / 2.0).z, 10.0) ); return; } \n" //
```
Vittalbabu: Do we ever set u_mode to 3?

GVRf/Framework/jni/engine/renderer/renderer.cpp
```diff
@@ -160,6 +162,49 @@ void Renderer::renderCamera(Scene* scene, Camera* camera, int framebufferId,
     // bone/weight/joint and other assimp data, we will put general model conversion
     // on hold and do this kind of conversion fist
 
+    //  LOGI(" Render Camera Framebuffer %d ", framebufferId);
+
+    DirectionalLight *cameraLight = scene->getDirectionalLight();
+
+    bool renderShadow = cameraLight; // TODO reader from scene;
+
```
Vittalbabu: What is the default behavior of getDirectionalLight()? Will it return NULL or an instance of DirectionalLight?

GVRf/Framework/jni/engine/renderer/renderer.cpp
```diff
@@ -160,6 +162,49 @@ void Renderer::renderCamera(Scene* scene, Camera* camera, int framebufferId,
...
+    if ((framebufferId != 0) && !renderShadow) {
```
Vittalbabu: We render to the non-shadow map rendering mode only if both of these checks are true. I understand the check for !renderShadow. Why do we have the other check?

Vittalbabu:
Overall looks nice but I have some questions regarding testing.

1. Need to validate that we do not accidentally invoke shadow map shader when that is not the intent. (eg) We can add a log message to ShadowShader::render() and run a few of our existing samples.
2. What is the performance delta of using these 3 passes? We should test with a scene of moderate GPU load and heavy GPU load - with shadow map on and off. Based on the results we can decide how to proceed.

fredduarte:
Hi, @Vittalbabu.

Prior to answering here some of your questions on the code,
i provided a feedback from me and our former work partner on this shadowmap solution.
Please check the issue [ https://github.com/Samsung/GearVRf/issues/86 ].

danke-sra:
@fredduarte Thanks for the contribution again. I also have some comments about the patch. 

1) I tested it but the framerate is about 30 FPS in the demo app. Do you have any idea about optimizing it to achieve 60 FPS?

2) We would need to use the same light types as Unity and Assimp to avoid confusion. In Unity and Assimp, the light types are: directional, spot, point and ambient (see https://github.com/Samsung/GearVRf/blob/master/GVRf/Framework/src/org/gearvrf/jassimp2/AiLightType.java). I will elaborate below.

3) The current light type is DirectionalLight (with PERSPECTIVE and ORTHOGONAL subtypes). The PERSPECTIVE subtype is similar to Spot lights in Unity and Assimp. The orthogonal subtype is similar to Directional lights but the origin is not at infinity. I would think it's better to first **rename the light type to SpotLight** and use PERSPECTIVE as default.

4) Consider other light types following the Unity/Assimp definitions: Directional (may reuse some code from the ORTHOGONAL type), Point and Ambient. Refactor common code out of the current (GVR)DirectionalLight into either a Light base class, or component classes.

Please correct me if I misunderstood anything. Also, please kindly let us know if you agree on these items. For checking in the code, I personally think we need to avoid confusion in the light type naming. Performance optimization can be done separately.

thomasflynn:
i get a compile error:
jni/shaders/posteffect/shadow_shader.cpp: In destructor 'gvr::ShadowShader::~ShadowShader()':
jni/shaders/posteffect/shadow_shader.cpp:337:9: error: 'gl_delete' was not declared in this scope
         gl_delete.queueVertexArray(vaoID_);

liaxim:
This is because of the change to remove some of the writable static data. Should be a straightforward thing to change.

danke-sra:
Is it possible to make keep the gl_delete change transparent? I think it was a rush solution but not sure if the changes can be hidden from developers. If possible, can we expedite that refactoring? If the use of the deleter is correct, we may need to use it in other new classes, and some finalized API would be helpful.

thomasflynn:
@fredduarte any eta on a build fix for this patch?

fredduarte:
Hi, Thomas.
Sorry for the delayed response.
i'm working on the suggested fixes for theses patches. I've just recently got back to it,
because from the time i first sent this, i was researching ways to improve this solution (in FPS terms)
and also on the last two weeks i helped Maginador with another fix which you guys know.
But, i'm back to it now. I'll send an updated version at least with small fixes as soon as i can.
I don't see @Vittalbabu 's lighting branch anymore.
Has it changed to this one : https://github.com/Vittalbabu/GearVRf/commits/light ?

liaxim:
@fredduarte I have a commit that shows how to fix the compilation error Tom got: https://github.com/liaxim/GearVRf/commit/44fdb9b3407978b4b688ddcddbd5cfc4a6f6e546

fredduarte:
Yes, @liaxim . I hadn't updated the shader to reflect the GLDeleter approach that is used now on GVRf.
I'll push this change right away. Thanks.

liaxim:
@fredduarte So what is the purpose of the "(framebufferId != 0)" check in renderer.cpp? It is leading to rendering with shadow mapping enabled on certain phones when it is actually not enabled and subsequently the app crashes.

PR #384, opened 2016-01-21
Adding support for gl_alpha_to_coverage

roshanch:
allow setting the coverage value.

### DIFF
--- a/GVRf/Framework/jni/engine/renderer/renderer.cpp
+++ b/GVRf/Framework/jni/engine/renderer/renderer.cpp
@@ -496,7 +496,7 @@ void Renderer::renderRenderData(RenderData* render_data,
             glDisable (GL_BLEND);
         }
        if( !render_data->alpha_to_coverage()) {
-        	glDisable(GL_SAMPLE_ALPHA_TO_COVERAGE);
+           glDisable(GL_SAMPLE_ALPHA_TO_COVERAGE);
         }
         if (render_data->mesh() != 0) {
             for (int curr_pass = 0; curr_pass < render_data->pass_count();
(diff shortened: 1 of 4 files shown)


thomasflynn:
we should also allow setting the coverage value:  https://www.khronos.org/opengles/sdk/1.1/docs/man/glSampleCoverage.xml

GVRf/Framework/jni/engine/renderer/renderer.cpp
```diff
@@ -495,6 +495,9 @@ void Renderer::renderRenderData(RenderData* render_data,
         if (!render_data->alpha_blend()) {
             glDisable (GL_BLEND);
         }
+       if( !render_data->alpha_to_coverage()) {
+           glDisable(GL_SAMPLE_ALPHA_TO_COVERAGE);
+        }
```
Vittalbabu: If prev state was set as disable for GL_SAMPLE_ALPHA_TO_COVERAGE how do we enable it when the flag is true? We depend on previous call to RenderRenderData to set the state properly. Prefer to explicitly enable/disable state.

thomasflynn:
please be sure to leave a note up on github when you upload a new patch.  the reviewers don't get an email when you upload a patch, just when you comment about it.

in any case, code looks good so far.  did you test it?  how did you test it?

roshanch:
I queried the values of GL_SAMPLE_BUFFERS and GL_SAMPLES. I will test it with the sample and then we can merge it.

thomasflynn:
ok, i modified the transparency-test sample to test this.  rather than setting the renderdata->renderingOrder to TRANSPARENT, i just enabled AlphaToCoverage, and set the number of multisamples in the xml file to 2.  worked well.

Issue #385, opened 2016-01-22
import .dae file with textures = no textures


thomasflynn:
I tried to import this model: http://tf3dm.com/download-page.php?url=r2d2-21038
and wound up with no textures.  Figuring maybe Android didn't load .tga files, i converted them to .png and modified the .dae accordingly.

for someone who knows the model importing part of this, could you take a look?  thanks.

danke-sra:
The model cannot be downloaded from the website (it doesn't start downloading). Can you please upload the original model and the modified one to another location? I will take a look as soon as I can get the model files. Thanks.

thomasflynn:
let's try this link:  http://tf3dm.com/3d-model/r2d2-21038.html

danke-sra:
Thanks for the new link. I've found a fix for this issue. Please check out the pull request: https://github.com/Samsung/GearVRf/pull/386

PR #386, opened 2016-01-22
Fix texture loading issue when texture contains path like File::///

danke-sra:
The URL File::/// confuses loader that the file is relative to the
localhost's root. However, it actually means the current directory.
Just ignore the initial '/' introduced by File::/// in texture loading.


danke-sra:
Fix for https://github.com/Samsung/GearVRf/issues/385

thomasflynn:
works now.  thanks!

PR #387, opened 2016-01-22
Add command-line interface for GVRf

danke-sra:
- Add a DebugServer which accepts connections from port 1645
- Add a CLI shell to execute commands
- Add 'lua' command to enter lua shell
- Support print rediction on lua commands. Users can see lua command
  output in the shell


danke-sra:
Optional pull request to add an interactive shell for GVRf.

Usage: add the following line to your onInit() method.
```
    Threads.spawn(new DebugServer(gvrContext));
```

The interactive shell can be connected to using
```
telnet <phone-ip> 1645
```

If you use putty, please enable "Implicit CR for every LF" and "Implicit LF for every CR".

danke-sra:
Add "js" command which enters the Javascript shell:

For testing:

```
telnet <phone-ip> 1645
gvrf> lua
lua> print(gvrf:getMainScene():getSceneObjects())
lua> exit
gvrf> js
js> print(gvrf.getMainScene().getSceneObjects())
js> exit
gvrf> exit
```

Note: Lua engine is a bit unstable. It may crash after detecting a syntax error. We can fix it in the JNI code. JS engine is expected to be faster and more stable.

thomasflynn:
ok, cherry-picked, tested it, haven't reviewed code yet.  I only tested javascript so far.

Some notes for anyone else testing:
- had to import org.gearvrf.utility.Threads; and import org.gearvrf.debug.DebugServer; in order to add that Threads.spawn() line mentioned in the commit.  Perhaps have a gvrContext.startDebugServer() that does the Threads.spawn() line?  maybe a stopDebugServer() that stops it?
- add to add the internet permission to the AndroidManifest.xml.  it's obvious in hindsight, but I missed it at first.

A suggestion, if possible:
- I'd like to be able to use the arrow keys if possible.  go back and forth within a line; 'up' and 'down' to go up and down through history like you would in a regular shell.

Possible bug related to this check-in:
- "gvrf> exit" didn't exit the session.  left my telnet session in a hung state of some sort.  have to either kill the telnet process or terminate the app on the phone to get my command line back.

Possible but not related to this check-in:
- tried a gvrf.loadModelFromURL();  It was that r2d2.dae file.  It downloaded and loaded the dae file, but not the textures.

danke-sra:
Thanks for the comments. I am working on these items. Most of them should be ready today including line editing and history support over telnet.

I could reproduce the texture loading issue with r2d2 with scripts. It seems to be a texture initialization issue. The code should run in GL thread, however, most of the script code shouldn't. So some investigation will be needed before finalizing the fix.

danke-sra:
Addressed reviewer comments.

Fixed issues:
1) New API startDebugServer() and stopDebugServer() have been added
2) After the last exit command, it disconnects the client
3) Add a fancy line editing / history layer that supports arrow keys
4) Add a Telnet layer to support telnet protocol

Issues to be addressed later:
1) The texture issue will need to be done in another patch -- prefer to find out the best solution (though running the script in a GL thread will work now).

2) Best telnet client is Linux. Windows Telnet is fully functional with some minor display issues. Putty is not yet supported. We don't have a fully functioning Telnet layer to use yet. Telnet-related issues will be done separately.

thomasflynn:
i'll probably test this tomorrow night.  i'm pretty exhausted now.  if anyone else is interested in reviewing, please do so now.

thomasflynn:
ok, cherry-picked it, built it, tested it.  looks good from a functionality point of view.

danke-sra:
thanks for the testing. there is some aesthetic issue if using Windows Telnet but functionality issues have been resolved.

thomasflynn:
ok, looks good.

Issue #391, opened 2016-01-28
NPE not reported inside onInit()


rahul27:
Recreation - Try calling a method on a null object inside GVRScript.onInit().

Guess the same is the case with onStep(). 

It looks like an issue with the GVREventManager.sendEvent() call for "onInit". The NPE is caught as InvocationTargetException inside the invokeMethod() call.

rahul27:
@danke-sra is this a bug?

danke-sra:
It should throw a NullPointerException to the method calling sendEvent(). If the exception is somehow caught with no side effect, it's also acceptable but better throw an exception in this case.

danke-sra:
I looked at the code: it looks like there is a type check on the 'target'. If it is null, it should throw another RuntimeException. Did you not get this exception either?

if (!eventsClass.isInstance(target)) {
            throw new RuntimeException(String.format("The target object does not implement interface %s",
                    eventsClass.getSimpleName()));
}

I haven't got time to run the code yet, but please let me know if this does not throw on null pointer. The framework should not rely on NullPointerException. Many methods may ignore invalid inputs and just return without throwing.

rahul27:
The target in my case is not null. I have a valid script, the scenario I am talking about is something like this:

```java
public class SampleScript extends GVRScript {

private GVRSceneObject test = null;

@Override
public void onInit(GVRContext gvrContext) {
  test.getTransform(); //This line should throw NPE, but does not.
}
```


Instead the NPE is caught here inside InvocationTargetException in invokeMethod():

```java
private void invokeMethod(Object target, Method method, Object[] params) {
        try {
            method.invoke(target, params);
        } catch (IllegalAccessException e) {
            e.printStackTrace();
        } catch (IllegalArgumentException e) {
            e.printStackTrace();
        } catch (InvocationTargetException e) {
            e.printStackTrace(); // NPE is caught here and not returned to the application
        }
    }
```

danke-sra:
I see. Can we re-throw the InvocationTargetException here?

rahul27:
I was thinking something along the lines of:

```java
 } catch (InvocationTargetException e) {            
            Throwable t = e.getCause();
            if (t instanceof RuntimeException) {
                throw t;
            } else {
                e.printStackTrace();
            }
        }
```

danke-sra:
That works better. Thanks.

PR #392, opened 2016-01-28
Rethrow the RuntimeException back to the application

rahul27:
Return the RuntimeException received as a result of the
InvocationTargetException caught by the GVREventManager
back to the application.


rahul27:
fix for #391.

danke-sra:
Looks good.

PR #395, opened 2016-01-30
Fix ant build.

dbrant:


thomasflynn:
i have the same change locally.  should've committed it.  good change.

PR #398, opened 2016-02-01
Catch exceptions from "user" callbacks

liaxim:
Keep the branch up-to-date.

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/GVRViewManager.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRViewManager.java
@@ -675,9 +675,13 @@ class GVRViewManager extends GVRContext implements RotationSensorListener {
 
             doMemoryManagementAndPerFrameCallbacks();
 
-            GVRViewManager.this.getEventManager().sendEvent(
-                    mScript, IScriptEvents.class,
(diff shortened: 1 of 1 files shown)


rahul27:
@liaxim what is the reason behind this change? and was "onInit" left out intentionally?

liaxim:
Not having random "user" exceptions escape uncontrollably on the rendering thread and inducing obscure crashes? In privileged apps it could lead to much weirder outcomes.

The onInit is an example how we should go about this - in that case finish the activity. Maybe we should do that too in other cases or bring up a common error dialog and shutdown gracefully. Before any of these, we can put up some defenses.

rahul27:
I see. But wouldn't this change make it even more difficult for developers to know something is wrong?Assume that you have an NPE in onStep(), it would go unnoticed by the app which would resume like nothing is wrong.

Unless like you mentioned, you are planning on implementing a dialog to let the user know that there is an error followed by a subsequent finish().

liaxim:
The exception is printed in logcat. I do agree with you but these exceptions if unhandled get into jni and can get harder to decipher.

I believe for now we should gracefully shutdown the app but something like this will require a discussion.

PR #399, opened 2016-02-01
Allow for capturer's garbage collection

liaxim:
Keep the branch up-to-date.

### DIFF
--- a/GVRf/Framework/jni/objects/components/texture_capturer.cpp
+++ b/GVRf/Framework/jni/objects/components/texture_capturer.cpp
@@ -47,9 +47,12 @@ void TextureCapturer::setCapturerObject(JNIEnv *env, jobject capturer) {
 
     if (mCapturerObject) {
         mJNIEnv->DeleteGlobalRef(mCapturerObject);
+        mCapturerObject = nullptr;
     }
 
-    mCapturerObject = env->NewGlobalRef(capturer);
+    if (nullptr != capturer) {
+        mCapturerObject = env->NewGlobalRef(capturer);
+    }
 }
 
 void TextureCapturer::setRenderTexture(RenderTexture *renderTexture) {
@@ -164,8 +167,10 @@ glm::mat4 TextureCapturer::getMvpMatrix(float half_width, float half_height) {
 }
 
 void TextureCapturer::callback(int msg, char *info) {
-    Java_org_gearvrf_NativeTextureCapturer_callbackFromNative(
-            mJNIEnv, mCapturerObject, msg, info);
+    if (nullptr != mCapturerObject) {
+        Java_org_gearvrf_NativeTextureCapturer_callbackFromNative(
+                mJNIEnv, mCapturerObject, msg, info);
+    }
 }
 
 }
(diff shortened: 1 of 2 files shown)


GVRf/Framework/jni/objects/components/texture_capturer.cpp
```diff
@@ -164,8 +167,10 @@ glm::mat4 TextureCapturer::getMvpMatrix(float half_width, float half_height) {
 }
 
 void TextureCapturer::callback(int msg, char *info) {
-    Java_org_gearvrf_NativeTextureCapturer_callbackFromNative(
-            mJNIEnv, mCapturerObject, msg, info);
+    if (nullptr != mCapturerObject) {
```
danke-sra: if mCapturerObject is r/w, a mutex is needed to protect it here. race condition can occur between this if-statement and Lines 48 - 55.

GVRf/Framework/src/org/gearvrf/GVRTextureCapturer.java
```diff
@@ -127,6 +127,9 @@ public void setCapture(boolean capture, float fps) {
     public void addListener(TextureCapturerListener l) {
         synchronized (mListeners) {
             mListeners.add(l);
+            if (1 == mListeners.size()) {
```
danke-sra: it might be more canonical to test for empty() before adding. but testing for 1 is totally correct.

GVRf/Framework/jni/objects/components/texture_capturer.cpp
liaxim: Feel free to make the proper fix in a subsequent commit. After all you know this code better. Also please fix the circular reference - the native side adds a reference to the java object. The java object cannot be ever collected because of that reference. Without collecting the java object the native can't be deleted either. Using the listeners is a kludge I had to invent to remove the leaks. There are no guarantees users will remove their listeners - if they don't, they will leak also whatever their listeners refer to.

danke-sra: Not sure why circular reference prevents GC?

liaxim: The native side makes a strong reference that it never releases (except in the special case with the listener I added). The Java object is hence never collected. If the java object is not collected, the native is not deleted either as we are dealing with a "hybrid" object.

danke-sra: That might be correct. But it can be fixed by passing a WeakReference to the native code, and that would not require hacking the listener methods.

liaxim: I had no idea of the assumptions in this code. It looked to me like the native side wanted a guarantee that the java object is alive at the time of the callback. A guarantee a weak reference cannot give. What I made felt like the safer fix. And after all, if nobody listens why send events?

danke-sra: The assumption is if the capturer is active, it should be reachable by its owner, which is an AppScene. BTW, does it require removing listeners when it is not used? If so, I should definitely look into the weakreference method.

liaxim: Ok, sounds good. Should I close this pull request?

danke-sra: Yeah, I will submit a new patch ASAP. thanks.

PR #401, opened 2016-02-01
Remove static state

liaxim:
- finalizer thread and related state
- gl_delete
- also add a tag object to the context to help users avoid static state themselves

--
With or without this change there are crashes in oculus when doing the dock/undock/dock test quickly enough; I believe it is actually the platform deviating from the expected behavior - a singleInstance activity should not be created more than once. In that special case I've seen numerous times where there are two instances though (first paused, second created then first destroyed). There is probably static state in Oculus that gets reset at unexpected time. Normal GVRF apps are fine.


GVRf/Framework/src/org/gearvrf/GVRHybridObject.java
```diff
@@ -35,7 +30,7 @@
  * Descendant classes all have native (JNI) implementations; this base class
  * manages the native lifecycles.
  */
-public abstract class GVRHybridObject implements Closeable {
+public abstract class GVRHybridObject {
```
liaxim: If anyone wonders - using this interface triggers very annoying static analysis error in Eclipse. The tool thinks this entity is some kind of a stream (which it is not) and if it doesn't see the corresponding close complains. And close on this class is generally not supposed to be called.

liaxim: Not to mention that "close" is not representative of what exactly happens; not to mention that after closing we need state checks as you can still call methods, etc..

GVRf/Framework/jni/gl/gl_render_buffer.h
```diff
@@ -32,11 +32,12 @@ namespace gvr {
 class GLRenderBuffer {
 public:
     GLRenderBuffer() {
+        deleter_ = getDeleterForThisThread();
```
danke-sra: why not try to keep deleter_ as a reference. Then, a lot of reference-to-pointer changes can be saved. Pointers themselves can also more error-prone too.

GVRf/Framework/src/org/gearvrf/GVRHybridObject.java
danke-sra: The "Closeable" has been a misuse of the interface, which has more bulit-in meaning than the interface itself. It's good to remove it or create our own version of that.

GVRf/Framework/src/org/gearvrf/GVRHybridObject.java
```diff
@@ -94,9 +89,7 @@ protected GVRHybridObject(GVRContext gvrContext, long nativePointer,
         mGVRContext = gvrContext;
         mNativePointer = nativePointer;
 
-        synchronized (sReferenceSet) {
-            sReferenceSet.add(new GVRReference(this, nativePointer, cleanupHandlers));
-        }
+        gvrContext.registerHybridObject(this,nativePointer, cleanupHandlers);
```
danke-sra: missing a space after "this,". It's a good time to fix it now if this patch is going to be updated.

GVRf/Framework/src/org/gearvrf/GVRHybridObject.java
```diff
@@ -372,45 +288,27 @@ public void run() {
...
-            if (mNativePointer != 0L) {
-                GVRReference reference = findReference(mNativePointer);
-                if (reference != null) {
-                    reference.close();
-                    mNativePointer = 0L;
-                }
-            }
-        }
+    public final void releaseNative() {
+        mGVRContext.releaseNative(this);
```
danke-sra: it's better to modularize the GVRContext now while adding new API. but this is totally optional.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -883,4 +885,24 @@ public GVREventManager getEventManager() {
...
         return mScriptManager;
     }
+
+
+    protected long mGlDeleterPtr;
+
+    @Override
+    public void finalize() throws Throwable {
+        try {
+            if (0 != mGlDeleterPtr) {
```
danke-sra: It looks like the DeletePtr requires explicit life-cycle management in user code, and is scatter in different places (a raw pointer, a constructor invocation, a destructor invocation, and using the raw pointer in API). This seems to be something for encapsulation.

liaxim: Yes, it is scattered as it came as an afterthought. This is what I could come up with quickly to eliminate at least some of the static state.

There must be a simple, more robust way but it needs thought. Like the gvrf finalizer thread managing the gl cleanup also. That whole thread-local stuff I don't like.

danke-sra:
The original solution exposes gl_delete and doesn't require creation/destruction. That's the good part. But of course, it doesn't work robustly.

If the solution is to use thread-local storage, we could hide internals in a new version of gl_delete. When it is first referenced by a thread, it can create an instance of it and put in TLS. We can also attach a destructor with the TLS key, so that the destructor doesn't need to be explicitly invoked from Java. Actually, I don't see why Java's finalize method has to do with TLS clean-up. Need to look at it more closely.

liaxim:
I don't want random threads creating "deleters". I want to "bind" certain entities to the right gl thread. Hence calls from Java to establish "context". This is a quick fix as I said before and I need the trees to be in sync.

A proper solution will take much more than that. Our custom finalizer thread knows when a sceneObject is to be destroyed. It could also manage the gl cleanup directly - needs to be seen what it means for it to have awareness of the gl thread, complexities and tradeoffs, etc.

Now sceneObjects get created in one way but get destroyed in another. No symmetry no good.

GVRf/Framework/jni/engine/memory/gl_delete.h
```diff
@@ -25,15 +28,48 @@
 #define GVR_INVALID 0
 
 namespace gvr {
+
+extern pthread_key_t deleter_key;
```
danke-sra: Add a few things that caught my eyes. This is internal data. Suggest removing from gl_delete.h.

GVRf/Framework/jni/engine/memory/gl_delete.h
```diff
@@ -25,15 +28,48 @@
...
 
 public:
+    /**
+     * Before using this class this method must be called once
+     * and only once per-process
+     */
+    static void createTlsKey() {
+        //corresponding pthread_key_delete missing intentionally as this is
+        //needed for as long as the process is alive
+        int err = pthread_key_create(&deleter_key, nullptr);
```
danke-sra: This function references an internal data "deleter_key". it can be moved to gl_deleter.cpp.

GVRf/Framework/jni/engine/memory/gl_delete.h
```diff
@@ -25,15 +28,48 @@
...
+    }
+
     GlDelete() {
-        pthread_mutex_init(&mutex, 0);
+        int err = pthread_mutex_init(&mutex, nullptr);
+        if (0 != err) {
+            LOGE("fatal error: pthread_mutex_init failed with %d!", err);
+            std::terminate();
+        }
+        err = pthread_setspecific(deleter_key, this);
```
danke-sra: same as above. can be moved to cpp.

GVRf/Framework/jni/engine/memory/gl_delete.h
```diff
@@ -69,7 +105,19 @@ class GlDelete {
     std::vector<GLuint> vertex_arrays_;
 };
 
-extern GlDelete gl_delete;
+/**
+ * The assumption is threads that do know they are supposed to have
+ * a deleter may only call this method.
+ */
+static GlDelete* getDeleterForThisThread() {
```
danke-sra: this static function in header file can increase code size. could be moved to cpp, and make it extern.

liaxim:
Sure, as a future improvement I may take care of all these. Ideally we should find a way not to need a clutch like it.

danke-sra:
For this patch, does it solve the crash? if it does, suggest clean up the code in a separate patch. If it doesn't, then we could find the root cause and fix that directly.

liaxim:
It does fix the original crash in gvrf reported in 19919.

PR #402, opened 2016-02-03
Fix the build

liaxim:
There are methods unused in the other tree but used here.


thomasflynn:
thanks.  talked to the build team this morning.  we'll get an automated build going here one way or another.

PR #404, opened 2016-02-03
Small improvements to GVRPickedObject.

johnspurlock:
- Make constructor visible, take struct fields as direct params.
 - Use standard naming for private fields.
 - Fix some javadocs.

GearVRf-DCO-1.0-Signed-off-by: John Spurlock
john.spurlock@gmail.com

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/GVRPicker.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRPicker.java
@@ -297,7 +297,7 @@ public class GVRPicker {
             for (final long pointer : pointers) {
                 final GVREyePointeeHolder holder = GVREyePointeeHolder.lookup(gvrContext, pointer);
                 if (null != holder) {
-                    result.add(new GVRPickedObject(holder));
+                    result.add(new GVRPickedObject(holder.getOwnerObject(), holder.getHit()));
                 }
             }
             return result;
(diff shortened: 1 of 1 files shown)


liaxim:
Looks good

johnspurlock:
Thanks - can you merge?  I don't have write access.

GVRf/Framework/src/org/gearvrf/GVRPicker.java
```diff
@@ -336,12 +336,12 @@ public static final float pickSceneObject(GVRSceneObject sceneObject,
...
     public static class GVRPickedObject {
-        private final GVRSceneObject sceneObject;
-        private final float[] hitLocation;
+        private final GVRSceneObject mHitObject;
+        private final float[] mHitLocation;
 
-        private GVRPickedObject(GVREyePointeeHolder holder) {
-            sceneObject = holder.getOwnerObject();
-            hitLocation = holder.getHit();
+        public GVRPickedObject(GVRSceneObject hitObject, float[] hitLocation) {
```
thomasflynn: this went from private to public.  i'm assuming there's a good reason?  and if so, all public methods of all public classes need javadoc.

GVRf/Framework/src/org/gearvrf/GVRPicker.java
```diff
@@ -362,22 +362,22 @@ public GVRSceneObject getHitObject() {
...
         }
 
         /** The x coordinate of the hit location */
         public float getHitX() {
-            return hitLocation[0];
+            return mHitLocation[0];
         }
 
-        /** The x coordinate of the hit location */
+        /** The y coordinate of the hit location */
```
thomasflynn: haha, good catch.  not sure how we missed these two copy-n-paste typos.  thanks.

thomasflynn:
needs javadoc.  otherwise, looks ok.  and only reviewers/maintainers can merge code:
http://wiki.gearvrf.org/bin/view/GearVRF/GearVRfWikiCommunity#ProjectGovernance

johnspurlock:
Ok, added javadoc for the ctor, please take another look.

Use case was modifying the list of hit objects returned during frame processing, removing some, adding some synthetic ones for related objects.   Noticed that this is basically an immutable struct / pair class, and it would be useful to reuse instead of creating another class.  Might also be useful for unit testing, but that's definitely not what I was doing.  : )

thomasflynn:
ok, fair enough.  looks good now.  and feel free to write unit tests if that's what you enjoy doing ;-)

PR #406, opened 2016-02-04
Refine and correct futureTexture setting in GVRMaterial

mingsong:
GVRMaterial spawns a new thread to set a future texture and the thread is blocked once
the future texture is not finish loading. However, if a texture is already loaded e.g.,
for the preloaded textures, it is not only unnecessary to spawn a new thread, but also
could lead to scheduling issue once the thread pool/system is crowed, such that even a texture
is set to a material, the setting thread is not being executed in the next frame but later.


mingsong:
This is a quick patch to fix preloaded texture setting, and a more fundamental refinement for async future texture loading is coming soon.

danke-sra:
This patch is good by itself. It sets the texture more reliably if it is ready. Should be good to go.

PR #407, opened 2016-02-04
Texture: delay GL ops until binding the texture

danke-sra:
Defer GL operations until the texture is used in a GL context.


danke-sra:
This patch allows textures to be constructed outside GL thread. It fixes the gl_texture, texture, and base_texture. If it looks good. Can we merge this, and work on the rest of textures in separate patches?

thomasflynn:
looks good.

Issue #376, opened 2016-01-13
Make input events accessible to applications


danke-sra:
Bring up this issue for discussion and some design / implementation efforts.

After the Input manager is in, we may need to work on application's handling of input events, in addition to the framework's handling of inputs.

What I have in mind is some steps below, but I am looking for inputs to revise them:

1) Define input events in an interface like the IScriptEvents in the script patch. For example, add an interface called IInputEvents, and define signatures of handlers, like this

```
interface IInputEvents {
    void onKeyEvent(KeyEvent event);
    void onMotionEvent(MotionEvent event);
}
```

The actual event class may be changed to GVR equivalents (when we have them). We may also add a source parameter, and some a return value.

2) Possibly modify the input manager to allow adding these listeners to corresponding. We might need to consider both application-level listener, and fine-grained scene-object-level listeners.

3) When invoking handling methods in the listeners, please use the GVREventManager. This allows for scripting support transparently.

After the above input event handling is in place, we can continue to add scripting support for input events. When input event support is done, we can extend it to other event groups, such as hit-test, visibility, collision, timers, etc.

@rahul27, can you please comment on this proposal? Do you have suggestions on its improvement / modification?

rahul27:
@danke-sra If my understanding of the proposal is correct then I would like to address the issue at two levels:

1) Application Level:

The current implementation of the input manager does not prevent the application from accessing the MotionEvents or KeyEvents generated by Android i.e. the app can do this by overriding the corresponding methods like below:

```java
public class SampleActivity extends GVRActivity {

    private static final String TAG = SampleActivity.class.getSimpleName();

    @Override
    protected void onCreate(Bundle icicle) {
        super.onCreate(icicle);
        setScript(new SampleViewManager(), "gvr_note4.xml");
    }

    @Override
    public boolean dispatchGenericMotionEvent(MotionEvent event) {
        // TODO Auto-generated method stub
        return true;
    }

    @Override
    public boolean dispatchKeyEvent(KeyEvent event) {
        // TODO Auto-generated method stub
        return true;
    }

    @Override
    public boolean dispatchTouchEvent(MotionEvent event) {
        // TODO Auto-generated method stub
        return true;
    }
}
```

This allows the application to override the Frameworks handling of the events and implement its own handling mechanism like how vrf_controls_sample handles input:

https://github.com/gearvrf/GearVRf-Demos/blob/master/gvrf_controls_sample/src/org/gearvrf/controls/MainActivity.java

[...]

PR #410, opened 2016-02-07
Adding the Gaze Cursor Controller

rahul27:
Use the setting useGazeCursorController to enable this feature.

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/io/GVRGazeCursorController.java
+++ b/GVRf/Framework/src/org/gearvrf/io/GVRGazeCursorController.java
@@ -1,0 +1,83 @@
+/* Copyright 2015 Samsung Electronics Co., LTD
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gearvrf.io;
+
+import org.gearvrf.GVRContext;
+import org.gearvrf.GVRDrawFrameListener;
+import org.gearvrf.GVRScene;
+import org.gearvrf.GVRSceneObject;
+import org.gearvrf.GVRTransform;
+import org.gearvrf.utility.Log;
+
+import android.opengl.Matrix;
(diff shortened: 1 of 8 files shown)


rahul27:
This is the first of many pull requests for https://github.com/Samsung/GearVRf/issues/376.

Before we can distribute key events to scene objects we need to add a gaze controller to detect collision events using the GearVR touchpad.

rahul27:
@danke-sra Here is a patch you can use to test it.

https://github.com/rahul27/GearVRf-Demos/commit/4ef12adbe3692b486df356054d6b5a19a2bffe53

GVRf/Framework/src/org/gearvrf/io/GVRGazeCursorController.java
```diff
@@ -0,0 +1,83 @@
...
+    public void setPosition(float x, float y, float z) {
+        this.x = x;
+        this.y = y;
+        this.z = z;
+        super.setPosition(x, y, z);
+    }
+
+    @Override
+    public void onDrawFrame(float frameTime) {
+        float[] position = new float[] { x, y, z, 1.0f };
```
danke-sra: I am taking a look at the patch now. Looks good generally but just comment on some minor things for your reference.

It may be better to save the allocation here if it is allocated as a data member at initialization. Allocation on onDrawFrame handlers should be kept minimal.

GVRf/Framework/src/org/gearvrf/io/GVRGazeCursorController.java
```diff
@@ -0,0 +1,83 @@
...
+        float[] modelMatrix = context.getMainScene().getMainCameraRig()
```
danke-sra: After the initial input manager patch, we have adopted JOML as a GL math library. You may consider using Vector3f for this computation (same allocation pattern is recommended). You might even consider storing x, y, z directly in a Vector3f data member. I think there is a getModelMatrix4f version that returns a Matrix4f. Just for your reference.

GVRf/Framework/src/org/gearvrf/io/GVRGazeCursorController.java
rahul27: I always assumed that the compiler would kick in to do that optimization. In any case I have changed this in the new patch.

GVRf/Framework/src/org/gearvrf/io/GVRGazeCursorController.java
rahul27: Done. Thanks for letting me know. Have seen a lot of variation in vector definition throughout the Framework. Vector3f should make things simpler going forward, an added bonus is that it has Quaternion definitions.

liaxim:
Looks good. Updated sample works fine.

thomasflynn:
@danke-sra any further feedback?

danke-sra:
It looks good.

PR #412, opened 2016-02-11
VRApi support + refactoring

liaxim:
List of notable changes:

- attempted to separate out the oculus-specifics at the java level (see VrapiActivityHandler)
- the framebufferobject class is ported over from the sample but the depth handling is ported from vrappfw
- broke down the C++ activity to few pieces, the main class is no more a template
- added few gl-specific logging macros and put them to use
- should fall back to monoscopic if oculus is not present but have not tested that; need to find a proper device
- brings up the oculus confirm-to-exit and global menu screens on back and long-press back; we may choose not to do it now
- decided to use the "showLoadingIcon" config option to disable/enable our logo; good idea/bad idea?; it is not applicable currently to the other branch and is only available to vrappfw apps as the loading icon is implemented in that library
- updated the xml parser to recognize xyz-parm and xyz-param tags so it can take config files from any branch

Monoscopic fallback (S4 mini, no head rotation possible as it has no Sensor.TYPE_GAME_ROTATION_VECTOR)
- found a crash on app shutdown - fixed
- gvrf logo never goes away, doubtful it has anything to do with this pull request; the splashScreen fadeOut animation gets scheduled but never finishes.
- with the logo disabled (do not forget to switch from getNextMainScene to getMainScene in the sample) the monoscopic renderer seems to work fine

Fixed known issues:
- force switching to monoscopic mode
- model-viewer rendering incorrectly models due to not picking the depth texture format correctly
- gvrsimplegallery; shows black screen after the logo; fixed when the default configuration on the java side is updated
- vr-controls; native crash somewhere in base_texture; it should be there in master too; fixed
- vrtop; docking/undocking frequently leads to crashes in vrapi; fixed by un-initializing vrapi from onPause instead
- gvr-vuforia; the camera quads goes all-white in less than a second; master's monoscopic mode exhibits the same behavior but stereo is fine; ensure GL_BLEND is not enabled when vuforia updates the camera

List of samples broken both here and in master:
- VRKeyboard; crashes somewhere in base_texture; the fix for vr-controls didn't take care of this one; fixed - a texture was not created on the gl thread
- TransparencyTest; probably not broken, just me not understanding it; exhibits the same (seemingly odd to me) behavior with and without this change
- gvrsample; right eye's cube is missing its left side - drawn black; while orange-ish for the left eye
- gvrlitshader; same issue as gvrsample


thomasflynn:
-1.  Can't explain why, but the model-viewer sample app is broken.  Looks fine intially, but when you click on each of the models, they are each missing parts.  like maybe the first mesh is being rendered??  can't say.  

i haven't tested with a non-VR phone yet.  i used to test with my S5, but the company installed KNOX on it and disabled USB debugging :(.  I'll have to find another device.

liaxim:
There is something wrong with the depth texture. Working on it.

---
I should have gone through each and every sample. Will do so after I fix this problem.

liaxim:
Fixed the model-viewer; root cause was not picking the right depth texture format.

liaxim:
Found a problem with gvrsimplegallery; will keep updating the list of known issues in the first comment from now on to minimize spamming.

GVRf/Framework/jni/gl/gl_texture.h
```diff
@@ -56,8 +56,10 @@ class GLTexture : public GLPendingTask {
         memcpy(texture_parameters_, texture_parameters, sizeof(int) * 5);
     }
 
-    ~GLTexture() {
-        deleter_->queueTexture(id_);
+    virtual ~GLTexture() {
+        if (GL_TASK_NONE == pending_gl_task_) {
```
danke-sra: Not sure the reason for this check. If it is to make sure id_ is valid, it should be initialized to an invalid value, and check against that.

GVRf/Framework/jni/oculus/activity.cpp
```diff
@@ -0,0 +1,233 @@
+/* Copyright 2015 Samsung Electronics Co., LTD
```
danke-sra: we might need to use 2016 now. and might need to update all headers to 2016 as other open source projects have done.

GVRf/Framework/jni/gl/gl_texture.h
liaxim: id_ is a GLuint. glGenTextures can give you any GLuint. There is no invalid value to set it to strictly speaking. Instead of a dedicated flag I chose what I thought was the lesser evil.

danke-sra: You can just check against 0. 0 is never a valid texture id. Checking GL_TASK_NONE to mean the task is done is dangerous. There are too many complications, 1) the task could have never started. 2) the task started and finished but ID is invalid.

liaxim: Ok, will do that.

thomasflynn:
if you've tested against all the samples, then you can also test against: https://github.com/dbrant/GearVR_planetarium

thomasflynn:
cool, was just able to run in monoscopic mode on a nexus 5.  first time that's been possible in a while.

liaxim:
GearVR_planetarium works fine. Couldn't spot any difference vs master.

liaxim:
Update on gvr_vuforia:
- master-stereo - works
- master-mono - goes white
- vrapi-stereo - goes white
- vrapi-mono - goes white
- the_other_development_repo-master-stereo - goes white

Probably we are messing up the gl state for vuforia; the only somewhat relevant post: https://developer.vuforia.com/forum/rendering-opengl-es/using-vuforia-our-own-custom-render-engine. The effect described is similar to what I am seeing. If I don't move the phone at all I can still make out the darkest areas.

---
I've sunk enough time in this. My suggestion is to deal with this separately from this pull request (as time permits) and first update to the latest Vuforia to hopefully pick up bug fixes. If it still doesn't work, continue the investigation.

liaxim:
Vuforia sample fixed too. No known to me issues with this pull request anymore.

thomasflynn:
anyone else have comments on this?

danke-sra:
Tested on S6 edge+. Mono and fullscreen work. But it still requires developer mode. Is there a way to remove this requirement? In mono mode, we don't need timewarp because it is for developer only.

liaxim:
I think on devices without VR service mono will just work.

---
Btw, we can finally debug Java code properly with this change in.

danke-sra:
I have no problem for merging this patch if others agree.

thomasflynn:
needs rebase

liaxim:
Rebased; merged a bunch of stuff in renderer.cpp. Shadowmap sample and few others work fine.

PR #414, opened 2016-02-18
Expose input key and motion events via Controllers

rahul27:
The patch includes the following changes:
* The app and query the controller for the latest key and
  motion events processed by the GVRCursorController
* Allow the user to enable/disable the GVRCursorController.By
  default all controllers are enabled. Use the setEnable(boolean)
  call to disable them.
* Use and explicit process scene call whenever the scene changes.
  So that the SensorEvents are always latest.
* Add listener to notify application whenever the controller
  data is invalidated.
* Replaced float arrays by JOML Vector3f in GVRCursorController
* General Cleanup of Mouse and Gamepad controllers.

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/GVRCursorController.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRCursorController.java
@@ -15,11 +15,17 @@
 
 package org.gearvrf;
 
+import java.util.ArrayList;
+import java.util.List;
+
 import org.gearvrf.io.CursorControllerListener;
 import org.gearvrf.io.GVRCursorType;
 import org.gearvrf.io.GVRInputManager;
+import org.joml.Vector3f;
 
-import android.opengl.Matrix;
+import android.util.Log;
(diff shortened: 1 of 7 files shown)


rahul27:
@danke-sra do you have an app in mind we could test this on?

danke-sra:
I have looked at the code but would need some app to test it. Any suggestions?

BTW, I would suggest supporting the volume keys by adding a default device controller. Also, please consider if a better name is needed for the GVRCursorController, because now it includes key event delivery. Can you please also make a sample about how to listen to key events using the input manager?

rahul27:
@danke-sra @thomasflynn I've uploaded a sample to test this pull request. It forwards the generated events to a GVRFrameLayout to detect button and key clicks.

The sample also supports the back and volume up/down hardware buttons. 

Regarding the name for GVRCursorController - I am fine with keeping it the way it is since it is essentially still a controller that controls a cursor scene object. But now with the added benefit of distributing key and motion events. We could talk more if you'd like.

PR #415, opened 2016-02-19
Remove libpng

liaxim:
---
Tested again with the shadows sample, eye picking, the modified gvrcockpit.


liaxim:
Tested again with the shadows sample, eye picking, the modified gvrcockpit.

PR #420, opened 2016-02-23
Fix couple of issues identified by Mark Hong

liaxim:
- memory leak in picker.cpp
- stream not explicitly closed in GVRAsynchronousResourceLoader.java

### DIFF
--- a/GVRf/Framework/jni/engine/picker/picker.cpp
+++ b/GVRf/Framework/jni/engine/picker/picker.cpp
@@ -100,14 +100,12 @@ glm::vec3 Picker::pickSceneObjectAgainstBoundingBox(
         const SceneObject* scene_object, float ox, float oy, float oz, float dx,
         float dy, float dz) {
     glm::mat4 model_matrix = scene_object->transform()->getModelMatrix();
-    Mesh* mesh = scene_object->render_data()->mesh()->getBoundingBox();
-    if (mesh != 0) {
-        EyePointData data = MeshEyePointee::isPointed(mesh, model_matrix, ox,
-                oy, oz, dx, dy, dz);
-        if (data.pointed()) {
-            return data.hit();
-        }
+    std::unique_ptr<Mesh> mesh(scene_object->render_data()->mesh()->createBoundingBox());
+    EyePointData data = MeshEyePointee::isPointed(*mesh, model_matrix, ox, oy, oz, dx, dy, dz);
+    if (data.pointed()) {
+        return data.hit();
     }
+
     return glm::vec3(std::numeric_limits<float>::infinity());
 }
 
--- a/GVRf/Framework/jni/objects/mesh.cpp
+++ b/GVRf/Framework/jni/objects/mesh.cpp
@@ -31,7 +31,7 @@
 #include "../gl/gl_program.h"
 
 namespace gvr {
-Mesh* Mesh::getBoundingBox() {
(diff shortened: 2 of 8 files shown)


GVRf/Framework/jni/engine/picker/picker.cpp
```diff
@@ -100,14 +100,12 @@ glm::vec3 Picker::pickSceneObjectAgainstBoundingBox(
...
         float dy, float dz) {
     glm::mat4 model_matrix = scene_object->transform()->getModelMatrix();
-    Mesh* mesh = scene_object->render_data()->mesh()->getBoundingBox();
-    if (mesh != 0) {
-        EyePointData data = MeshEyePointee::isPointed(mesh, model_matrix, ox,
-                oy, oz, dx, dy, dz);
-        if (data.pointed()) {
-            return data.hit();
-        }
+    std::unique_ptr<Mesh> mesh(scene_object->render_data()->mesh()->createBoundingBox());
```
mingsong: Could the mesh possibly become shared in some scenario? How about a shared_ptr?

GVRf/Framework/src/org/gearvrf/asynchronous/GVRAsynchronousResourceLoader.java
```diff
@@ -159,6 +164,8 @@ public void run() {
                         });
                     } catch (Exception e) {
                         callback.failed(e, resource);
+                    } finally {
+                        resource.closeStream();
```
mingsong: I will check the other streams in the GVRAndroidResource and AsyncBitmapTexture scope

GVRf/Framework/jni/engine/picker/picker.cpp
liaxim: - here the mesh is passed to isPointed which doesn't need to keep it and then discarded; shared_ptr would still work but it would imply possible sharing which is not the case here.
- shared_ptr has to be used everywhere applicable or it won't work

GVRf/Framework/jni/engine/picker/picker.cpp
```diff
@@ -100,14 +100,12 @@ glm::vec3 Picker::pickSceneObjectAgainstBoundingBox(
...
+    EyePointData data = MeshEyePointee::isPointed(*mesh, model_matrix, ox, oy, oz, dx, dy, dz);
+    if (data.pointed()) {
```
rahul27: Looks like something I introduced. Thanks for cleaning it up.

GVRf/Framework/jni/objects/mesh.cpp
```diff
@@ -31,7 +31,7 @@
 #include "../gl/gl_program.h"
 
 namespace gvr {
-Mesh* Mesh::getBoundingBox() {
+Mesh* Mesh::createBoundingBox() {
```
rahul27: This call creates a new mesh object every time it is called, when in reality it only needs to be created when the bounding box is updated. We could have a cached copy of the mesh and invalidate it when the bounding box is updated. 

Let me know if you want me to fix it.

liaxim: You can give it a try. You need to check for code that assumes it is getting a new instance and manages this instance's lifecycle.

danke-sra:
This patch looks good. +1

If Mesh allocations can be saved, that will be great, but it also can be done in a separate patch.

PR #426, opened 2016-02-25
Add Statistics line to log and GVRConsole

danke-sra:


danke-sra:
Can this be merged for use? There is a flag that can disable it when needed.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
```diff
@@ -83,6 +86,8 @@
 class GVRViewManager extends GVRContext implements RotationSensorListener {
 
     private static final String TAG = Log.tag(GVRViewManager.class);
+    private static final boolean DEBUG_STATS = true;
```
liaxim: Should this be false by default?

liaxim:
Squash maybe?

danke-sra:
No problem.

GVRf/Framework/src/org/gearvrf/GVRViewManager.java
jason2kim: Let's keep all debug flags off even if overhead is small. We can turn if on whenever we need to collect it.
    protected static final boolean DEBUG_STATS = false; // keep it off

danke-sra:
Updated.

PR #427, opened 2016-02-25
Sync with internal branch

danke-sra:


danke-sra:
Per Tom's request, sync with internal branch.

PR #430, opened 2016-02-25
Fix AsyncResource loading from URL

mingsong:
When a resource is coming from URL, the decoding in AsyncBitmapTexture is not stable and sometimes fails due to OOM error
even the inSampleSize is correctly set to downgrade the source file.
The root cause is the original java BufferInputStream used to wrap the url'ss inputStream  doesn't handle reset() properly
(it resets the offet of the internal buffer well) even the buffer provided is big enough to hold the resource we use in the
failed cases (possibly due to the internal buffer is cleared when resetting the offet).
This change provides a URLBufferedInputStream that resets the http connection everytime we do a reset to that inputStream,
preventing the incorrect offset in the http connection.
In addition, this change also provides an option to resouce loading api to choose between
1. only use internet streaming, and
2. download the resource in local cache,
so that the developer has a much more flexibility to control the consistency vs performance when
doing the resource loading.

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/GVRAndroidResource.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRAndroidResource.java
@@ -22,6 +22,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
 
+import org.gearvrf.utility.Log;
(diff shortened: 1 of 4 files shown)


liaxim:
This has been reviewed and merged to the other branch. Planning to merge here tomorrow.

PR #431, opened 2016-02-26
Refactor Throttler for dependency injection, type checking and readability

danke-sra:
Type inference is used in Throttler to avoid lots of type wildcards (e.g.,
? extends T). This allows static type checking, and avoids raw types and
unchecked casts. This refactoring makes the code more robust and more
readable.

Configurability is also improved. Currently, Throttler is hardcoded for loading
all resources. Though it is not a problem by itself, there are cases
where developers would like to use a different scheduling policy, or simply
load textures synchronously for development. In this patch, the API of the
Throttler is extracted as interface Scheduler, and the Scheduler is run-time
configurable. A sample scheduler SyncScheduler is provided, which loads
resources synchronously and can be used for debugging purposes.

Some static members and fields in the Throttler and Async* classes are
changed to non-static, and the classes are converted into Singletons. Therefore,
their initialization sequence can be better controlled. Now, they do not register
themselves until they are called by AsyncManager.

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/asynchronous/AsyncBitmapTexture.java
+++ b/GVRf/Framework/src/org/gearvrf/asynchronous/AsyncBitmapTexture.java
@@ -15,8 +15,11 @@
 
 package org.gearvrf.asynchronous;
 
-import static android.opengl.GLES20.*;
-import static org.gearvrf.utility.Threads.*;
+import static android.opengl.GLES20.GL_MAX_TEXTURE_SIZE;
+import static android.opengl.GLES20.GL_NO_ERROR;
+import static android.opengl.GLES20.glGetError;
+import static android.opengl.GLES20.glGetIntegerv;
+import static org.gearvrf.utility.Threads.threadId;
 
 import java.io.FileDescriptor;
 import java.io.FileInputStream;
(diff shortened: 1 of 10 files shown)


danke-sra:
Main reason for this patch is to allow a different Throttler to replace the current one when needed (dependency injection). It also makes it easier to allow custom loaders to be added to the framework.

mingsong:
If we need a blocking sync throttler, why not set Throttler::DECODE_THREAD_LIMIT to 1 that will limit the queue size of available threadMangers to at most 1 that only one asyncLoader proc will be running at a time?

danke-sra:
The goal is not just a scheduler with one worker thread -- it is to make it more configurable than that. BTW, if you make Throttler::DECODE_THREAD_LIMIT = 1, it is still not blocking, but queue the task for a background thread.

mingsong:
With DECODE_THREAD_LIMIT = 1, the throttler will only provides one threadManager task into the thread pool, and do a loop inside that task, only after the current loading proc finishes and doesn't time out, it picks another loading proc from the policy group to execute the next one.

danke-sra:
I agree with what you said but this is refactoring. Make it clearer and more robust so hidden bugs can be more easily fixed. I do know in-depth how it works...

mingsong:
Right for refactoring, this provides cleaner and more scalable way. 
There is one use case for a real sync loader like in this change, which is to ignore timeout on purpose. The orignal throttler will create another new threadManager and get rid of the old one if it timeouts, which then becomes a not real seqential executing,

danke-sra:
I still wonder why Throttler sometimes does not callback, and what happens for the pending task? Can we make its logic simpler so corner case can be prevented?

danke-sra:
I've carefully reviewed Throttler again. It doesn't have too many issues, but just two corner cases missed out: 1) does not interrupt non-terminating loaders after timeout, 2) does not handle converter exceptions well. The latter is fixed. If loaders behave well, it works as expected. The refactoring patch fixes its comments about "shameful unchecked warnings" and unsafe type casts.

thomasflynn:
any more comments on this check-in?

liaxim:
Looks good to me

PR #433, opened 2016-02-26
Refactored DirectionalLight into Light class

fredduarte:
- Moved all DirectionalLight code to the Light class;
- Review and check comment:
https://github.com/Samsung/GearVRf/pull/381#issuecomment-175739434

GearVRf-DCO-1.0-Signed-off-by: Frederico Duarte <frederico.sd@samsung.com>


danke-sra:
@fredduarte Everything looks good. I see the shadow parameters in light.h and the comment that they will be moved. It's fine for now, unless we have something to hold shadow configuration. Where do you think they should be moved to?

fredduarte:
We are thinking of moving it to the RenderData. It seems more fitting there because
they are mostly shadow parameters (not just light).

danke-sra:
@fredduarte Sounds good.

thomasflynn:
@vittal any comments?

GVRf/Framework/jni/objects/light.h
```diff
@@ -92,6 +92,155 @@ class Light: public HybridObject {
...
+    float shadow_smooth_size = 0.0;
+    float shadowmap_edges_length = 0.1;
+    float light_ambient_on_shadow = 0.1;
+    float shadow_gradient_center = 2.0;
+    float u_lighting_shade = 0.0;
+    float u_stratified_sampling = 1.0;
+    float u_shadow_smooth_distance = 1.0;
+    float bias = 0.001;
+    // ========== LIGHT parameters END ==========
+
```
Vittalbabu: Some design questions :- 

1. This design mixes two modes -- we have a map to store arbitrary parameters and we have explicitly named parameters. (eg) Position is stored in the map while direction is a named parameter. What controls this decision? As an example, if we decide to define only common parameters (that is used in all light types) as named parameters, then all the shadow specific parameters should be placed in the map. 
2. A higher level design issue is how we plan to use this class? We will have a switch stmt (switch (cameraLight->getLightType()) ) and each case would either get parameters from the map or the named parameters for that light type. As a app developer, I would not know what parameters are applicable for a particular light unless I look at comments in the Light.h file (assuming it is kept up to date with changes). Would it be preferable to derive the different Light types from the Light class so it is clear in the class definition what parameters are applicable for that Light or shadow or ...?

GVRf/Framework/jni/objects/light.h
```diff
@@ -92,6 +92,155 @@ class Light: public HybridObject {
...
+    ShadowMapHandlerMode shadowmap_handler = HIDE;
+    ShadowMapHandlerEdges shadowmap_handler_edges = HANDLER_EDGES_DARK;
+    float shadow_smooth_size = 0.0;
+    float shadowmap_edges_length = 0.1;
+    float light_ambient_on_shadow = 0.1;
+    float shadow_gradient_center = 2.0;
+    float u_lighting_shade = 0.0;
+    float u_stratified_sampling = 1.0;
+    float u_shadow_smooth_distance = 1.0;
+    float bias = 0.001;
```
Vittalbabu: Set these values in ctr.

liaxim: Actually the way it is makes merging so much easier. Why would you like the values set in the ctor?

thomasflynn:
@fredduarte any responses to @Vittalbabu's questions?

fredduarte:
@Vittalbabu 
I agree that it should be a Light class and the light types should
be derived from it and each type would have its specific parameters.
Comments ?

NolaDonato:
I agree there should be a common GVRLight class with subclasses for the various light types. Right now the directional and spot light seem to be merged. There are shader optimizations we can do if we know it is a pure directional light with no cone so we should probably have a separate GVRSpotLight class for the more complex case. Are you also going to make a GVRPointLight for a light with only position?

-Nola

Vittalbabu:
The shader for multiple lights is next on our list -- either @fredduarte or I will work on it. 

I prefer not to recompile shaders if the type or number of lights change due to the high cost in compiling a shader. We could write a shader that allows a max number of lights (a reasonable number like 10 or 20) and we send the parameters for the max number of lights via a uniform -- if we choose max of 10 lights the uniform will always be 10 x size for one light. The actual number of lights will be sent via uniform(s) and the shader will process the required lights.

If user decides to go beyond the limit on the number of lights, we will compile the shader and use it. We always assume a minimum number of lights to avoid recompiling shaders.

Vittalbabu:
A rough sketch of shader for multiple lights  --

(assume we pass in parameters for lights, number of lights for each type as uniforms)

color = vec4(0, 0, 0, 1);
for (i = 0 ; i < spotlightscnt; i++)
 color += spotlightcontribution(i);
for (i = 0; i < dirlightscnt; i++)
color += dirlightcontribution(i);
...

NolaDonato:
What is our performance for 10 lights? Can we sustain this number at 60Hz?

I am inclined to suggest that we have some “fast path” shaders which may take a small number of lights, say 2 or 3, and special case these paths. For example, two directional lights should be on the fast path.

-Nola

NolaDonato:
Have we benchmarked looping in shaders to determine how much performance degradation we get. My understanding is that doing any kind of branching is really slow.

Vittalbabu:
We haven't done any benchmarking -- the only working multiple lights implementation is in my branch which supports the lighting shader used in texture_shader.

If we have a point light and a spot light how would the shader look? My goal was to generalize this to a number like 10 (to allocate enough UBO memory to store parameters for these lights).

Once we have this we can benchmark (for 2 directional lights)

spotslightcnt = 0, dirlightscnt = 2

color = vec4(0, 0, 0, 1);
 for (i = 0 ; i < spotlightscnt; i++)
 color += spotlightcontribution(i);
 for (i = 0; i < dirlightscnt; i++)
 color += dirlightcontribution(i);

vs just dir lights loop

color = vec4(0, 0, 0, 1);
 for (i = 0; i < dirlightscnt; i++)
 color += dirlightcontribution(i);

vs unrolled loop

color = vec4(0, 0, 0, 1);
  color += spotlightcontribution(0);
 color += dirlightcontribution(1);

NolaDonato:
We looked at this at IMVU on low end tablets and found a big performance loss when doing looping or branching in shaders (which is why we did special cases for several of the two and three light combinations that were frequently used). We did not benchmark on mobile phones so I am not sure what you will find there.

-Nola

Vittalbabu:
@fredduarte To summarise, we are in agreement that the Light class should be changed. Your comment capture what we all agree on.

I agree that it should be a Light class and the light types should
 be derived from it and each type would have its specific parameters.

 The other discussion was about shaders for lighting which can be deferred to when lighting is implemented.

PR #442, opened 2016-03-03
Fix crash in monoscopic view applications. The first problem was a ze…

NolaDonato:
…ro value being passed to sqrt when computing the bounding volume. The second problem was a check for framebuffer id not zero in renderCamera which causes important code to be skipped in monoscopic view.

### DIFF
--- a/GVRf/Framework/jni/engine/renderer/renderer.cpp
+++ b/GVRf/Framework/jni/engine/renderer/renderer.cpp
@@ -194,7 +194,7 @@ void Renderer::renderCamera(Scene* scene, Camera* camera, int framebufferId,
 
     bool renderShadow = cameraLight; // TODO reader from scene;
 
-    if ((framebufferId != 0) && !renderShadow) {
+    if (!renderShadow) {
         renderCamera(scene, camera, framebufferId, viewportX, viewportY,
                 viewportWidth, viewportHeight, shader_manager,
                 post_effect_shader_manager, post_effect_render_texture_a,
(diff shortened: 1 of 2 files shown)


GVRf/Framework/jni/engine/renderer/renderer.cpp
```diff
@@ -569,7 +569,6 @@ bool Renderer::isShader3d(const Material* curr_material) {
     case Material::ShaderType::TEXTURE_SHADER:
     case Material::ShaderType::EXTERNAL_RENDERER_SHADER:
     case Material::ShaderType::ASSIMP_SHADER:
```
thomasflynn: looks like you need to rebase.  seems you're missing the lightmap shader that was added recently.

NolaDonato: I followed the instructions for doing the rebase. Do I need to do the pull request over again or am I good?

-Nola

GVRf/Framework/jni/engine/renderer/renderer.cpp
```diff
@@ -194,7 +194,7 @@ void Renderer::renderCamera(Scene* scene, Camera* camera, int framebufferId,
 
     bool renderShadow = cameraLight; // TODO reader from scene;
 
-    if ((framebufferId != 0) && !renderShadow) {
+    if (!renderShadow) {
```
thomasflynn: @fredduarte, do you remember why there was a check for framebufferId != 0?

fredduarte: There was an issue with the monoscopic renderer and the rendering with the shadow map itself when we first did this a while back, but, i guess it can be removed now, since the crash is a critical issue.
I'll work on a adaptation of the shadow map with this renderer.

NolaDonato: Does the gvr-shadows sample program exhibit the rendering problem with the shadow maps if I run it on a monoscopic view?

fredduarte: Yes, it does. I'm trying to figure out why.

PR #443, opened 2016-03-03
Fixed mesh check on scene objects for IO

rahul27:
Currently if an object does not have a mesh. All of its children
are ignored.

This fix now checks the bounding volume and then proceeds to
the children even if the parent object does not have a mesh.


danke-sra:
The change looks reasonable. +1

PR #445, opened 2016-03-04
Fix dock handling on S7

liaxim:


liaxim:
Will have to be adapted for the other repo.

PR #447, opened 2016-03-05
Fixed problem with LOD culling

NolaDonato:
Recent changes to frustumCull() broke LOD because culling exited if everything was inside the frustum instead of testing LOD children individually.

GearVRf-DCO-1.0-Signed-off-by Nola Donato nola.donato@samsung.com

### DIFF
--- a/GVRf/Framework/jni/objects/scene_object.cpp
+++ b/GVRf/Framework/jni/objects/scene_object.cpp
@@ -425,7 +425,7 @@ int SceneObject::frustumCull(Camera *camera, const float frustum[6][4],
                     name_.c_str());
         }
 
-        return 3;
+        if (!using_lod_) return 3;
     }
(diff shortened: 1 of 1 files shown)


mingsong:
The changes is good to fix the missed LOD culling test, but would be better if we can still skip the test for frustum culling for all the children inside the frustum but just do the LOD culling.

NolaDonato:
To be honest, LOD culling does not belong in the SceneObject at all. There should be  way to extend the frustum culling behavior of a scene object, maybe by supplying a callback, instead of adding the feature to the base class. I think LOD should be refactored but I am not sure if we should make this change before SDC.

-Nola

thomasflynn:
if it can be a small enough well enough contained change, we can refactor it.  probably in a separate commit though.

PR #448, opened 2016-03-07
Fix final frame in animation

danke-sra:
If an animation stops at the last frame, it might be warpped to the first frame.
This can be fixed by adding a small value to the duration (1e-6).

The common code in GVRSkinningController and GVRNodeAnimationController is pulled
into a base class, and the fixed is applied on that.

Rename some variables to distinguish time vs ticks in the animation controller(s).

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/animation/keyframe/GVRAnimationController.java
+++ b/GVRf/Framework/src/org/gearvrf/animation/keyframe/GVRAnimationController.java
@@ -1,0 +1,36 @@
+package org.gearvrf.animation.keyframe;
+
+public abstract class GVRAnimationController {
+    private static final float TOL = 1e-6f;
+    protected final GVRKeyFrameAnimation animation;
+
+    public GVRAnimationController(GVRKeyFrameAnimation animation) {
+        this.animation = animation;
+    }
+
+    /**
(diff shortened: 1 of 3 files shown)


liaxim:
Looks good to me

GVRf/Framework/src/org/gearvrf/animation/keyframe/GVRAnimationController.java
```diff
@@ -0,0 +1,36 @@
+package org.gearvrf.animation.keyframe;
+
+public abstract class GVRAnimationController {
+    private static final float TOL = 1e-6f;
+    protected GVRKeyFrameAnimation animation;
```
rahul27: I would make this final unless you plan to implement a setKeyFrameAnimation method.

rahul27:
Tested it. GVRRepeatMode.ONCE animations work fine now.

GVRf/Framework/src/org/gearvrf/animation/keyframe/GVRAnimationController.java
danke-sra: Yeah, I can add 'final'.

danke-sra:
Updated. Added 'final' to the animation variable.

rahul27:
:+1:

PR #451, opened 2016-03-08
Method to apply the light map to the scene

ragner:
Add logic to apply the light map and other texture atlas to the whole scene objects.
GVRTexture may be a texture atlas.

This pull request is result of this following comment:
https://github.com/gearvrf/GearVRf-Demos/pull/15#discussion_r54613966

To test it, I have updated the solar system at:
https://github.com/gearvrf/GearVRf-Demos/pull/15


thomasflynn:
code looks good.  works as advertised.

PR #452, opened 2016-03-08
Remove GLES2.0 from shaders

ragner:
Shaders cleanup removing GLES2.0


thomasflynn:
seems as though you could remove the ifdefs for USE_GLES3 as well

ragner:
Should I also remove the line <code> #define _GVRF_USE_GLES3_  1</code> from <code>jni/util/gvr_gl.h</code>?

thomasflynn:
theoretically, that define would no longer be necessary so it could be removed.

ragner:
Done.

thomasflynn:
ok, code looks like what i would expect now.  just needs testing.

thomasflynn:
builds.  tested with several samples.  things look fine.  nothing blew up :).

PR #453, opened 2016-03-08
Add version command

danke-sra:
- Add version command to print the current GVRf version
- Fix error message of some existing commands

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/debug/ShellCommandHandler.java
+++ b/GVRf/Framework/src/org/gearvrf/debug/ShellCommandHandler.java
@@ -20,6 +20,7 @@ import java.util.List;
 import javax.script.ScriptEngine;
 
 import org.gearvrf.GVRContext;
+import org.gearvrf.GVRVersion;
 import org.gearvrf.debug.cli.Command;
 import org.gearvrf.debug.cli.HelpCommandHandler;
 import org.gearvrf.debug.cli.Shell;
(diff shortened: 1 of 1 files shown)


GVRf/Framework/src/org/gearvrf/debug/ShellCommandHandler.java
```diff
@@ -42,25 +43,31 @@ public ShellCommandHandler(GVRContext gvrContext) {
...
         if (engine == null) {
-            mShell.getOutputConverter().convertOutput("Cannot find the language engine for " + language);
-            return;
+            return "Cannot find the language engine for " + language;
         }
 
         mScriptHandler = new ScriptHandler(mGVRContext, language, engine);
         mShell.setLineProcessor(mScriptHandler);
+
+        return null;
```
thomasflynn: you changed the return type to Object, but only return non-null when the engine is null.  correct?  this is just for outputting an error?

danke-sra: Right. I think return type Object is too broad. Maybe String is better?

thomasflynn: that might make things more clear.  i was initially expecting the engine or some other object to be returned.  at first, i thought it was always returning null until i looked at it closer.

thomasflynn:
otherwise, looks good; works as expected.

PR #455, opened 2016-03-09
Fixing bug for custom shader to avoid generating VBOs for every frame

roshanch:
Custom shader was calling setVertexAttribLoc() which makes vao_dirty -> true which causes generate VBOs for every frame. adding a flag to avoid this.


thomasflynn:
compiled.  tested.  no negative side-effects seen so far.

PR #456, opened 2016-03-09
Add runtime preferences

danke-sra:
For enabling stats without recompiling.


danke-sra:
To enable stats, please push the following file as /sdcard/.gvrf_prefs
```
debug_stats=true
debug_stats_period_ms=1000
stats_format=ANALYSIS
```

Then, "adb logcat -vtime | grep gvrf-stats" will get data for analysis.

danke-sra:
This is for test engineers to enable stats in framework level. After GVRf configuration is centralized, it can be added as an entry of the configuration -- representing the runtime prefs for gvrf.

thomasflynn:
+1 Verified.  worked as expected.

PR #457, opened 2016-03-09
Fixed bug with cylinder construction

NolaDonato:
I believe this to be a Java compiler bug. I fixed it by moving some initialization outside of the inner loop and using two variables instead of one but the code should have worked fine as written. My fix just makes the compiler generate something more correct.


liaxim:
Looks good to me. The fact is somewhat unnerving. I'd be curious to know what happens in the old version if the createBody method was a "private strictfp void" instead.

liaxim:
Actually, how does this bug manifest itself?

thomasflynn:
run the scene-objects example, tap a few times to get to either the cylinder or cone.  What you'll see is that there will be a cylinder (or cone), but with some of its triangles having a point way off in z somewhere.  Nola was seeing this on her kitkat (4.4) note4.

PR #459, opened 2016-03-10
Fix leaking GVRViewSceneObjects

liaxim:
Previously the registered drawFrameListener was keeping them around forever.


mingsong:
Looks good. Surprised that previously we always keep a surfaceTexture and surface around all the time once registered.

PR #460, opened 2016-03-10
Fix shadow mapping in monoscopic

liaxim:


GVRf/Framework/jni/shaders/posteffect/shadow_shader.cpp
```diff
@@ -492,4 +487,15 @@ void ShadowShader::render(const glm::mat4& mvp_matrix_cam,
 
 }
 
+void ShadowShader::updateViewportInfo(int width, int height) {
+    if (0 == fbo_light && 0 == fbo_camera) {
```
danke-sra: Does it mean if fbo_light and fbo_camera have been initialized (!= 0), then the viewportWidth and viewportHeight cannot be changed?

liaxim: Yes, I left the exercise for another time when I have a case that needs that. I rather study it first. Still better than before's hardcoded 1024x1024.

PR #462, opened 2016-03-10
Buffer overflow fix

liaxim:
Mark's issue "Gear-VR-Framework/GVRf/Framework/jni/objects/vertex_bone_data.cpp, line 53 (Buffer Overflow)"

### DIFF
--- a/GVRf/Framework/jni/objects/vertex_bone_data.cpp
+++ b/GVRf/Framework/jni/objects/vertex_bone_data.cpp
@@ -50,8 +50,15 @@ int VertexBoneData::getFreeBoneSlot(int vertexId) {
 }
 
 void VertexBoneData::setVertexBoneWeight(int vertexId, int boneSlot, int boneId, float boneWeight) {
-    boneData[vertexId].ids[boneSlot] = boneId;
-    boneData[vertexId].weights[boneSlot] = boneWeight;
+    if (BONES_PER_VERTEX <= boneSlot || 0 > boneSlot) {
+        FAIL("index out of bounds; boneSlot: %d", boneSlot);
+    }
+    if (MAX_BONES <= boneId || 0 > boneId) {
+        FAIL("index out of bounds; boneId: %d", boneId);
+    }
+    BoneData& boneDataElement = boneData.at(vertexId);  //"at" performs bounds-checking
+    boneDataElement.ids[boneSlot] = boneId;
+    boneDataElement.weights[boneSlot] = boneWeight;
 }
 
 void VertexBoneData::normalizeWeights() {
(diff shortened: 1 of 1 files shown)


danke-sra:
Because of GPS limitation, atmost 4 bones are used for each vertex. So normally this is not checked. Is there any model that uses more than 4?

thomasflynn:
4 bones is pretty common.  this turned up in the automated scan for security vulnerabilities.  while >4 may not be common, if we at least guard against it, then 1) there is no risk of buffer overflow 2) it won't show up in future scans.

GVRf/Framework/jni/objects/vertex_bone_data.cpp
```diff
@@ -50,8 +50,15 @@ int VertexBoneData::getFreeBoneSlot(int vertexId) {
...
 void VertexBoneData::setVertexBoneWeight(int vertexId, int boneSlot, int boneId, float boneWeight) {
-    boneData[vertexId].ids[boneSlot] = boneId;
-    boneData[vertexId].weights[boneSlot] = boneWeight;
+    if (BONES_PER_VERTEX <= boneSlot || 0 > boneSlot) {
+        FAIL("index out of bounds; boneSlot: %d", boneSlot);
+    }
+    if (MAX_BONES <= boneId || 0 > boneId) {
+        FAIL("index out of bounds; boneId: %d", boneId);
+    }
+    BoneData& boneDataElement = boneData.at(vertexId);  //"at" performs bounds-checking
```
danke-sra: use of 'at' could throw an exception, which is probably not what we want. and it is proven to be 5x slower than [] operator. (http://stackoverflow.com/questions/9376049/vectorat-vs-vectoroperator) 

I would think it's better to use an "unlikely(..)" compiler directive to check for the bound as other things.

GVRf/Framework/jni/objects/vertex_bone_data.cpp
```diff
@@ -50,8 +50,15 @@ int VertexBoneData::getFreeBoneSlot(int vertexId) {
 }
 
 void VertexBoneData::setVertexBoneWeight(int vertexId, int boneSlot, int boneId, float boneWeight) {
-    boneData[vertexId].ids[boneSlot] = boneId;
-    boneData[vertexId].weights[boneSlot] = boneWeight;
+    if (BONES_PER_VERTEX <= boneSlot || 0 > boneSlot) {
```
liaxim: It is natural to me.

danke-sra: There is some coding style blog about readability: https://theholyjava.wordpress.com/2012/12/15/what-is-clean-code-quotes. FYI. It'd better not make it less natural for anyone who might read this code.

GVRf/Framework/jni/objects/vertex_bone_data.cpp
liaxim: Done

danke-sra:
Looks good. Thanks.

PR #464, opened 2016-03-11
Splitting the renderRenderData function

ragner:
Hi all,
i would like to do a cleanup at the Renderer class to reduce its complexity and the size of some functions.

The number of shaders is growing up and the renderRenderData(...) function
is growing up also. So I made a initial cleanup on it, still on progress.

In my opinion, we should think how to manage the shaders to avoid a huge "switch case" in renderRenderData(...). Maybe this "switch case" should be replaced by something like:

curr_material->shader()->render(mvp_matrix, render_data, ...);

But, I think that to implement something like this we should have a abstract/base class for the shaders.

Does anybody have any comments about it?

### DIFF
--- a/GVRf/Framework/jni/engine/renderer/renderer.cpp
+++ b/GVRf/Framework/jni/engine/renderer/renderer.cpp
@@ -641,175 +641,195 @@ void Renderer::renderRenderData(RenderData* render_data,
         const glm::mat4& view_matrix, const glm::mat4& projection_matrix,
         int render_mask, ShaderManager* shader_manager, int modeShadow) {
 
-    if (render_mask & render_data->render_mask()) {
-        if (render_data->offset()) {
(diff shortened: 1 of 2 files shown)


NolaDonato:
I agree that cleanup is needed.  That switch statement is a mess. I think a base class for shaders is a good idea.

Why do we have custom shaders in a table and builtin shaders with specific routines? Why not just have a general table of shaders. This table can be pre-populated with the builtin shaders and we can use constant indices to look them up. Then you can ask the ShaderManager to render a shader based on ID.

Ideally there should be no built-in shaders. We should be able to design it so that whatever we do for custom shaders will allow us to create the existing ones that are built-in now.

-Nola

liaxim:
Looks good and like an improvement to me.

---

@NolaDonato Have no good answer for you. We should try do what you suggested.

PR #465, opened 2016-03-12
GVRTextViewSceneObject: Add REALTIME as a new frequency

danke-sra:
The GVRTextViewSceneObject is only refreshed at the specified frequency if
anything has changed.

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/scene_objects/GVRTextViewSceneObject.java
+++ b/GVRf/Framework/src/org/gearvrf/scene_objects/GVRTextViewSceneObject.java
@@ -40,11 +40,12 @@ import android.view.ViewGroup.LayoutParams;
 import android.widget.LinearLayout;
 import android.widget.TextView;
 
-@Deprecated
 public class GVRTextViewSceneObject extends GVRSceneObject {
+    private static final int REALTIME_REFRESH_INTERVAL = 1;
     private static final int HIGH_REFRESH_INTERVAL = 10; // frames
     private static final int MEDIUM_REFRESH_INTERVAL = 20;
     private static final int LOW_REFRESH_INTERVAL = 30;
+    private static final int NONE_REFRESH_INTERVAL = 0;
 
     private static final float DEFAULT_QUAD_WIDTH = 2.0f;
     private static final float DEFAULT_QUAD_HEIGHT = 1.0f;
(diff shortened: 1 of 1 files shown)


ragner:
Sorry, but in my opinion,
we could keep improving the GVRTextView to simplify the use of it and keep GVRTextViewSceneObject  deprecated.

danke-sra:
OK. I will check with others about the de-deprecation. I also prefer GVRTextView because it is lighter.

liaxim:
@ragner 

We keep falling back to the GVRTextViewSceneObject because it is just more straightforward and easy to use. This is the only reason; I am not fond of it either but was basically forced to use it in certain situations.

As long as the construction is a two-phase thing for GVRViewSceneObjects we will have to keep the GVRTextViewSceneObject around. Also I think we might have to view a textWidget somewhat differently than any-android-view (we may switch to an implementation that is not based on Android views for text for example). In a way the GVRViewSceneObject and GVRTextViewSceneObject are complementary, not competing with each other.

I personally think the de-deprecation is reasonable considering that for text specifically we keep using the deprecated one more frequently.

liaxim:
Overall the code looks good to me. @ragner Since you voiced your concern I will wait for us all to come to an agreement before merging it.

rahul27:
Think it would be cleaner to switch out the internals of one and replace it with the other (whichever is better). Having two ways to create a textview is not ideal.

liaxim:
More or less tried exactly that. We had apps that deadlocked (GLSurfaceView gets callbacks on the UI thread and obtains locks that its GL thread locks on too) if you just start waiting on the UI thread while on the GL thread which a swap like the suggested would lead to. Nowadays we can create everything off the GL thread but not all apps take advantage of that and onInit is still happening on the GL thread. Once we hide the GL thread from the apps things will be much easier.

ragner:
@liaxim  I agree that the use of GVRTextViewSceneObject is easier than GVRViewSceneObject.

With the merge of https://github.com/Samsung/GearVRf/pull/468 is possible to improve GVRViewSceneObject/GVRTextView to simplify its construction to one phase too.

liaxim:
@ragner Are you ok with merging this the way it is?

ragner:
Yes, it is ok.

PR #468, opened 2016-03-14
Allow GVRTextView to be instantiated in any thread

danke-sra:


danke-sra:
I think something like this may get GVRTextView to work on any thread.

ragner:
Looks good to me
:+1:

ragner:
Maybe after this patch,
we could improve GVRView's implementation adding new constructors
something like GVRTextView(GVRScene scene, float width, float height, CharSequence text)
 that, internally, it creates its own GVRViewSceneObject adding it to the scene.

GVRf/Framework/src/org/gearvrf/GVRActivity.java
```diff
@@ -319,12 +319,17 @@ boolean updateSensoredScene() {
      * @param view Is a {@link GVRView} that draw itself into some
      *            {@link GVRViewSceneObject}.
      */
-    public void registerView(View view) {
+    public void registerView(final View view) {
         /* The full screen should be updated
            otherwise just the children's bounds may be refreshed. */
         mRenderableViewGroup.setClipChildren(false);
```
rahul27: Don't you also have to add this line to the runnable?

danke-sra: probably, thanks for the catch.

ragner: I have tested it with RenderableView sample and it is working fine ...

liaxim: When in doubt, make the android view calls on the main thread. Since we don't really know on which thread registerView may be called, I vote for the main thread.

rahul27: It needs to go into the runnable. It crashes when called from the gl thread:

````java
             System.err  W  android.view.ViewRootImpl$CalledFromWrongThreadException: Only the original thread that created a view hierarchy can touch its views.
                         W      at android.view.ViewRootImpl.checkThread(ViewRootImpl.java:7648)
                         W      at android.view.ViewRootImpl.invalidateChildInParent(ViewRootImpl.java:1107)
                         W      at android.view.ViewRootImpl.invalidateChild(ViewRootImpl.java:1102)
                         W      at android.view.View.invalidateInternal(View.java:12959)
                         W      at android.view.View.invalidate(View.java:12923)
                         W      at android.view.ViewGroup.setClipChildren(ViewGroup.java:3743)
                         W      at org.gearvrf.GVRActivity.registerView(GVRActivity.java:325)
                         W      at org.gearvrf.scene_objects.view.GVRTextView.<init>(GVRTextView.java:33)
                         W      at org.gearvrf.scene_objects.view.GVRTextView.<init>(GVRTextView.java:61)
                         W      at org.gearvrf.scene_objects.view.GVRTextView.<init>(GVRTextView.java:46)

````

danke-sra:
Updated. In registerView, moved everything to UI thread.

liaxim:
Have a question about the title of the commit. Should it actually read "Allow GVRViewSceneObject to be instantiated in any thread"? Or am I misunderstanding something?

danke-sra:
I did test constructing GVRTextView from off the GL thread and off the UI thread and it worked. So it's not a mistake I guess.

liaxim:
Is this discussion outdated or not applicable: https://groups.google.com/forum/?fromgroups=#!topic/android-developers/e-Rz0qMy0K8? And more specifically the posts by Dianne Hackborn and Romain Guy.

danke-sra:
If possible, the user should try to create GVRTextView from the UI thread. But it seems not absolutely needed now.

liaxim:
Looks good to me. Allows creation of GVRViewSceneObjects off the GL thread plus other improvements.

I completely disagree with the statement that it is safe to create a GVRTextView on a thread other than the one associated with the view hierarchy. Key developers working on the Android UI framework explicitly say that you are not supposed to even instantiate views on other threads even though it may seem to work. The official javadoc issues a much stronger blanket statement: ""Note: The entire view tree is single threaded. You must always be on the UI thread when calling any method on any view."".

Will let others chime in. Planning to merge this in the afternoon.

danke-sra:
This patch actually leaves the single thread protection issue open. It doesn't claim to solve that but just fix existing thread-affiliation issue(s).

PR #469, opened 2016-03-16
Create a "common" project the framework depends on

liaxim:
A proposal first and foremost. I find it impossible to keep the two repos we have in sync. I am getting behind with my own commits and from what I see, pretty much everybody else is behind too.

I think an incremental approach has more of a chance to work vs a big refactoring by one or two people. Once the basic structure is in place it will be fairly easy for anybody who feels like it to chip in with a file or two. The two repos share a lot of code and whatever we can move without major refactoring, we should as soon as possible.

There are some fundamental classes like GVRContext, GVRScene* - once these are addressed somehow the rest will be much easier. I plan to introduce interfaces in the common project for GVRContext, GVRActivity as in some cases moving implementations is not going to be straightforward immediately. Still, the current process is not working and we have to do something.

Eventually we will reduce the framework project to what really needs major work and we will know better how much and what it is instead of suspecting certain classes. These can be addressed as the time permits or worst case, continue with the existing process but at a much smaller scale.


liaxim:
This clearly is not gathering traction so I will abandon it.

PR #470, opened 2016-03-17
Fix issue 461: LoadModel misses textures

mingsong:
The root reason was the flaw of future textures that causes thread pool deadlocked when
doing async resource loading. This change addresses this directly when loading a model
from JassimpLoader by using callback mechanism instead of previous future texture pending
mechanism. A more fundermental redesign and implementation of the whole 
async resource loading infrastructure and thread scheduling is ongoing and coming soon, which will address all types of problems and performance issues we previously have seen.

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/jassimp2/GVRJassimpSceneObject.java
+++ b/GVRf/Framework/src/org/gearvrf/jassimp2/GVRJassimpSceneObject.java
@@ -2,10 +2,10 @@ package org.gearvrf.jassimp2;
 
 import java.io.FileNotFoundException;
 import java.io.IOException;
-import java.util.concurrent.Future;
 
 import org.gearvrf.FutureWrapper;
 import org.gearvrf.GVRAndroidResource;
+import org.gearvrf.GVRAndroidResource.TextureCallback;
 import org.gearvrf.GVRContext;
 import org.gearvrf.GVRMaterial;
 import org.gearvrf.GVRResourceVolume;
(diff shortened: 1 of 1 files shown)


mingsong:
@rahul27 As I talked to you last Friday, there are two ways to bypass the problem in 461: one is to use sync loading in GVRJassimpSceneObject.java that runs the texture loading and  setting on glthread; and the other is like this change which uses callbacks instead of pending on future values.
Since async loading is error prone from previous experiences and performance also degrades some times, I am working on the comprehensive solution for the whole infrastructure. For now, this change will be enough to address 461.

liaxim:
Looks good to me. @rahul27 Please confirm this resolves the problem you are experiencing.

NolaDonato:
Mingsong,

I agree that using callbacks right now is the best way to fix this in the short term. SDC looms…

We can go back to using the Future infrastructure when you have stabilized it and it no longer deadlocks. Thanks for researching the asset loading and for finding a speedy fix.

-Nola

GVRf/Framework/src/org/gearvrf/jassimp2/GVRJassimpSceneObject.java
```diff
@@ -145,15 +146,39 @@ private GVRSceneObject createSubSceneObject(
...
-                    Future<GVRTexture> futureDiffuseTexture = getGVRContext()
-                            .loadFutureTexture(resource);
-                    meshMaterial.setMainTexture(futureDiffuseTexture);
+                    
+                    TextureCallback callback = new TextureCallback() {
+                        @Override
+                        public void loaded(GVRTexture texture,
+                                GVRAndroidResource ignored) {
+                            meshMaterial.setMainTexture(texture);
+                            Log.e(TAG,
```
rahul27: Log.i?

rahul27:
Verified using the sample app - It works.

I agree that we need to look into the threading long term because even with this fix I do see the occasional missing texture but am not able to reproduce it consistently (maybe a race condition). 

But for now now this should work.

PR #472, opened 2016-03-18
GVRBitmapTexture javadoc update

liaxim:
The previous pattern of calling recycle() immediately after construction
is actually incorrect now and would lead to black textures or other
unexpected results.


thomasflynn:
do we have any samples that use this and call recycle() afterwards?  if so, we should also change them.

liaxim:
Found only one: gvrsixaxissensortest_v1_5_0. Will make a pull request for it. It is also not a real sample and to me another candidate for either removal and to be moved under Framework/Tools/.

PR #473, opened 2016-03-20
Fixed recycling of motion events for cursor controllers

rahul27:
Calling recycle() multiple times on the same motion event results in
a runtime exception. This patch prevents the exception for certain
code paths.


liaxim:
Looks good to me. Planning to merge after lunch.

PR #475, opened 2016-03-21
Add life-cycle events for GVRSceneObject and GVRScene

danke-sra:
- Create ILifeCycle events which define interface for life-cycle events
  common to all event-receiving classes (GVRScript, GVRScene, GVRSceneObject, etc.)
- Add onStep handler for GVRScene to recursively deliver the event to scene objects.
- Clean up GVRSceneObject constructors: use a general constructor to handle all specific
  constructors, and only call super from one constructor. This ensures there is only one
  exit-point of all constructors, and any common code can be easily added.
- Allow GVRSceneObject to be scriptable and receive ISceneObjectEvents via GVREventReceiver.

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/GVRScene.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRScene.java
@@ -24,8 +24,19 @@ import org.gearvrf.debug.GVRConsole;
 import org.gearvrf.script.IScriptable;
 import org.gearvrf.utility.Log;
 
-/** The scene graph */
-public class GVRScene extends GVRHybridObject implements PrettyPrint, IScriptable, ISceneEvents, IEventReceiver {
+/**
+ * The scene graph.
+ *
+ * It receives events defined in {@link ISceneEvents}. To add a listener to these events, use the
+ * following code:
+ * <pre>
+ *     ISceneEvents mySceneEventListener = new ISceneEvents() {
+ *         ...
+ *     };
+ *     getEventReceiver().addListener(mySceneEventListener);
+ * </pre>
+ */
+public class GVRScene extends GVRHybridObject implements PrettyPrint, IScriptable, IEventReceiver {
     @SuppressWarnings("unused")
     private static final String TAG = Log.tag(GVRScene.class);
 
@@ -62,6 +73,8 @@ public class GVRScene extends GVRHybridObject implements PrettyPrint, IScriptabl
 
         setMainCameraRig(cameraRig);
         setFrustumCulling(true);
+
+        getEventReceiver().addListener(mSceneEventListener);
     }
 
     private GVRScene(GVRContext gvrContext, long ptr) {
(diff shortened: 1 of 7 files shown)


danke-sra:
Can be tested with https://github.com/gearvrf/GearVRf-Demos/pull/52

liaxim:
I believe we should not expand the public APIs unless absolutely necessary and we do mean it. Are these interfaces required to be implemented by the class itself? Is there no other option?

From a developer's perspective it will be inevitably confusing when they start seeing methods like getEventReceiver, onInit, etc being available on their objects. I doubt these are supposed to be called by end-users but the methods will be there, expanding the public API.

Is GVREventReceiver an implementation detail or entity that end-users can/should use for something?

Code looks pretty good otherwise but please try to hide the methods from the public API space.

danke-sra:
This is a big topic to discuss. The design considers many factors including extensibility, efficiency, reduction of boilerplate code, and also possible use and misuse of public classes and methods by developers. It is hard to go over all of them here. I can say I agree that public API should not be abused or confusing by developers.

Can you please specify an alternative implementation for discussion?

danke-sra:
For API abusing issue, let me propose another way of thinking.

1) Java language's "public" keyword is not designed to distinguish a framework's API and internal methods. So it's essentially impossible to make sure every public API is "public", and every internal class/method is non-"public". Consider a simple framework with two layers stored in package A and B. B is not designed to be accessible to developers but is accessible by package A. Should you declares the classes/methods in B as "public" or not "public"? If "public", would it be understood as public API? If not "public", how is it accessible by package A?

2) In multithreading environments, some public methods are protected by callers' thread ID / process ID and user privileges. They only govern interface level. If one abuses (or just re-use) some internal interface that is meant for system, it is not harmful, if they only use it for their apps, instead of gaining access into system resources.

3) Difference of misuse and abuse. Users might occasionally find a creative way to access a method to do something not recommended by framework designers. Sometimes, it is valid but we are not aware of them. Sometimes, it is not so reliable as using the recommended API. Developers do not always know what is the case. If the exposed API could be misused, it should be better documented (I doubt a library without documentation can guarantee all functionalities be properly used without misuse). Still, the latter 'misuse' case should be distinguished from 'abuse' or gives user privilege that they don't have, which is more serious. Occasional misuse of API happens, even in APIs that are supposed to be used by users. There could be issues in calling sequence, parameter range, thread affiliation, etc.

Back to the code, the API "getEventReceiver()" is a public API, for users to attach their own event handlers (I*Events interface) to customize a class without subclassing it.

liaxim:
Ok, for example ISceneEvents in GVRScene. Could be moved to an inner class. If we can't avoid the IEvents interface then we could add a method to it to query for specific interfaces: like IEvents getEventSink(InterfaceClass.class) or something. Returns an instance of InterfaceClass if applicable, manager calls into it with applicable events. GVRScene.getEventSink would return an instance of its inner class implementing ISceneEvents when asked for it.

We are already using reflection so the getEventSink method could be an implicit protocol to avoid tagging classes and exposing even IEvents. That could be considered too loose by some.

Or if this events framework is so fundamental maybe IEvents belongs to the base class.

Also the distribution of events seems a bit wide to me (the recursive part). There is a danger we may start spending more and more time broadcasting and propagating events.

danke-sra:
I think the inner class can solve some issues, which are possible access of the event handlers by developers. However, event handlers are sometimes not sensitive code to protect -- some times, users even want to call those event handlers to invoke some functionality -- like adding a key release event for working around some issues, etc.

When an inner class is added, developers might not be able to subclass those inner class and outer classes in a consistent and convenient way, which may make code hard to maintain. Also, the outer class "might" become a proxy for inner class, resulting in abundance of less useful code -- and make the working code hard to debug and checked.

So I doubt the inner class is needed here. However, if you could provide a prototype, we can look at it more thoroughly.

liaxim:
We should actively discourage third-party developers from deriving from key classes like GVRSceneObject and GVRScene.

I have noted what I had to. Will let other reviewers decide.

danke-sra:
Updated per code review

liaxim:
Looks nice. Tested with the sample.

Planning to merge it sometime in the afternoon if there are no objections by then.

PR #478, opened 2016-03-28
getVAOId parameter cleanup

liaxim:


Vittalbabu:
Looks good.

I am working on a different cleanup of vaoid -- avoid deleting it if the mesh becomes dirty.

Do we have any sample where we set the GVRMesh  vertices directly in the app? A grep couldn't find any and I am wondering if we have any old samples that do that.

liaxim:
I am not aware of any sample that uses setVertices. Since it is a public API some people may have used it..

Since Vittal gave the thumbs up I am planning to merge this lunch time.

PR #480, opened 2016-04-01
GVREventManager: Always check method name and parameter types in sendEvent

danke-sra:
If user code specifies a wrong method name, wrong number of parameters,
or a parameter of a wrong type, throw RuntimeException, instead of ignoring
it silently.

This was implemented in the initial version, but was skipped after listeners
are supported. Now a slightly different version is added that works in the
listener case as well as the subtyping case.

This patch also fixes issues when the method in the event interface contains
boxable types. Due to the use of a variadic method for sending events, Java
boxes arguments like boolean as Boolean. The event-checking code needs to
allow a value of type Boolean for the type boolean. There is no simple Java
API for doing that, so a method is added to handle it.

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/GVREventManager.java
+++ b/GVRf/Framework/src/org/gearvrf/GVREventManager.java
@@ -117,13 +117,11 @@ public class GVREventManager {
         // Set to true if an event is handled.
         boolean handledSuccessful = false;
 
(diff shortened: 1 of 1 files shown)


liaxim:
Looks good to me but please add the "GearVRf-DCO-1.0-Signed-off-by:" section.

PR #481, opened 2016-04-01
Add more android activity events and an early init event

danke-sra:
- Activity events can be used by framework plugins which need to respond to android activity callbacks without subclassing GVRActivity.

- onEarlyInit event can be used to perform tasks before application onInit. It can be used for initializing plugins.

- Since more events are added to the interfaces, null implementations are also provided as bases for easier extension.


liaxim:
Looks good to me but please add the "GearVRf-DCO-1.0-Signed-off-by:" section.

PR #484, opened 2016-04-04
Design, refactor and implementation of async resource loading

mingsong:
1. Fix the deadlock of thread pool by using callback mechanism instead of pending future textures in GVRJassimpSceneObject
2. Use a resouce pool to hold the exsiting GVRAndroidResource instead of creating redundant ones in GVRResourceVolume and GVRAndroidResource
3. Optimize GVRAsynchronousResourceLoader to only checking one-time of compressed loader for many requests on the same resource
4. Fix concurrent safty issue in Throttler and ResourceCache
5. Optimize growing thread pool for more efficiency

GearVRf-DCO-1.0-Signed-off-by: Mingsong Bi
mingsong.bi@samsung.com

### DIFF
--- a/GVRf/Framework/src/org/gearvrf/GVRAndroidResource.java
+++ b/GVRf/Framework/src/org/gearvrf/GVRAndroidResource.java
@@ -22,11 +22,12 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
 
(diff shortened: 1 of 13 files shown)


liaxim:
Needs the "GearVRf-DCO-1.0-Signed-off-by:" section.

mingsong:
Done.

PR #487, opened 2016-04-06
Fix a NPE in GVRAndroidResource when closing the input streams

mingsong:
In addition, change debugState to streamState as it is not just for debugging purpose.


mingsong:
It is ready for review.

danke-sra:
Looks good

PR #488, opened 2016-04-07
Multilight

NolaDonato:
Add support for multiple lights to GearVRF. New classes added are: GVRShaderTemplate, GVRLightTemplate, GVRPhongShader, GVRPhongLight, GVRPhongSpotLight, GVRPhongPointLight.. Now you can attach a GVRLightTemplate to any GVRSceneObject to put a light in the scene that will move with the scene object. The C++ custom shader implementation has been updated to support bone matrices, lights and user defined vertex attributes. GVRShaderTemplate makes generating custom shaders easier and more flexible. Shadow mapping with multiple lights is not supported yet.

GearVRf-DCO-1.0-Signed-off-by: Nola Donato nola.donato@samsung.com and Roshan Chaudhari roshan.c@samsung.com


GVRf/Framework/jni/objects/mesh.cpp
```diff
@@ -177,6 +177,34 @@ void Mesh::getTransformedBoundingBoxInfo(glm::mat4 *Mat,
     }
 }
 
+void Mesh::bindVertexAttributes(GLuint id)
```
Vittalbabu: I would rename id as programId. Makes clear what it represents.

GVRf/Framework/jni/objects/light.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+
+/*
+ * Loads the uniforms associated with this light
+ * into the GPU if they have changed.
+ */
+void Light::render(int program) {
+    auto it = dirty_.find(program);
+
+    if (it != dirty_.end() && !it->second)
```
Vittalbabu: Add braces in this file.

PR #489, opened 2016-04-07
Enable the use of custom shaders off the GL thread

liaxim:
Shouldn't be merged before Nola's multilight change. I will update it afterwards but will still follow the same approach. A good test case is the model-viewer sample which uses a bunch of custom shaders. I verified almost all samples that use custom shaders (except immersivepia, because it is enormous).

I need this for another change that moves onInit and onStep off the GL thread.

### DIFF
--- a/GVRf/Framework/jni/shaders/material/custom_shader.cpp
+++ b/GVRf/Framework/jni/shaders/material/custom_shader.cpp
@@ -29,91 +29,213 @@
 #include <sys/time.h>
 
 namespace gvr {
-CustomShader::CustomShader(std::string vertex_shader,
-        std::string fragment_shader) :
-        program_(0), u_mvp_(0), u_mv_(0), u_right_(0), u_view_(0),
-        texture_keys_(), attribute_float_keys_(),
-        attribute_vec2_keys_(), attribute_vec3_keys_(), attribute_vec4_keys_(),
-        uniform_float_keys_(), uniform_vec2_keys_(), uniform_vec3_keys_(),
-        uniform_vec4_keys_(), uniform_mat4_keys_() {
-    program_ = new GLProgram(vertex_shader.c_str(), fragment_shader.c_str());
-    u_mvp_ = glGetUniformLocation(program_->id(), "u_mvp");
-    u_view_ = glGetUniformLocation(program_->id(), "u_view");
-    u_mv_ = glGetUniformLocation(program_->id(), "u_mv");
-    u_mv_it_ = glGetUniformLocation(program_->id(), "u_mv_it");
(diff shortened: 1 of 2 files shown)


thomasflynn:
any comments on this change?

GVRf/Framework/jni/shaders/material/custom_shader.cpp
```diff
@@ -29,91 +29,213 @@
...
         const std::vector<Light*> lightList, Material* material) {
-    for (auto it = texture_keys_.begin(); it != texture_keys_.end(); ++it) {
-        Texture* texture = material->getTextureNoError(it->second);
-        // If any texture is not ready, do not render the material at all
-        if (texture == NULL || !texture->isReady()) {
-             return;
+    initializeOnDemand();
+
+    {
+        std::lock_guard<std::mutex> lock(textureVariablesLock_);
```
mingsong: Previously I added a iteration here in customershader's rendering pass over all textures for checking the readiness, and now combined with the lock for guarding the texture set against the addTextureKey(), there might be some unexpected overhead added to each frame. Originally the check here is to fast return if any texture is not  ready to avoid executing between old line 131 and line 186, with change maybe we can get rid of this early check for the whole list and put the individual check in f_find() in new line 101?

mingsong: Or we can use a global boolean to tell if all textures are ready in the material for a given customer_shader's textureVariables list, but we may also need to take care of the case when new texture keys are dynamically added in customer_shader.

liaxim: We need to make somehow the access to the std collection thread-safe. This change was the most conservative. We may even think about eliminating the loop itself and/or employing a copy-on-write scheme for the collection(s), etc.

The very same loop takes wall clock time in the range of 0.0015-0.003ms without locking and in the range of 0.006-0.02ms with the locking so definitely there is overhead, especially in the worst case. But I find this matter to be applicable to any place we use locks in the framework - we must ensure correctness first and worry about performance once a bottleneck is identified.

--

If the check is in f_bind then you can end up with some textures bound and some not; the current behavior is that if any texture is not ready we short-circuit the rendering so I am not sure the check can be moved to f_bind.

-

Let me know how big of a problem is this for you and is it something you want tackled as part of this pull request. I probably would try a copy-on-write variant where I would push the overhead to the mutator methods that presumably are not on the gl thread. And they will not be in the near future.

mingsong: Copy on write definitely is good which I was about to mention. For this change, I think we can keep this as it is since it keeps the correctness, we may discuss later how to optimize those per-frame overheads including other places as separate changes.

PR #492, opened 2016-04-11
Fix lua crash upon some syntax errors

danke-sra:


danke-sra:
Can be verified by typing a lua statement with syntax errors. Some pending exception was not checked in JNI in original implementation.

liaxim:
Where is the source tree that you built this .so from? Thanks.

danke-sra:
The source is currently hosted here: https://github.com/danke-sra/jnlua-android

PR #502, opened 2016-04-14
Simple fix that prevents the thread pool from getting exhausted...

liaxim:
if multiple gvractivities get created and destroyed very quickly.

Making the thread-pool per activity instance is going to be quite disruptive and will be examined further in the future.


mingsong:
Looks good to me.

PR #476, opened 2016-03-23
androidstudio migration: framework

jason2kim:

PR #504, opened 2016-04-15
androidstudio migration: framework; rebased

liaxim:
This is a rebase of https://github.com/Samsung/GearVRf/pull/476; as I can't directly modify Jason's branch I have to create a new pull request.

All that this does is:
1. Rebase Jason's commit - minor merge conflict in Eclipse project files resolved.
2. New commit that moves the light-related files.

Verified that the Android Studio and the Eclipse instructions from https://github.com/Samsung/GearVRf/pull/476 still work.


liaxim:
Rebased again: moved the gdx plugin related files.

thomasflynn:
@NolaDonato , @roshanch , could you guys please test this?  thanks.

liaxim:
The multilight sample builds and runs when using Eclipse with the new folder structure. Somehow icon.png is still missing from the sample though.

NolaDonato:
I pushed this to the multilight branch of my GearVRf-Demos repo. Do I have to issue another pull request for it to get into main?


-Nola

liaxim:
I guess something happened and it never updated the pull request. A new one will be necessary.

liaxim:
So what we know so far; before launching Android Studio you need these:

export ANDROID_NDK=<location>
export ANDROID_NDK_HOME=<location>
export PATH=$PATH:$ANDROID_NDK

and very importantly, manually download Gradle 2.2.1 and configure Android Studio to use it.

NolaDonato:
So far I have failed miserably to create a desktop launcher for Android Studio. You can add it to the applications menu and that will launch properly but if you right click on that icon and say "add to desktop" you will get a launcher that does not correctly set the environment variables.


I put them into my .bashrc and still the desktop launcher can't find them (but the application launcher menu does it right).


-Nola

liaxim:
In the desktop launcher try this (note the leading env that we didn't use):

"env VAR1=abcd VAR2=efg ..../studio.sh"

liaxim:
Tried Android Studio 2.0 - it applied automatically couple of changes and the sample and the framework built and run fine. I actually feel better about 2.0..

jason2kim:
I also think tool version up must be smooth anytime. Android Studio 2.1 Preview seems even coming.

jason2kim:
All 3 settings of these NDK paths happen to have the same for my ndk version at this time: ANDROID_NDK, ANDROID_NDK_HOME, and bin. We can either reuse with one literal definition for the 3 or as many as their original purposes.

thomasflynn:
sorry, i should've merged this first.  needs rebase.

aviadmini:
I have build the framework with Android Studio by changing compiler in Applcation.mk and adding build.gradle only. Might be not the cleanest build(folder structure is not as for usual Gradle projects) but it's perfectly compatible with Eclipse ant build and doesn't need any env vars (except Oculus SDK dir var which is either hardcoded into build.gradle or gradle.properties).

liaxim:
@aviadmini Thank you for sharing. What do you think about our approach where we changed the folder structure to match what Gradle wants? I would imagine AS users would like that better.

liaxim:
Rebased

aviadmini:
@liaxim I do like that and it's more than I did. I'm a bit worried about backwards compatibility with Eclipse for people who still want to use it but that should be solvable. I don't think AS-like folder structure is a must especially for a library so IMO both ways are good. My approach was designed to have minimum change to the repo and I'm quite happy with the result. Also I guess users would be extremely happy to see the framework as a single library dependency in their build.gradle and this way it might be easier to roll out new releases (automatically ofc)

liaxim:
Eclipse will still be supported, the project files just reside in a different location. @jason2kim has that covered in the first paragraph of the original pull request: https://github.com/Samsung/GearVRf/pull/476

aviadmini:
Good! 
It's also worth notice that 
```
tasks.withType(JavaCompile) {
    compileTask -> compileTask.dependsOn buildNative
}
```
should be changed to specific build task to depend on native build OR remove it at all:
`build.dependsOn buildNative`
The reasons are that it might break build of a project that has a project(module) dependency on the framework and removing might be an option because rebuilding native code takes some time which is unnecessary if there are no changes(not confirmed though)

thomasflynn:
fwiw, it doesn't look like one can download the 1.5 version of Android Studio anymore (unless someone has a magic link somewhere).  It seems you can only download 2.0.

PR #505, opened 2016-04-15
Fix native crash in libz with compressed texture loader and some further improvement

mingsong:
1. Native crash was due to the concurrent unsafety in stream ops in GVRAndroidResource when sniffing compressed texture loader.
2. Simply the compressed texture loading in async resource loading by introducing a new Async Compressed Texture Factory/Loader. Each compressed texture load request that misses in the cache will now go through Throttler rather than directly processed to prevent redundant loading of different requests.
3. Simply and improve the initialization of all async classes.
4. Offload the texture creating convertion from glthread to the current background thread in AsyncLoader's last step in Throttler.


liaxim:
Looks good. Planning to merge after lunch.

PR #507, opened 2016-04-19
Add back the hasUniform method lost after the last commit

liaxim:


NolaDonato:
I just noticed this last night when I pulled a fresh copy. Thanks for fixing it!


-Nola

PR #510, opened 2016-04-21
Fix rendering objects for one eye only

lassehe:
Skip execution only if render masks do not match.
This fixes regression introduced by commit 59920e2.


liaxim:
Thank you for the fix. Please sign your patch per http://gearvrf.org/bin/view/GearVRF/GearVRfWikiGetInvolved#SubmitPatch.

lassehe:
Sure! Commit message updated accordingly.

liaxim:
Change looks good. Sorry it took us so long but we were in the middle of something. We can merge this as soon as you rebase it - or I can do that in a separate pull request if you prefer. Thanks.

lassehe:
No problem. :) Change rebased.

Issue #511, opened 2016-04-23
GVRScene.clear() removes the camera rig's scene objects


thomasflynn:
if you call clear() on the scene, it also removes all the scene objects that the camera rig sets up (headTransform, left, right, center, etc.).  So, if you clear the scene, you cannot attach anything to the camera rig ever again.

danke-sra:
What's the suggested fix? Is it to keep scene objects in the camera rig?

thomasflynn:
as a hack, i named the camera rig objects and during clear(), i skipped the objects with those names.  i gave them all a prefix so it was easy.  this is what i came up with given approx 2ms worth of thought.  better ideas would be appreciated.

liaxim:
Tom, does the fix look good to you? Would you like to verify first or can we close the issue?

PR #519, opened 2016-04-27
Fix for "GVRScene.clear() removes the camera rig's scene objects"

liaxim:
Should fix https://github.com/Samsung/GearVRf/issues/511; verified using the remote-scripting sample; added a cursor, cleared, added a cursor back.

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRScene.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRScene.java
@@ -110,8 +110,11 @@ public class GVRScene extends GVRHybridObject implements PrettyPrint, IScriptabl
      * Remove all scene objects.
      */
     public void removeAllSceneObjects() {
+        getMainCameraRig().removeAllChildren();
         mSceneObjects.clear();
         NativeScene.removeAllSceneObjects(getNative());
+
+        addSceneObject(getMainCameraRig().getOwnerObject());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


danke-sra:
Looks good.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRScene.java
```diff
@@ -110,8 +110,11 @@ public void removeSceneObject(GVRSceneObject sceneObject) {
      * Remove all scene objects.
      */
     public void removeAllSceneObjects() {
+        getMainCameraRig().removeAllChildren();
         mSceneObjects.clear();
         NativeScene.removeAllSceneObjects(getNative());
+
+        addSceneObject(getMainCameraRig().getOwnerObject());
```
thomasflynn: i'm not fully following this.  probably because it is late.  i guess you remove all sceneobjects from the camera rig, except for the camera sceneobjects, then call clear(), then add back the camera rig?  and somehow clear() doesn't remove the camera sceneobjects that were added to the list of sceneobjects in the scene?

liaxim:
We'll have to discuss this in person tomorrow as I can't quite formulate an answer. What I did in the remote scripting sample testing-wise was: add the cursor, call clear, add the cursor again. Without this fix the cursor would never show up, with this fix it shows up.

thomasflynn:
yeah, I think talking in person will be faster.  sounds like it accomplishes what I was looking for.

thomasflynn:
took a closer look.  calling clear() is only removing the camera rig owner object, which you add back in afterwards.  ok, this looks good and is also cleaner than looking for specially named objects.

PR #520, opened 2016-04-27
Small tweak so the framework can be built from the command line

liaxim:
or from AS standalone

To build from the command line:
gradlew framework:clean
gradlew framework:assemble

These will produce the aar bundles:
./framework/build/outputs/aar/framework-debug.aar
./framework/build/outputs/aar/framework-release.aar

Dependent projects can just import the output .aar as a new
module and start using gvrf.

### DIFF
--- a/GVRf/Framework/settings.gradle
+++ b/GVRf/Framework/settings.gradle
@@ -1,3 +1,1 @@
-//include ':vrlibrary'
-
-//project(':vrlibrary').projectDir = new File('../../oculus_vr/VRLib/vRLib')
+include ':framework'


GVRf/Framework/settings.gradle
```diff
@@ -1,3 +1,4 @@
-//include ':vrlibrary'
-
-//project(':vrlibrary').projectDir = new File('../../oculus_vr/VRLib/vRLib')
+include ':framework'
+//include ':vrlibrary'
+
+//project(':vrlibrary').projectDir = new File('../../oculus_vr/VRLib/vRLib')
```
jason2kim: It looks good to me. And, the commented out lines can be cleaned anytime, possibly when I check in some later.

liaxim:
Comments cleaned up.

PR #521, opened 2016-04-28
GVRFrameLayout need to be clear before draw it

xdestiny110:
Canvas need to be clear before draw it. Otherwise android TextView or ImageView would overlap on it.


liaxim:
Thank you for the fix. It looks like it is the proper fix for a bug we experienced in one of our samples. But please first format the code according to http://gearvrf.org/bin/view/GearVRF/GearVRfWikiGetInvolved#CodeGuide; we use only spaces for indentation. Afterwards please sign your contribution according to http://gearvrf.org/bin/view/GearVRF/GearVRfWikiGetInvolved#SubmitPatch.

PR #522, opened 2016-04-28
Execute onInit, onStep, event callbacks on a framework main thread instead on the GL thread

liaxim:
Just looking for feedback; also raises an issue.

All of the heavy-weight samples I tried work with this, so far there is no sample I am aware that does not. But the need to have a way to know when all the textures are ready is even more apparent with this change. You tend to get some textures loaded later than others in immersivepedia for example. This is its own issue outside of the scope of this change probably but nevertheless a serious one. Nola has already brought it up on few occasions too. All that async loading and framework is not that useful if we end up having textures pop into view one after another.

Maybe we need a special GVRScript stage where the app can return a list of textures it needs and then have the fw call the GVRScript when all these textures are ready for use. Whatever it is, there is a pattern so we can and should do more on the fw side.

A certain important app partially works with a small tweak. Have to examine it closer because it is a rather unique case.


liaxim:
Ok, that pop in effect with immersivepedia is really the app's fault. It is easy to deal with because the app is using the blocking loaders. An app that resorts to our Future APIs will have harder time knowing when it is ready for rendering.

GVRf/Framework/framework/src/main/jni/objects/textures/base_texture.h
```diff
@@ -20,66 +20,39 @@
...
-        javaVm_ = nullptr;
-        pending_gl_task_ = GL_TASK_INIT_PIXELS_PARAMS;
-    }
+    void setJavaOwner(JNIEnv& env, jobject javaObject) {
+        env.GetJavaVM(&javaVm_);
 
-    explicit BaseTexture(int* texture_parameters) :
-            Texture(new GLTexture(TARGET, texture_parameters)), pending_gl_task_(GL_TASK_NONE) {
+        gvrTextureClass_ = GetGlobalClassReference(env, kTextureClassName);
+        javaMethodIsAvailable_ = GetMethodId(env, gvrTextureClass_, kIdAvailableMethodName,
```
danke-sra: Is this method "isAvailable" or "idAvailable"?

GVRf/Framework/framework/src/main/jni/objects/textures/base_texture.h
```diff
@@ -98,41 +71,18 @@ class BaseTexture: public Texture {
     virtual void runPendingGL() {
```
danke-sra: Is it possible to keep getId() and runPendingGL() unchanged while adding the new thread?

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRContext.java
```diff
@@ -2679,4 +2685,19 @@ public Object getTag() {
         return mTag;
     }
 
+    private final HandlerThread mHandlerThread;
+    private final Handler mHandler;
+
+    public void runOnTheFrameworkThread(final Runnable runnable) {
```
danke-sra: Looks like the framework thread is only designed as a queue. I think maybe its behavior and interaction with GL thread needs to be designed too?

danke-sra:
I think it's a good idea to separate framework thread and GL thread. But haven't clearly seen what the actual role of the new thread is, except that it runs events. Maybe need to review other 3D engines and design the threads with more details?

GVRf/Framework/framework/src/main/jni/objects/textures/base_texture.h
liaxim: Old getId assumes being called on the GL thread. If we want to hide the threading from developers it has to change somehow.

runPendingGL I changed because it was not being actually used as intended in this repo. In this repo the things these code handles are already handled on the Java side, see update, updateCall. Not sure if any is better but this was easier to remove.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRContext.java
liaxim: Maybe. While working on this patch I did not identify requirements that demand that.

liaxim:
This is an incremental improvement, that is all. To see the impact and learn something for the future. Regular GVRF apps seem to quite easily accept this change and take some advantage of it (mostly during application startup). On the other hand certain important app shows that we may have an API problem - if we truly want to hide the threading from end developers we most likely need to redesign the api.

GVRf/Framework/framework/src/main/jni/objects/textures/base_texture.h
liaxim: Corrected isAvailable to idAvailable.

danke-sra:
I know it is incremental. However, for the minimalist implementation, a concern is the asynchronicity between events and rendering. I feel it might have changed the behavior for apps?

liaxim:
It does change, even more so if apps (rightly so) assume things about which thread is calling onInit, onStep. Our samples took the change quite well, but they are just samples. Simply changing the thread calling these methods is probably inadvisable. Especially since I saw the impact on that certain important app.

My suggestion is to deprecate GVRScript, preserve the semantics and the threading of its APIs. Come up with a new base class, with a better name, probably taking most of the APIs from GVRScript and maybe adding few others and have none of these called on the GL thread from the get go.

In the meantime this can be used as a case study or if the implications are considered acceptable, considered for merging (after whatever additional improvements requested are made of course).

liaxim:
Planning to close this after few days. It is still usable and can be experimented with but we will come up with an alternative to GVRScript that is not called on the GL thread from day one.

Some additional findings that came out, in no particular order:
- recursivelySendSimpleEvent uses GVRSceneObject.rawGetChildren(); I think there is potential for concurrent modification of the collection while iterating
- GVRTexture.getId() must be called on the GL thread; needs to be somehow addressed if we want to simplify the threading for end users
- base_texture.h and GVRBitmapTexture: I think the java side takes care of some or all the things base_texture does in runPendingGL

liaxim:
Since Tom mentioned it, I made a change to add a GVRMain class in addition to GVRScript. The only difference at the moment is that GVRMain is called on the fw thread and GVRScript is called on the GL thread like in the past.

NolaDonato:
We need an event that GearVRF and the application can respond to that indicates the asset AND ALL OF ITS TEXTURES are loaded. If one or more textures fails to load after a period of time, we can issue the event with an error response indicating what did not load. When GearVRF receives this event it can bind the appropriate shaders for the new content. We should not add the asset to the scene until we get the event.

thomasflynn:
@liaxim, do you think this is ready for merging?  (assuming no further comments?)

liaxim:
I think it is ok. Someone else needs to test it with some samples and The App.

thomasflynn:
If i understand the way you've done this commit, the internal reasearch app can continue to use GVRScript.  Correct?

liaxim:
Yes, The App is unaffected.

liaxim:
Squashed the commits

thomasflynn:
needs manual rebase.  Mihail is out on vacation so I rebased it in my own branch and created a new pull request:  https://github.com/Samsung/GearVRf/pull/560

PR #523, opened 2016-04-29
GVRFrameLayout need to be clear before draw it

xdestiny110:
I am not familiar with the useage of git, so I close the old pull request #521 and make a new one.

GearVRf-DCO-1.0-Signed-off-by: Youwen Zhuang
<bitzyw@126.com>


liaxim:
Verified this is the fix we needed on the fw side for the remotescripting sample problem. @xdestiny110 Sorry but there is just one more thing - the "import android.graphics.PorterDuff;" is missing.

liaxim:
I will add the missing import.

PR #529, opened 2016-05-02
Improvements to the IO Framework

rahul27:
* Fixed key and motion events in the Controller. Previously
all events between process cycles were skipped. Now the
controller returns all key and motion events processed.
* The controller returns the vendor, product and name of the
device that controls it.
* Corrected Javadocs
* Restored the invalidate() method for cases where the scene
graph changes but the controller data does not.
* Renamed GVRCursorType to GVRControllerType

r.rudradevan@samsung.com


liaxim:
Tried gvr-events. I don't see the blue dot for the gaze cursor with this change but I see it with the old code. Running on S7, developer mode, not docked and with a bluetooth mouse connected.

rahul27:
There were some conflicts on rebase. I've fixed them, try now.

liaxim:
Looks good to me. The sample works the same. If there are no objections will merge in the afternoon.

---
In the future, you might want to take a look at GVRSceneObject.getComponent and consider exposing the sensor as  a component through it.

rahul27:
Sure. Will look into it.

PR #530, opened 2016-05-03
Clear bounding volume before calculating bounds for mesh

NolaDonato:
Mesh::getBoundingVolume now clears the bounding box before expanding it with new vertices.


liaxim:
Closing. Created https://github.com/Samsung/GearVRf/pull/532 instead.

PR #532, opened 2016-05-03
Clear bounding volume before calculating bounds for mesh

liaxim:
Nola's fix; can't cherry-pick it

### DIFF
--- a/GVRf/Framework/framework/src/main/jni/objects/mesh.cpp
+++ b/GVRf/Framework/framework/src/main/jni/objects/mesh.cpp
@@ -106,7 +106,7 @@ const BoundingVolume& Mesh::getBoundingVolume() {
     if (have_bounding_volume_) {
         return bounding_volume;
     }
-
+    bounding_volume.reset();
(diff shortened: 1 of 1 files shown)


thomasflynn:
we'll need to ensure frustum culling and hierarchical frustum culling are still working as precisely after this change.  that'd be the only thing i'm worried about this affecting.

liaxim:
Tried immersivepedia with stats enabled. Collected information when looking at each individual dinosaur, when focused on the dinosaur menu in the menu scene. The number of draw calls and triangles looks to be the same with or without this change.

Any suggestions how to study the impact further?

thomasflynn:
yeah, two things i can think of:
1) check the accuracy and draw calls for solar-system
2) ditto for the jassimp sample with AstroBoy

liaxim:
With jassimp's AstroBoy I am getting the same number of draw calls and triangles when looking at the same parts of the model with and without this change.

thomasflynn:
and i checked the solar-system sample.  awesome.  thanks for checking.

Issue #531, opened 2016-05-03
OESVerticalStereo shader does not work


xdestiny110:
Using OESVerticalStereo shader only render top half of the texture in left and right eye.

I think variable uniforms.u_right should be initialized correctly in renderer.cpp at line 793.

liaxim:
Pull request https://github.com/Samsung/GearVRf/issues/531 should fix the issue. Let us know what you think. Thanks.

xdestiny110:
The problem is solved.

PR #533, opened 2016-05-03
Use the correct eye in vertical/horizontal shaders

liaxim:
Should fix issue https://github.com/Samsung/GearVRf/issues/531


liaxim:
I tested by using the gvr-sceneobjects sample and changing the shader used by the GVRCameraSceneObject.

PR #545, opened 2016-05-06
Save and restore texture states during rendered texture capture to avoid flickering

salgubel:
GearVRf-DCO-1.0-Signed-off-by: Srinivasa Algubelli s.algubelli@samsung.com


danke-sra:
Looks good. +1

Issue #541, opened 2016-05-05
Need to expose BoundingVolume to the java layer


thomasflynn:
We need to work out exposing the BoundingVolume to the java layer so applications can use it.

PR #546, opened 2016-05-06
(in progress) Expose getBoundingVolume Java API

liaxim:
Tom, is this something like what you have in mind in issue https://github.com/Samsung/GearVRf/issues/541?


thomasflynn:
that may work.  though people may also choose to explicitly set the bounding volume as well.

thomasflynn:
oh, and you have this on just the mesh.  at the c++ level, each scene object has a bounding volume as well.

liaxim:
Updated to expose relevant APIs at the GVRSceneObject level.

thomasflynn:
this looks good, though the javadoc needs more than a 'todo'.  i may merge this and fix the javadoc myself.

PR #547, opened 2016-05-07
Add a method to invoke the loaded script.

thomasflynn:
Add a method to invoke the loaded script.  Previously, there was only a method to invoke a function within the script.  This will execute the whole script.

tom.flynn@samsung.com


liaxim:
Looks good to me.

PR #548, opened 2016-05-09
GVRView should support unregister view from activity.

xdestiny110:
GearVRf-DCO-1.0-Signed-off-by: Youwen Zhuang
<bitzyw@126.com>


liaxim:
I see.. We automatically register the view but then we do not unregister automatically. When do you need to unregister? Do you think we can hide this in a finalizer so we can do it behind the scenes? Thanks.

xdestiny110:
Sometimes I need to create GVRView dynamically so I need a unregister to avoid memory leak.
In android framework, it supply unregister method rather than hiding it in a finalizer. So I think exposing interfaces to user is proper.

liaxim:
Actually this is probably the only way considering that the mRenderableViewGroup in GVRActivity holds a strong reference. Never mind my suggestion.

Still, how about updating the Javadoc to let people know that they need to use GVRActivity.unregisterView to get rid of the views they no longer need? We have the API already in place but it is not clearly communicated to what it applies and when to use it. I am not sure why we need to expose this API in the views in addition to GVRActivity.

xdestiny110:
Maybe you are right. But there is another thing that puzzled me. In GVRTextView, `context` register the `mTextViewContainer`. How can I unregister `mTextViewContainer` out of GVRTextView?

liaxim:
You need to have a reference to your GVRActivity-derived class. For example GVRContext.getActivity gives you that. Once you have it you can call GVRActivity.unregisterView. How does that sound?

xdestiny110:
Well, maybe I didn't catch your idea because of my pool English. Indeed, `GVRActivity.unregisterView` satisfies my demand so my commits is unnecessary. Thank you for your illustration.
By the way, maybe 'GVRTextView' should expose his member `mTextViewContainer` so that user can unregister it?

PR #549, opened 2016-05-10
Better error message for phones without rotation sensor

NolaDonato:
Phones without a rotation sensor (like the Droid) will crash cryptically when asked to run a GearVRF application. I added a more descriptive error message that indicates why the app is crashing. I know it would be better to actually implement sensor fusion on such phones but this is a lot of work.


NolaDonato:
fixed my master branch back up.  will do a new pull request without all these other commits.

PR #550, opened 2016-05-10
Adding a log message if we do not have a rotation sensor.

NolaDonato:
Adding a log message to indicate we do not have a rotation sensor to work.

GearVRf-DCO-1.0-Signed-off-by: Nola Donato
nola.donato@samsung.com


liaxim:
Looks good to me.

PR #554, opened 2016-05-10
Add simulated head position data

mkeblx:
Noticed it was missing. Using the OVR head model. Matches standard Oculus Unity integration/plugin where enabled by default so think makes sense to do the same.

Test: readily apparent in a sample with near field objects. (Or stop by.)

Minor, but could rename head_rotation_provider.h/cpp to head_pose_provider.h/cpp to be more accurate.

Signed-off-by: m.blix <m.blix@samsung.com>

### DIFF
--- a/GVRf/Framework/framework/src/main/jni/objects/components/camera_rig.cpp
+++ b/GVRf/Framework/framework/src/main/jni/objects/components/camera_rig.cpp
@@ -191,6 +191,11 @@ void CameraRig::setRotation(const glm::quat& transform_rotation) {
     }
 }
 
+void CameraRig::setPosition(const glm::vec3& transform_position) {
+	Transform* transform = getHeadTransform();
+	transform->set_position(transform_position);
+}
+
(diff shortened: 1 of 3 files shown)


thomasflynn:
@danke-sra, please review.  thanks.

danke-sra:
@mkeblx We have a head-model implementation in place. Do you think we can use that one instead of the head pose provided by Oculus?

In the current change, it looks like both our head model and Oculus head model are applied. Our head model holds the CameraRig and already handles the position. However, we can review it too.

mkeblx:
@danke-sra Interesting, I can notice no head model in effect, only seeing rotating around center-eye point. It should be quite noticeable when looking for it.

I see now head-model-parms in VR app settings XML, but could not see effect when setting to various values.

danke-sra:
It may be broken due to other changes. I plan to check it today or tomorrow.

danke-sra:
Sorry I just got a chance to look at this. The code looks good. +1

PR #558, opened 2016-05-14
Downgrade MSAA to maximum available samples

NolaDonato:
The gvr-controls sample crashed in FrameBuffer::create because it requested 8x MSAA but the phone only supports 4X MSAA. I changed activity.cpp to check the maximum MSAA samples before creating the framebuffer and not request more than available.


danke-sra:
Looks good. Most Android devices support up to 4x MSAA. Checking the max value will make it most adaptable.

PR #561, opened 2016-05-17
Fixed problem with GVRSceneViewObject invalidation

NolaDonato:
Invalidating just the GVRView associated with the GVRViewSceneObject
incorrectly set the clip rectangle to just that view. To fix this,
we have to create a full screen android View and invalidate this
to restore the clip rectangle. This full screen view is only created
when necessary and only one of them is needed.

GearVRf-DCO-1.0-Signed-off-by:  Nola Donato nola.donato@samsung.com

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/VrapiActivityHandler.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/VrapiActivityHandler.java
@@ -27,6 +27,7 @@ import javax.microedition.khronos.opengles.GL10;
 
 import org.gearvrf.utility.Log;
 import org.gearvrf.utility.VrAppSettings;
+import org.joml.Vector2f;
 
 import android.graphics.PixelFormat;
 import android.opengl.EGL14;
@@ -53,6 +54,7 @@ class VrapiActivityHandler implements ActivityHandler {
     private EGLSurface mPixelBuffer;
     private EGLSurface mMainSurface;
     boolean mVrApiInitialized;
+    private Vector2f mScreenDimensions = null;
 
     VrapiActivityHandler(final GVRActivity activity,
             final ActivityHandlerRenderingCallbacks callbacks) throws VrapiNotAvailableException {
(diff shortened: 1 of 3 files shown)


GVRf/Framework/framework/src/main/java/org/gearvrf/VrapiActivityHandler.java
```diff
@@ -150,19 +152,27 @@ public void onSetScript() {
...
-            Log.v(TAG, "--- width: %d", framebufferWidth);
-            Log.v(TAG, "--- height: %d", framebufferHeight);
-            //a different resolution of the native window requested 
-            holder.setFixedSize(framebufferWidth, framebufferHeight);
-            Log.v(TAG, "----------------------------");
+        mScreenDimensions = new Vector2f(appSettings.getFramebufferPixelsHigh(), appSettings.getFramebufferPixelsWide());
+        if ((-1 != mScreenDimensions.y) && (-1 != mScreenDimensions.x)) {
+            if ((screenWidthPixels != mScreenDimensions.x) && (screenHeightPixels != mScreenDimensions.x)) {
+                Log.v(TAG, "--- window configuration ---");
+                Log.v(TAG, "--- width: %d", mScreenDimensions.x);
```
rahul27: The formatter should now use %f since the int has been changed to float.

GVRf/Framework/framework/src/main/java/org/gearvrf/VrapiActivityHandler.java
```diff
@@ -150,19 +152,27 @@ public void onSetScript() {
...
-        final int framebufferWidth = appSettings.getFramebufferPixelsWide();
-        if (-1 != framebufferHeight && -1 != framebufferWidth && screenWidthPixels != framebufferWidth
-                && screenHeightPixels != framebufferHeight) {
-            Log.v(TAG, "--- window configuration ---");
-            Log.v(TAG, "--- width: %d", framebufferWidth);
-            Log.v(TAG, "--- height: %d", framebufferHeight);
-            //a different resolution of the native window requested 
-            holder.setFixedSize(framebufferWidth, framebufferHeight);
-            Log.v(TAG, "----------------------------");
+        mScreenDimensions = new Vector2f(appSettings.getFramebufferPixelsHigh(), appSettings.getFramebufferPixelsWide());
```
rahul27: should we be using the DIsplayMetrics values instead of the framebuffer? Say someone sets the values to 100x100 in using gvr.xml then it would only create a view of 100x100.

GVRf/Framework/framework/src/main/java/org/gearvrf/scene_objects/GVRViewSceneObject.java
```diff
@@ -83,13 +86,20 @@ public void run() {
...
+                        View surfView = gvrContext.getActivity().getFullScreenView();
+                        if (surfView != null)
+                            surfView.invalidate();
                     }
                 });
             }
         });
     }
 
+    public void setParent(GVRView parentView) {
```
rahul27: is this required?

NolaDonato: No I removed it

GVRf/Framework/framework/src/main/java/org/gearvrf/VrapiActivityHandler.java
NolaDonato: Yes, I changed it to use the DisplayMetrics values always.

GVRf/Framework/framework/src/main/java/org/gearvrf/VrapiActivityHandler.java
NolaDonato: I changed this back to int again. I restored the old code.

rahul27:
:+1: Looks good. Verified this using the sample app.

PR #565, opened 2016-05-18
Add support for ETC1

thomasflynn:
Turns out you can have an ETC1 file in a PKM file.  We need to check for
that as well when checking the header.

tom.flynn@samsung.com


danke-sra:
Looks good. Do you think the file name is still OK?

thomasflynn:
oh, because the filename is EricssonTextureCompression2.java?

danke-sra:
Yeah, now since it supports ETC1 and ETC2, it can be renamed EricssonTextureCompression.java. Anyway, it's not an issue.

thomasflynn:
ok, yeah, we can rename it later.

PR #567, opened 2016-05-18
Generalized scene object components

NolaDonato:
The scene object now keeps a list of components on the Java and C++ side. This generalizes the component concept, making them a real extension mechanism for scene objects. Each scene object can now have any arbitrary number of GVRComponent instances. Components have a type that is used to retrieve them from the scene object. A scene object can only have one component of a particular type.
Added GVRSceneObject.attachComponent, GVRSceneObject.detachComponent, GVRSceneObject.getComponent, GVRComponent.getComponentType


GVRf/Framework/framework/src/main/java/org/gearvrf/GVRMesh.java
```diff
@@ -465,6 +457,30 @@ private void checkVectorLengthWithVertices(String parameterName,
         }
     }
 
+    /**
+     * @todo
```
thomasflynn: i'm still seeing this BoundingVolume class here in the diff.  should i?

PR #572, opened 2016-05-19
Changes for 3D Cursor

rahul27:
* Fixed set position for the mouse
* Allow multiple gamepads to be used at once
* Gaze now allows depth scrolling
* Fixed keying in the input manager

r.rudradevan@samsung.com


rahul27:
The 3D Cursor library requires a few modifications to the framework. This pull request takes care of that.

PR #576, opened 2016-05-20
Fixed crash in OESHorizontalStereoShader

NolaDonato:
The shader was not binding the vertex array


thomasflynn:
yeah, that seems like an important line.  how did this even work before??

NolaDonato:
I suspect it got “merged out” somehow…

PR #577, opened 2016-05-23
Changes to VAO to reuse VAO iD and VBO Ids and allow interleaving of …

Vittalbabu:
Changes to VAO to reuse VAO iD and VBO Ids and allow interleaving of the 3 builtin attributes.


NolaDonato:
The change looks good. I tested with multilight and jassimp samples so it works with skinning and lighting together.

My only comment is that we should go further and make all the vertex attributes interleaved. We need to separate vertices and indices for X3D. Perhaps we can interleave all the vertex attributes at that time.

Vittalbabu:
Making all vertex attributes interleaved is in progress.

GVRf/Framework/framework/src/main/jni/objects/mesh.cpp
```diff
@@ -208,61 +208,120 @@ void Mesh::bindVertexAttributes(GLuint programID)
...
+    if (attribSize == 0) {
         std::string error = "no vertex data yet, shouldn't call here. ";
         throw error;
         return;
     }
 
-    glGenVertexArrays(1, &vaoID_);
+    if (vao_dirty_ || vaoID_ == GVR_INVALID)
+    {
+        glGenVertexArrays(1, &vaoID_);
```
liaxim: In case of dirty, should the vaoId be glDeleteVertexArrays-ed?

liaxim:
@Vittalbabu Should this be closed?

Vittalbabu:
This can be closed.

PR #578, opened 2016-05-23
Call bindShaders from addSceneObject

NolaDonato:
When a scene object is added at the top level, GVRScene.bindShaders is automatically called if necessary. Now the remote scripting demo doesn't have to explicitly call bindShaders.
clear light list when clearing entire scene
return if texture is null in custom shader (instead of crashing)


NolaDonato:
Adding Missing DCO GearVRf-DCO-1.0-Signed-off-by: Nola Donato nola.donat@samsung.com

GVRf/Framework/framework/src/main/java/org/gearvrf/IAssetEvents.java
```diff
@@ -0,0 +1,56 @@
+package org.gearvrf;
```
thomasflynn: needs copyright header

NolaDonato: removed this file, not needed

thomasflynn:
fails to build.  Apparently GVRContext and GVRImporter reference IAssets in a number of places.

NolaDonato:
I reverted GVRContext, GVRImportSettings, GVRImporter to master, it builds now and I tested with gvr-remote-scripting demo which no longer needs bindShaders

PR #579, opened 2016-05-24
Allow zip files to be loaded by GVRf

rahul27:
r.rudradevan@samsung.com

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/ZipLoader.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/ZipLoader.java
@@ -1,0 +1,94 @@
+package org.gearvrf;
+
+import android.content.Context;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.Future;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+
+/**
+ * This is a small convenience class that makes it easy to unzip a file and load entries as
+ * {@link GVRAndroidResource}s into the GVRf framework.
+ *
+ * For cases where the number of files in an applications gets too large to handle, consider
+ * zipping them and using the {@link ZipLoader} to process them one at a time.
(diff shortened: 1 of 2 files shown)


GVRf/Framework/framework/src/main/java/org/gearvrf/ZipLoader.java
```diff
@@ -0,0 +1,94 @@
+package org.gearvrf;
```
danke-sra: We might need a copyright header here

danke-sra:
Please add a missing copyright header and should be good.

GVRf/Framework/framework/src/main/java/org/gearvrf/ZipLoader.java
```diff
@@ -0,0 +1,94 @@
...
+     * @return a list of processed zip file entries.
+     * @throws IOException this function returns an {@link IOException} if there are issues
+     *                     processing the provided zip file.
+     */
+    public static <T> List<T> load(GVRContext gvrContext, String zipFileName, ZipEntryProcessor<T>
+            processor) throws IOException {
+        Context context = gvrContext.getContext();
+        InputStream inputStream = context.getAssets().open(zipFileName);
+        ZipInputStream zipInputStream = new ZipInputStream(inputStream);
+        List<T> listTextures = new ArrayList<T>();
```
rahul27: The name of this list also needs to change. The load call is is not only for textures. Will fix this.

GVRf/Framework/framework/src/main/java/org/gearvrf/ZipLoader.java
rahul27: Sure. Will add one.,

rahul27:
Addressed the comments. 

@NolaDonato I've also modified a sample to test the ZipLoader:

https://github.com/gearvrf/GearVRf-Demos/pull/162

thomasflynn:
@NolaDonato thumbs up/down?

NolaDonato:
I tested this with the updated gvr-performance sample. It is good to go.

NolaDonato:
I tested both of Rahul’s pull requests for the Zip load (GearVRf and GearVRf-Demos) and the feature works.

PR #584, opened 2016-05-25
Batching working with texture shader

roshanch:
--adding batching support for texture shader based on same material.
--modified pickObjects in picker.cpp to skip object outside view frustum
--added API to get sceneobject is culled or not for each frame
--added API to set batching status for renderdata (by default its enabled), for headtracker or textobject we need to set it to disabled.

@NolaDonato @dankex @thomasflynn @liaxim @vitttal please review this


GVRf/Framework/framework/src/main/java/org/gearvrf/GVRScene.java
```diff
@@ -593,8 +593,9 @@ public void onStep() {
             }
         }
 
-        private void recursivelySendSimpleEvent(GVRSceneObject sceneObject, String eventName) {
-            getGVRContext().getEventManager().sendEvent(
+        private void recursivelySendSimpleEvent(GVRSceneObject sceneObject, String eventName) {           
+            if(!sceneObject.isCulled())
+                getGVRContext().getEventManager().sendEvent(
```
Vittalbabu: Add brackets

GVRf/Framework/framework/src/main/jni/engine/renderer/batch.cpp
```diff
@@ -0,0 +1,141 @@
...
+    mesh = new Mesh(); // should perhaps have a pool of meshes of a certain size
+}
+
+Batch::~Batch() {
+    vertices_.clear();
+    normals_.clear();
+    tex_coords_.clear();
+    indices_.clear();
+    matrices_.clear();
+    matrix_indices.clear();
```
Vittalbabu: Place these calls in a new method (cleanup()?) and call from both locations. Not sure if cleanup is required in ctor.

roshanch:
Pushed a patch for static and dynamic meshes. Still, needs to handle the case where renderdata properties are different in same batch.

roshanch:
reverted back commit.. opening new pull request

PR #586, opened 2016-05-26
Fixed a problem with point light attenuation

NolaDonato:
fixed light source disable logic


NolaDonato:

PR #590, opened 2016-06-01
3DCursor: change width of toggle button in UI

parthmehta209:


parthmehta209:
@rahul27 made a minor UI change, to make the text of the toggle button fit.

PR #594, opened 2016-06-01
adding batching support for texture shader

roshanch:
--render data is sorted according to following attributes: rendering order, shader type, material, render data properties (states)
--modified pickObjects in picker.cpp to skip object outside view frustum
--added API to get sceneobject is culled or not for each frame
--mesh can be set dynamic for creating separate batches for dynamic batches

@NolaDonato @danke-sra @thomasflynn @liaxim @Vittalbabu  please review this


thomasflynn:
a couple of quick, minor comments.
- spacing was really inconsistent.  look at the diffs here on github.  code at the same level should be indented at the same level.  maybe your tabs are tabs and not spaces?
- there are some commented-out code (i think the sort() routine, maybe in other places).  removed commented out code.

roshanch:
@thomasflynn needs to fix formatting in my eclipse.

NolaDonato:
In Batch::add you are allocating memory by creating a new RenderData. Why can't you just store a pointer to the existing one? It is really inefficient to allocate memory in the render loop.

NolaDonato:
In Batch::update you pick 3 because the others are reserved. Test to make sure this will work with skinning or lightmapping. There may be a location collision.

NolaDonato:
Why does Batch::add need the renderdata and the material? The material can be oibtained from the renderdata, avoiding passing an additional argument.

NolaDonato:
Using std::map internally is very inefficient. We should optimize this in a subsequent pull request.

NolaDonato:
In Renderer::renderbatches why are we restricting batching to just the texture shader. This implies that any assets loaded with Jassimp will never do batching.

NolaDonato:
RenderData is getting pretty big. We should look at whether using bit fields instead of separate Booleans for each thing is a worthwhile optimization. This depends on the relative speed of accessing a byte vs a bit on mobile platforms.

NolaDonato:
Computation of the hash code is really inefficient. It involves string manipulation which involves memory allocation which is slow in the inner loop. Why not add up all the values and use that as a base for a hash?

NolaDonato:
Why do you make a copy of the children in Transform::invalidate? This is inefficient because it allocates memory. It would be better to have SceneObject::setTransformDirty lock the children and update them.

PR #596, opened 2016-06-02
Move processing off the GL Thread

rahul27:
* Process input data on the thread that the call is made on - does
  not need to be the GL Thread.
* Also made bug fixes to support GVRMain

r.rudradevan@samsung.com


GVRf/Framework/framework/src/main/java/org/gearvrf/SensorManager.java
```diff
@@ -31,11 +31,19 @@
...
     // Create a HashMap to keep reference counts
     private final Map<GVRBaseSensor, Integer> sensors;
+    private static SensorManager instance;
 
-    SensorManager() {
+    private SensorManager() {
         sensors = new HashMap<GVRBaseSensor, Integer>();
     }
 
+    static SensorManager getInstance(){
```
liaxim: Just curious, why it had to become a singleton?

rahul27: The input manager previously created an instance to run a pick on all the controllers using the GL thread. This is the code that was removed:

```
-    private GVRDrawFrameListener drawFrameListener = new GVRDrawFrameListener() {		
 -        @Override		
 -        public void onDrawFrame(float frameTime) {		
 -            for (GVRCursorController controller : controllers) {		
 -                controller.update(sensorManager, scene);		
 -            }		
 -        }		
 -    };
```

Now that it is a singleton, any given controller can get the instance to run a pick.

PR #597, opened 2016-06-02
Wait for textures to be loaded before adding model to the scene

NolaDonato:
1. Added IAssetEvents interface to listen for asset load events 
- onTextureLoaded: raised when texture successfully loaded
- onTextureError: raised when texture fails to load
- onModelLoaded: raised when scene graphi successfully loaded
- onModelError: raised when model fails to load
- onAssetLoaded: raised when model and all textures are loaded
2. Modify GVRContext to use GVRImporter.loadModel
3. GVRImporter.loadModel takes a GVRScene as an argument. If it is present, the model will be added to the scene after all the textures have been loaded


GVRf/Framework/framework/src/main/java/org/gearvrf/GVRContext.java
```diff
@@ -193,6 +194,19 @@ public GVRActivity getActivity() {
...
     /**
+     * Get the {@link GVRImporter} to use for loading assets.
+     * 
+     * A {@code GVRImporter} loads models asynchronously from your applications's
+     * local storage or the network.
+     * 
+     * 
+     * @return The asset loader associated with this context.
+     */
+    public GVRImporter getAssetLoader() {
```
rahul27: The function description says that the call would return an asset loader but it returns an importer. We would decide on a name either rename the call to getImporter or rename GVRImporter to AssetLoader.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRContext.java
```diff
@@ -193,6 +194,19 @@ public GVRActivity getActivity() {
     }
 
     /**
+     * Get the {@link GVRImporter} to use for loading assets.
+     * 
+     * A {@code GVRImporter} loads models asynchronously from your applications's
```
rahul27: typo here applications's

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRContext.java
```diff
@@ -524,9 +537,23 @@ public GVRSceneObject getAssimpModel(String assetRelativeFilename)
      */
     public GVRSceneObject getAssimpModel(String assetRelativeFilename,
             EnumSet<GVRImportSettings> settings) throws IOException {
-        return GVRImporter.getAssimpModel(this, assetRelativeFilename, settings);
+        return getAssimpModel(assetRelativeFilename, settings, null);
```
rahul27: passing a null argument to IAssetEvents would result in an NPE inside the function call.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRContext.java
```diff
@@ -669,6 +722,9 @@ public GVRSceneObject loadModelFromURL(String urlString) throws IOException {
      *                  load from the cache in future uses;
      *           false: Don't keep local copy and only do online streaming for loading everytime
      *           
+     * @param handler
```
rahul27: handler seems to be missing.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRContext.java
NolaDonato: I will rename it GVRAssetLoader then. I think that is a better name. It was a private class until now so nobody was using it externally.

rahul27: Also, there seems to be two conflicting ways of getting an importer object - using this call getAssetLoader or creating one using GVRImporter importer = new GVRImporter(context). This might be a little too confusing. We need to decide on one or the other.

GVRf/Framework/framework/src/main/java/org/gearvrf/IAssetEvents.java
```diff
@@ -0,0 +1,70 @@
...
+ * Interface for processing asset load events.
+ * The asset loader raises an event whenever a texture
+ * or a model is loaded. It also raises an event when
+ * the entire asset (models and textures) is loaded.
+ * This interface lets you observe  these events.
+ */
+public interface IAssetEvents extends IEvents
+{
+    /**
+     * Called when a model is successfully loaded.
```
rahul27: The javadocs are identical for onAssetLoaded and onModelLoaded. Makes it harder to understand what is called when.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRContext.java
NolaDonato: I am removing this function. I noticed it has already been deprecated so we don’t need yet another version of it.

GVRf/Framework/framework/src/main/java/org/gearvrf/IAssetEvents.java
```diff
@@ -0,0 +1,70 @@
...
+     * This event will not be raised until the model and all of
+     * its textures are loaded.
+     * @param context   GVRContext used to load the model
+     * @param model     GVRSceneObject root of the model scene graph, null if model did not load.
+     * @param filePath  File path or URL of the model.
+     * @param errors    String with loading errors or null if successful
```
rahul27: Also having the errors param here makes it difficult to implement error handling. We would need to look at this value in addition onModelError and onTextureError which may not be ideal.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRContext.java
NolaDonato: Yes, we do. The preferred way will be to use the asset loader. For ALL assets, even textures. I took a deep look into the texture loading issues because we have some bugs there. I have a plan to fix it by making the whole texture loading thing a lot simpler but that will take a while.

So I am doing it in several stages. The first stage is handling the model loading. If we use the new API which waits for all the textures to be loaded  before adding the model to the scene, it will fix some of the problems.

The second stage will be to move texture loading from GVRContext to the asset loader. This is a lot harder because I don’t want to just propagate the existing API. I would like to make something cleaner that will immediately return a GVRTexture (not a future texture or any callback thing). This texture will “magically” load behind the scenes without the user having to do anything. When all the textures for a scene object actually have data present, the lower level layer will render it. Of course, this requires changes to the C++ layer (lots of them) and changes to the implementation of the Java API. I have started this work and will release that in stages too. It is in the “texturereform” branch of my personal repo if you want to look at it.

The final stage will be to change all the samples to use the newer Java API and completely deprecate the old one.

Be patient and it will get less confusing…

GVRf/Framework/framework/src/main/java/org/gearvrf/IAssetEvents.java
NolaDonato: I changed the JavaDoc to explain onModelLoaded does not wait for textures while onAssetLoaded does.

GVRf/Framework/framework/src/main/java/org/gearvrf/IAssetEvents.java
NolaDonato: I need to have these functions because the asset loader actually uses them internally. This is how it figures out whether all the textures loaded or not.

I don’t understand why having a parameter makes error handling harder. Can you explain that one.

My thinking behind the “errors” parameter for onAssetLoaded is for the end case. It is possible that the model or some of the textures do not load. Internally we might want to catch that case and do something about it. If you have no errors it implies your asset and everything it referenced loads correctly. If the model is null, it didn’t load. The errors string will have the accumulation of all the texture load errors.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
```diff
@@ -45,23 +43,137 @@
...
  */
-final class GVRImporter {
-    private GVRImporter() {
-    }
+public final class GVRAssetLoader implements IEventReceiver {
+    /**
+     * Listens for texture load events and raises the "onAssetLoaded"
+     * event after all textures have been loaded.
+     */
+    static public class AssetEventListener implements IAssetEvents
```
liaxim: Is AssetEventListener supposed to be used by end users? If not, how about making it non public?

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
```diff
@@ -45,23 +43,137 @@
...
+        protected IAssetEvents      mUserHandler = null;
+        protected String            mFileName;
+        protected String            mErrors;
+        
+        public AssetEventListener(GVRAssetLoader importer, GVRScene scene) {
+            mScene = scene;
+            mImporter = importer;
+            mNumTextures = 0;
+        }
+        public void beginLoadTexture()
```
liaxim: Doesn't seem to be used. Is it supposed to be in a future pull request?

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
```diff
@@ -45,23 +43,137 @@
...
+        {
+            ++mNumTextures;
+        }
+        
+        public void beginLoadAsset(String filePath, IAssetEvents userHandler)
+        {
+            mImporter.getEventReceiver().addListener(userHandler);
+            mFileName = filePath;
+            mNumTextures = 0;
```
rahul27: set userhandler to mUserHandler here?

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
```diff
@@ -45,23 +43,137 @@
...
  * available through instances of {@link GVRAssimpImporter}.
  * <p>
  * Supports importing models from an application's resources (both
  * {@code assets} and {@code res/raw}) and from directories on the device's SD
  * card that the application has permission to read.
  */
-final class GVRImporter {
-    private GVRImporter() {
-    }
+public final class GVRAssetLoader implements IEventReceiver {
```
liaxim: Maybe the loadModel api should be exposed by GVRContext directly.. Then this class can be hidden. Now the users can use GVRContext.getAssetLoader or simply create their own AssetLoader. They shouldn't be able to the same thing in more than one way. I personally vote for exposing the API in GVRContext so we can control the number of instances of GVRAssetLoader.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
```diff
@@ -125,91 +237,191 @@ static GVRAssimpImporter readFileFromResources(GVRContext gvrContext,
...
+     *            
+     * @return A {@link GVRModelSceneObject} that contains the meshes with textures and bones
+     * and animations.
+     * @throws IOException 
+     *
+     */
+    public GVRModelSceneObject loadModel(String filePath, GVRResourceVolume.VolumeType volumeType,
+            GVRScene scene) throws IOException
+    {
+        if (scene != null)
```
liaxim: Firing the events seems to me generally useful for users regardless whether they intent to add the model the scene. The users might want to add the model to a sceneObject or whatever. How about relaxing the scene instance requirement?

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
liaxim: In your final future version, how are users going to get the assetLoader? Are there going to be multiple instances of the assetLoader?

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
```diff
@@ -45,23 +43,137 @@
...
+                    generateLoadEvent(context);
+                }
+            }           
+        }
+        
+        /**
+         * Called when a model cannot be loaded.
+         * @param error error message
+         */
+        public void onModelError(GVRContext context, String error, String modelFile)
```
liaxim: It is more common in Java to pass exceptions in case of an error. The exception may just wrap a String or might carry a lot more information.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
```diff
@@ -125,91 +237,191 @@ static GVRAssimpImporter readFileFromResources(GVRContext gvrContext,
...
+     * @param scene
+     *            If present, this asset loader will wait until all of the textures have been
+     *            loaded and then adds the model to the scene.
+     *            
+     * @return A {@link GVRModelSceneObject} that contains the meshes with textures and bones
+     * and animations.
+     * @throws IOException 
+     *
+     */
+    public GVRModelSceneObject loadModel(String filePath, GVRResourceVolume.VolumeType volumeType,
```
liaxim: This is kind of cosmetic but I'd say the final API will be cleaner if we just expose loadModelFromUrl, loadModelFromAssets, loadModelFromSdcard, etc instead of passing the volumeType. It would make for easier to read code and possibly make it a bit less error-prone. Also we might request URL instance from the user in the case of loadModelFromUrl so we know we get valid input, etc.

--
For example the existing API would return a null model instance in case of an invalid URL. The user needs to know what to look for in logcat to find out that.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
liaxim: Looks like this will happen in the future. It won't make a difference in the existing code.

liaxim:
The direction is good. Isn't the title of this particular pull request misleading though? Please squash the commits at some point.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
NolaDonato: Yes it does not need to be public

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
NolaDonato: We will eventually support passing in an asset listener so the caller can trap when models and textures are loaded and modify them before adding them to the scene. Since our asset loading API is already so complex and nobody has actually asked for this yet I did not expose the asset listener to the user.
Sometimes you want to do something like flip all the textures in the Y direction for a set of assets. Exposing the asset listener feature would allow it. I would like to simplify the asset loading first before we add new stuff.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
NolaDonato: The asset loader is not a heavyweight class and doesn't keep a lot of stuff so there is no reason why you could not have multiple instances of it. I am not sure why we want to control the number of instances of this particular class. I plan to deprecate all the asset load functions on the context and have GVRAssetLoader provide the only API for loading. GVRContext has an asset loader for convenience - so there is always a way to load an asset from anywhere.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
NolaDonato: The reason we have the scene is so that we can send the events somewhere (they go to the scene's event receiver) and so that we can add the model after all the textures are loaded. Which event receiver would be appropriate for the events if we don't have a scene?

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
NolaDonato: What about for texture loads? Should this be an exception? The reason I did not do an exception is that it would not allow the programmer to correct the error and continue (by substituting a different model or texture).

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
NolaDonato: I think it would be even better if we got rid of the volume type altogether and just put a prefix on the filename like HTTP does, maybe "SD:" for the SD card and "ASSETS:" for the assets directory? I just preserved the arguments of the existing API but moved the functions to GVRAssetLoader because I didn't want to break existing applications.

NolaDonato:
Squashing commits and reopening new pull request #601

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
liaxim: I wouldn't mind at all removing the volume type or at least hiding it from public view. Prefixing as suggested could work.

PR #600, opened 2016-06-06
3DCursor: Compressed the larger texture files

rahul27:
Also added support for compressed textures in the
3D cursor library

r.rudradevan@samsung.com


parthmehta209:
verified.

rahul27:
This one is good to go.

PR #601, opened 2016-06-06
Exposes GVRAssetLoader class for loading assets - replaces PR #597

NolaDonato:
New asset loader API which waits for all the textures to be loaded
before adding the model to the scene

Added IAssetEvents interface to listen for asset load events
onTextureLoaded: raised when texture successfully loaded
onTextureError: raised when texture fails to load
onModelLoaded: raised when scene graphi successfully loaded
onModelError: raised when model fails to load
onAssetLoaded: raised when model and all textures are loaded
Renamed GVRImporter to GVRAssetLoader, moved asset loading functions
from GVRContext to GVRAssetLoader


liaxim:
Looks good to me. Verified with few samples. Will merge tomorrow lunch-time.

PR #606, opened 2016-06-10
Add GVRBehavior, GVRSwitch and expose GVRSceneObject visibility

NolaDonato:
GVRBehavior is a new component that allows users to add new components
to a scene object that have per-frame behavior.
The GVRSwitch component derives from GVRBehavior and selects one of a
scene object's children to display.
Added GVRSceneObject.setVisible, GVRSceneObject.isVisible
Made GVRSceneObject.attachComponent and GVRSceneObject.detachComponent public


GVRf/Framework/framework/src/main/java/org/gearvrf/GVRSwitch.java
```diff
@@ -0,0 +1,146 @@
...
+    {
+        int i = 0;
+        GVRSceneObject owner = getOwnerObject();
+        if (owner == null)
+        {
+            return;
+        }
+        for (GVRSceneObject child : owner.children())
+        {
+            if (child.getName() == childName)
```
rahul27: You might want to change this to child.getName().equals(). 

== is a reference comparison, .equals() would do a value comparison.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRSwitch.java
```diff
@@ -0,0 +1,146 @@
...
+ * it sets all but one of the scene object's children to be invisible.
+ * You can control which child is visible by setting the switch index.
+ * 
+ * @see GVRBehavior
+ */
+public class GVRSwitch extends GVRBehavior
+{
+    static private long TYPE_SWITCH = (System.currentTimeMillis() & 0xfffffff);
+    protected int mSwitchIndex = -1;
+    protected GVRSceneObject mSelected = null;
```
rahul27: mSelected is not used.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRSwitch.java
```diff
@@ -0,0 +1,146 @@
...
+ * When the GVRSwitch component is attached to a GVRSceneObject
+ * it sets all but one of the scene object's children to be invisible.
+ * You can control which child is visible by setting the switch index.
+ * 
+ * @see GVRBehavior
+ */
+public class GVRSwitch extends GVRBehavior
+{
+    static private long TYPE_SWITCH = (System.currentTimeMillis() & 0xfffffff);
+    protected int mSwitchIndex = -1;
```
rahul27: Should mSwitchIndex be initialized to 0?

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRSwitch.java
NolaDonato: Now I have to go back and look at where else I did this. I have done much more C# programming than Java, I did not realize that == was just a reference comparison only in Java (In C# it actually compares the strings).

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRSwitch.java
NolaDonato: I removed it. Originally I was going to track the currently selected one but then I realized it would not be very resilient against changes to the switch children array.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRSwitch.java
NolaDonato: Yes, probably. Originally I thought that it should show nothing unless you selected something but I think it should probably default to displaying the first child.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRBehavior.java
```diff
@@ -0,0 +1,160 @@
...
+        mIsListening = false;
+        mHasFrameCallback = isImplemented("onDrawFrame", float.class);
+    }    
+
+    static public long getComponentType()
+    {
+        return TYPE_BEHAVIOR;
+    }
+    
+    public void enable()
```
liaxim: start&stopListening probably needs some synchronization to ensure thread-safety.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRComponent.java
```diff
@@ -109,17 +121,40 @@ public boolean hasOwnerObject() {
     }
     
     /**
+     * Enable or disable this component.
+     * @param flag true to enable, false to disable.
+     * @see enable
+     * @see disable
+     * @see isEnabled
+     */
+    public void setEnable(boolean flag) {
```
liaxim: Can this be a private method? Seems most places settled on enable/disable public APIs.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRSceneObject.java
```diff
@@ -803,6 +806,30 @@ public boolean isColliding(GVRSceneObject otherObject) {
...
+    }
+ 
+    /**
+     * Designates whether the scene object should be displayed or not.
+     * If a scene object is not enabled, none of its children
+     * will be displayed either.
+     * 
+     * @param visible if true the object will be displayed, if false it will not be.
+     */
+    public void setEnable(boolean enable) {
```
liaxim: How about enable/disable public APIs instead?

GVRf/Framework/framework/src/main/jni/objects/components/component.h
```diff
@@ -30,31 +30,34 @@ class SceneObject;
...
     virtual ~Component() {
+        set_owner_object(NULL);
     }
 
     SceneObject* owner_object() const {
         return owner_object_;
     }
 
-    void set_owner_object(SceneObject* owner_object) {
+    virtual void set_owner_object(SceneObject* owner_object) {
```
liaxim: Why virtual?

GVRf/Framework/framework/src/main/jni/objects/components/component.h
```diff
@@ -66,16 +69,26 @@ class Component: public HybridObject {
...
+    }
+
 private:
     Component(const Component& component);
     Component(Component&& component);
     Component& operator=(const Component& component);
     Component& operator=(Component&& component);
 
-private:
+protected:
```
liaxim: Is the "protected" change necessary?

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRSwitch.java
```diff
@@ -0,0 +1,145 @@
...
+     */
+    public void selectByName(String childName)
+    {
+        int i = 0;
+        GVRSceneObject owner = getOwnerObject();
+        if (owner == null)
+        {
+            return;
+        }
+        for (GVRSceneObject child : owner.children())
```
liaxim: I don't think iterating like this over owner.children() is going to be thread-safe. The underlying collection can be modified by other threads.

liaxim:
As a side note I do prefer giving interfaces to end users but considering how the framework already is, it is going to be a tall order even if I could've convinced everybody. For example an end user can easily forget to call super.onAttach from their override. Can be coded in a different way to ensure super is called but it is a hassle. What you could do is determine whether some methods are not supposed to be overridden and make them final. You could consider making the GVRSwitch class final too.

liaxim:
The only real issue I think I raised is the thread-safety concern when iterating over the owner's children.

GVRf/Framework/framework/src/main/jni/objects/components/component.h
NolaDonato: Because other components in the C++ layer override this function.

NolaDonato: Overriding set_owner_object is the way subclasses of Component can do something when they are attached or detached from a scene object. I tound this capability useful when implementing Collider components because I need to keep a special list of them to optimize picking.

GVRf/Framework/framework/src/main/jni/objects/components/component.h
NolaDonato: No, I will keep it private.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRComponent.java
NolaDonato: You would think so but in scanning the code I discovered enable / disable, setEnable, setEnabled, etc. I put setEnable in for backwards compatibility because some components (like GVREyePointeeHolder) have that in their API. I would like all components to at least use the same native function to enable and disable themselves instead of each component having a different way.

I agree we should settle on something. The reason I don't like enable() and disable() is that, if I am coming from an API that passes a flag, I have to write code that tests the flag and calls one of two functions. If I start out with an underlying API that takes a flag, it is easy to implement enable and disable.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRSwitch.java
NolaDonato: I will change the GVRSceneObject.mChildren array to be a CopyOnWriteArrayList. We read the scene graph a lot more than we write it. This change will make iteration thread-safe at the expense of copying the list when it is written.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRBehavior.java
NolaDonato: I added synchronization in these functions

NolaDonato:
I believe I have addressed all the comments. Is there other feedback or can we merge this?

liaxim:
Will merge after lunch.

PR #608, opened 2016-06-13
Support for external material file from Android asset folder

deepakrawat22:
The support was added previously by Danke in his repository, just picked the appropriate changes and tested with the cube model.

deepak.rawat@samsung.com


liaxim:
This would have been much quicker to review and get in if there was a way to avoid the introduction of GVRObject just because of the byte array wrapper.

I think the ResourceCache can be split into a base class that accepts any T (and used for the byte array wrapper) and another specialized one that deals with wrapCallback and stuff and caches hybrid objects.

NolaDonato:
I tested this by modifying gvr-cockpit to use MTL files and loadModel() instead of separately applying the textures to the mesh. When I made these changes the models were displayed without textures. I am sending you a zip file with my application.

liaxim:
@deepakrawat22 Please try to remove the GVRObject. I think there is a much more localized change that can be made to ResourceCache as I suggested.

GVRf/Framework/framework/src/main/jni/contrib/jassimp2/memory_file.cpp
```diff
@@ -0,0 +1,90 @@
+/*
```
danke-sra: We should probably add a Copyright header. Either the original MIT license or Samsung Apache should be OK, as they are compatible.

GVRf/Framework/framework/src/main/jni/contrib/jassimp2/memory_file.h
```diff
@@ -0,0 +1,20 @@
+/*
```
danke-sra: We should probably add a Copyright header. Either the original MIT license or Samsung Apache should be OK, as they are compatible.

GVRf/Framework/framework/src/main/java/org/gearvrf/utility/ResourceCache.java
```diff
@@ -40,7 +39,7 @@
  * 
  * @since 2.0.2
  */
-public class ResourceCache<T extends GVRObject> {
+public class ResourceCache<T extends GVRHybridObject> extends ResourceCacheBase {
```
danke-sra: The issue with GVRHybridObject is that it assumes everything that is cacheable is a hybrid object, which can be too restrictive. However, we can address such issues separately. It's OK here.

deepakrawat22:
@liaxim Pushed new changes to get rid of GVRObject. 
@NolaDonato The mtl file in gvr-cockpit had ./ in then, added a fix for that.

danke-sra:
Looks good with me. Thanks for the updates.

liaxim:
Looks good. Verified with few samples. Will wait for Nola's thumbs up before merging.

liaxim:
Will merge lunch-time

PR #615, opened 2016-06-17
Batching with texture shader

roshanch:
-Batching support for texture shader.
-only small meshes are batched, right now size of each batch is 400 indices, mesh with larger size wont be batched.
-removing getTexture() from custom_shader.cpp, as texture unit enums can be incremented. 
-- i created a sample with around 225 objects all of them have texture and uses texture shader, without batching it gives 30 fps whereas with batching it gives 60 fps.


GVRf/Framework/framework/src/main/jni/engine/renderer/batch.cpp
```diff
@@ -0,0 +1,187 @@
...
+    glm::vec3 init_vec3(0,0,0);
+    glm::vec2 init_vec2(0,0);
+    for(int i=0;i<no_vertices;i++){
+        vertices_.push_back(init_vec3);
+        tex_coords_.push_back(init_vec2);
+        matrix_indices.push_back(0);
+        normals_.push_back(init_vec3);
+        indices_.push_back(0);
+    }
+    for(int i=0;i<60;i++)
```
liaxim: Should this 60 here be replaced with BATCH_SIZE?

liaxim:
Please add a high-level description of your definition of a static and dynamic mesh somewhere. Should there be API restrictions on either of these?

roshanch:
@NolaDonato  
@liaxim yes, i will add it somewhere. I don't think there should be API restrictions, user should take care while modifying mesh. let me know your thoughts on it.

liaxim:
We might have to spell out what "take care" means.

roshanch:
@liaxim  Can we add in documentation the instructions for dynamic mesh?

liaxim:
Sure. Btw, you need to rebase your change.

roshanch:
Creating new pull request.

liaxim:
You don't need a new pull request. If you want we can work together next time so you don't have to go through this; and lose comments and history.

PR #617, opened 2016-06-18
Fixed material transparency in phong shader

NolaDonato:
The Phong shader was ignoring the alpha in the diffuse color of the
material. Now it is correctly applying it.


liaxim:
@roshanch ?

roshanch:
@NolaDonato I think we should have support for the opacity component of material and not just a opacity texture. We have opacity component in material coming from the assimp, just need to add it into shader.

NolaDonato:
This pull request adds support for material opacity. It is in the alpha component of the diffuse material color.

NolaDonato:
Pull request #624 addresses Roshan's comment about exporting opacity.

liaxim:
Will merge lunch time.

PR #621, opened 2016-06-20
Picking Reform

NolaDonato:
1. Added different collider types derived from base component  GVRCollider in Java and C++
2. Make GVRPicker a component that generates pick events
3. Optimize C++ picking to not transform vertices but transform the ray instead
This patch should be backwards compatible with the existing picking API.


liaxim:
This needs more than one pair of eyes.

thomasflynn:
at an initial skim, it looks fine.  it'll need someone to go over it closer than I did.  it'll also need some thorough testing.

rahul27:
Already on it. This change could benefit some of the IO modules.

GVRf/Framework/framework/src/main/jni/objects/components/mesh_collider.cpp
```diff
@@ -0,0 +1,200 @@
...
+             */
+            if (distance <= 0)
+            {
+                continue;
+            }
+            glm::vec4 p = model_view * glm::vec4(hitPos, 1);
+            glm::vec3 distVec = rayStart - glm::vec3(p);
+
+            distance = distVec.length();
+            LOGD("Picker: hit mesh %f\n", distance);
```
rahul27: remove this log, it floods logcat.

NolaDonato: I removed the printout

GVRf/Framework/framework/src/main/jni/engine/picker/picker_jni.cpp
```diff
@@ -25,38 +25,77 @@
...
+        const ColliderData& data = *it;
+        jlong collider = reinterpret_cast<jlong>(data.ColliderHit);
+        long_colliders.push_back(collider);
+    }
+    jlongArray jcolliders = env->NewLongArray(
+            long_colliders.size());
+    env->SetLongArrayRegion(jcolliders, 0,
+            long_colliders.size(),
+            reinterpret_cast<jlong*>(long_colliders.data()));
+    colliders.clear();
```
rahul27: Small optimization - long_colliders.clear() here.

rahul27:
For the immersivepedia app the rotation works only if you try to swipe at the head of the rotating dino.

NolaDonato:
I made a few changes to the picking and now it works a lot better with immersivepedia. I wasn’t able to get the planetarium app to build. I will check that one tomorrow.

-Nola

GVRf/Framework/framework/src/main/jni/engine/picker/picker_jni.cpp
NolaDonato: I figured out how to get rid of long_colliders. That was a stupid intermediate array. I can directly store the pointers into the java array.

NolaDonato:
I fixed the issues with the internal collider list and scene switching. This is why the play button wasn't pickable and the dino wouldn't rotate when you swipe.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRPicker.java
```diff
@@ -21,29 +21,256 @@
...
+    }
+    
+    public void onDrawFrame(float frameTime)
+    {
+        if (getOwnerObject() != null)
+        {
+            GVRTransform trans = getOwnerObject().getTransform();
+            Matrix4f worldmtx = trans.getModelMatrix4f();
+            worldmtx.getTranslation(mRayOrigin);
+            mRayDirection.x = 0;
```
rahul27: Why is the ray direction being overwritten here? If I were to set a ray direction using setPickRay(), these lines would reset them.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRPicker.java
```diff
@@ -277,36 +466,48 @@ public static final float pickSceneObject(GVRSceneObject sceneObject,
...
+     */
+    public static final List<GVRPickedObject> findObjects(GVRScene scene, float ox, float oy, float oz, float dx,
+            float dy, float dz) {
+        return Arrays.asList(pickObjects(scene, ox, oy, oz, dx, dy, dz));
+    }
+    
+    /**
+     * Internal utility to help JNI add hit objects to the pick list.
+     */
+    public static GVRPickedObject makeHit(long colliderPointer, float distance, float hitx, float hity, float hitz)
```
rahul27: Make this private since it is an internal utility?

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRSceneObject.java
```diff
@@ -639,19 +671,23 @@ public void detachEyePointeeHolder() {
...
     /**
      * Simple, high-level API to enable or disable eye picking for this scene
      * object.
      * 
-     * The {@linkplain #attachEyePointeeHolder(GVREyePointeeHolder) low-level
+     * The {@linkplain #attachCollider low-level
      * API} gives you a lot of control over eye picking, but it does involve an
-     * awful lot of details. Since most apps are just going to use the
-     * {@linkplain #attachEyePointeeHolder() simple API} anyhow, this method
+     * awful lot of details. This method
```
rahul27: Should mention that using this method would attach a sphere collider as default rather than a mesh collider.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRPicker.java
NolaDonato: If you attach the picker to a scene object it will automatically compute the ray for you. setPickRay is for when you want to generate pick events from a particular ray but it is not derived from a scene object. For example, if you want to generate a ray based on the mouse position you can have a GVRPicker which will generate pick events even though it is not in the scene.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRPicker.java
NolaDonato: If I make it private can I still call it from JNI?

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRSceneObject.java
NolaDonato: I updated the documentation to add this.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRPicker.java
rahul27: Yup, it should work.

NolaDonato: It is no longer public

rahul27:
:+1: Looks good. This one is good to go.

liaxim:
I pray we don't have to cherry-pick this change. It could have been squashed.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVREyePointeeHolder.java
```diff
@@ -184,11 +101,11 @@ public void setEnable(boolean enable) {
      *            The {@link GVREyePointee} to add
      * 
      */
-    public void addPointee(GVREyePointee eyePointee) {
+    public void addCollider(GVRCollider eyePointee) {
```
liaxim: Since this whole class is deprecated does it deserve having the new add/removeCollider apis?

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRPicker.java
```diff
@@ -21,29 +21,256 @@
...
+        mRayOrigin.z = oz;
+        mRayDirection.x = dx;
+        mRayDirection.y = dy;
+        mRayDirection.z = dz;
+        mHasChanged = true;
+    }
+    
+    public void onDrawFrame(float frameTime)
+    {
+        if (getOwnerObject() != null)
```
liaxim: Should you be using the protected "owner" field instead? getOwnerObject() doesn't return null but throws in direct contradiction to its javadoc.

GVRf/Framework/framework/src/main/jni/objects/scene.cpp
```diff
@@ -23,9 +23,15 @@
 #include "objects/scene_object.h"
 
 namespace gvr {
+
+Scene* Scene::main_scene_ = NULL;
```
liaxim: Could this writable static be a concern in an environment where the user has multiple mainScenes and adds sceneObjects with colliders to some of them that not necessarily are in the active mainScene?

There could be subtle errors. For example the first time the app launches the ctor below will set the main scene but in subsequent app starts this may not be the case as the process frequently lingers on after app exit. It is all covered seemingly by the GVRViewManager but then why have this in the ctor?

If possible do not use writable static data. In general, what do you think about having fixed one and only mainScene?

liaxim:
Did the planetarium app work?

rahul27:
Yup. It worked.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVREyePointeeHolder.java
NolaDonato: The GVREyePointee class is deprecated, it is replaced by GVRCollider

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRPicker.java
NolaDonato: getOwnerObject should not throw. I will change that.

GVRf/Framework/framework/src/main/jni/objects/scene.cpp
NolaDonato: One single main scene would be far better than the confusion we have now. I know there may be subtle errors which is why I programmed around them (I think). When the app launches main_scene will be the initial scene. When the app calls getNextMainScene the main_scene will be that one. When the app calls setMainScene the main_scene will be that one. I am trying to cover all the cases of when the app changes the scene.
GVRSceneObject.attachComponent will attach a collider component to the main scene. When the main scene changes, there is a one time overhead of gathering all the colliders from the scene graph and saving them in the "current" scene.
We need some way for the C++ layer to know the scene has changed so that it can do one-time initialization for the scene graph (like gathering the colliders).

I hate the whole getMainScene / getNextMainScene thing. This is silly. The app should make a GVRScene, attach their stuff and return that scene as the result of onInit. The view manager can then do what it needs to to make this the current scene.

The whole reason I need this collider list at all is to optimize the case where most of your objects are not pickable and to allow the possibility of picking stuff you can't see. This isn't important for the camera picking but when we start to use hand controllers it might be a case we encounter.

NolaDonato:
It kinda worked for me. The OKHttp library causes a crash "could not find class okio.Buffer when I tap on something that should bring up the web view. I think I have the wrong library or something.

liaxim:
Will merge after lunch. Everything seems to work as before, including the planetarium.

GVRf/Framework/framework/src/main/jni/engine/renderer/renderer.cpp
```diff
@@ -348,14 +348,17 @@ void addRenderData(RenderData *render_data) {
 void Renderer::occlusion_cull(Scene* scene,
         std::vector<SceneObject*>& scene_objects, ShaderManager *shader_manager,
         glm::mat4 vp_matrix) {
-
+    scene->lockColliders();
```
rahul27: Is it okay to have the overhead of performing a lock/unlock and pick check if there are no colliders or pickers in the scene?

Just curious since this is the renderer's code.

NolaDonato: I could change it to not lock if the allColliders list is empty. I am not sure this is 100% safe. It assumes colliders cannot be added during rendering. We don’t lock the scene graph during cull. Instead we call SceneObject::children which looks like:

    std::vector<SceneObject*> children() {
        std::lock_guard < std::mutex > lock(children_mutex_);
        return std::vector<SceneObject*>(children_);
    }
Apparently we make a copy of the child list as we do culling. This can’t be efficient either. I think the whole cull thing needs some rework.

liaxim:
@NolaDonato Should we merge as-is or are you planning further changes for this pull request?

NolaDonato:
No further changes now unless there are bugs.

PR #622, opened 2016-06-20
adding batching for texture shader

roshanch:

-Batching support for texture shader.
-only small meshes are batched, right now size of each batch is 400 indices, mesh with larger size wont be batched.
-removing getTexture() from custom_shader.cpp, as texture unit enums can be incremented. 
-- i created a sample with around 225 objects all of them have texture and uses texture shader, without batching it gives 30 fps whereas with batching it gives 60 fps.


roshanch:
@danke-sra @liaxim @NolaDonato @vittalbabu can you review this.

liaxim:
Needs rebase

liaxim:
@roshanch Do you plan to rebase this soon?

roshanch:
@danke-sra can you please review this.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRBitmapTexture.java
```diff
@@ -268,6 +268,8 @@ public Boolean call() {
     private boolean updateCall(Bitmap bitmap) {
         glBindTexture(GL_TEXTURE_2D, getId());
         GLUtils.texImage2D(GL_TEXTURE_2D, 0, bitmap, 0);
+        android.opengl.GLES20.glGenerateMipmap(GL_TEXTURE_2D);
```
danke-sra: Minor coding style comment. Please import the statics from the GLES20 package, and reduce this line to glGenerateMapmap(...) or GLES20.gl*.

GVRf/Framework/framework/src/main/jni/objects/components/render_data.h
```diff
@@ -25,17 +25,30 @@
```
danke-sra: Please apply GearVRf formatting style -- http://www.gearvrf.org/bin/view/GearVRF/GearVRfWikiGetInvolved#CodeGuide

GVRf/Framework/framework/src/main/jni/objects/components/render_data.h
```diff
@@ -51,13 +64,56 @@ class RenderData: public Component {
...
-                    render_mask_(DEFAULT_RENDER_MASK), rendering_order_(
-                    DEFAULT_RENDERING_ORDER), offset_(false), offset_factor_(
+            Component(RenderData::getComponentType()), mesh_(0), light_(0), use_light_(false), use_lightmap_(false),batching_(false),
+                    render_mask_(DEFAULT_RENDER_MASK), batch_(nullptr),rendering_order_(
+                    DEFAULT_RENDERING_ORDER), hash_code_dirty_(true), offset_(false), offset_factor_(
                     0.0f), offset_units_(0.0f), depth_test_(true), alpha_blend_(
                     true), alpha_to_coverage_(false), sample_coverage_(1.0f), invert_coverage_mask_(GL_FALSE), draw_mode_(GL_TRIANGLES), texture_capturer(0) {
     }
 
+    void copy(RenderData& rdata){
```
danke-sra: Please make the parameter rdata const. Then, you can use this function to impl. the copy constructor without duplicating the code.

GVRf/Framework/framework/src/main/jni/objects/components/render_data.h
```diff
@@ -152,7 +213,21 @@ class RenderData: public Component {
     void set_rendering_order(int rendering_order) {
         rendering_order_ = rendering_order;
     }
-
+    Batch* getBatch(){
```
danke-sra: naming consistency. use underscore instead of camel case.

GVRf/Framework/framework/src/main/jni/objects/components/render_data.h
```diff
@@ -245,18 +328,44 @@ class RenderData: public Component {
...
+    		render_data_string.append(to_string(alpha_to_coverage_));
+    		render_data_string.append(to_string(sample_coverage_));
+    		render_data_string.append(to_string(invert_coverage_mask_));
+    		render_data_string.append(to_string(draw_mode_));
+
+    		hash_code = render_data_string;
+    		hash_code_dirty_ = false;
+
+    	}
+    	return hash_code;
```
danke-sra: Please use a local string for creating the hash code. There shouldn't be a data member called hash_code. It is also not cleared. C++ will optimize on the return value. So there is no performance concern.

GVRf/Framework/framework/src/main/jni/objects/components/render_data.h
```diff
@@ -265,9 +374,13 @@ class RenderData: public Component {
     static const int DEFAULT_RENDER_MASK = Left | Right;
     static const int DEFAULT_RENDERING_ORDER = Geometry;
     Mesh* mesh_;
+    Batch* batch_;
+    bool hash_code_dirty_;
+    std::string hash_code;
```
danke-sra: please remove this data member.

GVRf/Framework/framework/src/main/jni/objects/components/render_data.h
```diff
@@ -329,31 +442,38 @@ static inline bool compareRenderDataByOrderDistance(RenderData* i, RenderData* j
...
-            // if it is a transparent object, sort by camera distance from back to front
-            if (i->rendering_order() >= RenderData::Transparent
-                    && i->rendering_order() < RenderData::Overlay) {
-                return i->camera_distance() > j->camera_distance();
-            }
-            // otherwise sort from front to back
-            return i->camera_distance() < j->camera_distance();
-        }
 
+        	if(i->material(0) == j->material(0)) {
```
danke-sra: This impl. is too complicated. Please consider revising it later. (it's ok for this patch)

GVRf/Framework/framework/src/main/jni/objects/mesh.cpp
```diff
@@ -32,73 +32,81 @@
...
-
-    mesh->indices_.push_back(2);
-    mesh->indices_.push_back(7);
-    mesh->indices_.push_back(3);
-    mesh->indices_.push_back(2);
-    mesh->indices_.push_back(6);
-    mesh->indices_.push_back(7);
-
-    return mesh;
+    if(bounding_box_mesh_ == nullptr || !have_bounding_volume_)
```
danke-sra: Suggestion. Return on the "false" case to reduce nesting.

GVRf/Framework/framework/src/main/jni/objects/textures/cubemap_texture.h
```diff
@@ -138,7 +140,7 @@ class CubemapTexture: public Texture {
                             "AndroidBitmap_lockPixels () failed! error = " + ret;
                     throw error;
                 }
-
+                LOGE("cubemap ????");
```
danke-sra: this line seems unintended?

danke-sra:
Generally OK. Please re-format some code and refer to inline comments.

danke-sra:
Can you please also fix the line " LOGE("cubemap ????");" ?

danke-sra:
I think the code is OK to go once the extra LOGE line is removed. Thanks.

roshanch:
@liaxim  can we merge this ?

liaxim:
@roshanch Which samples have been verified as working?

roshanch:
Planeterium and I modified simple sample where I added around 200 scene objects.

Vittalbabu:
What is the fps gain for Planeterium?

roshanch:
previously it was 30 fps now its 45-60 fps as this app uses custom shaders as well.

danke-sra:
I think the code is OK to go. Thanks roshan.

liaxim:
Will run tests tomorrow and merge

NolaDonato:
I have accumulated a set of test models and have run them on master. I have screenshots of them. I will test them all against Roshan’s and Vittal’s changes today. I will send out a test report at the end of the day when I finish.

NolaDonato:
I tested this with a set of jassimp and x3d assets. For jassimp, the behavior is the same as master. It can read the entire set of test models. I pulled over the X3D parser and tried a set of X3D models with the same test program. The sample crashes after the second model.

LOGCAT has this mysterious message:
07-07 20:28:44.789: D/gvrf(13222): SHADER: setVertexAttrib a_matrix_index
07-07 20:28:44.789: A/libc(13222): Fatal signal 6 (SIGABRT) at 0x000033a6 (code=-6), thread 13270 (Thread-31994)

I think we should push this change and test again when the X3D parser has been integrated. We should expect some issues here.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRSceneObject.java
```diff
@@ -1219,6 +1219,13 @@ public final BoundingVolume expandBoundingVolume(
     public final BoundingVolume expandBoundingVolume(final Vector3f center, final float radius) {
         return expandBoundingVolume(center.x, center.y, center.z, radius);
     }
+    
+    /*
+     * Check if the object is culled in the frame or not
```
liaxim: What good is this method to be exposed to end users? If that was to be exposed it probably would take an event-based mechanism.

GVRf/Framework/framework/src/main/jni/objects/components/render_data.h
liaxim: Please use spaces instead of tabs.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRMesh.java
```diff
@@ -361,7 +361,14 @@ public GVRMesh getBoundingBox() {
     public List<GVRBone> getBones() {
         return mBones;
     }
-
+    /**
+     * Sets the status of mesh, whether static or dynamic, by default it is static.
```
liaxim: Please provide examples of when people should use dynamic, when static; advantages/disadvantages. We can't force them trying to figure what this does by reading the source code.

GVRf/Framework/framework/src/main/jni/engine/renderer/batch.h
```diff
@@ -0,0 +1,94 @@
...
+#include <vector>
+#include <string>
+#include <set>
+namespace gvr{
+
+class RenderData;
+class Material;
+class Mesh;
+class Batch {
+public:
```
liaxim: This is all-tabs. Please get your formatter configured according to the coding style.

GVRf/Framework/framework/src/main/jni/engine/renderer/batch.h
```diff
@@ -0,0 +1,94 @@
...
+	}
+	bool isBatchDirty();
+	unsigned int getIndexCount(){
+		return index_count_;
+	}
+private:
+	std::map<RenderData*,int>matrix_index_map;
+	std::set<RenderData*>render_data_set_;  // use it later if we want to modify meshes
+
+	Mesh *mesh;
```
liaxim: Please use consistent names: mesh -> mesh_, renderdata -> renderdata_. Why pointers? Having to new and delete is just more work.

GVRf/Framework/framework/src/main/jni/engine/renderer/batch.cpp
```diff
@@ -0,0 +1,197 @@
...
+            renderdata->set_batching(true);
+            mesh->set_vertices(vertices_);
+            mesh->set_normals(normals_);
+            mesh->set_tex_coords(tex_coords_);
+            mesh->set_indices(indices_);
+
+            // if render data has light, custom attribute will have location:3 if it does not have light,
+            // it will have location 2
+            if (renderdata->light_enabled() && renderdata->light()->enabled()) {
+                mesh->setVertexAttribLocF(3, "a_matrix_index");
```
liaxim: Where do these 2 and 3 come from exactly? Is there glBindAttribLocation or layout(location=) somewhere?

GVRf/Framework/framework/src/main/jni/objects/components/render_data.h
```diff
@@ -245,18 +319,44 @@ class RenderData: public Component {
...
         texture_capturer = capturer;
     }
-
+    // TODO: need to consider texture_capturer in hash_code ?
     TextureCapturer *get_texture_capturer() {
         return texture_capturer;
     }
 
+    std::string getHashCode(){
+    	if(!hash_code_dirty_){
```
liaxim: All tabs instead of spaces too.

GVRf/Framework/framework/src/main/jni/engine/renderer/renderer.cpp
```diff
@@ -121,7 +130,212 @@ void Renderer::state_sort() {
...
+                    std::unordered_map<Batch*, int>::iterator it =
+                            batch_map.find(current_batch);
+                    if (it != batch_map.end()) {
+                        int index = it->second;
+                        batch_set.erase(batch_set.begin() + index);
+                    }
+
+                    delete current_batch;
+                    Batch* new_batch = new Batch(MAX_INDICES, MAX_INDICES);
+                    render_data->setBatch(current_batch);
```
liaxim: Should this be setBatch(new_batch)?

GVRf/Framework/framework/src/main/jni/objects/mesh.cpp
```diff
@@ -32,8 +32,18 @@
 
 namespace gvr {
 Mesh* Mesh::createBoundingBox() {
+    if (bounding_box_mesh_ != nullptr && have_bounding_volume_) {
+        return bounding_box_mesh_;
```
liaxim: There is code that expects createBoundingBox to return a new instance always. The Java side does that (see GVRMesh.getBoundingBox); Java assumes it owns the native ptr and will delete it at some point. There is also Picker::pickSceneObjectAgainstBoundingBox. You might have to create a special version of this method that does what you want in the cases where it matters for your change.

liaxim:
You might want to check out the SIGABRT Nola encountered, despite that it is seems to be X3D-specific. It shouldn't be crashing like this no matter what since X3D is virtually all Java..

In general, impossible to tell whether the memory is properly managed by this change. It is a change parts of which could have been unit tested actually.

At least the serious issues and the javadoc must be addressed.

liaxim:
gvr-immersivepedia crashes on launch; s7 mali:

```
07-08 08:50:52.038  1656  2320 D gvrf    : SHADER: setVertexAttrib a_matrix_index
07-08 08:50:52.058  1656  2320 E gvrf    : Could not compile shader 35633:
07-08 08:50:52.058  1656  2320 E gvrf    : 2:35: L0002: No matching function for call to 'inverse'
07-08 08:50:52.058  1656  2320 F libc    : Fatal signal 11 (SIGSEGV), code 1, fault addr 0x3c in tid 2320 (GLThread 1477)
```

GVRf/Framework/framework/src/main/jni/objects/mesh.cpp
roshanch: If that's the case i will revert this. I am not using this method.

GVRf/Framework/framework/src/main/jni/engine/renderer/batch.cpp
roshanch: If we dont call glBindAttribLocation(), locations are assigned by GL. so when we have light, it will take 3rd location always and 2nd location when we don't have lights. first 3 are hard-coded by calling glBindAttribLocation.

Vittalbabu: @roshanch, @liaxim  With attribute interleaving, we will not have any pre-defined locations for attributes. Let's get this change merged and then merge my attribute interleaving code -- I will remove these lines in my merge.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRSceneObject.java
roshanch: May be we could use it where we are traversing though all the scene-objects in a scene.

GVRf/Framework/framework/src/main/jni/engine/renderer/renderer.cpp
roshanch: good catch.

GVRf/Framework/framework/src/main/jni/engine/renderer/batch.cpp
liaxim: Just curious, are the locations guaranteed by the spec to monotonically increase like this if you don't use glBingAttribLocation?

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRSceneObject.java
liaxim: I'd say when the need arises we expose it. Otherwise we just confuse people. Unless you can give a specific example where application developers can find it useful? And how to correctly use it since this could be called by any thread.

GVRf/Framework/framework/src/main/jni/engine/renderer/batch.cpp
roshanch: the spec does not guaranty that but i haven't see the case where it is not. When we will have the interleaving code in, it will remove this dependency.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRSceneObject.java
roshanch: Okay I will remove it for now.

GVRf/Framework/framework/src/main/jni/engine/renderer/batch.cpp
Vittalbabu: This might be driver dependent but normally attribute locations are sequential unless overridden by location or explicit binding to location.

As a side note, matching out/in variables between stages ((eg) out of Vertex shader to in of Fragment shader) is trickier as there are both implicit and explicit rules.

Vittalbabu: That said, it is usually preferable to query via GL calls for location of an attribute but this issue will go away with attribute interleaving.

liaxim:
You need to rebase also

GVRf/Framework/framework/src/main/jni/engine/renderer/batch.cpp
```diff
@@ -0,0 +1,197 @@
...
+    indices_.clear();
+    matrices_.clear();
+    matrix_indices_.clear();
+    render_data_set_.clear();
+
+    vertices_.reserve(no_vertices);
+    indices_.reserve(no_indices);
+    normals_.reserve(no_vertices);
+    tex_coords_.reserve(no_vertices);
+    matrix_indices_.reserve(no_vertices);
```
Vittalbabu: Can we change no_indices, no_vertices to num_indices, num_vertices? no_indices could mean *no* indices as opposed to number of indices

GVRf/Framework/framework/src/main/jni/engine/renderer/batch.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+Batch::Batch() :
+        draw_count_(0), vertex_count_(0), index_count_(0), vertex_limit_(400),
+        indices_limit_(400), renderdata_(nullptr), index_offset_(0), not_batched_(false) {
+    vertices_.clear();
+    normals_.clear();
+    tex_coords_.clear();
+    indices_.clear();
+    matrices_.clear();
+    matrix_indices_.clear();
```
Vittalbabu: If common inititalisation in done in the 2 ctors, better to move them to a function called in both ctors.

Also, it this necessary?

GVRf/Framework/framework/src/main/jni/engine/renderer/batch.cpp
```diff
@@ -0,0 +1,197 @@
...
+   // mesh = new Mesh();
+    glm::vec3 init_vec3(0, 0, 0);
+    glm::vec2 init_vec2(0, 0);
+    for (int i = 0; i < no_vertices; i++) {
+        vertices_.push_back(init_vec3);
+        tex_coords_.push_back(init_vec2);
+        matrix_indices_.push_back(0);
+        normals_.push_back(init_vec3);
+        indices_.push_back(0);
+    }
```
Vittalbabu: What is meant to be stored in vertices_ vector (and other vectors)? What is this used for? What happens if the other ctor is called and we set this data in the mesh.

GVRf/Framework/framework/src/main/jni/engine/renderer/renderer.cpp
```diff
@@ -77,23 +81,28 @@ void Renderer::frustum_cull(glm::vec3 camera_position, SceneObject *object,
 
         cullVal = object->frustumCull(camera_position, frustum, planeMask);
         if (cullVal == 0) {
+            object->setCullStatus(true);
```
Vittalbabu: Prefer enums instead of magic values like 0, 2 and 3.

GVRf/Framework/framework/src/main/jni/engine/renderer/renderer.cpp
```diff
@@ -121,7 +130,212 @@ void Renderer::state_sort() {
...
+                if (existing_batch
+                        && existing_batch->getNumberOfMeshes() < BATCH_SIZE) {
+                    // add failed because mesh is large try with next batch
+                    if (!existing_batch->add(render_data)) {
+                        Batch* new_batch = new Batch(MAX_INDICES, MAX_INDICES);
+                        new_batch->add(render_data);
+                        render_data->setBatch(new_batch);
+                        batch_set.push_back(new_batch);
+                        batch_map[new_batch] = batch_set.size() - 1;
+                        existing_batch = new_batch;
```
Vittalbabu: Same Batch creation code in 6 locations -- for easier maintenance move this to a function.

GVRf/Framework/framework/src/main/jni/engine/renderer/batch.h
```diff
@@ -0,0 +1,95 @@
...
+#include <string>
+#include <set>
+namespace gvr{
+
+class RenderData;
+class Material;
+class Mesh;
+class Batch {
+public:
+	Batch();
```
Vittalbabu: Why have this ctor? Is this being called?

GVRf/Framework/framework/src/main/jni/engine/renderer/renderer.cpp
```diff
@@ -121,7 +130,212 @@ void Renderer::state_sort() {
...
+                current_batch->UpdateModelMatrix(render_data,
+                        render_data->owner_object()->transform()->getModelMatrix());
+            }
+
+            if (!render_data->mesh()->isDynamic()) {
+                if (batch_map.find(current_batch) == batch_map.end()) {
+                    batch_set.push_back(current_batch);
+                    batch_map[current_batch] = batch_set.size() - 1;
+                }
+            } else { // mesh is dynamic
```
Vittalbabu: Is the code for dynamic mesh the same as in the previous if block? If it meant to be same refactor it into a common function.

GVRf/Framework/framework/src/main/jni/engine/renderer/renderer.cpp
```diff
@@ -121,7 +130,212 @@ void Renderer::state_sort() {
...
+        } else { // batch is not null
+
+            // update the transform if model matrix is changed
+            if (render_data->owner_object()->isTransformDirty()
+                    && render_data->owner_object()->transform()) {
+                current_batch->UpdateModelMatrix(render_data,
+                        render_data->owner_object()->transform()->getModelMatrix());
+            }
+
+            if (!render_data->mesh()->isDynamic()) {
```
Vittalbabu: Stylistic comment -- either have a function isStatic() and use it instead of negating isDynamic() or move the dynamic case first. Sometimes it is easy to overlook the negation in the if stmt clause while looking at this code months later. If you have both if and else, don't use negation in the if clause.

GVRf/Framework/framework/src/main/jni/engine/renderer/renderer.cpp
```diff
@@ -161,17 +375,115 @@ void Renderer::cullFromCamera(Scene *scene, Camera* camera,
...
+    if (!(rstate.render_mask & render_data->render_mask()))
+        return;
+
+    if (render_data->offset()) {
+        GL(glEnable (GL_POLYGON_OFFSET_FILL));
+        GL(
+                glPolygonOffset(render_data->offset_factor(),
+                        render_data->offset_units()));
+    }
+    if (!render_data->depth_test()) {
```
Vittalbabu: Do you need to have an else clause to enable Depth test, blend, etc or is the assumption that the render state will have these states set?

roshanch: by default all the states are enabled.

GVRf/Framework/framework/src/main/jni/engine/renderer/batch.cpp
roshanch: vertices_ used to store vertex data in batch. will remove other constructor.

GVRf/Framework/framework/src/main/jni/objects/mesh.cpp
```diff
@@ -207,7 +217,86 @@ void Mesh::bindVertexAttributes(GLuint programID)
...
+void Mesh::updateVBOs(unsigned int vtx_count, unsigned int idx_start, unsigned int indices_count, std::vector<float>& matrix_indices,
+		const std::vector<glm::vec3>& vertices, std::vector<unsigned short>& indices, const std::vector<glm::vec2>& texcords,
+		const std::vector<glm::vec3>& normals){
+
+	if(vert_vboID_ == GVR_INVALID || norm_vboID_ == GVR_INVALID
+            || tex_vboID_ == GVR_INVALID || triangle_vboID_ ==GVR_INVALID){
+        std::string error = "VAO is not created ";
+        throw error;
+        return;
+    }
```
Vittalbabu: This should not be necessary -- with the attribute interleaving code and https://github.com/Samsung/GearVRf/pull/685 VAO/VBO details/state are internal to the Mesh instance

roshanch:
Added support for multiview extension for texture and custom shader.

thomasflynn:
is this ready for review again?

liaxim:
Yes. I am not sure why the multiview had to be a part of it. We will have to test and review again.

roshanch:
I tested with simple-sample, planetarium and jassmiploader on note4 and s6 edge.

roshanch:
Added patch for user input for multiview from gvr.xml file. If user enables flag and does not provide proper shader, error is thrown on native side. all the comments are addressed. @NolaDonato @thomasflynn @liaxim @Vittalbabu it is ready for review.

roshanch:
user has to give useMultiview="true" in xml file. by default it is disabled

NolaDonato:
objects that do not have textures on them do not display at all. I tried using the texture shader in tutorial lesson 2 (modifying the code as follows) and nothing shows up:
    GVRSceneObject makeBalloon(GVRContext context)
    {
        GVRSceneObject sphere = new GVRSphereSceneObject(context, true);
        GVRRenderData rdata = sphere.getRenderData();
        GVRMaterial mtl = new GVRMaterial(context, GVRMaterial.GVRShaderType.Texture.ID);
       // mtl.setDiffuseColor(1.0f, 0.0f, 1.0f, 0.5f);
        mtl.setColor(1, 0, 0);
        sphere.setName("balloon");
        //rdata.setShaderTemplate(GVRPhongShader.class);
        //rdata.setAlphaBlend(true);
        rdata.setMaterial(mtl);
        //rdata.setRenderingOrder(GVRRenderingOrder.TRANSPARENT);
        sphere.getTransform().setPositionZ(-3.0f);
        return sphere;
    }

I am not sure if this bug is due to the batching changes or not but if you fix it I can test the batching with the particle system from the tutorial. As it stands now I can't do that test.

NolaDonato:
I tested all the jassimp assets that I have with multiview and it works OK (except for the area light which we know fails on master too)

roshanch:
if scene object does not have any texture, we are supposed to use assimp_shader or custom shader type, there is check on renderer, if texture is null, it returns, hence we don't see anything.

roshanch:
Tested with tutorial-5 sample, works fine.

NolaDonato:
What kind of performance improvement do you get with it? You can change the particle system emission rate to be higher and get a lot more balloons. The framerate should be lots better with batching because this is exactly the case it was designed for.

NolaDonato:
Did you remember to comment out the lines that make it use the Phong shader?
I pulled over your patch again and it still does not show anything.


    GVRSceneObject makeBalloon(GVRContext context)
    {
        GVRSceneObject balloon = new GVRSceneObject(context, mSphereMesh);
        GVRRenderData rdata = balloon.getRenderData();
        GVRSphereCollider collider = new GVRSphereCollider(context);
        Random rand = new Random();
        int mtlIndex = rand.nextInt(mMaterials.size() - 1);

        balloon.setName("balloon");
//        rdata.setAlphaBlend(true);
        rdata.setMaterial(mMaterials.get(mtlIndex));
//        rdata.setRenderingOrder(GVRRenderingOrder.TRANSPARENT);
        collider.setRadius(0.8f);
        balloon.attachComponent(collider);
        return balloon;
    }

roshanch:
If you want to use texture shader, scene object should have texture with its material, if it does not have, it  won't be rendered.

NolaDonato:
That is ridiculous. No user will expect this behaviour - this will always be a surprise. It still surprised me and I have seen it before. This is why we need to eliminate the texture shader



roshanch:
That's why we have different types of shaders to support different use-cases (prior to custom shader). I agree that we should remove all the shader types and just have custom one. For this particular use-case we can't use texture shader with current architecture. We can again add one more #ifdef in texture shader to take color from material but according to me that doesn't makes sense as we already have custom and assimp shaders which support this. And again its going to complicate texture shader with one more case as we already have 8 different cases with #ifdefs.

NolaDonato:
I am not suggesting we change it. You are right – making something we are going to take out more complex is not a good idea. I just hate the way it works, don’t you?

I wanted to test your batching with the particle system. I will add a texture to the balls so I can do that.

-Nola

liaxim:
@NolaDonato You could try commenting out the call to checkTextureReady in Renderer::renderMaterialShader. Should end up rendering with the error shader (green).

GVRf/Framework/framework/src/main/backends/oculus/org/gearvrf/GVRViewManager.java
```diff
@@ -663,6 +663,7 @@ public void beforeDrawEyes() {
             if (mActivity.getAppSettings().showLoadingIcon) {
                 mSplashScreen = mScript.createSplashScreen();
                 if (mSplashScreen != null) {
+                    getMainScene().setUseMultiview(mActivity.getAppSettings().isMultiviewSet());
```
liaxim: There is a separate path where there is no splash screen requested and this flag is not going to be set properly in that case. And we shouldn't be using the scene for this flag since the scenes can be instantiated all the time. We don't want to keep track of all instances and try to set the flag for all.

GVRf/Framework/framework/src/main/backends/oculus/org/gearvrf/utility/VrAppSettings.java
```diff
@@ -618,6 +621,23 @@ public static void setShowDebugLog(boolean showDebugLog) {
     }
 
     /**
+     * Set if user wants to use multiview or not
+     * 
+     * @param useMultiview
+     *            
+     */
+    public void setUseMultiview(boolean useMultiview){
```
liaxim: Please make the method private.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRScene.java
```diff
@@ -113,6 +112,25 @@ private GVRScene(GVRContext gvrContext, long ptr) {
     }
 
     /**
+     * Sets the useMultiview flag in scene
```
liaxim: We should find a better place for the flag; a place that is completely hidden from end users.

liaxim:
@roshanch Do you think you could separate out the multiview part in a different pull request? The batching might and probably is ready but I think we will need to polish the multiview a bit more. Now we have to hold up the pull request because of it..

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRShaderTemplate.java
```diff
@@ -14,6 +14,9 @@
  */
```
liaxim: No real changes to this file. Please revert to previous version. You can do that with ```git checkout HEAD~1 GVRShaderTemplate.java```

GVRf/Framework/framework/src/main/jni/backends/oculus/activity.cpp
```diff
@@ -140,10 +140,24 @@ void GVRActivity::onSurfaceChanged(JNIEnv& env) {
...
+        const char* extensions = (const char*)glGetString(GL_EXTENSIONS);
+        if(multiview && std::strstr(extensions, "GL_OVR_multiview2")!= NULL){
+            use_multiview = true;
+        }
+        if(multiview && !use_multiview){
+            std::string error = "Multiview is not supported by your device";
+            LOGE(" Multiview is not supported by your device");
+            throw error;
+        }
+
```
liaxim: This multiview flag is in too many places. We agreed that using a static is unavoidable at the moment. But it could be that there is a static in the Renderer class; a static that is set from here and this should be it. Or not?

roshanch:
@liaxim apart from rebase do we need any code change? We have tested this with x3d models, as there is issue with parser some of them haven't been tested.

liaxim:
Let's get this merged.

liaxim:
Will merge in the evening if there are no objections.

NolaDonato:
Yes merge

PR #625, opened 2016-06-21
Make bindShaders(GVRSceneObject) public

rahul27:
r.rudradevan@samsung.com

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRScene.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRScene.java
@@ -405,11 +405,11 @@ public class GVRScene extends GVRHybridObject implements PrettyPrint, IScriptabl
      * bindShaders may need to be called again to regenerate the correct shaders
      * for the new lighting conditions. This function is called whenever a scene
      * object is added at the root of the scene.
-     * @see GVRRenderData.bindShader
+     * @see GVRRenderData#bindShader(GVRScene)
      * @see GVRShaderTemplate
-     * @see GVRScene.addSceneObject
+     * @see GVRScene#addSceneObject(GVRSceneObject)
      */
-    protected void bindShaders(GVRSceneObject root) {
+    public void bindShaders(GVRSceneObject root) {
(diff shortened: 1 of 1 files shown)


liaxim:
Why do you need this? It looks like an API that ideally the users should not know about. @NolaDonato ?

rahul27:
Here is what happens: You add a scene object "A" to the main scene. GVRf internally calls bindShaders(). Next you load a model scene object "B" and add it as a child of "A". "B" does not show up as bindShaders has not been called in this case. 

Here I could make an explicit scene.bindShaders() on the entire scene graph or choose to only call bindShaders on object "A" using the new API.

liaxim:
Sound exactly like the thing the framework should do for you.

rahul27:
In that case the framework would have to be smart about all modifications made to the scene graph and make this call only for cases where it is required. 

Since the app has knowledge of the scene graph and knows about all objects added/removed it would be more efficient to force this call only for cases where necessary.

NolaDonato:
It would be great if the framework could do that for you. For that to happen the call to GVRSceneObject.addChildObject would need to know the current scene under ALL conditions. It would need to work from onInit, from onStep, from everywhere. Adding a child to a scene object would then incur the overhead of calling bindShaders whether you need it or not. This overhead involves examining all of the scene objects being added to see if there is a light. If a new light is found, ALL the shaders on all the scene objects in the hierarchy could potentially need to be recompiled so the entire hierarchy is scanned. If there are no new lights then it just has to check all the new objects.

Another potential fix for this is to call GVRScene.bindShaders during the cull stage of rendering when we first encounter a scene object without a shader. Since bindShaders could potentially compile one or more shaders I was not clear that putting that into our main rendering loop would be a good idea. But that is the only place where we scan the entire hierarchy every frame - we do not do this on the Java side.

liaxim:
I personally vote for keeping the API as-is and have the user call scene.bindShaders() in special circumstances until we have a proper way of dealing with the problem Rahul described. I assume Rahul wants the API exposed. So let us know Nola should we merge this or not.

NolaDonato:
Yes merge it

PR #629, opened 2016-06-21
Don't allow a scene object to be added twice 

NolaDonato:
GVRSceneObject.addChildObject now throws an exception if you try to add
a scene object that already has a parent.
Replaced GVRScene children array with single root node. Added
GVRScene.getRoot().


GVRf/Framework/framework/src/main/java/org/gearvrf/GVRScene.java
```diff
@@ -128,6 +130,14 @@ public void clear() {
...
+     * @return top level scene object.
+     */
+    public GVRSceneObject getRoot() {
+        return mSceneRoot;
+    }
+    
+    /**
      * The top-level scene objects.
      * 
      * @return A read-only list containing all the 'root' scene objects (those
```
liaxim: Javadoc probably needs update since now we have a root sceneObject and not a list of "root objects".

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRScene.java
```diff
@@ -439,15 +413,6 @@ public void bindShaders(GVRSceneObject root) {
...
-                if (l == light) {
-                    return false;
-                }
-                if (l.getClass().equals(light.getClass())) {
-                    ++classIndex;
-                }
-            }
             Integer lightIndex = mLightList.size();
             String name = "light" + lightIndex.toString();
             mLightList.add(light);
```
liaxim: You could be using the return value of Set.add to determine whether the light was actually added or not and return the real result to the caller.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRSceneObject.java
```diff
@@ -699,11 +699,20 @@ public GVRSceneObject getParent() {
...
      *            object.
+     * @return true if child was added, else false
      */
-    public void addChildObject(GVRSceneObject child) {
+    public boolean addChildObject(GVRSceneObject child) {
+        if (child.mParent == this) {
+            return false;
+        }
+        if (child.mParent != null) {
+            Log.e("gvrf", "Cannot add child - it has another parent");
```
liaxim: How can you tell why false was returned? There are two separate conditions that are quite different that can give you that result.

If the operation of adding a child that already has a different parent is not allowed/an error then we should throw an exception. It seems to me like an abnormal operation and the users should know clearly it is an error they need to fix.

GVRf/Framework/framework/src/main/jni/objects/scene_object.cpp
```diff
@@ -134,6 +143,14 @@ SceneObject* SceneObject::getChildByIndex(int index) {
     }
 }
 
+void SceneObject::getDescendants(std::vector<SceneObject*>& descendants) {
+    for (auto it = children_.begin(); it != children_.end(); ++it) {
```
liaxim: Should this be locking the mutex also?

liaxim:
Very nice change.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRSceneObject.java
NolaDonato: I did that and it broke immersivepedia. I agree that it should throw an exception. I figured out how to fix immersivepedia too so I will do that also.

GVRf/Framework/framework/src/main/jni/objects/scene_object.cpp
NolaDonato: Yes, it should. I will fix that. Thanks for keeping me thread-safe.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRScene.java
NolaDonato: This code has been simplified.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRScene.java
NolaDonato: I updated the documentation in GVRScene with respect to the root node, adding & removing children from it.

liaxim:
This needs rebase&merge now

NolaDonato:
I did rebase and merge

liaxim:
Github says this still needs a rebase.

liaxim:
I rebased your change here: https://github.com/liaxim/GearVRf/tree/addtwice_2 and also squashed the commits.

The conflicts are in GVRScene, scene.h and scene.cpp; my branch doesn't compile since I didn't update the collider-related code. You can diff. It is pretty annoying to rebase a change that contains a long list of commits.

There are some articles about doing rebase with tortoise:
http://stackoverflow.com/questions/12528854/how-to-perform-rebase-squash-using-tortoisegit
https://tortoisegit.org/docs/tortoisegit/tgit-dug-rebase.html

We need to figure out a workflow for you that works. I am no tortoise user but we should sit down together and come up with something.

NolaDonato:
I am using GitHub desktop. Tortoise stopped working on my machine. To make that branch I did “git merge –squash”. Is that the wrong thing?

Nola

liaxim:
This is what works for me:

1. Configure "upstream" remote
https://help.github.com/articles/configuring-a-remote-for-a-fork/
2. Fetch "upstream"; this will give you the most up-to-date master of "upstream"
```
git fetch upstream
```
3. Rebase your branch on top of upstream/master
```
git checkout <your_branch>
git rebase upstream/master
```
Generally using "git merge" is undesirable.

NolaDonato:
Replaced by PR #663 which rebases, resolves conflicts and squashes commits

PR #630, opened 2016-06-21
API to set viewport for a scene

deepakrawat22:
API to set viewport. Usage:

mainScene.getMainCameraRig().setViewport(x, y, width, height);

deepak.rawat@samsung.com

### DIFF
--- a/GVRf/Framework/framework/src/main/jni/oculus/activity.h
+++ b/GVRf/Framework/framework/src/main/jni/oculus/activity.h
@@ -35,6 +35,7 @@ public:
 
     bool updateSensoredScene();
     void setCameraRig(jlong cameraRig);
+    void setViewport(int x_, int y_, int width_, int height_);
 
     GVRViewManager viewManager_;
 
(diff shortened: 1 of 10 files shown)


liaxim:
Ok, how about instead of punching this hole in the APIs we add viewportX, viewportY, viewportWidth and viewportHeight as configuration parameters to eyeBufferParms?

deepakrawat22:
Having the API in getMainCameraRig() allows to change the viewport at run-time, although can't say if its a useful feature to change viewport during run-time.

liaxim:
Well, where does this API really belong? It is something highly specific so until we know why we want it dynamic I'd say make it a static, somewhat hidden, configuration. Would your vuforia changes work across devices with a static config?

deepakrawat22:
Yes vuforia changes works well with static config. Ok will move the viewport to config file. Thanks.

deepakrawat22:
Set viewport in config file

GVRf/Framework/framework/src/main/jni/oculus/activity.h
```diff
@@ -35,6 +35,7 @@ class GVRActivity
 
     bool updateSensoredScene();
     void setCameraRig(jlong cameraRig);
+    void setViewport(int x_, int y_, int width_, int height_);
```
liaxim: Unused. Remove?

liaxim:
Will merge in the afternoon

Issue #631, opened 2016-06-21
GVRVideoSceneObject memory leak


liaxim:
```
                mVideo = new GVRVideo(mediaPlayer, texture);
                gvrContext.registerDrawFrameListener(mVideo);
```

The drawFrameListener is not unregistered from anywhere. The GVRVideo instance will be alive forever keeping a SurfaceTexture and MediaPlayer instance around, potentially.

deepakrawat22:
@liaxim I will work on this. Thanks.

liaxim:
Awesome. You can use GVRTextViewSceneObject as an example of one way how to deal with this problem. See the weakRefs and the internal drawFrameListener.

PR #633, opened 2016-06-22
Video scene unregisterDrawFrameListener fix for #631

deepakrawat22:
deepak.rawat@samsung.com

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/scene_objects/GVRVideoSceneObject.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/scene_objects/GVRVideoSceneObject.java
@@ -363,19 +363,21 @@ public class GVRVideoSceneObject extends GVRSceneObject {
          */
         public void release() {
             if (mMediaPlayerRef != null) {
-                if (mMediaPlayerRef.get() != null) {
-                    mMediaPlayerRef.get().release();
+                MediaPlayer mediaPlayer = mMediaPlayerRef.get();
+                if (mediaPlayer != null) {
+                    mediaPlayer.release();
                 }
             }
         }
 
         @Override
         public void onDrawFrame(float drawTime) {
-            if (mMediaPlayerRef.get() != null && mActive) {
+            MediaPlayer mediaPlayer = mMediaPlayerRef.get();
+            if (mediaPlayer != null && mActive) {
                 mSurfaceTexture.updateTexImage();
             }
 
-            if (mMediaPlayerRef.get() == null) {
+            if (mediaPlayer == null) {
                 mContext.unregisterDrawFrameListener(this);
             }
         }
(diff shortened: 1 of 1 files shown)


GVRf/Framework/framework/src/main/java/org/gearvrf/scene_objects/GVRVideoSceneObject.java
```diff
@@ -359,17 +362,22 @@ public long getTimeStamp() {
          * {@link MediaPlayer}
          */
         public void release() {
-            if (mMediaPlayer != null) {
-                mMediaPlayer.release();
-                mMediaPlayer = null;
+            if (mMediaPlayerRef != null) {
```
liaxim: You need to get() and hold on to the reference via a strong one. Theoretically by the time you make the second get() call it may return null.

GVRf/Framework/framework/src/main/java/org/gearvrf/scene_objects/GVRVideoSceneObject.java
```diff
@@ -95,7 +96,7 @@ public GVRVideoSceneObject(final GVRContext gvrContext, GVRMesh mesh,
             @Override
             public void run() {
                 // Because texture.getId() is called, this needs to run in GL thread
-                mVideo = new GVRVideo(mediaPlayer, texture);
+                mVideo = new GVRVideo(gvrContext, mediaPlayer, texture);
                 gvrContext.registerDrawFrameListener(mVideo);
```
liaxim: The drawFrameListener needs to be unregistered at some point. Otherwise the GVRVideo instance will be still around for as long as the app runs.

GVRf/Framework/framework/src/main/java/org/gearvrf/scene_objects/GVRVideoSceneObject.java
```diff
@@ -254,8 +255,9 @@ public long getTimeStamp() {
 
     private static class GVRVideo implements GVRDrawFrameListener {
 
+        private final GVRContext mContext;
         private SurfaceTexture mSurfaceTexture = null;
```
liaxim: The surfaceTexture can still be leaked because the GVRVideo instance can be leaked.

liaxim: Ok, noticed the unregister. Scratch the previous comment.

liaxim:
Btw, did you test whether this works?

liaxim:
Please update the use of get() as noted.

deepakrawat22:
@liaxim I updated the usage of get() as per review comments. As per my test unregisterDrawFrameListener() is getting called once the instance MediaPlayer is no more.

liaxim:
Planning to merge after lunch

PR #634, opened 2016-06-23
fixed problem with shader transparency 

NolaDonato:
This fixes a regression introduced by the transparency fix. The multilight and eyepicking samples should be fixed by this.


liaxim:
Please if possible remove the unnecessary commits. The history is difficult enough to deal with as is.

NolaDonato:
I reverted one commit from another branch that didn't belong. The rest is from rebasing. I only changed three files.

liaxim:
If you were rebasing there would not be merge commits.

NolaDonato:
I am using GitHub desktop and doing “Sync”. I thought that did a rebase but I am not really sure.

PR #636, opened 2016-06-24
3DCursor: Use the setEnable call to switch assets

rahul27:
* Exposed a setVisible() call to hide the Cursor
* Also call bindShaders() after scene objects have been added
to a root node

r.rudradevan@samsung.com

### DIFF
--- a/GVRf/Extensions/3DCursor/3DCursorLibrary/src/main/java/org/gearvrf/io/cursor3d/AnimatedCursorAsset.java
+++ b/GVRf/Extensions/3DCursor/3DCursorLibrary/src/main/java/org/gearvrf/io/cursor3d/AnimatedCursorAsset.java
@@ -23,6 +23,7 @@ import org.gearvrf.GVRHybridObject;
 import org.gearvrf.GVRMaterial;
 import org.gearvrf.GVRMesh;
 import org.gearvrf.GVRRenderData;
+import org.gearvrf.GVRSceneObject;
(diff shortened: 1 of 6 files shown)


NolaDonato:
I suggest making CursorAsset inherit from GVRBehavior. This will allow you to attach it to a scene object. Then you can eliminate set() and reset().

MeshCursorAsset looks an awful lot like a GVRSceneObject. I don't understand why you don't just load multiple models and use a GVRSwitch to select the one you want.

In a number of places you override attachRenderData and detachRenderData. This is a bad idea because it won't actually work in all cases. You can call GVRSceneObject.attachComponent to attach a render data component to a scene object. Again, multiple scene objects each with their own render data is a better way to express what you are trying to do than attaching and detaching render data repeatedly. This approach defeats some of our internal optimizations, like batching similarly shaded objects together.

rahul27:
This change is the first step to getting there. You will see that this patch gets rid of the attach/detach renderdata calls and also does away with the array of renderdata objects.

Am waiting for the assets for the actual cursor itself, when that is ready. I would just load them in and attach a GVRSwitch to switch cursor states.

NolaDonato:
OK then this one looks good to go.

liaxim:
Will merge lunch-time.

Also it sounds to me like detach/attachRenderData should be final. It is another example of why we should discourage inheritance and be providing interfaces for stuff the users want to plug into the framework.

PR #640, opened 2016-06-24
Fixed material transparency in phong shader (again)

NolaDonato:
This pull request replaces #634. It is the same with the commits squashed.


liaxim:
Looks good to me. Will merge after lunch.

PR #642, opened 2016-06-25
Bugfix for GVRBehavior getType

parthmehta209:
mType does not need to be declared as a field in GVRBehavior, it is
already a field in GVRComponent. The bug was causing
GVRSceneObject.getComponent(long type) to not work correctly


NolaDonato:
good catch

PR #646, opened 2016-06-28
Assigning mType in GVRBehavior and subclasses

parthmehta209:
Each subclass of GVRBehavior needs a unique type value. Previous approach
of using System.getCurrentTimeMillis() was creating clashes between multiple subclasses this adds a hashcode of the class with the time stamp to create a unique type.

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRSwitch.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRSwitch.java
@@ -32,7 +32,8 @@ import java.util.List;
  */
 public class GVRSwitch extends GVRBehavior
 {
-    static private long TYPE_SWITCH = (System.currentTimeMillis() & 0xfffffff);
+    static private long TYPE_SWITCH = ((long)GVRSwitch.class.hashCode() << 32) & (System
+            .currentTimeMillis() & 0xffffffff);
     protected int mSwitchIndex = 0;
 
     public GVRSwitch(GVRContext gvrContext)


parthmehta209:
@NolaDonato do you think this is a good way to assign mType for all GVRBehavior and its subclasses ?

liaxim:
There is native code that comes up with a component type in a different, independent, way. This code is also not thread-safe. We have to think about it first.

NolaDonato:
We should put login in GVRBehavior to generate a unique ID and then call this from the individual getComponentType() functions. I did not realize milliseconds would not be unique.

If you do it in GVRBehavior it won’t interfere with the native components.

parthmehta209:
I changed it to use a simple counter to assign a new type.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRSwitch.java
```diff
@@ -32,13 +31,13 @@
  */
 public class GVRSwitch extends GVRBehavior
 {
-    static private long TYPE_SWITCH = (System.currentTimeMillis() & 0xfffffff);
+    static private long TYPE_SWITCH = 0;
```
liaxim: If we need a quick fix I guess you could shift GVRSwitch.class.hashCode() to the left and & it with currentTimeMillis.

liaxim:
The current code is not thread-safe. Looks like GVRBehavior should be abstract.

parthmehta209:
changed patch to use hashcode and time stamp.

liaxim:
Looks good to me. Will merge tomorrow lunch-time.

PR #651, opened 2016-06-29
GVRInputManager: fix touchpad detection in Note4

parthmehta209:
The touch pad was not recognized if it has InputDevice.SOURCE_MOUSE and was being
wrongly identified as a mouse.


liaxim:
@rahul27 ?

rahul27:
:+1: Good to go.

PR #652, opened 2016-06-30
Fixed problem in bounding box computation 

NolaDonato:
Was not checking for null radius and infinite corners


liaxim:
Will merge after lunch

danke-sra:
Wouldn't it be nice to squash the commits? Here is some info: https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/

apjagdale:
Tested with Model viewer 2.
Works fine.

Issue #650, opened 2016-06-29
CameraRig.reset() et al not working as intended?


imuli:
I want to be able to manually perform a reorientation, claim that the direction I'm currently looking in is the normal forward. It looks like the code to do this is something like `mGVRContext.getMainScene().getMainCameraRig().reset()` (or resetYaw for just the yaw). Calling this doesn't appear to have any effect.

Test app: GearVRf-Demo/gvr-modelviewer with the following addition to actually call onBackPressed upon pressing back. (Similar lack of behavior exhibited by the other two test apps that call resetYaw.)

    @Override
    public boolean onKeyUp(int keyCode, KeyEvent event) {
        if(event.getKeyCode() == KeyEvent.KEYCODE_BACK){
            onBackPressed();
            return true;
        }
        return super.onKeyUp(keyCode, event);
    }

(changes also at https://github.com/imuli/GearVRf-Demos/tree/reset-bug-example)

liaxim:
Thank you for the catch. Fix is in https://github.com/Samsung/GearVRf/pull/655.

PR #655, opened 2016-06-30
Fix camera rig's reset while docked

liaxim:

### DIFF
--- a/GVRf/Framework/framework/src/main/jni/objects/components/camera_rig.cpp
+++ b/GVRf/Framework/framework/src/main/jni/objects/components/camera_rig.cpp
@@ -139,19 +139,6 @@ void CameraRig::predict(float time) {
 }
 
 void CameraRig::predict(float time, const RotationSensorData& rotationSensorData) {
-    long long clock_time = getCurrentTime();
-    float time_diff = (clock_time - rotationSensorData.time_stamp())
-            / 1000000000.0f;
-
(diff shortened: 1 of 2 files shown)


imuli:
Certainly fixes #650, thank you!

yjlintw:
The code would work without changing things in camera_rig.cpp

liaxim:
Without this fix it would only work if you use developer mode and do not dock into a headset.

yjlintw:
Weird, because it works fine with my app in non-developer mode and the phone is on the dock. Thanks for the timely fix anyway, it saves my life.

liaxim:
@yjlintw Weird. It definitely didn't work for me and I can't see how it could have worked unless you are using the phone's internal sensors all the time. Please try the fix and let us know if it causes problems for you.

liaxim:
I have confirmation this fixes the issue. Will merge tomorrow morning.

yjlintw:
@liaxim Sorry that I was out of town the past week and didn't have the chance to test the code until now. Can you tell me that how can I tell if I am using the phone's internal sensors or not. I'm using the innovator version, does that matter?

liaxim:
@yjlintw Hi. There is no direct way but the experience would be terrible if you were using the phone sensors while wearing the headset. There is a "handleOnDock" printout in logcat but this is pretty much it.

I said that as the camera rig's reset method could only possibly work previously if you were using the phone sensors. There is a different path in the code for that case.

PR #624, opened 2016-06-21
Asset loader: expose events, fix opacity export

NolaDonato:
Modified asset loader to check filename prefix to determine volume type:
"sd:" indicates SD card, "http:" and "https:" indicate URL, otherwise
assets directory.

Expose asset event handler for users to handle asset loading events

Make Jassimp export opacity correctly (addresses issue in PR #617)

PR #658, opened 2016-06-30
Asset loader: expose events, fix opacity export

NolaDonato:
Modified asset loader to check filename prefix to determine volume type:
"sd:" indicates SD card, "http:" and "https:" indicate URL, otherwise
assets directory.

Expose asset event handler for users to handle asset loading events

Make Jassimp export opacity correctly (addresses issue in PR #617)
cleaned up version of PR #624

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
@@ -17,6 +17,7 @@ package org.gearvrf;
 
 import java.io.BufferedInputStream;
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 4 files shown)


GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
```diff
@@ -121,86 +237,144 @@ public void onTextureLoaded(GVRContext context, GVRTexture texture, String texFi
...
-        public void onAssetLoaded(GVRContext context, GVRSceneObject model, String filePath, String errors)
+        public void onAssetLoaded(GVRContext context, GVRSceneObject model, String modelFile, String errors)
         {
             mContext.getEventReceiver().removeListener(this);
         }
         
         private void generateLoadEvent()
         {
-            if ((mScene != null) && (mModel != null))
+            if ((mScene != null) && (mModel != null) && (mErrors == ""))
```
liaxim: Please do not compare references to strings; ``"".equals(mErrors)`` would work instead; or if you know mErrors can't be null - ``mErrors.isEmpty()``.

liaxim:
Looks good. Will merge tomorrow morning. Minor things we can take care of later.

PR #663, opened 2016-07-01
Don't add an object to the scene twice (replaces PR #629)

NolaDonato:
Cleaned up version of #629 (squashed commits and rebased)

GVRSceneObject.addChildObject will throw an exception if you try to add
a scene object that already has a parent.
Also made a common root object in GVRScene instead of the mChildren
array. Added GVRScene.getRoot() to get the hierarchy root. Perhaps we
can use the root constructively to attach the camera rig, etc.
Made a common root object in Scene (C++ layer)


liaxim:
The changes to collider.h do not seem to belong and are unused. If you plan to soon use these we can merge as-is I guess.

NolaDonato:
No the changes to collider are a result of merge stupidity. I will remove them.

NolaDonato:
I have removed the changed to collider.h

liaxim:
You probably need to push the change to github since I do not see any new commits.

NolaDonato:
Yeah, you’re right ☺ I pushed the commits now.

Issue #665, opened 2016-07-04
App crashes when changing GVRBitmapTexture/GVRSceneObject multiple times


QKnows:
I try to update my main scene by adding a new GVRBitmapTexture as an GVRSceneObject for the left and right screen, the GVRBitmapTexture uses a Android.Bitmap. 

`final GVRBitmapTexture textureLeft = bitmapLeft == null ? null : new GVRBitmapTexture(this.gvrContext,
                    bitmapLeft);
            final GVRBitmapTexture textureRight = bitmapRight == null ? null : new GVRBitmapTexture(this.gvrContext,
                    bitmapRight);`

I am adding the texture to the object as followd:

` final GVRSceneObject leftScreen = new GVRSceneObject(this.gvrContext, this.mesh, textureLeft);
            leftScreen.getRenderData().setRenderMask(GVRRenderMaskBit.Left);

            final GVRSceneObject rightScreen = new GVRSceneObject(this.gvrContext, this.mesh, textureRight);
            rightScreen.getRenderData().setRenderMask(GVRRenderMaskBit.Right);`

I also removing the old scene objects before adding new objects.

`
this.mainScene.removeSceneObject(this.leftScreenObject);
this.mainScene.removeSceneObject(this.rightScreenObject);

this.leftScreenObject = leftScreen;
this.rightScreenObject = rightScreen;`

This works for nearly 30 images, and then the app crashes. The RAM usage seems to be ok, but the GC must work a lot:

`I/art: Background partial concurrent mark sweep GC freed 139(4KB) AllocSpace objects, 2(43MB) LOS objects, 14% free, 92MB/108MB, paused 2.068ms total 397.845ms`

I can not find an exception or error in the log, just an information shortly before my app crashes:

` I/ActivityManager: Process com.oculus.horizon (pid 15848)(adj 0) has died(87,92)`

And then I get the following log:

` D/ActivityManager: isAutoRunBlockedApp:: com.oculus.horizon, Auto Run ON
07-04 11:15:31.841 3474-4605/? D/GraphicsStats: Buffer count: 2
07-04 11:15:31.851 3474-4485/? I/WindowState: WIN DEATH: Window{461bef2 u0 d0 com.myapp/com.myapp.TestActivity}
07-04 11:15:31.861 3474-4474/? E/Sensors: Mag old sensor_state 4, new sensor_state : 0 en : 0
[...]

PR #672, opened 2016-07-06
Do not keep strong references in the EventManager's cache

liaxim:
- was causing leaks; see https://github.com/Samsung/GearVRf/issues/665


liaxim:
@danke-sra Please review

danke-sra:
Looks good to me. FYI, there are other caches in GVRScriptManager that can be changed to weak references too.

liaxim:
Ok, will check them out. Thanks.

PR #673, opened 2016-07-07
Null check for owner to fix SIGSEGV in immersivepedia 

deepakrawat22:
deepak.rawat@samsung.com


danke-sra:
Looks good.

PR #677, opened 2016-07-07
fix exception in GVRScene.clear, check max light count 

NolaDonato:
The exception for adding an object twice was triggered by GVRScene.clear
(camera rig owner was parented twice).
The scene now checks for maximum light number and wont generate
duplicate light  names.


GVRf/Framework/framework/src/main/java/org/gearvrf/GVRScene.java
```diff
@@ -445,13 +448,22 @@ public void bindShaders(GVRSceneObject root) {
...
-            return true;
+        if (light == null)
+        {
+        	return false;
+        }
+        Integer lightIndex = mLightList.size();
+            
+        if (lightIndex >= MAX_LIGHTS)
+        {
+        	return false;
```
liaxim: It looks like it should be throwing an exception in this case.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRScene.java
```diff
@@ -445,13 +448,22 @@ public void bindShaders(GVRSceneObject root) {
      * @see GVRScene.getLightList
      */
     protected boolean addLight(GVRLightBase light) {
```
liaxim: Btw, since people could derive from GVRScene addLight should be either final or package private. Can't see anything good coming out of end users messing with this method.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRScene.java
```diff
@@ -445,13 +448,22 @@ public void bindShaders(GVRSceneObject root) {
...
-        if (light != null) {
-            Integer lightIndex = mLightList.size();
-            String name = "light" + lightIndex.toString();
-            mLightList.add(light);
-            NativeLight.setLightID(light.getNative(), name);
-            NativeScene.addLight(getNative(), light.getNative());
-            return true;
+        if (light == null)
+        {
+        	return false;
```
liaxim: In case of null I'd say either let it throw the NPE later or throw an IllegalArgumentException from here. Checking for null seems better left to the caller.

liaxim:
Will merge after lunch

NolaDonato:
Don’t merge yet – I haven’t fixed the issues you brought up.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRScene.java
NolaDonato: This would be an awkward place to have an exception because addSceneObject may indirectly call this function. I think I should emit an error event when that PR makes it in so the application can trap this error and modify their light usage.

NolaDonato:
OK you can merge now. I responded to comments.

PR #682, opened 2016-07-07
Fix wrong usage of == for strings

NolaDonato:
This bug causes the asset loader to miss textures


GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
```diff
@@ -284,7 +284,7 @@ public void onAssetLoaded(GVRContext context, GVRSceneObject model, String model
         
         private void generateLoadEvent()
         {
-            if ((mScene != null) && (mModel != null) && (mErrors == ""))
+            if ((mScene != null) && (mModel != null) && (mErrors.equals("")))
```
danke-sra: I think Java's recommended way is isEmpty()

GVRf/Framework/framework/src/main/java/org/gearvrf/jassimp2/GVRJassimpSceneObject.java
```diff
@@ -206,7 +206,7 @@ private void loadTextures(GVRAssetLoader.AssetRequest assetRequest, AiMaterial m
             if(texType != AiTextureType.UNKNOWN)
             {
                 final String texFileName = material.getTextureFile(texType, 0);
-                if ((texFileName != null) && (texFileName != ""))
+                if ((texFileName != null) && !texFileName.equals(""))
```
danke-sra: same as line 287 above

liaxim: The two checks could be reduced to ``!"".equals(texFileName)``

liaxim:
Will merge after lunch

NolaDonato:
Not yet – I haven’t fixed the issues…

NolaDonato:
OK to merge now I have used the clever "".equals way of comparing

PR #685, opened 2016-07-08
Cleanup of generateVAO() to remove special casing of TextureShader. Reuse VAO and VBOids and make generateVAO() private.

Vittalbabu:
Discussed with @roshanch the following cleanup of generateVAO() usage. Tested shadow, jassimp, simplescene, litshader, multilight samples.

Remove the special case for TextureShader to call generateVAO(). Now generateVAO() will be called just before the glBindVertexArray(mesh->getVAOId(programId)) call and after the shader->render() call.

TextureShader's render method sets the proper program instance for the light and no_light case and the programId will be valid after the call to render().

### DIFF
--- a/GVRf/Framework/framework/src/main/jni/engine/renderer/renderer.cpp
+++ b/GVRf/Framework/framework/src/main/jni/engine/renderer/renderer.cpp
@@ -695,10 +695,10 @@ void Renderer::renderMaterialShader(RenderState& rstate, RenderData* render_data
     Mesh* mesh = render_data->mesh();
 
     GLuint programId = -1;
+    ShaderBase* shader = NULL;
 
     try {
          //TODO: Improve this logic to avoid a big "switch case"
-        ShaderBase* shader = NULL;
         if (rstate.material_override != nullptr)
             curr_material = rstate.material_override;
          switch (curr_material->shader_type()) {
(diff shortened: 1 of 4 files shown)


GVRf/Framework/framework/src/main/jni/engine/renderer/renderer.cpp
```diff
@@ -772,6 +767,11 @@ void Renderer::renderMaterialShader(RenderState& rstate, RenderData* render_data
                 error.c_str());
         shader_manager->getErrorShader()->render(&rstate, render_data, curr_material);
     }
+    programId = shader->getProgramId();
+    if (programId != -1)
```
roshanch: what would be the case when programId is -1 ?

Vittalbabu: It is a sanity check. This case should not occur. I can leave this as is or trigger an error in generateVAO() for invalid programId.

Vittalbabu:
1. Reuse VAO and VBO ids when attributes change in mesh.cpp
2. Mesh::getVAOId() will now invoke generateVAO() as needed -- there is no need for the Renderer class to invoke this method. generateVAO() is now private.
3. Added method to delete VAO and VBO ids for a particular programId. This is not used for now.

roshanch:
i think title is bit misleading. its more about deleting existing VBO and VAO ids when we the mesh changes. Code looks good to me. Needs to test it.

liaxim:
Needs DCO

Vittalbabu:
GearVRf-DCO-1.0-Signed-off-by: Vittalbabu Srinivasan
vittalbab.s@samsung.com

Vittalbabu:
Tested with Shadows, jassimp, litshader and a modified version of polyline where a click changes the vertex data of the polyline.

liaxim:
Will merge in the afternoon

PR #690, opened 2016-07-11
Add a show toast api; was lying around in another branch anyway so

liaxim:
this is a little bit of api reconciliation too.

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRContext.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRContext.java
@@ -29,6 +29,8 @@ import org.gearvrf.GVRAndroidResource.TextureCallback;
 import org.gearvrf.GVRHybridObject.NativeCleanupHandler;
 import org.gearvrf.animation.GVRAnimation;
 import org.gearvrf.animation.GVRAnimationEngine;
+import org.gearvrf.animation.GVRMaterialAnimation;
+import org.gearvrf.animation.GVROnFinish;
 import org.gearvrf.asynchronous.GVRAsynchronousResourceLoader;
 import org.gearvrf.asynchronous.GVRCompressedTexture;
 import org.gearvrf.asynchronous.GVRCompressedTextureLoader;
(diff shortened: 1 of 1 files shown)


GVRf/Framework/framework/src/main/java/org/gearvrf/GVRContext.java
```diff
@@ -2769,4 +2774,73 @@ public void finalize() throws Throwable {
...
+    public void showToast(final String message, float duration) {
+        final float quadWidth = 1.2f;
+        final GVRTextViewSceneObject toastSceneObject = new GVRTextViewSceneObject(this, quadWidth, quadWidth / 5,
+                message);
+
+        toastSceneObject.setTextSize(6);
+        toastSceneObject.setTextColor(Color.WHITE);
+        toastSceneObject.setGravity(Gravity.CENTER_HORIZONTAL|Gravity.TOP);
+        toastSceneObject.setBackgroundColor(Color.DKGRAY);
+        toastSceneObject.setRefreshFrequency(GVRTextViewSceneObject.IntervalFrequency.LOW);
```
liaxim: Instead of LOW this should have been NONE but NONE is currently broken.

danke-sra: Is it possible to use REALTIME here?

liaxim: It is possible but it is just an extra overhead. Since the toast text never changes drawing to the canvas once should be enough.

danke-sra: REALTIME is optimized. It does not refresh unless the content has been changed.

liaxim: Didn't notice that. Will use it and maybe remove NONE altogether.

liaxim:
Using REALTIME instead of LOW, per Danke.

thomasflynn:
+1, looks good to me.

PR #691, opened 2016-07-11
New gradle tasks for the GitHub releases

liaxim:

### DIFF
--- a/GVRf/Framework/framework/build.gradle
+++ b/GVRf/Framework/framework/build.gradle
@@ -15,17 +15,43 @@ android {
         }
     }
 
-    buildTypes {
-        release {
-            minifyEnabled false
-            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-project.txt'
+    //copy tasks duplicated since I had trouble getting the same task executed more
+    //than once; and the copying must happen for both release and debug otherwise
+    //the releaseToGitHub may delete the files; should be possible to be done much
+    //better but I left it as a future exercise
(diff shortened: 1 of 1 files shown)


liaxim:
Added a releaseToGitHub build target that should produce the aar without files we can't re-distribute.

Added a eclipseReleaseForGitHub task that produces a zipped Android Library project for Eclipse users.

liaxim:
Am not liking the scripts but they are functional. Prettifying can wait.

liaxim:
Btw, this change will make it so that the Oculus dependencies will be automatically copied *if* the gradle.properties is modified according to the wiki. In any case we should stop relying on the PATH and the rest of the environment variables. This has been proven finicky for most people to get right.

Planning to merge this lunch-time. It is a stepping stone towards the nightly build release. I am pretty sure it will break the automated build but I need the change in to find out what else needs work.

jason2kim:
If the team decides to go with no prebuilt binary for Eclipse users, I think it is time to drop the gvr video staticlib, too.

liaxim:
Whatever the team decides we currently cannot have a single file prebuilt binary for Eclipse. It is just not possible.

Whether we want to make code changes to somehow achieve that and whether we should is a different topic altogether.

I personally think we shouldn't waste too much time with Eclipse.

jason2kim:
Talked with team members on this. People agreed to go without libs-gvrf.zip for Eclipse. And thus, gvr-video-staticlib.

liaxim:
Thank you. Will remove the staticlib sample.

PR #698, opened 2016-07-13
Add a convenience getGVRContext method to GVRScript and GVRMain

liaxim:


liaxim:
@rahul27 What do you think about this? We can eliminate redundant code from pretty much all of the apps.

rahul27:
What is the redundant code that this helps with?

liaxim:
Everybody creates a member variable in their GVRScript/GVRMain to store the context in, for example.

rahul27:
Ahh, I see. Looks good :+1:

PR #668, opened 2016-07-06
Added GVRScriptBehavior scripting component

NolaDonato:
Allows a script to be attached as a component. Sets up event listeners
for picking and draw frame events and directs them to the script.
Directs all events on the scene object to the script.

Modified script manager to attach GVRScriptBehavior component when a
script bundle is loaded. Added onAttach, onDetach script events.
GearVRf-DCO-1.0-Signed-off-by: Nola Donato nola.donato@samsung.com

Tested with gvr-javascript

PR #699, opened 2016-07-13
Added GVRScriptBehavior scripting component - replaces PR #668

NolaDonato:
Allows a script to be attached as a component. Sets up event listeners
 for picking and draw frame events and directs them to the script.
 Directs all events on the scene object to the script.

Modified script manager to attach GVRScriptBehavior component when a
 script bundle is loaded. Added onAttach, onDetach script events.

Tested with gvr-javascript gvr-remotescripting


liaxim:
Will verify tomorrow morning and merge.

liaxim:
I am not sure about Strings as error messages. We will have to localize them for example. You can't decide whether it might be safe to ignore an error or is it fatal. We might have to look at this area again in the future.

liaxim:
We probably shouldn't be invoking onStep and the likes on the GL thread but this is again something we can examine closer in the future.

PR #701, opened 2016-07-14
Fixed problem with addLight

NolaDonato:
The return code was not being handled properly
Also upgraded framework gradle version


GVRf/Framework/framework/src/main/java/org/gearvrf/GVRScene.java
```diff
@@ -134,6 +134,7 @@ public void removeAllSceneObjects() {
         head.getParent().removeChildObject(head);
```
liaxim: Should the removeAllSceneObjects method lock on mLightList too?

liaxim:
Needs rebase

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRScene.java
NolaDonato: Yes I think it should. If bindShaders gets called from two different threads it could create a problem. I think the asset loader can do this.

liaxim:
Btw, this needs rebase

NolaDonato:
I synchronized the light list in clear and rebased

liaxim:
Still says that needs rebsae.

NolaDonato:
Replaced by PR #718 (squashed and rebased)

Issue #692, opened 2016-07-13
A scene object cannot be added to a scene after clear()


danke-sra:
UnsupportedOperationException is raised. The error message is "GVRSceneObject cannot have multiple parents".

NolaDonato:
Yes, this is because a scene object cannot have two parents. We were not checking for this before and it can cause a log of problems internally. For example, if the object has two parents, what is it’s model matrix? Which parent do you use to calculate it?

Now it will throw an exception if you try to add a scene object to another hierarchy if it is already in one. The simple fix is to remove it from the existing hierarchy and then add it to the new one:

sceneObject.getParent().removeChildObject(sceneObject)

will remove it from the hierarchy.



-Nola

danke-sra:
Thanks. So it's should be easy to fix. I will try setting the parents to null when the scene is cleared.

liaxim:
@danke-sra Can we close then?

danke-sra:
GVRScene.clear() needs to be fixed to set the parents of the removed scene objects to null. Should be easy. I will submit a change tomorrow.

NolaDonato:
This is fixed in the scriptbehavior pull request #668. Perhaps you can test it.

danke-sra:
FYI, I tested it but this is still not fixed by https://github.com/Samsung/GearVRf/pull/699.

PR #705, opened 2016-07-14
Clear parent when removing all scene objects from scene

danke-sra:
Fixes  https://github.com/Samsung/GearVRf/issues/692


liaxim:
Looks good to me. Will merge lunch time.

PR #708, opened 2016-07-15
Added visitor for scene objects and descendants

NolaDonato:
This pattern allows you to iterate thru the hierarchy without making a
copy of it


liaxim:
Looks good to me but the public APIs need javadoc.

liaxim:
We can add the javadoc later.

NolaDonato:
javadocs added in PR #722

Issue #695, opened 2016-07-13
Bad quality with texture on GVRSphereSceneObject 


dlockh3ll:
I'm working at a simple 360 photo viewer with some POIS to click but I'm seeing bad quality 360 photos even with very big and high quality pictures (10000 x 5000)

I used the code from the "gvr-360photo" sample and added a bit of scale but it looks very pixellated.

```
Future<GVRTexture> texture = gvrContext.loadFutureTexture(new GVRAndroidResource(gvrContext, R.drawable.provahires));
        sphereObject = new GVRSphereSceneObject(gvrContext, false, texture);
        sphereObject.getTransform().setScale(50, 50, 50);
        sphereObject.getRenderData().setRenderingOrder(1);
        scene.addSceneObject(sphereObject);

```
Is there something i'm missing to show the picture in good quality?

thomasflynn:
The images are automatically getting scaled down.  I encountered this myself two weeks back or so.  I have some tweaks in my tree that i haven't yet converted into a pull request.  I'll do that today.  Even given those tweaks though, i still ran into a memory limit on android.  Even when you request a large heap, you still are only given 512 megs to use (a normal heap is 256 megs).  Your image will decompress to width x height x 4 bytes.  If that's larger than what you have to work with, the image will need to be scaled down to fit into memory.  Thankfully, the image i was working with (originally 16k x 8k, scaled down to 8k x 4k) still looked very good at the scaled down resolution.

dlockh3ll:
Thank you I will check your fork!

thomasflynn:
I haven't committed the changes to any of my branches yet.  Again, i'll do that in a few hours.  It's actually sitting as local changes on my master branch on my machine at work (i'm currently st home in bed on my phone :).  If you're impatient, i think i can remember two of the key changes.  Let me browse the repo real quick...

dlockh3ll:
Don't worry, take your time, just please let me know when you have commited the changes so i could download it and try it. Thanks in advance.

thomasflynn:
Ok, in GVRAsyncBitmapTexture.javs in the asynchronous/ directory:
1) There's a bug with our detection and handling of MAX_TEXTURE_SIZE.  When i was debugging it, we actually detect it correctly, but it somehow gets overwritten by the default we set (1024).  I changed the default to the max texture size on s7 (16384) since that's what i was debugging on.  This needs to be fixed properly.

[...]

PR #709, opened 2016-07-15
fixes for supporting larger images

thomasflynn:
fixes for supporting larger images.

1) use getLargeMemoryClass() if largeHeap is set in the application's
AndroidManifest.xml.

2) ensure pending gl runnables are run before onInit.  this is necessary
for ensuring glMaxTexureSize gets set correctly.

3) increase the MAXIMUM_IMAGE_FACTOR

this will address #695 

tom.flynn@samsung.com


liaxim:
Looks good to me. Will merge after lunch.

PR #712, opened 2016-07-15
Gradle tweaks:

liaxim:
- print whether the OVR_ gradle property has been defined
- enable creation of a zipped Android Library project with prebuilt binaries


liaxim:
Will merge since this is the fastest way to get it to the build machine and test there.

PR #714, opened 2016-07-15
Support to copy AAR files

deepakrawat22:
For now only added the support in Framework so that whenever build framework by opening the Framework project and do build it will copy the generated AAR file to the path specified in LIBS_DIRECTORY

I want to do similar change in Extensions projects.

deepak.rawat@samsung.com


liaxim:
Looks good to me. Will merge lunch time.

PR #717, opened 2016-07-18
Final reconcilation between the trees

liaxim:
- The common java files are identical now
- A lot of the native files are either reconciled or in the corresponding backend


liaxim:
Can be reviewed and considered for merging. Next:

- split the project into a framework and backend subprojects; will require some refactoring since the framework has dependencies to the backend currently
- we can merge google daydream before or after the split
- dynamically discovering, loading and switching backends will come next
- google vr will become its own backend
- monoscopic should become a separate backend too

liaxim:
Will merge by the end of Friday, if there are no objections.

PR #729, opened 2016-07-20
Move the VRHackathons to the Demos repo

liaxim:


liaxim:
@thomasflynn this pull request removes the VRHackathons from the framework repo. Is this what you wanted?

thomasflynn:
Yes, thank you.  It'll eliminate some confusion.  We've had people think those were our only samples.

PR #733, opened 2016-07-20
Copy libs by default

parthmehta209:
1. This change makes it easier to build the extensions by defining the LIBS_DIRECTORY by default
2. Changes permission on gradlew files to make commandline build easier.


rahul27:
Should we just add prebuilt libraries for the extensions to the libs folder? It would be one less step to build the samples.

parthmehta209:
Not sure if we want to checkin binaries.

parthmehta209:
Not sure if we want to checkin binaries.

rahul27:
Having to build and copy these files definitely does not make it easy for either the developer or the end user to build applications.

parthmehta209:
Eventually we want all the binaries framework and extension aar to be
hosted on github, So as a user if you are only looking for samples you dont
need to build anything other than the sample itself.

liaxim:
Yep. All the prebuilt binaries should be available either from an official release or the nightly build release. This is still work in progress; it will get better and easier with time.

liaxim:
@parthmehta209 If the LIBS_DIRECTORY is defined but the directory doesn't really exist the build will fail. We need to make changes to the build.gradle files to just skip copying if the directory doesn't exist. What do you think?

parthmehta209:
Oh yes. Thats probably better than failing. Will make that change.

parthmehta209:
@liaxim changed the build.gradle file to not copy aar files if the directory does not exist.

PR #736, opened 2016-07-21
Make backends "switchable"

liaxim:
Depends on #717 

- quotes because it is rather simplistic but doable without extensive code changes; those will come later as the time permits; still using one module for the project

- GVRF_BACKEND in gradle.properties has to specify the backend of choice; so we could have multiple backends side-by-side and switch between them statically by editing the gradle.properties file


GVRf/Framework/framework/build.gradle
```diff
@@ -73,7 +76,7 @@ android {
     }
 
     sourceSets.main {
-        java.srcDirs = ['src/main/java', 'src/main/backends/oculus']
+        java.srcDirs = ['src/main/java', '../' + rootProject.property("GVRF_BACKEND")+'/src/main/java']
```
rahul27: Does this enforce backend 2 to keep the same file names to support all apps that work with backend 1?

rahul27: Also if Google VR is backend 2, we would need a way to compile both backend 1 an 2 and load backend 2 if backend 1 is not available.

rahul27: Finally does this work with Eclipse?

liaxim: - Yes, backend 2 needs the same folder structure
- Dynamic loading and switching will be in the future; for now we will probably bundle Google VR into the Oculus backend; like monoscopic. Eventually both monoscopic and Google VR will be separate.
- No

liaxim:
Have to come up with something that can handle the Android.mk variation in the different backends.

rahul27:
This is good to go. I've tested it and it works fine.

GVRf/Framework/gradle.properties
```diff
@@ -15,6 +15,7 @@
 #Thu May 05 12:55:40 PDT 2016
 android.useDeprecatedNdk=true
 org.gradle.jvmargs=-Xmx2048M
+GVRF_BACKEND=backend_oculus
```
rahul27: You might want to add this line to the Extension projects as well.

liaxim:
Will handle the Android.mk by just copying the backend's one for now instead of something more sophisticated. The sooner this is ready the sooner we can:

- have the development environment sorted out and redundant classes removed from the other repo
- have GoogleVR support merged in

GVRf/Framework/gradle.properties
liaxim: I am not sure why the Extension projects need that. If they use an .aar it shouldn't matter. Or does it?

liaxim:
Had the other backend pretty easily plug in by editing the gradle.properties. Have to make some changes to the other repo though. Then you would clone it and create a link under Framework to backend_whatever, specify backend_whatever in gradle.properties and you can go.

The mechanism is somewhat clunky but works. Will keep improving it until we have true modules for the backends which will be built separately and discovered and loaded dynamically.

@rahul27 This is it for this pull request. Once this is in we can see what is needed for GoogleVR so it gets merged sooner rather than later.

GVRf/Framework/gradle.properties
rahul27: They don't use the aar. Should they be using the aar?

liaxim: I am tempted to say they should but for now I will make the change you requested.

liaxim:
Defaults to backend_oculus so Extensions do not require changes.

rahul27:
:+1: Looks good.

liaxim:
Will merge late afternoon.

liaxim:
Eclipse support is not there out of the box. Workaround to get it to work:

1. Copy backend_oculus/../Android.mk to framework/../jni
2. Define a GVRF_BACKEND variable in Eclipse and set it to the full path to your backend_oculus dir

Not sure if and when this will be made more user-friendly.

liaxim:
One more thing for Eclipse:
- a new source folder; use link source and specify the complete path to .../backend_oculus/src/main/java

rahul27:
@liaxim I get the following error when I try to run a clean from Android Studio

````
:framework:copyDebugLint UP-TO-DATE
:framework:mergeDebugProguardFiles UP-TO-DATE
:framework:packageDebugRenderscript UP-TO-DATE
:framework:buildNative
make: Entering directory 'GearVRf/GVRf/Framework/framework/src/main'
Android NDK: There is no Android.mk under ./jni    
android-ndk-r10e/build/core/add-application.mk
Error:(214) *** Android NDK: Aborting...    .  Stop.
Android NDK: If this is intentional  please define APP_BUILD_SCRIPT to point   
Android NDK: to a valid NDK build script.    
make: Leaving directory 'GearVRf/GVRf/Framework/framework/src/main'
:framework:buildNative FAILED
Error:Execution failed for task ':framework:buildNative'.
> Process 'command 'ndk-build'' finished with non-zero exit value 2
Information:BUILD FAILED
Information:Total time: 0.597 secs
Information:2 errors
Information:0 warnings
Information:See complete output in console
```

Seems like it is skipping :framework:copyAndroidMk before buildNative. Do you see the same.

./gradlew assembleDebug works fine

PR #745, opened 2016-07-25
Fix the Eclipse framework project

liaxim:
- include the backend's source folder

### DIFF
--- a/GVRf/Framework/framework/src/main/.classpath
+++ b/GVRf/Framework/framework/src/main/.classpath
@@ -1,10 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
 	<classpathentry kind="src" path="libs"/>
+	<classpathentry kind="src" path="backends/oculus"/>
 	<classpathentry kind="con" path="com.android.ide.eclipse.adt.ANDROID_FRAMEWORK"/>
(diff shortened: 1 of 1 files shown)


GVRf/Framework/framework/src/main/.classpath
```diff
@@ -1,10 +1,12 @@
...
-	<classpathentry exported="true" kind="con" path="com.android.ide.eclipse.adt.LIBRARIES"/>
-	<classpathentry exported="true" kind="con" path="com.android.ide.eclipse.adt.DEPENDENCIES"/>
-	<classpathentry kind="src" path="java"/>
-	<classpathentry kind="src" path="gen"/>
-	<classpathentry kind="output" path="bin/classes"/>
-</classpath>
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="src" path="libs"/>
+	<classpathentry kind="src" path="backends/oculus"/>
```
rahul27: Wouldn't this path change after https://github.com/Samsung/GearVRf/pull/736?

liaxim: Yes, but in the meantime Eclipse would work. I am having issues with Android.mk for #736 so finishing up #736 could take some time.

rahul27: Ah, this is good to go then.

PR #748, opened 2016-07-25
tga file format support

deepakrawat22:
Can read tga file format for texture.


deepakrawat22:
To review the change need a model with tga file as texture and use loadModel. Attached is one.

[bulbasaur-pokemon-go-obj.zip](https://github.com/Samsung/GearVRf/files/382737/bulbasaur-pokemon-go-obj.zip)

GVRf/Framework/framework/src/main/java/org/gearvrf/asynchronous/AsyncBitmapTexture.java
```diff
@@ -343,13 +344,155 @@ protected AsyncLoadTextureResource(GVRContext gvrContext,
...
 
+    static Bitmap decodeStreamTGA(InputStream stream) {
+        Bitmap bitmap = null;
+        try {
+            byte[] headerTGA = new byte[18];
+
+            stream.read(headerTGA);
+            int type = headerTGA[2] & 0xFF;
+            // uncompressed RGB tga
+            if (type == 2) {
```
liaxim: How about writing a warning to logcat in case of an unsupported type?

liaxim:
Looks good to me. Will merge in the afternoon.

PR #751, opened 2016-07-26
Allocate proper number of texture coordinates for cylinder

Vittalbabu:
Allocate proper number of texture coordinates for cylinder


liaxim:
Looks good to me. Will merge in the afternoon.

thomasflynn:
Good catch.  +1

thomasflynn:
Perhaps ahould double-check sphere and cube as well.

Vittalbabu:
@thomasflynn The other two cases are fine.

PR #752, opened 2016-07-27
 Fix for the external renderer

liaxim:
the custom shader could leave another texture unit active leading
to confusion in the external renderer


liaxim:
Rebased; merged a conflict in renderer.cpp.

liaxim:
If there are no objections will merge this Monday afternoon

PR #754, opened 2016-07-28
Fix for "gradlew clean assemble"

liaxim:
can't get the same task to run more than once so as a quick fix
just duplicated the copyAndroidMk task


liaxim:
@rahul27 this fixes your problem in AS

rahul27:
:+1: Works

PR #756, opened 2016-07-29
Fix extensions builds that also build the framework

liaxim:


liaxim:
Parth, this should fix your build problem. Please make sure you are starting from a clean state. Verified it locally in AS.

rahul27:
:+1: This works

PR #759, opened 2016-07-30
adding support for egl_image for different vendors in oes shader

roshanch:
This is to support different vendors for GL_OES_EGL_image_external. 
If multiview is disabled, it will use GLSL 100. 
Simple gallery still has problem with multiview, we need to run it without multiview. However, other sample works with multiview + OES shader. 
Simple gallery is giving different errors on different vendors with multiview.


liaxim:
Samples work. Will merge in the afternoon.

Shouldn't you enable multiview support for all the samples you know that work with it?

roshanch:
do you mean by default in xml file ?

liaxim:
yes

roshanch:
Okay, I will add that.

PR #760, opened 2016-07-30
Fix issue with running immersivepedia 

Vittalbabu:
Fix issue with running immersivepedia -- in the non multiview case, extension GL_OES_EGL_image_external_essl3 is needed.


roshanch:
GL_OES_EGL_image_external_essl3  only works for Adreno, for Mali, you need GL_OES_EGL_image_external.

Vittalbabu:
Good point. Let's figure out a cleaner solution than this.

Vittalbabu:
How about having two shaders -- instead of using #ifdef we pick the appropriate source. We use the old shader that worked for the non-multiview case.

liaxim:
@Vittalbabu Should we close this?

Vittalbabu:
Yes

PR #761, opened 2016-07-31
Fix issue in SceneObject::isColliding()

Vittalbabu:
Handle cases in SceneObject::isColliding() when either source or target mesh is NULL.


roshanch:
code seems fine to me.

liaxim:
Will merge after lunch

Issue #762, opened 2016-07-31
View tracking not a as single say


arianaa30:
Puting it simple, is there any ways to define gaze as a box, say a rectangle that might intersect with **multiple** objects? Basically I want to detect whatever objects are currently within user's view and user can see; not necessarily a single point as the end of a ray. With a single point I can't really achieve that.

NolaDonato:
Are you interested in obtaining the list of currently visible objects, everything the user can see? Or do you want to further constrain it to a smaller gaze box, a subset of what is visible?

arianaa30:
I want to experiment both! I understand we have 96 degree of view in GearVR, so for a box it can be a 70ish degree or sth maybe.

arianaa30:
Any thoughts on this? It is a bit urgent for me to achieve this.

NolaDonato:
I can add something to GVRPicker to pick against the camera's view frustum. It will return the list of currently visible objects. Will that work?

arianaa30:
Sure. It would be best to make it as a configurable box (given a height and width) centered on the gaze.

NolaDonato:
Pull request #765 adds GVRPicker.pickVisible which will return the list of visible objects. Please test this pull request and see if it does what you want. If so we can do further refinement.

arianaa30:
Thanks. Just that I'm not very familiar with the pull request. I went to your repository and downloaded the whole 'pickvisible' branch. Then I replaced the common files in my current Framework with that of the downloaded ones (but didn't remove the existing ones), but when I am trying to compile it, it has errors! 
Aren't there any plans to update the master branch of Gearvrf so I can download the whole thing from scratch or something?!

arianaa30:
OK, I replaced only the changed GVRPicker files and fixed an error due to my API being a bit older and things compiled Ok. 
I was using the `pickVisible()` in a similar way as `pickObjects()`, but it is not working, and in my example, the non-visible objects get selected! The selection also lags a lot, and blinks. Is it fully tested to make sure it works, or maybe my test is problematic?

I do this in `onStep()` :


```
		for (GVRPickedObject pickedObject : GVRPicker.pickVisible(mGVRContext.getMainScene())) {
			for (int i = 0; i < objectsRed.length; i++) {
				if (pickedObject.getHitObject().equals(objectsRed[i])) {
					if (!objects[i].isEnabled()) {
						objectsRed[i].setEnable(false);
						objects[i].setEnable(true);
						break;
					}
[...]

PR #765, opened 2016-08-02
Adding GVRPicker.pickVisible

NolaDonato:
The picker internally has access to the list of visible objects compiled
by the renderer each frame. This function provides access to that list from Java.

This feature is added in response to issue #762 

GearVRf-DCO-1.0-Signed-off-by: Nola Donato nola.donat@samsung.com


rahul27:
I assumed that we were planning to move away from the polling model proposed in this change to an event based model for picking. 

If we were to implement this change to fit the events model then it would make sense to create a new type of picker like a box or frustum picker to the camera rig to listen for IPickEvents.

NolaDonato:
OK I added GVRFrustumPicker which will generate the same pick events as GVRPicker

rahul27:
:+1: Looks good. Tried this with the eye picking sample, changed the GVRPicker to GVRFrustumPicker to test.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRFrustumPicker.java
```diff
@@ -0,0 +1,262 @@
...
+                {
+                    GVRSceneObject sceneObj = hit.hitObject;
+                    GVRSceneObject.BoundingVolume bv = sceneObj.getBoundingVolume();
+                    center.set(bv.center, 1);
+                    center.mul(view_matrix);
+                    dir.z = bv.radius;
+                    dir.mul(view_matrix);
+                    if (!mCuller.isSphereInsideFrustum(center.x, center.y, center.z, dir.length()))
+                    {
+                        Log.d("Picker", "Picker: outside %s (%f, %f, %f) %f", sceneObj.getName(), center.x, center.y, center.z, dir.length());
```
liaxim: We should remove the logging. Can be done later of course.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRFrustumPicker.java
```diff
@@ -0,0 +1,262 @@
...
+        }
+        else
+        {
+            mCuller = new FrustumCuller(projMatrix);
+        }
+    }
+
+    public void onDrawFrame(float frameTime)
+    {
+        if (isEnabled())
```
liaxim: I think we should unregister the drawFrameListener if the component is disabled instead. Not something that needs immediate attention.

liaxim:
Should we document that probably people want to add the component to ``mainScene.getMainCameraRig().getHeadTransform().getOwnerObject()``? Considering how obfuscated that looks, should we "api"-fy the common use-case?

liaxim:
If I modify the eye-picking sample accordingly and use ``setFrustum(90, 1, znear, zfar)`` what should I be seeing? znear and zfar are retrieved from the main camera rig. I kind of expect all boards and bunnies to be red but that is not what happens.

liaxim:
Looks good. Will merge Thursday afternoon if there are no further comments.

NolaDonato:
setFrustum(90, 1, near, far) should show all the bunnies as red. If that is not happening it is a bug. I will look at it.

NolaDonato:
GVRCameraRig.getHeadTransformObject is protected. I will make it public. That is the accessor for the head transform owner object.

PR #766, opened 2016-08-02
Fix use of compressed textures

liaxim:
Per glGenerateMipmap docs:
GL_INVALID_OPERATION is generated if the levelbase array is stored in a compressed internal format.


liaxim:
@roshanch Please review

roshanch:
good to go..

PR #767, opened 2016-08-03
Added function clearLights to remove all lights from scene

apjagdale:
This would remove the obsolete lights from scene.


liaxim:
Looks good to me.

@NolaDonato Any comments?

NolaDonato:
I will test today.

liaxim:
@NolaDonato Is this change ok to go in? Thanks.

NolaDonato:
Yes. I tested this one with gvr-multilight, gvr-jassimp, gvr-testmodels with x3d content and gvr-shadow and they all work.

Issue #741, opened 2016-07-23
X3D parser does not handle texture coordinates correctly


NolaDonato:
Looking at the code it appears the X3D parser is directly using the texture coordinate list supplied by the file. This is not correct because X3D does not guarantee that the normals and texture coordinates will match. It allows for separate index tables for texture, positions and normals.

The parser needs to reconstruct the texture coordinates so there is a texcoord for every vertex. Same with the normals. The arrays must be of the same size so they can use the same index table.

NolaDonato:
pull request #768 fixes this

PR #768, opened 2016-08-03
Fixed X3D vertex order problem, async texture load errors 

NolaDonato:
During async loading we ignore errors from
GVRAndroidResource.openStream. It printed a stack trace  but did not
issue any exception or event. I now let openStream throw an exception
and try to handle it in the loaders. This change may cause IOExceptions
to be thrown where they were not before, I might not have caught all the cases.

Fixes issue #741 


GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAndroidResource.java
```diff
@@ -295,9 +295,15 @@ public GVRAndroidResource(GVRContext context, URL url,
      */
     public synchronized final InputStream getStream() {
         if (streamState != StreamStates.OPEN) {
-            openStream();
+            try {
```
liaxim: There is a lot of code that doesn't expect this method to return null. Please let the exception escape getStream. There will be methods that will need to be updated. It is better knowing the root cause of the error instead of trying to deal with a vague null reference value.

liaxim:
Ideally should have been split in two pull requests. We could have merged the x3d fix by now.

NolaDonato:
OK I have separated out the X3D stuff into pull request #774. I will change getStream to throw the exception and do another pull request for the texture changes.

PR #769, opened 2016-08-03
Fix external renderer state

liaxim:
Couple of batching crashes worked around; nullptr checks may not be the most correct fix
Minor cleanup of oes_shader


liaxim:
@roshanch Please review

liaxim:
The internal app works ok without batching; will keep looking into problems with batching enabled.

roshanch:
looks good.

PR #771, opened 2016-08-04
gradle: prevent the native code from being rebuilt on every build

liaxim:
gradle's file copy doesn't copy the timestamp.. and can't be configured to...


liaxim:
If there are no comments will merge after lunch

roshanch:
Verified..

PR #773, opened 2016-08-04
Restore the splash timeout feature

rahul27:
r.rudradevan@samsung.com


liaxim:
Looks good and works as expected. Will merge after lunch if there are no further comments.

PR #775, opened 2016-08-04
Adding support for disabling render-data with batching

roshanch:
Few minor changes:
1. Adding render pass into batching.
2. If user disables render data in onStep(), batch will be updated and batching will be disabled for that renderdata
3. Segregate batching code from renderer.cpp to batch_manager


NolaDonato:
I tested this patch with gvr-testmodels (using x3d assets), gvr-multilight and gvr-shadows. I verified that it does not generate more or different shaders than before (lights are removed correctly). We can merge this.

NolaDonato:
I tested this with gvr-tutorial-lesson6. The balloons sometimes come back after you pop them. The balloon will disappear and then reappear. This happens with the custom shader but happens more with the texture shader. I modified the sample to use the texture shader and a white texture. I tried to attach the modified sample here but it wouldn't let me. I will email it to you. If you set the variable MainView.mUseTextureShader to true the sample will use the texture shader. Otherwise it uses the custom shader.

liaxim:
@roshanch How was this tested? Which samples have been verified as working at least as well as before?

@NolaDonato With or without this change, the balloons do not disappear for me. I need to go back to the commit just before batching to have the balloon disappear consistently on tap. What I discovered though is that the picker doesn't seem to work as expected - with the latest code a lot of the balloons do not get picked (I added logging to BalloonMain.PickHandler.onPick). With the code prior to batching everything is getting picked as expected. Does this make sense? So it is not necessary the case that setEnable doesn't work.

liaxim:
@roshanch confirmed what I see. Looks like we have a picking problem. Have not yet identified which commit causes it.

roshanch:
yeah but its due to latest picking change in 2 weeks. If you just take my repo, it works, but if I pull latest picker change, its broken.

liaxim:
gvr-eyepicking doesn't work either. And yes, it is not related to this patch.

liaxim:
Btw, you should make sure you test your changes on top of master. Otherwise the test results don't mean much.

NolaDonato:
Pull request #812 fixes it.

roshanch:
it works with #812 and gvr-tutorial6 sample.

liaxim:
I tested with video,vuforia,shadows,multilight,accessibility,immersivepedia but accessibility,video,immersivepedia crash due to unrelated problem. Will wait for this to be fixed first before re-testing and re-reviewing this.

GVRf/Framework/framework/src/main/jni/engine/renderer/batch.cpp
```diff
@@ -87,47 +128,51 @@ bool Batch::add(RenderData *render_data) {
...
+    matrices_.clear();
+    tex_coords_.clear();
+    vertices_.clear();
+    normals_.clear();
+}
+bool Batch::isRenderDataDisabled(){
+     bool update_vbo = false;
+     for(auto it= render_data_set_.begin();it!=render_data_set_.end();){
+        if(!(*it)->enabled() || !(*it)->owner_object()->enabled()){
+            (*it)->set_batching(false);
```
liaxim: Are set_batching(false) and setBatchNull related? Like if we do set_batching(false) should batch_ automatically be set to nullptr?

GVRf/Framework/framework/src/main/jni/engine/renderer/renderer.cpp
```diff
@@ -334,7 +169,12 @@ void Renderer::cull(Scene *scene, Camera *camera,
     // Note: this needs to be scaled to sort on N states
     state_sort();
 
-    BatchSetup();
+    if(do_batching){
+        if(batch_manager == nullptr){
```
liaxim: In a future change you should consider eliminating this lazy initialization. The batch_manager is needed always by default.

liaxim:
Good change. Tested with numerous samples. No (new) problems. Will merge in the afternoon together with the gvr-outline fix.

GVRf/Framework/framework/src/main/jni/engine/renderer/batch.cpp
roshanch: Not related. Actually first we check whether batch is null for a renderdata if it is then we try to add into available or new batch. So in add() we return if we have disabled batching.

PR #776, opened 2016-08-04
Remove onStep call for scripting

roshanch:
Discussed with @NolaDonato  @thomasflynn , we don't need this anymore. Nola has added alternate code for this.


liaxim:
Looks good to me but please remove the code instead of commenting it out

liaxim:
gvr-javascript doesn't work like it used, I think. On tapping the text should start animating but with this change in, it no longer does.

1) should we remove the onStep handler in text.js?
2) is there an alternate way of doing the same?

NolaDonato:
I am sure there is another way of doing it. I will take a look at gvr-javascript and see why it is no longer animating.

NolaDonato:
I have fixed part of the problem. I changed “script.js” in the gvr-javascript sample to attach the script to the “text” scene object programmatically. Now if a sensor event is sent, the script’s onSensorEvent function will be called. However, tapping the screen will not cause it to animate. Tapping is not a sensor event that appears to qualify as a key event or it is not passed through to this function. I am not sure where to look for how tapping generates a sensor event. Perhaps Rahul or Parth can help here. My changes to gvr-javascript are in PR #294.


function onSensorEvent(event) {
    var keyEvent = event.getCursorController().getKeyEvent();
    if (keyEvent == null) {
        return;
    }

    if (keyEvent.getAction() == 0x0 /* ACTION_DOWN */) {
       if (++colorIdx >= colors.length) {
            colorIdx = 0;
       }

       // Change text color
       var textSceneObject = event.getObject();
       textSceneObject.setTextColor(0xFF << 24 | colors[colorIdx]);

       // Local animation
       localAnimation = 60;
    }
}

liaxim:
Will merge after lunch if there are no objections. Will deal with the onSensorEvent separately.

JulianaFigueira:
Hi, 
I'm new here...
What is the alternate code for onStep? I`m trying to use it on GVRMain, but with this change it doesn`t work anymore.

liaxim:
@JulianaFigueira This problem has been fixed in the latest master. You can also register a drawFrameListener (see GVRContext.registerDrawFrameListener).

JulianaFigueira:
Thank you!

PR #777, opened 2016-08-04
gradle: fix the timestamp in the clean task

liaxim:


liaxim:
Merging since gradlew clean would fail otherwise

PR #778, opened 2016-08-05
Put in workarounds to stop some tasks from executing for backends

liaxim:
other than oculus


liaxim:
Merging since the work on one of the automated builds is blocked otherwise

PR #780, opened 2016-08-05
Fix the type of exception thrown by the shader_manager

liaxim:
We catch std::string


NolaDonato:
Looks pretty safe and now the message is more informative.

liaxim:
If there are no other comments will merge after lunch.

PR #784, opened 2016-08-07
Add a bind method to GVRRenderTexture.

thomasflynn:
In a recent hackathon, this was all that needed to be added for that
team to do some fancy render to texture stuff.

tom.flynn@samsung.com


liaxim:
Looks good to me. If there are no comments will merge after lunch.

PR #788, opened 2016-08-08
Bring up the VR volume ui automatically

liaxim:
Based off Rahul's code

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRActivityBase.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRActivityBase.java
@@ -30,6 +30,7 @@ import org.gearvrf.utility.VrAppSettings;
 import android.app.Activity;
 import android.content.pm.ActivityInfo;
 import android.content.res.Configuration;
+import android.media.AudioManager;
 import android.os.Bundle;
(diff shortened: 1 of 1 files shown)


liaxim:
@rahul27 Please let me know what you think. Thanks.

liaxim:
Reacting to action_down instead.

rahul27:
:+1: Looks good.

deepakrawat22:
@liaxim Why use AudioManager.STREAM_MUSIC, what if no music is playing in background ?

liaxim:
@deepakrawat22 Good question; two reasons:
- that is what other people use (see ovr_mobile_sdk/VrAppFramework/java/com/oculus/vrappframework/VrActivity.java:adjustVolume
- if the app was to play music I'd imagine that would be the type of audio stream in use so this is the volume you want to change regardless of the current status

What do you think?

liaxim:
Had a chat with @deepakrawat22 . This is better than what we had before. The standard volume UI shows up in addition if the phone is not docked (if docked everything works as expected). @deepakrawat22 will have a look and maybe fix that or create an issue so we don't lose track of this quirk.

liaxim:
Will merge after lunch if there are no comments

PR #794, opened 2016-08-10
Use as many parallel compile jobs as it makes sense for

liaxim:
the machine


liaxim:
@mwitchwilliams After this is merged I expect you will no longer have to patch your local gradle files.

@Vittalbabu Some of the random errors during native compilation should happen less frequently or disappear

@mwitchwilliams had a very severe case of the random compilation errors during native compilation on our corporate windows laptops. Limiting the number of jobs to 1 (I expect 4 would work too) the errors disappeared but the compilation obviously slowed down.

If I have to guess it has something to do with our security software.

liaxim:
If there are no comments will merge after lunch.

PR #797, opened 2016-08-11
Throttler must inform in case of failure

deepakrawat22:
Also added all the type of texture name supported by Assimp

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/asynchronous/Throttler.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/asynchronous/Throttler.java
@@ -198,7 +198,7 @@ class Throttler implements Scheduler {
                     }
                 } else {
                     // loadResource() returned null
-                    callback.failed(null, resource);
+                    callback.failed(new Throwable("Not able to load resource"), resource);
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


NolaDonato:
Please test this with cube_wrongtex.fbx and let me know if it handles the missing texture properly. I will email you this file.

liaxim:
@deepakrawat22 What were the results of the test? Thanks.

deepakrawat22:
@liaxim There was a crash after I commented return statement. Will investigate more the reason behind the carsh.

deepakrawat22:
@NolaDonato With cube_wrongtex.fbx it gave file not found exception as the png file was not there.

NolaDonato:
Yes, the asset is missing a texture intentionally to test this case. GVRAssetLoader.loadModel should refuse to put this asset in the scene if the file not found exception is thrown. Is that what it does now?

liaxim:
@deepakrawat22 What is the status of this? Thanks.

deepakrawat22:
This pull request was to refactor Throttler code that it must inform in case its not able to load a resource file, earlier it used to throw exception. I didn't made any changes to GVRAssetLoader.loadModel.

Another change was I added all the types of texture supported by assimp, in the textureMap

liaxim:
Ok. So from your point of view there is no more work left for this pull request?

deepakrawat22:
Earlier TextureRequest failed() method was not getting called, now it will be called in case a texture file is not supported.

deepakrawat22:
@liaxim yes no more work on this pull request.

liaxim:

--
Not sure the callbacks copying is necessary. Adding a callback happens only if the thread owns the pendingRequests monitor so just clearing the callbacks member should have been enough. Also not so sure new threads will not be spawned; at least theoretically seems possible to me and inevitable since we remove the pendingRequest from the requests array.

Hopefully we will get to streamline and simplify this code; it is doing most likely more than what we actually need.

PR #799, opened 2016-08-11
Allow shader version to be specified 

NolaDonato:
Added GVRShaderTemplate constructor variant which lets the user pass the
GLSL version to use. Previously all custom shaders were restricted to be
GLSL 300. This change permits old custom shaders to be used as input to
GVRShaderTemplate without modification.


liaxim:
Looks good. Will merge after lunch if there are no comments.

roshanch:
There is bug with driver when it uses "samplerExternalOES" sampler with the video and GLSL 300. It throws some glErrors on some of the drivers. But using GLSL 100 works well with Mali and Adreno.

liaxim:
@roshanch This bug affects the oes_* shaders. This change is for the generated shaders so it shouldn't matter.. What do you think?

roshanch:
Yes this change is not going to affect it. I just mentioned as it is relevant. The intent was to keep this in mind when we develop any video sample.

NolaDonato:
It matters if a user subclasses GVRShaderTemplate to make their own shader. If they use OES extensions it won’t work now because we hardcode version 300 into our shaders.

I ran into this when I tried to make gvr-video use GVRShaderTemplate to make the shader and it didn’t work. The version 300 shader produces incorrect colors due to a driver bug.

PR #803, opened 2016-08-11
Fix cubemap loading

parthmehta209:
1. Do not throw an error if there are extra files in the .zip file for cubemap texture.
2. Throw an error if any one of the six cubefaces do not have a texture.


liaxim:
Looks good. Will merge after lunch if there are no comments.

PR #804, opened 2016-08-11
Fixes for 3DCursorLibrary and org.gearvrf.io

parthmehta209:
org.gearvrf.io:
SensorManager.java
1. Changed sensors map to a ConcurrentHashMap. This is needed when sensors are added/removed to the scene while iterating/handling a sensor event.

GVRCursorController.java
1. Expose getNear/FarDepth() calls for use in enabling and disabling settings cursor in CursorManager.

GVRGazeCursorController.java
1. Reduced the TAP_TIMEOUT to make clicking more responsive and enable double clicking on objects.
2. Added a HandlerThread to process events on the GazeCursorController. Previously events were being processed on the Main thread(when receiving
    touch events) as well as the GlThread(onDrawFrame)

3DCursorLibrary:
CursorManager.java
1. Added enable/disableSettingsCursor calls to enable applications to use
    a settings cursor to interract with a GVRViewSceneObject.
2. In cursorEventListener Traverse SceneGraph to find Selectable/MovableBehavior components instead of just checking the parent of the object with CursorEvent.
3. Added a convinience method getGVRContext() to avoid having to pass GVRContext as well as CursorManager.

CursorSceneObject.java
1. Added a lock arround methods accessing meshSceneObject

MovableBehavior.java
1. Added locks arround selected variable as it can be accessed from multiple threads

ObjectCursor.java
1. Added destroyIoDevice call to set active to false for a cursor which is removed from the scene when active=true

SelectableBehavior.java
1. Added behavior and cursor arguments to StateChangeListener
2. Setting state before calling handleClickEvent and handleClickReleased

### DIFF
--- a/GVRf/Extensions/3DCursor/3DCursorLibrary/src/main/java/org/gearvrf/io/cursor3d/SelectableBehavior.java
+++ b/GVRf/Extensions/3DCursor/3DCursorLibrary/src/main/java/org/gearvrf/io/cursor3d/SelectableBehavior.java
@@ -1,14 +1,21 @@
 package org.gearvrf.io.cursor3d;
 
+import android.os.Handler;
+import android.os.Handler.Callback;
+import android.os.Message;
 import android.util.SparseArray;
 import android.view.KeyEvent;
 
 import org.gearvrf.GVRBehavior;
 import org.gearvrf.GVRComponent;
+import org.gearvrf.GVRMaterial;
+import org.gearvrf.GVRPhongShader;
 import org.gearvrf.GVRSceneObject;
+import org.gearvrf.GVRSceneObject.BoundingVolume;
 import org.gearvrf.GVRSwitch;
 import org.gearvrf.GVRTransform;
 import org.gearvrf.utility.Log;
+import org.gearvrf.scene_objects.GVRSphereSceneObject;
 
 import java.util.HashMap;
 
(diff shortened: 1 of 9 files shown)


liaxim:
@rahul27 What do you think?

rahul27:
Will review this today.

GVRf/Framework/framework/src/main/java/org/gearvrf/io/GVRGazeCursorController.java
```diff
@@ -172,5 +167,64 @@ void close() {
...
+                            setKeyEvent(keyEvent);
+                            break;
+                    }
+                    return false;
+                }
+            });
+        }
+
+        public void setPosition(float x, float y, float z) {
+            Message msg = Message.obtain(gazeEventHandler,SET_POSITION,new float[]{x,y,z});
```
rahul27: You can reuse a Vector3 to prevent a float array being created every time setPostion is called.

GVRf/Extensions/3DCursor/3DCursorLibrary/src/main/java/org/gearvrf/io/cursor3d/SelectableBehavior.java
```diff
@@ -399,7 +415,7 @@ void onCursorDeactivated(Cursor cursor) {
             states.remove(cursorId);
             if (currentState == state) {
                 ObjectState highestPrioritState = getHighestPriorityState();
```
rahul27: Minor comment : typo on priority

rahul27:
Looks good. Tested it with the 3D Cursor app and it worked fine.

Just need to address the comments and this PL is good to go.

parthmehta209:
Addressed comments on the patch.

liaxim:
If there are no further comments, will merge after lunch.

rahul27:
Noticed a few things while working on the scene editor. Can we hold off on merging this.

liaxim:
Sure

rahul27:
Tested this patch at it worked fine. GTG.

Must have noticed an issue with the scaling patch then.

PR #805, opened 2016-08-11
Adding the SceneSerializer extension

parthmehta209:
This library helps export/import scene information to/from a json file in the sdcard.


liaxim:
Any comments? If not, will merge in the afternoon.

PR #810, opened 2016-08-12
Make sure GVRScript.onStep is invoked as before

liaxim:
- the scripting is not going to be called but we do need to
call the Java side


liaxim:
If there are no comments will merge Monday morning.

thomasflynn:
i thought this was removed for performance reasons?

liaxim:
Without this we are not invoking GVRScript.onStep anymore.

roshanch:
We need this one. We need to remove onStep() for mScene. I will open the pull request for that.

Issue #800, opened 2016-08-11
Eyepicker + TextViewSceneObjects


Nathipha:
I'm currently trying to get the following to work:
Two to four buttons (TextViewSceneObjects) with the one you look at changing color. Later I also want to make them clickable to start different things depending on which one got chosen.

Here's my code (half taken from the "gvrpickandmove" sample):
```
public class PickHandler implements IPickEvents {
    public void onEnter(GVRSceneObject sceneObj, GVRPicker.GVRPickedObject pickInfo) {
        if(sceneObj instanceof GVRTextViewSceneObject) {
            ((GVRTextViewSceneObject) sceneObj).setBackgroundColor(Color.GREEN);
            picked = sceneObj;
        }
    }
    public void onExit(GVRSceneObject sceneObj) {
        if(sceneObj instanceof GVRTextViewSceneObject) {
       	    ((GVRTextViewSceneObject) sceneObj).setBackgroundColor(Color.BLACK);
       	}
    }
    public void onNoPick(GVRPicker picker) {
        picked = null;
    }
    public void onPick(GVRPicker picker) { }
    public void onInside(GVRSceneObject sceneObj, GVRPicker.GVRPickedObject pickInfo) { }
}

@Override
public void onInit(GVRContext gvrContext) {
    .....

    GVRSceneObject headTracker = new GVRSceneObject(gvrContext,
        new FutureWrapper<GVRMesh>(gvrContext.createQuad(0.1f, 0.1f)),
        gvrContext.loadFutureTexture(new GVRAndroidResource(
            gvrc, R.drawable.crosshair)));
    headTracker.getTransform().setPosition(0.0f, 0.0f, -1.0f);
    headTracker.getRenderData().setDepthTest(false);
    headTracker.getRenderData().setRenderingOrder(100000);
    scene.getMainCameraRig().addChildObject(headTracker);
    headTracker.attachComponent(new GVRPicker(gvrContext, scene));
    pickHandler = new PickHandler();
    scene.getEventReceiver().addListener(pickHandler);


    scene.getMainCameraRig().getLeftCamera().setBackgroundColor(1.0f, 0.0f, 0.0f, 1.0f);
    scene.getMainCameraRig().getRightCamera().setBackgroundColor(1.0f, 0.0f, 0.0f, 1.0f);
    GVRTextViewSceneObject b1 = new GVRTextViewSceneObject(gvrContext, 2f, 0.5f, "test 1");
[...]

PR #812, opened 2016-08-15
Fixed picking bug (fixes issue #800)

NolaDonato:
The wrong matrix was being used for transforming the picking ray causing
it to pick the wrong thing.
Now the picker actually looks at the object it is attached to and gets
its position and orientation from that. Before it was incorrectly always
using the camera view matrix no matter what it was attached to. Some of
the demos have bugs because they attach the picker to the wrong object
and that used to work. I am doing a pull request for these demos.
Removed logging and exposed GVRCameraRigBase.getHeadTransformObject.

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRCameraRigBase.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRCameraRigBase.java
@@ -103,7 +103,7 @@ abstract class GVRCameraRigBase extends GVRComponent implements PrettyPrint {
     }
 
     protected abstract void addHeadTransformObject();
-    protected final GVRSceneObject getHeadTransformObject() {
+    public final GVRSceneObject getHeadTransformObject() {
         return headTransformObject;
     }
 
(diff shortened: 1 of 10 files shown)


roshanch:
tested with planetarium, and works fine.

roshanch:
tetsted with gvr-tutorial6 sample, works fine.

liaxim:
Does not fix a crash on launch of gvr-immersivepedia. I think the same reason is why gvr-video and gvr-accessibility crash on tap.

The crash was introduced in the pick visible pull request. It was my mistake to merge it early but please do mark pull requests not meant for merging in the future. How can we tell otherwise?

deepakrawat22:
In new video I received crash at:

_ZN3gvr9Transform14getModelMatrixEv
/home/deepak.rawat/Documents/GearVRf/GearVRf/GVRf/Framework/framework/src/main/jni/objects/components/transform.cpp:69


will check in more details.

deepakrawat22:
@NolaDonato Even with this new pull request both gvr-accessibility and gvr-immersivepedia are crashing at startup. I was wrong earlier regarding gvr-immersivepedia, forgot to rebuild it.

NolaDonato:
gvr-accessibility crashes on startup for me with master too. It runs out of memory during initialization.

NolaDonato:
I found out what the crash on startup was in gvr-accessibility and gvr-immersivepedia and fixed it

NolaDonato:
I updated pull request #812 to fix this problem

liaxim:
Looks good to me. Went through all the samples and the picker works and the crash due to the null matrix is gone. Will merge after lunch if there are no comments.

PR #813, opened 2016-08-15
Remove onStep() for mainScene

roshanch:
This onStep() traverses through all the sceneObjects in the scene. We don't need this one as we don't have any use case where we override the onStep() on sceneObject.

GearVRf-DCO-1.0-Signed-off-by: Roshan Chaudhari  roshan.c@samsung.com


liaxim:
Removing the invocation is not enough. We have to remove the code that is no longer used and correct the events interface (since onStep is not going to be called anymore).

1. Need a new ISceneEvents that derives directly from IEvents and defines onInit and onAfterInit only.
2. Remove obsolete onStep-related code from GVRScene.java.

roshanch:
Made changes as per @liaxim comments.

liaxim:
Looks good to me. If there are no comments will merge after lunch.

Issue #734, opened 2016-07-21
Working with COLLADA or FBX


QKnows:
Hello,
I try to import some complex scene (gvrContext.loadModel("scene.fbx")) and I have no so much success.
What's the best way to work with GVRModelSceneObject, which fbx or COLLADA format is supported, is there some specification available? Some tip which application should be used, to work with fbx or Collada files? All formats I tried have some issue with shader or something else, each scene doesn't look nice as it should. So I would be really thankful for some idea...

NolaDonato:
We do support Collada and FBX but not everything in those file formats. Would it be possible for you to send us a test asset showing the problem? Maybe we can figure out what is in your file that does not import correctly.

liaxim:
@QKnows Attaching some logs might help us identify the problem.

QKnows:
Hello, I attached a log from my application, try to load an fbx file, which have converted with the autodesk converter, I read about the converter in another issue.

The FBX file has the following header:
; FBX 7.3.0 project file
; Copyright (C) 1997-2010 Autodesk Inc. and/or its licensors.
; All rights reserved.
; ----------------------------------------------------

I can't provide that scene file, because it has some copyright issues.

I tried many exports from DAZ 3D, and the most scenes [just simple scene, sphere with a mapped texture works] doesn't work because of some shader problems.

So any (complex-)scene which has been created with DAZ 3D and some lights will be shown wrong in GearVR, I am not sure about copyright of provided DAZ 3D stuff - so I am afraid to share it here...

http://www.daz3d.com

[load_scene.txt](https://github.com/Samsung/GearVRf/files/376331/load_scene.txt)

liaxim:
Ok, I see couple of these which is abnormal:

```
07-21 17:01:49.883 718-772/com.test.app.scene.player I/gvrf: !! 0xd7e0c640 Prog Id -- -1 
07-21 17:01:49.883 718-772/com.test.app.scene.player I/gvrf: after renderMesh::renderMaterialShader() glError (0x501)
```

We will look into it. Thank you for letting us know.

QKnows:
Hello @liaxim - are you still working on that isseu? Or should I try to find another way to use 3D scenes?

liaxim:
We have been fixing model loading bugs. I will have somebody double-check the current status when trying to import models from DAZ 3D.

QKnows:
Thank you @liaxim for still working on that issue...

deepakrawat22:
Hi All,

[...]

PR #814, opened 2016-08-16
Should free jNode when not needed

deepakrawat22:
This fixes the JNI table overflow issue for models which have large hierarchy of nodes. Models exported as FBX from DAZ studio has very large hierarchy, so they can be used to review this pull request.


liaxim:
Looks good. Will merge in the afternoon if there are no comments.

@deepakrawat22 Do you think this fixes https://github.com/Samsung/GearVRf/issues/734 or are there other issues with DAZ-exported models?

deepakrawat22:
@liaxim The model file I am using still has issue with bone index out of bound problem. I am looking separately into that. So for now we can keep #734 open.

deepakrawat22:
@roshanch told me the bone index out of bound is a known problem and for now the limit is 60. I tried another fbx file from DAZ studio which does not have any bones and it worked. I exported to FBX 2012 ASCII with texture in a folder.

PR #816, opened 2016-08-16
updating assimp library to support layered textures

roshanch:
updating assimp library for:
1. Storing all the layered textures and blend mode.
2. Some of the FBX files has "SpecularFactor" with specular texture, adding support for that.


liaxim:
What samples have been verified with the updated assimp?

roshanch:
I test with gvr-testmodels with assimp samples and with the model given by Mike, and modified framework to see whether all textures are imported or not.

roshanch:
Test model is shared at \\corpdata_sj\SHARE FOLDER\Roshan\lightmap_test

liaxim:

roshanch:
assimp needs more modification, which i will update today or tomorrow. I will upload the new patch here.

liaxim:
Should be merged together with or prior to https://github.com/Samsung/GearVRf/pull/870

liaxim:
@roshanch Whenever you get the chance, could you please add an so for arm64-v8a? Thanks.

liaxim:
And again, when you have the time, please try to build with optimizations and no debug info. The so is 11MBs; after stripping it goes down to 7MBs. Every once in a while we get asked about binary sizes.. Are all changes made to assimp in our gerrit?

liaxim:
No problems encountered with the samples. Will merge tomorrow morning if there are no comments.

roshanch:
Do we need .so with target= "arm64-v8a" ? I tried with -O3 flag but still we are getting same size. I need to try more flags.

liaxim:
An arm64 build is a nice-to-have. We can produce a 64 bit build with certain backends. The 64 bit assimp so in the repo is very old.

roshanch:
I remember few months back I tried with the arm64 target and it crashed. That's the reason I built it with other. But I do have .so for arm64. 
And this code is not updated in gerrit yet.

PR #820, opened 2016-08-16
Fixed problem with disabling lights 

NolaDonato:
The uniform value in the shader was not getting correctly set


liaxim:
Looks good to me. Will merge Thursday morning if there are no further comments.

PR #821, opened 2016-08-17
Fix a crash due to a mesh without normals

liaxim:
gvr-controls known to be affected

### DIFF
--- a/GVRf/Framework/framework/src/main/jni/engine/renderer/batch.cpp
+++ b/GVRf/Framework/framework/src/main/jni/engine/renderer/batch.cpp
@@ -63,7 +63,6 @@ bool Batch::add(RenderData *render_data) {
     matrices_.push_back(model_matrix);
     render_data->owner_object()->setTransformUnDirty();
 
-
     // if it is not texture shader, dont add into batch, render in normal way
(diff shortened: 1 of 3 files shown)


liaxim:
@roshanch Please review

Vittalbabu:
What happens if tex_cords are empty? Do we need to pull that out of the vertices loop like normals or will that never occur?

roshanch:
@liaxim  I added this check in #775 .

roshanch:
@Vittalbabu We are batching only with texture shader. I would add again error message if texCords are not present.

thomasflynn:
the change looks sane.

liaxim:
Will merge after lunch.

Vittalbabu:
More of an FYI -- the normal size and vertices size have to match else the Mesh::generateVAO code will not generate a mesh in some circumstances. Adding a LOG stmt if the sizes do not match might help.

liaxim:
@Vittalbabu Added the requested log trace to mesh.cpp. Let me know whether this is ok.

Vittalbabu:
Leave aLooks good. comment

liaxim:
Thanks. Will merge in the evening.

PR #822, opened 2016-08-17
Support for N7 and the new headset

liaxim:
- default fov-y to 90 for the old headset and 93 for the new
- handle the back key on the new headset differently
- gvrconfigurationmanager moved to framework; backends customize it

### DIFF
--- a/GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrActivityDelegate.java
+++ b/GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrActivityDelegate.java
@@ -127,30 +127,51 @@ final class OvrActivityDelegate implements GVRActivity.GVRActivityDelegate {
 
     @Override
     public boolean onKeyDown(int keyCode, KeyEvent event) {
-        if (KeyEvent.KEYCODE_BACK == keyCode) {
-            event.startTracking();
-            return true;
-        }
         return false;
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


liaxim:
Ready for review and testing. Few notes:

1. Use https://github.com/gearvrf/GearVRf-Demos/pull/302 to update the demos so the default fov-y gets picked by the framework
2. Pay attention to the short and long back key press; short back key press doesn't do anything on the new headset - the dedicated home key gets you to Oculus home; long press should still bring up the global menu as before.
3. It is quite easy to not dock properly Note7 in the headset..
4. And most importantly:  I don't think things looks good with fov-y of 101 on the new headset and 96 on the old one. Seems like 5 degrees less is better. But will need input from other observers on that.

liaxim:
There is some stuttering on Note7. Not seeing anything in this pull request so far. But I do notice S7 to be significantly faster doing the same things as Note7. Launching an app, bringing up the global menu.. Is Note7 underclocked or what..

liaxim:
The stuttering is there on Note7 and not present on S7. We should investigate separately. I don't think it is down to this change.

liaxim:
Rebased

thomasflynn:
tested, looks much better now.  back key seems to work appropriately as well.

one interesting thing i noticed:  if you have the device in developer mode, run an app, place it in the new headset, then take it out and place it in the old headset, the fov is noticeably wrong.  i realize this isn't a normal use case, but it was worth noting.

PR #823, opened 2016-08-18
Fix for the texture shader when its light support and batching are en…

liaxim:
…abled

fixes the crash in gvr-cardboard-audio on launch


liaxim:
@roshanch Please review

roshanch:
with this fix, how does it get mv (modelView) matrix when light and batching enabled?

liaxim:
@roshanch Updated according to your comment; in case both BATCHING and LIGHT are enabled mv is ``"mv = u_view * model_matrix;\n"``

The litshader sample works because LIGHT and BATCHING are not enabled together when it runs. For whatever reasons in cardboard-audio they are both enabled hence we were creating a shader without u_model defined but were trying to refer to it. The shader compilation failure is handled less than gracefully so we crash some time later.

--

Noticed that if batching and lighting are both enabled on startup, trying to disable the light on the renderData has no effect (the light stays enabled always). Can you look into that?

It would be really nice if you build a test app that goes through the various scenarios possible with batching and allows for quick and focused verification. It will be a must for the custom shader batching so the sooner you think about this for the texture shader the better.

roshanch:
so the fix should be replace u_model with model_matrix so 
```
"#else\n"
        "mv = u_view * model_matrix ;\n
  "#endif\n"
```

roshanch:
And i will look into other issue you mentioned.

liaxim:
Yes, the fix in case both BATCHING and LIGHT are defined is to do "mv = u_view * model_matrix;\n"

liaxim:
Will merge in the afternoon if there are no further comments.

liaxim:
@roshanch  Made the update you requested.

roshanch:
thanks. good to go.

NolaDonato:
Disabling the light on the renderdata works with the custom shader if batching is off. The gvr-multilight sample tests this.

roshanch:
@NolaDonato  problem is with texture shader. I found the fix, will push it today afternoon.

PR #824, opened 2016-08-19
Fixed a bug in GVRFrustumCuller

NolaDonato:
I had accidentally swapped the field of view and aspect ratio arguments.
That is fixed and frustum culling works now.


GVRf/Framework/framework/src/main/java/org/gearvrf/GVRFrustumPicker.java
```diff
@@ -217,6 +213,7 @@ public void doPick()
                     p.sub(center, p);
                     p.w = 0;
                     radius = p.length();
+                    boolean pointIn = mCuller.isPointInsideFrustum(center.x, center.y, center.z);
```
liaxim: This is unused. Is it intentional?

liaxim:

PR #826, opened 2016-08-19
Adding GVRObjectPicker, fixed bug with hit position 

NolaDonato:
This patch fixes a bug with the calculation of the hit location for
GVRMeshCollider
It add GVRObjectPicker which picks against the bounding box of the scene
object the picker is attached to.

Updated documentation on picking.


liaxim:
Please consider squashing the commits.

NolaDonato:
Replaced by PR #829 with commits squashed

liaxim:
Fyi, you don't have to close a pull request to update it after squashing or any other git operation.

You can be on any local branch (any name) and then do something like:
```git push origin HEAD:<branch-name-you-use-for-the-pull-request> --force```

This will overwrite the remote branch ``<branch-name-you-use-for-the-pull-request>.``

PR #829, opened 2016-08-23
Adding GVRObjectPicker, fixed bug with hit position

NolaDonato:
This is the same as PR #826 with commits squashed
This patch fixes a bug with the calculation of the hit location for
 GVRMeshCollider
 It add GVRObjectPicker which picks against the bounding box of the scene
 object the picker is attached to.

Updated documentation on picking.


rahul27:
I still haven't found the time to try this out. I should get to it next week. 

Am okay if someone else wants to try this out and :+1: it.

rahul27:
@NolaDonato did you use the test framework to test this code or did you modify an existing sample?

I was planning on writing some code to try it, would help if you already have something.

NolaDonato:
I modified a sample but I did not test really thoroughly. I will write some unit tests when I am done fixing shadow mapping.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRObjectPicker.java
```diff
@@ -0,0 +1,134 @@
...
+     * This function is called automatically by
+     * the picker if it is attached to a scene object.
+     * You can instantiate the picker and not attach
+     * it to a scene object. In this case you must
+     * manually set the pick ray and call doPick()
+     * to generate the pick events.
+     * @see IPickEvents
+     * @see GVRFrustumPicker.pickVisible
+     */
+    public void doPick()
```
liaxim: Regarding the javadoc: if not attached to a scene object, wouldn't it still pick against the list of visible colliders? How the pick ray comes into play in this case?

liaxim:
Looks good to me

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRObjectPicker.java
NolaDonato: If not attached to a scene object, all of the pickers act like they are attached to the owner of the camera rig. The pick ray is always relative to that. Maybe we should just get rid of the pick ray. The reason I did not do that is in case the IO devices needed something different than that.

liaxim:
Any comments? Will merge tomorrow otherwise.

rahul27:
I modified the eye picking sample to test the object picker. 

Picking works fine with the bunnies, but does not seem to work well with the quads in the scene. Here is the modified sample I used to test:

https://github.com/rahul27/GearVRf-Demos/commit/e5cc9442b9bd37629e0e0e740b93ef86f563b170

Use the swipe to move the cursor forward/backward.

liaxim:
@NolaDonato Can this be reviewed and tested once more?

liaxim:
@rahul27 I see a number of commits since your comment. Would you mind running your test case again? Thanks.

rahul27:
Yup. Still see the issue.

NolaDonato:
I fixed the problem Rahul found. I think there might be a problem with our bounding box computation logic but using GVRSceneObject.isColliding seems to get around it.

liaxim:
This now has 55 commits. A lot of them that do not belong. Impossible to review. Please shrink the number of commits down the actual changes. Thank you.

NolaDonato:
Yeah I guess I made a mess of it when I committed from my personal laptop. I forgot my power adapter last night and I wanted to fix it so I did it on a different machine. I thought I pulled over everything before I started so I'm not sure what happened. I used the method you suggested to force push just mychanges  (git push origin HEAD:objectpicker2 --force)

rahul27:
@NolaDonato I've reverted the picker to use the bounding volume and use the modified matrix transformation here https://github.com/Samsung/GearVRf/pull/862.

Can we close this pull request?

NolaDonato:
yes

PR #832, opened 2016-08-24
Refactor the framework and the backend so they are separate

liaxim:
gradle modules and the framework has no dependencies to the
backends anymore

the framework produces its own aar including libgvrf.so
the oculus backends produces its own aar including libgvrf-oculus.so

Some things will be left out (fallback to backends, monoscopic
backends, google vr, other backends and more) to meet a certain deadline


liaxim:
I will go through testing all the samples once more but I think this is review-able and can be tried out.

Needs the following Demos pull requests:

Gradle changes to use the backend aar
https://github.com/gearvrf/GearVRf-Demos/pull/308
Uncomment in common.gradle the ``compile(name: 'backend-debug', ext: 'aar')`` line

Fix for gvr-keyboard for what I believe is an application bug
https://github.com/gearvrf/GearVRf-Demos/pull/309

This is needed so modelviewer2 starts using the common.gradle file
https://github.com/gearvrf/GearVRf-Demos/pull/310


Actually some samples that didn't work all too well happen to work now. Namely:
gvr-meshanimation (the trex was frozen, now animating)
gvr-accessibility (used to render nothing, now it seems fully working; very odd)
gvr-sixaxissensor (i am now seeing things i haven't seen ever before - some additional readings)

Probably there is/was a race condition during startup and/or order of execution issue.

liaxim:
Ok, I think this is ready for review and considered for merging.

Misc notes:

Added the com.samsung.android.hmt.permission.READ_SETTINGS permission to the framework's manifest; all apps will get it. Oculus is trying to use VRHelper according to the logs so this would help with their headset detection, I presume.

Added android:largeHeap="true" to the framework manifest; again all apps will get it; seems reasonable for vr apps. gvr-video is using a lot memory and can run out during onInit; leaves it running but not quite functional (the pause/resume buttons do not show for example). With this pull request the memory usage is just a little bit more and maybe it is enough to go over the threshold.

GoogleVR and monoscopic fallbacks have been verified as working.

liaxim:
One last thing: it is possible after pause/resume to end up with the whole view squashed in the left eye. This happens on master too and affects the use case where developer mode is in use.

GVRf/Framework/backend_oculus/build.gradle
```diff
@@ -0,0 +1,216 @@
...
+        commandLine ndkbuild, '-C', file('src/main').absolutePath, '-j', Runtime.runtime.availableProcessors(), 'clean'
+    }
+
+    clean.dependsOn 'cleanNative'
+
+    tasks.withType(JavaCompile) {
+        compileTask -> compileTask.dependsOn buildNative, copyOculusFiles
+    }
+
+    task eclipseZip(type: Zip) {
```
rahul27: Is it safe to delete this task? Seems like the directory structure and packing rules don't apply anymore.

GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrActivityDelegate.java
```diff
@@ -0,0 +1,192 @@
...
+    @Override
+    public void setViewManager(GVRViewManager viewManager) {
+        mActiveViewManager = (OvrViewManager)viewManager;
+    }
+
+    @Override
+    public void onInitAppSettings(VrAppSettings appSettings) {
+        if(mUseFallback){
+            // This is the only place where the setDockListenerRequired flag can be set before
+            // the check in GVRActivityBase.
```
rahul27: Small change : GVRActivityBase is now GVRActivity

GVRf/Framework/backend_oculus/build.gradle
liaxim: I will delete it and re-add it in a functional state later, possibly.

rahul27:
We might have to make a note to copy the oculus libs to the backend library in the future for correctness. Although it can also pick it up from the framework library.

liaxim:
The oculus backend's build.gradle does copy the oculus libraries. The core framework does not.

GVRf/Framework/framework/src/main/AndroidManifest.xml
```diff
@@ -25,9 +25,12 @@
         android:minSdkVersion="19"
         android:targetSdkVersion="19" />
 
+    <uses-permission android:name="com.samsung.android.hmt.permission.READ_SETTINGS" />
```
rahul27: Is this required by oculus?

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRActivity.java
```diff
@@ -183,8 +210,6 @@ protected void onResume() {
     protected void onDestroy() {
         android.util.Log.i(TAG, "onDestroy " + Integer.toHexString(hashCode()));
         if (mViewManager != null) {
-            mViewManager.onDestroy();
```
rahul27: Was this skipped intentionally?

GVRf/Framework/framework/src/main/jni/objects/components/component_types.h
```diff
@@ -0,0 +1,30 @@
...
+namespace gvr {
+    static const long long COMPONENT_TYPE_TRANSFORM          = 10001;
+    static const long long COMPONENT_TYPE_LIGHT              = 10002;
+    static const long long COMPONENT_TYPE_BONE               = 10003;
+    static const long long COMPONENT_TYPE_BONE_WEIGHT        = 10004;
+    static const long long COMPONENT_TYPE_CAMERA             = 10005;
+    static const long long COMPONENT_TYPE_CAMERA_RIG         = 10006;
+    static const long long COMPONENT_TYPE_COLLIDER           = 10007;
+    static const long long COMPONENT_TYPE_RENDER_DATA        = 10008;
+    static const long long COMPONENT_TYPE_TEXTURE_CAPTURER   = 10009;
```
rahul27: Is this how we are going to assign component types going forward? There are a bunch of components at the Java level that makes use of timestamps/class hashcode to determine the type.

GVRf/Framework/framework/src/main/AndroidManifest.xml
rahul27: Can this go into the AndroidManifest of the backend?

GVRf/Framework/framework/src/main/jni/objects/components/component_types.h
liaxim: I will probably reserve a range for Java components and have something similar implemented on the Java side.

GVRf/Framework/framework/src/main/AndroidManifest.xml
liaxim: It could. Currently it would be in all backends that we have. I rather request the permission for Oculus since I can see that they are trying to detect the model. The permission will not be required by more recent GearVR services but there are versions out there that need it.. If you insist I can move it to the backends.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRActivity.java
liaxim: No. This is a mistake. Will upload a fix.

GVRf/Framework/framework/src/main/AndroidManifest.xml
rahul27: If it is required by all the backends then it is fine. We could revisit it when we work on adding more backends in the future.

GVRf/Framework/framework/src/main/jni/objects/components/component_types.h
NolaDonato: This is only for native component types. These were always handled differently than the Java components. The Java code does not need to change. The native components used the address of a static object as the type. Now they use integers. It is less open-ended but safer.

liaxim:
Rebased

rahul27:
Looks good :+1:  Tested it with quite a few apps and it works fine.

liaxim:
Thank you. Will merge tomorrow morning if there are no further comments.

liaxim:
Actually this can only be merged together with the other pull request. Have to wait on it.

NolaDonato:
I wrote some tests for picking in the GearVRf-Tests repo in pull request #4. Some of them test the exact hit position so if the camera is different maybe the picking test will show it.

liaxim:
I will try the tests. Thanks.

PR #835, opened 2016-08-26
Fix for change in material and render-data properties with batching

roshanch:
1. Creates a pool of batches which is created initially, batches are taken from this pool.
2. Adding event listener for renderdata as part of mesh, material, render-pass which can be shared by multiple render-data
3. Disables batching for the render-data if mesh, material or any of the property changes in OnStep().

Tested with jassimp, gvr-testmodels, gvr-tutorial-6, litshader, outline sample, gvr-video, gvr-controller, planetarium and custom sample where light, material, mesh is modified in onStep().


GVRf/Framework/framework/src/main/jni/engine/renderer/batch.cpp
```diff
@@ -164,68 +160,52 @@ bool Batch::isRenderDataDisabled(){
...
-    bool update_vbo = isRenderDataDisabled();
-
-    if(render_data_set_.size() ==0)
-        return;
-
-    if(update_vbo)
-        regenerateMeshData();
+void Batch::resetBatch(){
+    clearData();
+    delete renderdata_;
```
liaxim: Wouldn't it be safer if renderdata_ was also set to nullptr? In the future the dtor may actually execute and the flow may be such that this renderdata_ may be deleted second time.

GVRf/Framework/framework/src/main/jni/engine/renderer/batch.cpp
```diff
@@ -164,68 +160,52 @@ bool Batch::isRenderDataDisabled(){
...
+    Renderer::batch_manager->deallocateBatch(this);
```
liaxim: You may consider a more canonical name for a pool like "free". deallocate could create a wrong impression.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRMesh.java
```diff
@@ -362,14 +362,6 @@ public GVRMesh getBoundingBox() {
         return mBones;
     }
     /**
```
liaxim: So what change makes the need to distinguish static vs dynamic mesh not needed anymore? How is the dynamic case handled now?

GVRf/Framework/framework/src/main/jni/engine/renderer/batch.h
```diff
@@ -20,13 +20,13 @@
 #include "objects/components/render_data.h"
 #include <map>
```
liaxim: This file uses tabs mostly for indentation; please fix it at some point.

liaxim:
Looks good and I can't find a sample broken by this change. Will merge in the afternoon if there are no other comments.

--
Having said all that I do think we need to start the unit tests for the batching. The sooner the better as it will be easier to write and expand them as batching grows in complexity. The fact the samples work doesn't mean that much - there is a lot of branching that the samples will not cover. Also we can't really tell how effective batching is by just looking at the samples. Not to mention ensuring memory management is correct. It is impossible code this complex not to have bugs and it will be much harder for us to find and deal with those bugs as they are discovered in apps by us or end users.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRMesh.java
roshanch: now any if any of the mesh changes, batching will be disabled for that mesh. so there is no more "dynamic"/ "static" now. same applies for material/ renderdata.

GVRf/Framework/framework/src/main/jni/engine/renderer/batch.cpp
roshanch: Will change that.

GVRf/Framework/framework/src/main/jni/engine/renderer/batch.h
roshanch: ok, I need to change my formatting in studio. will fix it.

roshanch:
Sure.! So I have sample which I modified for different cases for the batching. I will include that in Nola's test suit.

GVRf/Framework/framework/src/main/jni/engine/renderer/batch.cpp
roshanch: okay. will make that change.

PR #836, opened 2016-08-29
Fixed build issues with the IO template device

rahul27:
Also set the default connected state to true for convenience

r.rudradevan@samsung.com


liaxim:
Looks good. If there are no comments will merge after lunch.

PR #839, opened 2016-08-31
Refactor renderer, adding vulkan renderer

roshanch:
Making GL and Vulkan Renderer seperate.
Tested with jassimp and modelview sample.

### DIFF
--- a/GVRf/Framework/backend_oculus/src/main/jni/monoscopic/monoscopic_renderer_jni.cpp
+++ b/GVRf/Framework/backend_oculus/src/main/jni/monoscopic/monoscopic_renderer_jni.cpp
@@ -46,8 +46,8 @@ void Java_org_gearvrf_OvrNativeMonoscopicRenderer_renderCamera(JNIEnv * env,
     RenderTexture* post_effect_render_texture_b =
             reinterpret_cast<RenderTexture*>(jpost_effect_render_texture_b);
 
-    Renderer* renderer = Renderer::getInstance();
-    renderer->renderCamera(scene, camera, viewportX, viewportY, viewportWidth,
+    gRenderer = Renderer::getInstance();
+    gRenderer->renderCamera(scene, camera, viewportX, viewportY, viewportWidth,
                viewportHeight, shader_manager, post_effect_shader_manager,
                post_effect_render_texture_a, post_effect_render_texture_b);
 
@@ -59,11 +59,9 @@ void Java_org_gearvrf_OvrNativeMonoscopicRenderer_cull(JNIEnv * env,
     Scene* scene = reinterpret_cast<Scene*>(jscene);
     Camera* camera = reinterpret_cast<Camera*>(jcamera);
     ShaderManager* shader_manager = reinterpret_cast<ShaderManager*>(jshader_manager);
-    renderer = Renderer::getInstance();
-    if(renderer)
-        renderer->cull(scene, camera, shader_manager);
-    else
-        LOGE("renderer is null");
+    gRenderer = Renderer::getInstance();
+    gRenderer->cull(scene, camera, shader_manager);
+
 }
 
 }
--- a/GVRf/Framework/backend_oculus/src/main/jni/monoscopic/monoscopic_renderer_jni.cpp
+++ b/GVRf/Framework/backend_oculus/src/main/jni/monoscopic/monoscopic_renderer_jni.cpp
@@ -46,9 +46,11 @@ void Java_org_gearvrf_OvrNativeMonoscopicRenderer_renderCamera(JNIEnv * env,
     RenderTexture* post_effect_render_texture_b =
             reinterpret_cast<RenderTexture*>(jpost_effect_render_texture_b);
 
-    Renderer::renderCamera(scene, camera, viewportX, viewportY, viewportWidth,
-            viewportHeight, shader_manager, post_effect_shader_manager,
-            post_effect_render_texture_a, post_effect_render_texture_b);
+    Renderer* renderer = Renderer::getInstance();
+    renderer->renderCamera(scene, camera, viewportX, viewportY, viewportWidth,
+               viewportHeight, shader_manager, post_effect_shader_manager,
+               post_effect_render_texture_a, post_effect_render_texture_b);
+
 
 }
 
@@ -57,8 +59,8 @@ void Java_org_gearvrf_OvrNativeMonoscopicRenderer_cull(JNIEnv * env,
     Scene* scene = reinterpret_cast<Scene*>(jscene);
     Camera* camera = reinterpret_cast<Camera*>(jcamera);
     ShaderManager* shader_manager = reinterpret_cast<ShaderManager*>(jshader_manager);
-
-    Renderer::cull(scene, camera, shader_manager);
+    Renderer* renderer = Renderer::getInstance();
+    renderer->cull(scene, camera, shader_manager);
 }
 
 }
--- a/GVRf/Framework/backend_oculus/src/main/jni/ovr_view_manager.cpp
+++ b/GVRf/Framework/backend_oculus/src/main/jni/ovr_view_manager.cpp
@@ -29,11 +29,9 @@ void Java_org_gearvrf_OvrViewManager_cull(JNIEnv * jni, jclass clazz,
     Scene* scene = reinterpret_cast<Scene*>(jscene);
     Camera* camera = reinterpret_cast<Camera*>(jcamera);
     ShaderManager* shader_manager = reinterpret_cast<ShaderManager*>(jshader_manager);
-    renderer = Renderer::getInstance();
-    if(renderer)
-        renderer->cull(scene, camera, shader_manager);
-    else
-        LOGE("renderer is null");
+    gRenderer = Renderer::getInstance();
+    gRenderer->cull(scene, camera, shader_manager);
+
 }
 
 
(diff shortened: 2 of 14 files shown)


Vittalbabu:
Quick question -- should we have Composition as opposed to inheritance for the GL and Vulkan implementations? (ie) Should the Renderer class instance hold an instance of one of the GL or Vulkan implementation instance or should Renderer class be an interface that is implemented by the GL and Vulkan variants? 

At one level this depends on the state we have in Renderer class and how to give access to this to the GL and Vulkan variants.

(eg) this old article

http://www.artima.com/cppsource/codestandards3.html

NolaDonato:
Whether or not the Renderer is implemented in Vulkan, GL, DirectX or a combination, the public API should be the same (in both the C++ and Java layers).

If we need data structures that maintain rendering-specific information, these should be in the Renderer too. For example, the list of all currently active shaders should be there. The memory pools containing the items that will be rendered each frame for all of the active render targets should be there and we should use these items over and over each frame. These items will refer to the RenderDatas that are now in a static list. The geometry batches could also be associated with the Renderer.

I think there will be a common core implementation that is shared between all renderer implementations – culling, render state sorting, batching. The common implementation will call GL or Vulkan specific functions implemented either by a derived class from some other class. If you use a derived class there is a single memory allocation for the object and the GL or Vulkan specific implementations can easily access the common data. If you use another separate class, it will require two memory allocations and extra indirections to get to the renderer-specific functions. This seems more complex and less efficient to me. Given that this is a real time system where performance is important, I would favor inheritance.

-Nola

liaxim:
Calling virtual methods all the time can be considered another form of overhead. I'd say go for the design that makes sense the most right now and do not worry about micro optimizations. The design inevitably will evolve with time.

Vittalbabu:
It boils down to do we have a is-a or has-a relationship for the Renderer class. I think it is a has-a relationship which would favor composition.

Nothing prevents the composited components to implement inheritance if it is the right thing to do. 

I am a bit influenced by Herb Sutter's thoughts on this -- inheritance is a far tighter coupling/relationship than composition and has drawbacks.

thomasflynn:
looks straight-forward and looks like it should work fine.  i still need to cherry-pick and test...

thomasflynn:
needs rebase

roshanch:
Rebase completed.

GVRf/Framework/backend_oculus/src/main/jni/monoscopic/monoscopic_renderer_jni.cpp
```diff
@@ -57,8 +59,11 @@ void Java_org_gearvrf_OvrNativeMonoscopicRenderer_cull(JNIEnv * env,
     Scene* scene = reinterpret_cast<Scene*>(jscene);
     Camera* camera = reinterpret_cast<Camera*>(jcamera);
     ShaderManager* shader_manager = reinterpret_cast<ShaderManager*>(jshader_manager);
-
-    Renderer::cull(scene, camera, shader_manager);
+    renderer = Renderer::getInstance();
+    if(renderer)
```
thomasflynn: braces

GVRf/Framework/backend_oculus/src/main/jni/ovr_view_manager.cpp
```diff
@@ -29,7 +29,11 @@ void Java_org_gearvrf_OvrViewManager_cull(JNIEnv * jni, jclass clazz,
     Scene* scene = reinterpret_cast<Scene*>(jscene);
     Camera* camera = reinterpret_cast<Camera*>(jcamera);
     ShaderManager* shader_manager = reinterpret_cast<ShaderManager*>(jshader_manager);
-    Renderer::cull(scene, camera, shader_manager);
+    renderer = Renderer::getInstance();
+    if(renderer)
```
thomasflynn: braces

GVRf/Framework/backend_oculus/src/main/jni/ovr_view_manager.cpp
```diff
@@ -38,7 +42,11 @@ void Java_org_gearvrf_OvrViewManager_makeShadowMaps(JNIEnv * jni, jclass clazz,
     Scene* scene = reinterpret_cast<Scene*>(jscene);
 
     ShaderManager* shader_manager = reinterpret_cast<ShaderManager*>(jshader_manager);
-    Renderer::makeShadowMaps(scene, shader_manager, width, height);
+    renderer = Renderer::getInstance();
+    if(renderer)
```
thomasflynn: braces

GVRf/Framework/backend_oculus/src/main/jni/ovr_view_manager.cpp
```diff
@@ -121,10 +129,12 @@ void GVRViewManager::renderCamera(Scene* scene,
     float fps = 0.0;
     gettimeofday(&start, NULL);
 #endif
-
-    Renderer::renderCamera(scene, camera, shader_manager,
-            post_effect_shader_manager, post_effect_render_texture_a,
-            post_effect_render_texture_b);
+    if(renderer)
+        renderer->renderCamera(scene, camera, shader_manager,
```
thomasflynn: braces

GVRf/Framework/framework/src/main/jni/engine/renderer/renderer.cpp
```diff
@@ -37,37 +37,37 @@
...
+            instance = new GLRenderer();
+        std::atexit(resetInstance);      // Destruction of instance registered at runtime exit
+    }
+    return instance;
 }
-
-int Renderer::getNumberDrawCalls() {
-    return numberDrawCalls;
+Renderer::Renderer():numberDrawCalls(0), numberTriangles(0), batch_manager(nullptr){
+    if(do_batching)
```
thomasflynn: braces

GVRf/Framework/framework/src/main/jni/objects/light.cpp
```diff
@@ -249,7 +249,11 @@ bool Light::makeShadowMap(Scene* scene, ShaderManager* shader_manager, int texIn
     rstate.uniforms.u_view = lightcam.getViewMatrix();
     rstate.uniforms.u_proj = lightcam.getProjectionMatrix();
     rstate.render_mask = 1;
-    Renderer::renderShadowMap(rstate, &lightcam, fboId_, scene_objects);
+    renderer = Renderer::getInstance();
+    if(renderer)
```
thomasflynn: braces

GVRf/Framework/framework/src/main/jni/objects/scene.h
```diff
@@ -66,17 +66,19 @@ class Scene: public HybridObject {
     void clearLights();
 
     void resetStats() {
-        if (!statsInitialized) {
-            Renderer::initializeStats();
-            statsInitialized = true;
-        }
-        Renderer::resetStats();
+        renderer = Renderer::getInstance();
+        if(nullptr!= renderer)
```
thomasflynn: braces

thomasflynn:
need to put {} around your if-else statements.  will cherry-pick and test in the meantime...

liaxim:
I see backend-specific changes. You need to update the other backend in the other repo before this can be merged.

GVRf/Framework/framework/src/main/jni/engine/renderer/renderer.cpp
```diff
@@ -37,37 +37,38 @@
...
+
+#include "gl_renderer.h"
+#include "vulkan_renderer.h"
 #define MAX_INDICES 500
 #define BATCH_SIZE 60
 bool do_batching = true;
 
 namespace gvr {
-BatchManager* Renderer::batch_manager = nullptr;
+Renderer* renderer = nullptr;
```
liaxim: This being a global variable needs a better, more identifiable, name.

GVRf/Framework/backend_oculus/src/main/jni/ovr_view_manager.cpp
```diff
@@ -29,7 +29,13 @@ void Java_org_gearvrf_OvrViewManager_cull(JNIEnv * jni, jclass clazz,
     Scene* scene = reinterpret_cast<Scene*>(jscene);
     Camera* camera = reinterpret_cast<Camera*>(jcamera);
     ShaderManager* shader_manager = reinterpret_cast<ShaderManager*>(jshader_manager);
-    Renderer::cull(scene, camera, shader_manager);
+    renderer = Renderer::getInstance();
```
liaxim: Can Renderer::getInstance() really return nullptr? I think new should throw; in that case the checks and the logs are unnecessary.

thomasflynn:
did some testing, especially with pause and resume.  looks good so far.

GVRf/Framework/framework/src/main/jni/engine/renderer/renderer.cpp
```diff
@@ -208,58 +206,10 @@ void Renderer::cullFromCamera(Scene *scene, Camera* camera,
...
-    }
-    if (render_data->alpha_to_coverage()) {
-        GL(glDisable(GL_SAMPLE_ALPHA_TO_COVERAGE));
-    }
-}
 
 void Renderer::renderRenderDataVector(RenderState &rstate) {
 
-    if (!do_batching) {
+    if (!do_batching || (dynamic_cast<GLRenderer*>(renderer) == nullptr) ) {
```
liaxim: No need for dynamic cast and enabling rtti just for this. You could have the Renderer interface define a method that lets implementations tell you do they support batching or not.

roshanch:
removed rtti will do changes in other backend.

liaxim:
Looks good. The samples run fine.

NolaDonato:
I suggest that we make a separate class for rendering statistics instead of adding these functions to the renderer. In the future we will probably want to keep track of more things like the number of state changes, the number of shader changes, average batch size, etc. Since we are refactoring it would be more open-ended if we collected statistics and processed them in a separate class.

liaxim:
@NolaDonato Do you want this rendering statistics class done as part of this pull request or maybe somewhat later?

NolaDonato:
It can be done the next time we add a new statistic maybe. I have also noticed that specular reflections are broken in this patch. They might have been broken before this patch - not sure. The shadowmap patch will fix this once it is merged because reflections work correctly there.

PR #840, opened 2016-09-01
Initial phase of vulkan support

Vittalbabu:
Checking in files to allow ongoing development.


GVRf/Framework/framework/src/main/jni/vulkan/vulkanCore.cpp
```diff
@@ -0,0 +1,809 @@
+//
+// Created by vittalbab.s on 8/24/2016.
+//
+
```
thomasflynn: this should have the samsung copyright that we have on the other .cpp files.

GVRf/Framework/framework/src/main/jni/vulkan/vulkanCore.cpp
```diff
@@ -0,0 +1,809 @@
...
+#include "vulkanCore.h"
+#include "util/gvr_log.h"
+#include <assert.h>
+#include <cstring>
+
+CoreVulkan* CoreVulkan::theInstance = NULL;
+
+#define QUEUE_INDEX_MAX 99999
```
thomasflynn: where does 99999 come from?

apjagdale: We need to assign UINT32_MAX instead 99999.

Vittalbabu:
Two things --

1. We could use a bool to signal if the check succeeds as opposed to checking against the initial value to ensure we have not found the queue family. That would be a cleaner solution.
2. What should happen if we do not get the queue family? Is using the queue index of UINT32_MAX  useful or should we error out and declare that Vulkan support is absent on this device?

    uint32_t queueIndex = QUEUE_INDEX_MAX;
    for (uint32_t i = 0; i < queueFamilyCount; i++) {
        if ((queueProperties[i].queueFlags & VK_QUEUE_GRAPHICS_BIT) != 0) {
            if (supportsPresent[i] == VK_TRUE) {
                queueIndex = i;
                break;
            }
        }
    }

    delete [] supportsPresent;
    delete [] queueProperties;

    if (queueIndex == QUEUE_INDEX_MAX) {
        GVR_VK_CHECK("Could not obtain a queue family for both graphics and presentation." && 0);
    }

    // We have identified a queue family which both supports our android surface,
    // and can be used for graphics operations.
    m_queueFamilyIndex = queueIndex;

thomasflynn:
needs rebase

Vittalbabu:
Rebase completed.

GVRf/Framework/framework/src/main/jni/vulkan/vulkanCore.cpp
```diff
@@ -0,0 +1,820 @@
...
+ */
+
+#include "vulkanCore.h"
+#include "util/gvr_log.h"
+#include <assert.h>
+#include <cstring>
+
+VulkanCore* VulkanCore::theInstance = NULL;
+
+#define QUEUE_INDEX_MAX 99999
```
thomasflynn: what was the final deal with this magic number?

Vittalbabu: Removed the magic number and use a test against queueFamilyCount + 1 to check if graphics queue is present.

liaxim:
Looks good.

PR #841, opened 2016-09-04
Added GVRAssetLoader.loadScene, removed unused code 

NolaDonato:
Added asset loader call to replace the entire scene with the loaded
asset. X3D needs this for the anchor node. Added new capabilities to
GVRModelSceneObject: an initial filename and the ability to load that
asset.
GearVRf-DCO-1.0-Signed-off-by: Nola Donato nola.donato@samsung.com

PR #842, opened 2016-09-08
Added GVRAssetLoader.loadScene, removed unused code (replaces #841)

NolaDonato:
Added asset loader call to replace the entire scene with the loaded
 asset. X3D needs this for the anchor node. Added new capabilities to
 GVRModelSceneObject: an initial filename and the ability to load that
 asset.

Trying to do some advanced git stuff I hopelessly screwed up pull request #841. I am submitting this one in place of 841.


GVRf/Framework/framework/src/main/java/org/gearvrf/scene_objects/GVRModelSceneObject.java
```diff
@@ -43,6 +48,49 @@ public GVRModelSceneObject(GVRContext gvrContext) {
...
+     * If the model is already in the scene it will be
+     * disabled until all of the textures are available.
+     * If no file name was given on the constructor, this
+     * function does nothing. If the asset has already
+     * been loaded, this function does nothing.
+     * @throws IOException if the asset does not exist
+     * @see GVRModelSceneObject
+     * @see GVRModelSceneObject.getFileName
+     */
+    public void load() throws IOException
```
liaxim: Why do we need to expose yet another way of loading a model? Why don't we have all the loading go through the assetManager?

GVRf/Framework/framework/src/main/java/org/gearvrf/jassimp2/GVRJassimpModelHelper.java
```diff
@@ -3,62 +3,110 @@
...
 import org.gearvrf.scene_objects.GVRModelSceneObject;
+import org.joml.Matrix4f;
 import org.joml.Quaternionf;
+import org.joml.Vector3f;
 
-public class GVRJassimpSceneObject extends GVRModelSceneObject {
-    private static final String TAG = GVRJassimpSceneObject.class.getSimpleName();
-    protected AiScene scene;
-    protected GVRResourceVolume volume;
+public final class GVRJassimpModelHelper
```
liaxim: Could be moved to the org.gearvrf package and we could have one less internal class exposed as public. Just a note, we will revisit the packages some time in the future.

liaxim:
I am not so sure about the changes to GVRModelSceneObject. In my opinion all the loading APIs should be in the AssetManager.

Should the AssetManager start using the new ctor where the file path is passed in?

--
Other than that I made modifications to gvr-meshanimation and it all worked fine.

NolaDonato:
GVRModelSceneObject needs to at least keep the filename so it can act as an X3D Anchor node. This means that when you click on it, the scene is replaced with the asset it references. The loading API is still in the asset manager, GVRModelSceneObject just invokes it at the right time.

liaxim:
Since GVRModelSceneObject is unfortunately a public class we need to be careful with its apis.

1. A user can discover GVRModelSceneObject and its new load() method. He or she can (should) discover getGVRContext().getAssetLoader().loadModel(); not to mention that he or she can discover that you can create a GVRAssetLoader and call load on that too. This is not simple.

2. Since the new ctor and load in GVRModelSceneObject are added just for internal purposes they do not have to be exposed. The new ctor may be unavoidable but whatever the GVRModelSceneObject.load does, x3d can do itself. I guess yet another overload for GVRAssetLoader.load is necessary so you can pass both the modelSceneObject and the fileName.

3. All GVRModelSceneObjects that are produced as a result of loading a file from somewhere should return the filename. From what I can tell this is not the case.

4. Sooner rather than later we should move the x3d classes to the org.gearvrf package so we can avoid leaking internal methods as public.

5. Instead of adding a new method to GVRModelSceneObject why don't add a meta-data component that X3D can query for from any GVRSceneObject? The assetLoader can produce the metadata component for everything it loads. We should remove GVRModelSceneObject in the future, in my opinion.

NolaDonato:
Mihail requested components instead of GVRModelSceneObject. I agree so I reduced dependency on GVRModelSceneObject (so we can take it out) and introduced two new components.
GVRAnimator contains a list of animations (to replace the list that was in GVRModelSceneObject) but it has more functionality. It can operate on the animation list as a whole (set the repeat count, mode, start animations)
GVRExternalScene is a component that contains a filename and is capable of loading an asset at runtime. If you put this into the scene it references an external asset that can be loaded programmatically, X3D
needs this to implement the anchor node.
This pull request is currently broken because getting the camera rig crashes - do not merge yet!

liaxim:
Change in X3DObject::X3DObject lines 301-305 from

```
      this.mainCamera = new GVRSceneObject(gvrContext);
      this.mainCamera.setName("MainCamera");
      cameraRigAtRoot = GVRCameraRig.makeInstance(gvrContext);
      this.mainCamera.attachComponent(cameraRigAtRoot);
      cameraRigAtRoot.setOwnerObject(this.mainCamera);
```
to
```
      cameraRigAtRoot = GVRCameraRig.makeInstance(gvrContext);
      this.mainCamera = cameraRigAtRoot.getOwnerObject();
      this.mainCamera.setName("MainCamera");
```

Creating a camera rig instance creates an owner for it automatically since the camera scene objects have to be set up for the camera rig. For that an owner is required. For now assume you can't change the owner of a camera rig once it is created.

Can't easily enforce the restriction in the code. There are two paths to guard - GVRCameraRig.setOwnerObject - easy to deal with; and GVRSceneObject.attachComponent - not so easy without hacks.

Will make a note to look at this: improve it or at least have it error out appropriately and write some tests.

NolaDonato:
I have fixed the problem with the camera rig, written JavaDocs for the new components and written unit tests which pass. This one is ready for review again.

GVRf/Framework/framework/src/main/java/org/gearvrf/scene_objects/GVRModelSceneObject.java
```diff
@@ -15,17 +15,21 @@
...
 import org.gearvrf.GVRContext;
+import org.gearvrf.GVRScene;
 import org.gearvrf.GVRSceneObject;
 import org.gearvrf.PrettyPrint;
 import org.gearvrf.animation.GVRAnimation;
 import org.gearvrf.utility.Log;
 
 /**
  * {@linkplain GVRSceneObject Scene object} that holds a loaded model.
+ * You can specify the path to an asset and it will automatically
```
liaxim: This javadoc probably is not correct after the last changes.

liaxim:
Looks good. Will merge in the afternoon if there are no comments.

PR #843, opened 2016-09-08
Change all extensions so they use the framework aar instead of

liaxim:
building the framework themselves

- also adds a top-level build file for all extensions so we can build them all easier


thomasflynn:
@rahul27 ?

rahul27:
:+1: Looks good.

PR #844, opened 2016-09-09
Fixed problm specifying GLSL shader version

NolaDonato:
This fix applies only to shaders derived from GVRShaderTemplate using the constructor that specifies the GLSL version.


thomasflynn:
@deepakrawat22 ?

liaxim:
The "es" part is required for 300. Shader compilation will fail for those shaders. I'd say just omit the version if 100 is needed.

NolaDonato:
I changed the default version to 100.  GVRPhongShader now explicitly specifies 300.

liaxim:
Looks good.

liaxim:
@deepakrawat22 Does this solve your problem?

PR #845, opened 2016-09-09
GVRVideoSceneObject: changes to support video players other than the Android MediaPlayer

liaxim:
Note: pausing and resuming playback because the activity has been paused/resumed should be handled automatically by the videoSceneObject; it currently relies on the user to do it. Some time in the future this will be fixed.


thomasflynn:
gvr-sceneobjects and gvr-opacityanigallery builds are broken with this change.  will need appropriate changes for those apps.

thomasflynn:
opacityanigallery has a problem starting the video i think, but it is not related to this check-in.  i see the same issue on master.  everything else looks good and works well.

PR #859, opened 2016-09-19
Fix the matrix transformation of the bounding volume

rahul27:
Generate the correct min and max values of the
bounding volume using the object transform.

r.rudradevan@samsung.com


rahul27:
Will merge this tomorrow if there are no comments.

liaxim:
I assume this change is ok with Nola?

NolaDonato:
Yes this change is good. It puts back the original object picker code which actually worked if the bounding box is correct and fixes the problem with the bounding box.

After this is merged all the picker tests will work. Right now the tests that look at the hit position fail. This pull request also fixes that issue.

mwitchwilliams:
Verified this fix works.  There had been a repeatable bounding box problem when using a specific object rotation in X3D.  But it works now!

PR #860, opened 2016-09-19
Copy all publishable artifacts to GVRf/gearvrf-libs

liaxim:
Needed to simplify the automated build; a zip file will be produced with all
aars in that directory at the end of a successful build.


liaxim:
Will merge tomorrow morning if there are no comments.

PR #862, opened 2016-09-20
Adding the Object Picker

rahul27:
r.rudradevan@samsung.com

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRObjectPicker.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRObjectPicker.java
@@ -14,17 +14,6 @@
  */
 
 package org.gearvrf;
-
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import java.util.concurrent.locks.ReentrantLock;
-
-import org.joml.FrustumCuller;
-import org.joml.Matrix4f;
-import org.joml.Vector3f;
-import org.joml.Vector4f;
-
 /**
  * Finds the scene objects that intersect the scene object the
  * picker is attached to.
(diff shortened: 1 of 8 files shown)


rahul27:
This PL should replace https://github.com/Samsung/GearVRf/pull/829.

It also requires PL https://github.com/Samsung/GearVRf/pull/859.

Can be tested using https://github.com/rahul27/GearVRf-Demos/commit/f174ce44d494c6d90b7de22fdd5d6841d0a869ab

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRObjectPicker.java
```diff
@@ -0,0 +1,134 @@
...
+            GVRSceneObject.BoundingVolume bv1 = owner.getBoundingVolume();
+
+            for (int i = 0; i < picked.length; ++i)
+            {
+                GVRPickedObject hit = picked[i];
+
+                if (hit != null)
+                {
+                    GVRSceneObject sceneObj = hit.hitObject;
+                    GVRSceneObject.BoundingVolume bv2 = sceneObj.getBoundingVolume();
```
liaxim: Might be beneficial if getBoundingVolume was updated not to always create a new Java object.

liaxim:
@NolaDonato ?

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRObjectPicker.java
rahul27: I can push the intersect check to native.

rahul27:
@liaxim @NolaDonato I've made the changes.

NolaDonato:
My overall opinion here is that BoundingVolume should be a native class whose operations are all in C++ which we access from Java. The C++ SceneObject should contain one and keep it up to date. When the user requests the bounding volume, they will always get the same pointer back in C++ and Java. We can do this in a separate pull request.

liaxim:
Looks good to me. If @NolaDonato approves then we can merge.

rahul27:
I found an issue with IPickEvents. Lets hold off on this merging this.

liaxim:
Looks good to me. If @NolaDonato approves then we can merge.

NolaDonato:
Yes, merge it.

PR #866, opened 2016-09-23
Separate out a Daydream backend

liaxim:
+ some viewManager refactor


liaxim:
Basic work done; lots of testing left.

We will have three backends; order of loading is determined by the asset/backend_x.txt file where x is the priority. This file gives a factory class name we use to bring up the backend and hook it up.

The monoscopic renderers are still bundled.

Significant refactoring of viewManager but there is more left when it comes to dealing with screenshots, statistics.

liaxim:
Smoke testing showed only a problem with shadow mapping.

liaxim:
The problem with gvr-shadows was caused by another change. Fix merged in master already.

liaxim:
gvr-lodtest doesn't work; doesn't work prior to this change too. Looks to be caused by the recent bounding volume change. Will look into it anyway.

liaxim:
All samples works as well as before. I will experiment with the existing unit tests but this is ready for review already.

lodtest is a separate thing that will have to covered by another pull request.

rahul27:
Will take a look.

GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrSurfaceView.java
```diff
@@ -93,12 +93,6 @@ public OvrSurfaceView(Context context,
     }
 
     @Override
-    public boolean onKeyDown(int keyCode, KeyEvent event) {
```
rahul27: Was this intentionally removed?

GVRf/Framework/backend_oculus/build.gradle
```diff
@@ -136,8 +132,6 @@ android {
 dependencies {
     compile fileTree(include: ['*.jar'], dir: 'libs')
     compile fileTree(include: ['*.jar'], dir: 'src/main/libs')
-    compile(name: 'google-vr-base', ext: 'aar')
-    compile(name: 'google-vr-common', ext: 'aar')
     compile project(':framework')
 }
```
rahul27: Cant comment on the next file, will add it here instead:

Why is the oculus backend being loaded at order 2?

GVRf/Framework/framework/src/main/AndroidManifest.xml
```diff
@@ -31,8 +31,6 @@
         android:minSdkVersion="19"
         android:targetSdkVersion="19" />
 
-    <uses-permission android:name="com.samsung.android.hmt.permission.READ_SETTINGS" />
```
rahul27: Has this been moved to the oculus backend. Don't see it there.

liaxim:
I got caught by .gitignore. Will add the daydream backend files.

GVRf/Framework/framework/src/main/AndroidManifest.xml
liaxim: It was duplicated in this file itself.

GVRf/Framework/backend_oculus/build.gradle
liaxim: Leaving room for another backend to have higher loading priority.

GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrSurfaceView.java
liaxim: Yes. It is going to the daydream backend module.

liaxim:
Ran framework-tests; I don't think they are yet in a state that allows conclusive decisions.

GVRf/Framework/backend_daydream/src/main/java/org/gearvrf/DaydreamViewManager.java
```diff
@@ -28,52 +28,49 @@
...
+        updateSensoredScene();
     }
 
     public void onDrawEye(Eye eye) {
         if (eye.getType() == Eye.Type.LEFT) {
-            OvrMonoscopicRenderer.cull(mMainScene, mMainScene.getMainCameraRig().getCenterCamera
-                    (), mRenderBundle);
-            renderCamera(mActivity.getNative(), mMainScene, mMainScene
-                    .getMainCameraRig().getLeftCamera(), mRenderBundle);
+//            OvrMonoscopicRenderer.cull(mMainScene, mMainScene.getMainCameraRig().getCenterCamera
```
liaxim: Removed commented out code from the daydream viewManager. Culling (and shadow mapping) is done by the base viewManager for all backends.

liaxim:
Monoscopic and Nexus is something we can't handle right now. It is a rather special case since Oculus is not there but the monoscopic renderer is bundled with Oculus.

This needs work of separating out the monoscopic renderer(s). @thomasflynn Should this be handled as part of this change or could it be done in the future?

thomasflynn:
future

liaxim:
Rahul is ok with merging this. Will merge tomorrow morning if there are no further comments.

NolaDonato:
I tested this today with the experimental gradle build and the normal build. It works OK for the Oculus backend. I don't have a Daydream phone so I did not test that option.

PR #871, opened 2016-09-26
DepthShader: specify the correct glsl version

liaxim:


liaxim:
gvr-shadows doesn't work correctly otherwise

PR #872, opened 2016-09-26
LOD fixes

liaxim:
- BoundingVolume::transform: without the expand calls the center and
the radius would not be recalculated
- SceneObject::frustumCull: crude workaround to force lod check for
all children scene objects; disables optimization but only in case
any scene object uses lod

### DIFF
--- a/GVRf/Framework/framework/src/main/jni/objects/bounding_volume.cpp
+++ b/GVRf/Framework/framework/src/main/jni/objects/bounding_volume.cpp
@@ -61,11 +61,16 @@ void BoundingVolume::expand(const glm::vec3 point) {
         max_corner_[2] = point[2];
     }
 
+    updateCenterAndRadius();
+}
+
+void BoundingVolume::updateCenterAndRadius() {
     center_ = (min_corner_ + max_corner_) * 0.5f;
-    if (min_corner_ == max_corner_)
-    	radius_ = 0;
-    else
-    	radius_ = glm::length(max_corner_ - min_corner_) * 0.5f;
+    if (min_corner_ == max_corner_) {
+        radius_ = 0;
+    } else {
+        radius_ = glm::length(max_corner_ - min_corner_) * 0.5f;
+    }
 }
 
 /*
@@ -182,9 +187,8 @@ void BoundingVolume::transform(const BoundingVolume &in_volume,
             max_corner_.z += a;
         }
     }
-    // expand with the new corners
-    expand(min_corner_);
-    expand(max_corner_);
+
+    updateCenterAndRadius();
 }
 
 bool BoundingVolume::intersect(glm::vec3& hitPoint, const glm::vec3& rayStart, const glm::vec3& rayDir)  const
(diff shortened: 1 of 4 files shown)


liaxim:
Have not checked the other samples yet for side effects.

GVRf/Framework/framework/src/main/jni/objects/bounding_volume.cpp
```diff
@@ -182,6 +182,9 @@ void BoundingVolume::transform(const BoundingVolume &in_volume,
             max_corner_.z += a;
         }
     }
+    // expand with the new corners
+    expand(min_corner_);
```
rahul27: why are these two lines required? 

min_corner_ and max_corner_ are set by this function itself.

liaxim: Matches the side-effects of the old implementation.

liaxim:
For whatever it is worth: org.gearvrf.tester.PickerTests#canPickWithFrustum passes with this change but not with master.

Two other tests fail no matter what:
org.gearvrf.tester.PickerTests#canPickObjectWithRay
org.gearvrf.tester.PickerTests#testPickObjects

GVRf/Framework/framework/src/main/jni/objects/bounding_volume.cpp
rahul27: I see. I assumed the bounding radius was not used.

In any case you could add the following lines in place of expand(), will eliminate the need for the 6 if conditionals in the expand method

```
    center_ = (min_corner_ + max_corner_) * 0.5f;
    if (min_corner_ == max_corner_)
    	radius_ = 0;
    else
    	radius_ = glm::length(max_corner_ - min_corner_) * 0.5f;
```

liaxim: Expand calls eliminated

rahul27:
:+1: LGTM

liaxim:
@NolaDonato Any thoughts?

liaxim:
Will merge Thursday morning if there are no further comments.

GVRf/Framework/framework/src/main/jni/objects/scene_object.h
```diff
@@ -180,7 +180,7 @@ class SceneObject: public HybridObject {
     std::vector<SceneObject*> children_;
     float lod_min_range_;
     float lod_max_range_;
-    bool using_lod_;
+    static bool using_lod_;
```
thomasflynn: just curious, why was changing this to static necessary?

liaxim: Previously, if any object uses lod but the parent is completely inside the frustum the lod check would not be performed because as an optimization the children would not be further evaluated.

I didn't want to overdo this pull request hence the crude solution where if any object uses lod then this global flag disables the aforementioned optimization to ensure the lod check is performed always.

The lod-test has been broken for quite some time (few months) now that I understand what exactly it is doing.

thomasflynn: got it

PR #873, opened 2016-09-28
Experimental gradle plugin support

liaxim:
Experimental gradle plugin support:
- framework module converted
- oculus backend converted
- one sample converted

Import GVRf/tools/Framework; C++ debugging just works


NolaDonato:
I LOVE this pull request. It enabled C++ debugging in Android Studio for me. Great work!

liaxim:
There is a problem referring to libgvrf.so. First a build needs to be made with the regular build files and only then this would work. Both builds have to be off the very same source tree.

liaxim:
- Open tools/Framework instead
- Referring to libgvrf.so from the oculus backend Android.mk is problematic; works but it is a hack

- Not seeing the app getting paused for no reason whether it is just being run or being debugged; @NolaDonato you will have to show me what happens on your phone.

NolaDonato:
You are right. I pulled it over again and tested it and this time there is no onPause event. Not sure what was going on there.

liaxim:
Will merge tomorrow morning if there are no comments. @NolaDonato You might consider creating a pull request for your tests changes.

NolaDonato:
this latest update enabled C++ debugging for me in the gvr-cubemap example. before the last change I wasn't able to set breakpoints reliably.

PR #876, opened 2016-10-03
Remove GVRScript; PostEffectShader changes to support removal of GVRScript

liaxim:
Demos pull request: https://github.com/gearvrf/GearVRf-Demos/pull/339


liaxim:
Ready for review

thomasflynn:
this looks like more than just the PostEffectShader changes.  it also has all the code to remove GVRScript.  was that intended?

liaxim:
Yes, intended. Should I split the change in two?

thomasflynn:
I think so.  I'd be fine with removing GVRScript in the next release.  For now, I want to keep things stable until we get a few more bugs fixed and then do a release.

liaxim:
The change to the PostEffectShader can wait too I think. We could have it in after the release. It is a prerequisite to the removal of GVRScript more than anything.

NolaDonato:
My patch that removes C++ builtin shaders makes some changes to the PostEffectShader stuff too.

liaxim:
@NolaDonato If you patch for removing the builtin shaders will come soon enough then this pull request will wait for it.

NolaDonato:
Don’t wait. They are now merging my patch with uniform buffers. This will take some time.

liaxim:
This is ready for review, testing and merging eventually.

No problems with the samples have been found. Works equally well for all backends we support.

All the automated framework tests that do not involve image comparison pass.

liaxim:
@rahul27 Do you think you could have a look at this change? Thanks.

rahul27:
@liaxim looking at it now.

rahul27:
@liaxim are you able to build the extensions using command line?

liaxim:
Yes, the extensions build for me from the command line.

rahul27:
Do you use `./gradlew assembleDebug` ?

I get `Observed package id 'ndk-bundle' in inconsistent location `.

liaxim:
Yes. This is the one. There is something off with your local Android SDK installation. Usually it is not fatal.. the build machine has been getting these forever.

rahul27:
gvr-jassimp and meshanimation crash on launch for me with this change.

liaxim:
Do you have stacktraces?

liaxim:
Tried on Note5 and S7 - worked for me. ``git clean -fdx`` might be advisable.

rahul27:
Fixed the command line build for the extensions. I had the wrong path to ANDROID_HOME in my bash file. It now builds correctly.

Reg. gvr-jassimp it still crashes. I don't see a visible crash trace on logcat.

rahul27:
@liaxim gvr-jassimp works when the wifi is on but crashes when it is off. 

This does not happen on master.

liaxim:
gvr-jassimp: made a small tweak to the app itself.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRActivity.java
```diff
@@ -240,10 +240,10 @@ protected void onDestroy() {
     }
 
     /**
-     * Links {@linkplain GVRScript a script} to the activity; sets the version;
+     * Links {@linkplain GVRMain a script} to the activity; sets the version;
```
liaxim: Good point. Will make the changes.

liaxim:
Will merge tomorrow after lunch

PR #880, opened 2016-10-06
Add physics support

ragner:
Integration of a Physics simulation component into the GearVRf framework in order to add realistic movement and interaction between 3D scene objects.

Physics bullet third party engine is used to provide support to this solution.

 * We are still working on collision events and Picker based on physics engine.
* Check bullet demo at https://github.com/ragner/GearVRf-Demos/tree/bullet_support

GearVRf-DCO-1.0-Signed-off-by: Frederico Duarte <frederico.sd@samsung.com>
GearVRf-DCO-1.0-Signed-off-by: JulianaFigueira <juliana.f@samsung.com>


liaxim:
First question, mostly for @thomasflynn. Shouldn't this be an extension instead of a part of the core framework? It is rather large and not necessarily needed by all users.

ragner:
@liaxim yes, we need do it as extension but we need your help about it :).

liaxim:
If we want this to become an extension we can't have the framework reference anything from it. Luckily this looks to require very little work.

GVRScene: no references to GVRWorld, just expose attach/detachComponent from the rootSceneObject as methods.
GVRSceneObject: similar situation; we should rely only on the component model, no references to rigidBody, world, etc.

Let me know if you want me to prototype something.

ragner:
@liaxim I got it. Could I make it as a new commit on top?

At GVRScene I have made a setPhysicsEnabled(..) function to make easier to enable or disable the physics. Should I just remove it or do you have other suggestion about it?


'
public void setPhysicsEnabled(boolean enabled) {
        if (enabled == getPhysicsEnabled()) {
            return;
        }

        if (enabled) {
            mSceneRoot.attachComponent(new GVRWorld(getGVRContext()));
        } else {
            mSceneRoot.detachComponent(GVRWorld.getComponentType());
        }
    }'

liaxim:
New commit would be just fine.

For now please remove setPhysicsEnabled. We could try to come up with some helpers like this after the code is moved to be an extension. The helpers have to reside in the extension.

ragner:
Hi @liaxim, could you check my last commit?

liaxim:
You can cherry-pick https://github.com/liaxim/GearVRf/commit/313359dc536eeb5c99fa6d51d0962ab05c334922 from my repo to get the changes to make this an extension module.

``git fetch git@github.com:liaxim/GearVRf.git bullet_physics&&git cherry-pick 313359dc536eeb5c99fa6d51d0962ab05c334922``

Import GVRf/Extensions to get all the extensions, including gvrf-physics.

NolaDonato:
I agree setPhysicsEnabled is not needed. You can call setEnable on the GVRWorld to turn physics on and off.

NolaDonato:
All components have a setEnable function to enable and disable them. If a GVRRigidBody is disabled, it should not continue to participate in physics. A rigid body should only react if it is enabled and its ancestor has a physics world.

NolaDonato:
if GVRBoxCollider is attached to a scene object that has a mesh, could we have it calculate the bounds of the box from the mesh bounding volume? This might make it easier for the user.

NolaDonato:
Can we expose a GVRHullCollider object in the physics package to Java. This component would automatically compute the convex hull from the mesh attached to the scene object. I notice you already have the code for this in C++.

ragner:
Hi @liaxim,
After your commit I am getting:
/Users/ragner/Workspace/Devtools/android/ndk/build/core/prebuilt-library.mk
Error:(45) *** Android NDK: Aborting    .  Stop.
Android NDK: Check that jni/prebuilt/armeabi-v7a/libBullet.so exists  or that its path is correct   
make: Entering directory `/Users/ragner/Workspace/SIDIA/GearVRf/GearVRf/GVRf/Extensions/gvrf-physics/src/main'
make: Leaving directory `/Users/ragner/Workspace/SIDIA/GearVRf/GearVRf/GVRf/Extensions/gvrf-physics/src/main'
:gvrf-physics:buildNative FAILED
Error:Execution failed for task ':gvrf-physics:buildNative'.
> Process 'command '/Users/ragner/Workspace/Devtools/android/ndk/ndk-build'' finished with non-zero exit value 2

Looks that libBullet.so was removed. May I add it back?

liaxim:
.so-s are gitignored so I didn't notice. Just add it back to GVRf/Extensions/gvrf-physics/src/main/libs/armeabi-v7a/.

ragner:
Hi @liaxim, I am still getting that error even after add it back to GVRf/Extensions/gvrf-physics/src/main/libs/armeabi-v7a/

in Android.mk there is the following line:
LOCAL_SRC_FILES := prebuilt/$(TARGET_ARCH_ABI)/libBullet.so

liaxim:
Yeah. The right place for the .so is GVRf/Extensions/gvrf-physics/src/main/prebuilt/armeabi-v7a/.

ragner:
@liaxim, worked! My last commit puts back the shared library after your commit
and looks that physics is working as extension now.

ragner:
Hi @NolaDonato,
I will work on your comments and give your some feedback. 
Thank you.

ragner:
Hi @NolaDonato,
I think that if we remove a SceneObject with RigidBody of the Scene 
it should be removed of the physics world too.

Do you have any idea how could I implement it? I mean that the RigidBody should be notified that its owner was removed of the scene.

NolaDonato:
The issue we are facing is that a component or a scene object doesn't get informed when it is added or removed from the scene. GVRSceneObject.removeChildObject should notify all the scene object components that the scene object has been removed from the scene. All we have to do is make a new function GVRComponent.onRemove and call it from removeChildObject for all the components. This will make it easy to determine when to remove the RigidBody from the physics World.

Knowing when a component been added to the scene is much trickier. GVRSceneObject.addChildObject has no information on whether the parent object is in a scene or not. onInit only knows about the scene objects added before it returns. Nothing tracks the scene objects added afterwards. I think we need some kind of notification telling a GVRSceneObject and its components that they have been added to a scene.I am not sure how to efficiently implement this however. Only the C++ side traverses the scene graph every frame so the notification would probably have to come from there.

The issue with adding the RigidBody in onAttach is that you can attach a component to a scene object that is not in any scene.

ragner:
Hi all,
I am facing a small issue.
The SceneObject should be removed from the physics world when it is removed from the Scene.
Does anyone have a suggestion how to fix it?

ragner:
Hi all,
I implemented a logic to enable/disable the physics of a SceneObject
at my last commit "RigidBody just reacts to physics if it it is enabled" and also I did some fix of code style.

It is still missing a logic to disable the physics when the SceneObject is removed from the Scene.

ragner:
Hi @NolaDonato,
about GVRHullCollider.
So, what should be the difference between GVRMeshCollider and GVRHullCollider?
I am asking just because we are converting GVRMeshCollider to btConvexHullShape internally to improve the performance. 
Should we convert GVRMeshCollider to concave mesh like btBvhTriangleMeshShape?

Thank you.

NolaDonato:
If you change GVRMeshCollider to a convex hull, will it still work without the physics extension?

NolaDonato:
pull request #899 removes the rigid body from the physics world when its owner is removed from the scene.

ragner:
Hi @NolaDonato,
Without the physics extension GVRMeshCollider may represent a concave mesh. 
I think we could add GVRHullCollider at some new pull request.

I just wold like to fix the main issues of this pull request to make it ready to merge,
if you agree.

JulianaFigueira:
Hi, I just pushed two commits, one adding collision events and another to apply some code style fixes

liaxim:
@ragner Not sure how onOwnersParentChanged would work. Maybe you could prototype something that we can try.

@NolaDonato Why not start with a simple implementation for the notification when a scene object is added to the active scene so we can unblock this pull request?

- since the scene object has a reference to its parent then it is straightforward to reach the root of the hierarchy and check whether it matches the mainScene's or the nextMainScene's root then issue notifications to the added scene object and its children.

- if the scene is not the mainScene or the nextMainScene and is GVRContext.setMainScene-ed we have to pay one time cost for GVRScene.getWholeSceneObjects and issue the notifications to the flattened list.

ragner:
Hi @liaxim,
please check my last two commits about notification of parent changes.

You need update ragner:bullet_support from Demo to test it.
https://github.com/ragner/GearVRf-Demos/tree/bullet_support

I think that now it looks better.

NolaDonato:
I do not love this approach because it adds extra overhead every time you change the hierarchy. However I think you should go with it for this release because I don't want to introduce something that might be destabilizing until after the release.

I believe the long term answer here is to track on the C++ side whether or  not the scene object has been visited. The first time a scene object is visited the C++ renderer can start a task on the Java thread to call listeners for that event. I am not even sure how to do this yet and I have to consult Mihail on it but I think having the general ability for the C++ side to issue events to the Java side would solve a lot of issues. What I don't want is for the C++ thread to block while some expensive Java function is called. I would like the ability to schedule Java work to be done on the Java thread - not for the Java function to be called from within the renderer.

liaxim:
I don't think it would cover the case where a new GVRScene is created that is not the main scene and the scene objects are added to it. Then some time later this scene is made the main one.

But this version would do for now. Didn't find any side-effects caused by this change. If Nola is ok with merging, then we will go ahead and merge.

NolaDonato:
Yes, lets merge this

NolaDonato:
Doesnt getNextMainScene make a new scene?



liaxim:
getNextMainScene creates a new one, yes. But it becomes the main one as soon as the splash screen is gone so it is all good.

You could create your own GVRScene though (few sample apps do that), add scene objects to it, keep a reference then decide some time later it should be set as the main one. What happens with the rigid bodies added to the previous main scene, what happens with the rigid bodies added to a scene that is not the main one, etc. We will look into it in the future.

PR #884, opened 2016-10-10
Fix invalidateFrameBuffer() for multisample

dnvasilev:
After multisample resolve, invalidateFrameBuffer() was discarding resolved color buffer instead of original multisample framebuffer.


roshanch:
good catch.!!

PR #885, opened 2016-10-10
Replace GlDelete with a more generic Context entity that

liaxim:
doesn't have the same threading restrictions
- minor change in how the renderer is instantiated


NolaDonato:
A big improvement over the GlDeleter thing. It makes more sense for the C++ objects to be associated with a context and have that context own them than assigning a deleter later.

liaxim:
Superseded to large extent by https://github.com/Samsung/GearVRf/pull/969 which is also much smaller in scope.

PR #887, opened 2016-10-11
Allow including extra modules from possibly another repos without

liaxim:
having to update the master settings.gradle.

For example, to include a backend xyz create extra_settings.gradle in
the directory where the GearVRf/ directory is, with the following contents:

include ':backend_xyz'
project(':backend_xyz').projectDir=new File('../../../SomeOtherRepo/GVRf/Framework/backend_xyz')


liaxim:
Updated to work with the experimental-gradle based project.

liaxim:
Will merge after lunch if there are no objections.

PR #888, opened 2016-10-11
Experimental gradle: framework-test tweaks

liaxim:
Changes so the unit tests can be run.


liaxim:
Will merge after lunch if there are no objections.

PR #889, opened 2016-10-12
Adding WearTouchPad extension

parthmehta209:
WearTouchPad is a GearVRf extension. It is an app that should be installed on the phone.
This app has a companion Android wear app. The mobile app gets all the touchpad events from
the companion Android wear app. This extension is used by the GVRAndroidWearTouchpad for getting touchpad events from the Android wear watch and delivering it to the GVRActivity.
The GVRf/Extensions/WearTouchPad/WearTouchPad.apk apk file is a combined apk file that will install the app on the phone as well as the connected Android wear watch. This is included for easily installing the app on the watch and the phone.


thomasflynn:
fwiw, it turns out i needed to upgrade to java 8 to compile.  now to actually try it...

thomasflynn:
ok, good, i can confirm this is functional now.  +1, looks good to me.

GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrXMLParser.java
```diff
@@ -106,6 +106,10 @@
                                 settings.setUseGazeCursorController(Boolean
                                         .parseBoolean(xpp.getAttributeValue(i)));
                             } else if (attributeName
+                                    .equals("useAndroidWearTouchpad")) {
```
liaxim: You will need to update the parsers for the other backends to recognize the new setting. Clearly we need a common parser but for now they are separate.

GVRf/Framework/framework/src/main/java/org/gearvrf/io/GVRAndroidWearTouchpad.java
```diff
@@ -0,0 +1,134 @@
+package org.gearvrf.io;
```
liaxim: Needs copyright note.

liaxim:
Do you want something integrated with the automated build? Any publishable build artifacts?

--

It is not really an extension in the usual terms. If it were a real extension (like a separate .aar) it would be the best if the feature gets automatically enabled as long as the aar is included in the app. No need for extra setting in the xml file. Should we try to make it a real extension? What do you think?

I know there will be a bit of work involved since we haven't had that type of extension before.

GVRf/Extensions/WearTouchPad/mobile/build.gradle
```diff
@@ -0,0 +1,45 @@
...
+        release {
+            minifyEnabled false
+            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+        }
+    }
+}
+
+dependencies {
+    compile fileTree(dir: 'libs', include: ['*.jar'])
+    wearApp project(':wear')
```
liaxim: What does ``wearApp project(':wear')`` do?

GVRf/Framework/framework/src/main/java/org/gearvrf/io/GVRAndroidWearTouchpad.java
parthmehta209: Done.

GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrXMLParser.java
parthmehta209: Yes. I will make the change in other backends.

parthmehta209:
@Liahim I am not sure how we can make an aar. The reason I made a separate app is that each app that has a companion wear app has to have the same application ID as the wear app. As far as I know we cannot have a wearable app inside an aar. But I can investigate further into making it more like a regular extension. I have added the .apk file in this checkin. So I we would not need a build artifact for this extension. Or we can have the automated build create a signed apk and have it as one of the artifacts uploaded to github.

liaxim:
@parthmehta209 and I had a chat. We are not going to make big changes for this pull request. Will look into the extension thing in the future.

GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrXMLParser.java
parthmehta209: added the check to the day dream parser and verified by running the gvr-video sample on a day dream phone.

liaxim:
Will merge Tuesday after lunch unless there are comments.

PR #890, opened 2016-10-12
Adding support for embedded textures

NolaDonato:
FBX files may have texture data embedded in them. ASSIMP supports this but GearVRF did not. This patch adds support for embedded textures. It also fixes missing textures by supplying a small 16x16 white texture in place of textures that do not load.


liaxim:
1. Now the texture cache is per AssetLoader instance and we can (and encourage to?) have more than one instance of it. Isn't this a bit of a problem?
2. Why do we have to expose public cache APIs? What is the benefit to end users? I believe the cache should be transparent and people do not and should not do anything with it directly.

NolaDonato:
I have hidden the texture cache. To do this I needed to move all the texture load routines from GVRContext to GVRAssetLoader and move GVRJassimpAdapter into org.gearvrf.

This is one small step in reforming texture load. I would like to eventually get rid of future textures but not before the next release. I would like to get rid of all the asset loading functions from GVRContext but first we have to change all the samples...

liaxim:
Sounds like a good change. We probably will have to move a lot more classes (async, x3d) for the very same reason.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
```diff
@@ -407,33 +430,281 @@ public void loaded(GVRTexture texture, GVRAndroidResource ignored)
     }
 
     protected GVRContext mContext;
-    public GVRAssetLoader(GVRContext context) {
+    protected static final ResourceCache<GVRTexture> mTextureCache = new ResourceCache<GVRTexture>();
```
liaxim: Please look at the use of GVRContext.addResetOnRestartHandler to recreate the instance. Ideally we would not have static state but since we do, then you need to reset the instance.

The problem is that once the app is backgrounded, the process usually is not killed. If the app is started again the cache will stick around but there will be a new gl context created - the cache contents will be invalid and there will be cache hits.

GVRf/tools/Framework/gradle.properties
```diff
@@ -16,7 +16,7 @@
 org.gradle.jvmargs=-Xmx2048M
 #ANDROID_NDK_HOME=
 # Un-comment and add the path to ovr directory
-#OVR_MOBILE_SDK=
+OVR_MOBILE_SDK=d:/projects/android/GearVRf/ovr_sdk_mobile
```
liaxim: Please revert this change.

liaxim: You could define properties in the global gradle.properties; under Linux, modify ~/.gradle/gradle.properties.

liaxim:
Where do we have models with embedded textures?

liaxim:
The 3DCursorLibrary needs couple of minor changes to compile.

liaxim:
Good number of samples require changes to compile.

NolaDonato:
I will make a pull request in GearVRf-Tests from this branch https://github.com/NolaDonato/GearVRf-Tests/tree/fbxtests for the embedded texture test. I made a specific test for it there.

NolaDonato:
The 3D cursor library did not compile because I removed the "throws IOException" from the GVRAncroidResource constructor. It might once have thrown an exception but it does not now. I will put the throws clause back for compatibility. We should remove it in the next version.

NolaDonato:
GVRContext.loadMesh SHOULD throw an IOException when the file is not there. GearVRf was dropping the IOException on the floor and ignoring it. For backwards compatibility I will compromise - loadMesh will catch the exception, send the onModelError event and return null for the mesh.

NolaDonato:
I believe I have responded to all the feedback. I have made GVRAndroidResource and GVRContext.loadMesh backwards compatible again so the samples should all compile (I checked them). I added the GVRContext.addResetOnRestartHandler to clear the cache on restart.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
```diff
@@ -407,33 +430,296 @@ public void loaded(GVRTexture texture, GVRAndroidResource ignored)
...
+     * since all of the GL textures have been deleted.
+     */
+    static
+    {
+        GVRContext.addResetOnRestartHandler(new Runnable() {
+
+            @Override
+            public void run() {
+                mTextureCache = new ResourceCache<GVRTexture>();
+            }
```
liaxim: I think mDefaultTexture needs to set to null for the same reason.

liaxim:
gvr-cubemap crashes; doesn't without this pull request. There isn't much info:

```
10-18 13:43:33.977  3135  3135 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
10-18 13:43:33.977  3135  3135 F DEBUG   : Build fingerprint: 'samsung/heroltexx/herolte:6.0.1/MMB29K/G930FXXU1APB3:user/release-keys'
10-18 13:43:33.977  3135  3135 F DEBUG   : Revision: '6'
10-18 13:43:33.977  3135  3135 F DEBUG   : ABI: 'arm'
10-18 13:43:33.977  3135  3135 F DEBUG   : pid: 11282, tid: 11327, name: GLThread 78248  >>> org.gearvrf.cubemap <<<
10-18 13:43:33.977  3135  3135 F DEBUG   : signal 11 (SIGSEGV), code 2 (SEGV_ACCERR), fault addr 0xd32d0000
10-18 13:43:34.007  3135  3135 F DEBUG   :     r0 d0504bc4  r1 d32d0000  r2 00000400  r3 d0504800
10-18 13:43:34.007  3135  3135 F DEBUG   :     r4 d0504fc4  r5 d32d0400  r6 0000000c  r7 d0505000
10-18 13:43:34.007  3135  3135 F DEBUG   :     r8 000007d0  r9 00000800  sl 00000000  fp 00000001
10-18 13:43:34.007  3135  3135 F DEBUG   :     ip 80000000  sp dd6677f0  lr e797fc74  pc e7668454  cpsr 200f0010
10-18 13:43:34.007  3135  3135 F DEBUG   : 
10-18 13:43:34.007  3135  3135 F DEBUG   : backtrace:
10-18 13:43:34.007  3135  3135 F DEBUG   :     #00 pc 00259454  /system/vendor/lib/egl/libGLES_mali.so
```

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
NolaDonato: will do

NolaDonato:
the cubemap sample works for me now. I was calling the wrong texture load routine for compressed cubemaps.

liaxim:
gvr-pickandmove doesn't work. The sphere is not visible.

NolaDonato:
gvr-pickandmove is fixed now

liaxim:

Issue #886, opened 2016-10-11
Some objects are not drawn


lassehe:
My application stopped drawing some objects (or their textures) after commit d02474726d39cef500310006ebe36540389ef210 "Fix the matrix transformation of the bounding volume" was merged. Depending on the direction looked at all or none of the objects are missing. Later commits 979bf2df9f13fae870146d461c5a01d2c06d90ed and 8513f039e9f2a814fde3ba09bdb0d4cb8f933cf5 related to this do not fix the issue.

Any ideas what is causing this? I am not sure if there is a problem in this change, or had I been doing something wrong before.

rahul27:
The fix corrects the bounds of the bounding volume for a given object. Could you share an example of what you are seeing?

lassehe:
Here are screenshots made using GVRf version 3de46c46aa0cb43e71dae5b2aae9ee50f2d5ae80  where Mihail's patches related to this were also merged. A complete sphere surface, assembled from smaller tiles, should be present here, but now some GVRSceneObjects are not drawn, based on the exact direction which is being looked at. For example, when looking directly towards top or bottom this looks just fine.

I came to conclusion my issue has something to do with that bounding volume fix by using git bisect. Of course it is possible that the problem is in my code, but earlier this worked fine.

![screenshot1](https://cloud.githubusercontent.com/assets/15195342/19300317/511a498e-9062-11e6-8d4a-f22f9384a78b.jpg)
![screenshot2](https://cloud.githubusercontent.com/assets/15195342/19300319/51201eea-9062-11e6-871e-d1da758c548c.jpg)

rahul27:
@lassehe Are you using the GVRPicker to determine picked objects? 

Could you let me know the exact GVRf calls that fail? Also could you share a sample code (if possible) so that I can recreate and track down the issue faster?

Finally, when you say it could be a problem with your code are you doing something different than the Picker?

rahul27:
Hi @lassehe, just following up on my question above.

lassehe:
Thanks for your help! Perhaps I'm missing something here, but why would GVRPicker affect rendering? All these GVRSceneObjects should be visible. In screenshots the camera is in the center of the sphere. The app used GVRPicker for certain purposes, but removing it didn't fix this drawing issue.

However, I probably did ran into that GVRPicker bug earlier. Now I tested editing head model params in the configuration XML, but didn't notice any difference with this rendering.

[...]

PR #891, opened 2016-10-12
Fixed a bug found by the picking tests

rahul27:
pickSceneObject() used the view matrix instead of the
model matrix, causing the wrong distance to be reported in
the tests.

r.rudradevan@samsung.com


rahul27:
@NolaDonato here are the corrected tests for picking:

https://github.com/gearvrf/GearVRf-Tests/pull/17

liaxim:
Could this also fix issue https://github.com/Samsung/GearVRf/issues/886?

rahul27:
I am not sure if the issue is caused by the picker, I can follow up to find the exact issue.

NolaDonato:
The change you made looks good and makes sense.

PR #892, opened 2016-10-12
Daydream: fix for gvr-shadows showing black screen

liaxim:
daydream doesn't like changing the currently bound fbo on it


liaxim:
@rahul27 I also disabled the daydream intro screen (setTransitionViewEnabled(false)). Do you think this would be a problem? The screen gets really annoying after a while..

rahul27:
I always assumed that this was the call to prompt the user to place the phone into the headset. Somewhat like the screen shown to the user with the Oculus SDK.

We can add it back if it is required for Daydream apps.

liaxim:
Yes, it is that prompt. But it sits there forever unless you start moving the phone..

liaxim:
Will merge tomorrow after lunch if there are no comments.

PR #895, opened 2016-10-14
Daydream: fix crash in gvr-widgetviewer and gvr-modelviewer2

liaxim:


liaxim:
Will merge Tuesday morning unless there are comments.

PR #896, opened 2016-10-14
Daydream: fix for the vuforia sample; looks like daydream

liaxim:
enables gl_blend at some point

- we had problems in the past with this and at the end of
renderCamera we do disable it; still with daydream we get
called back with it enabled
- vuforia with this backend (and others) has a problem due
to the viewport configuration being supported only in the
oculus backend; still the sample is usable


liaxim:
Will merge Tuesday morning unless there are comments.

PR #899, opened 2016-10-18
Remove rigid body from physics world

NolaDonato:
Add GVRComponent.onRemoveFromScene which is called when the component's
owner is removed from the scene. GVRRigidBody overrides it to remove the
object from the physics world.

This pull request requires PR #880.


ragner:
Hi @NolaDonato,
I think we could make some GVRComponent.onOwnersParentChanged(GVRSceneObject parent).

So, I could implement a better logic to add/remove the RigidBody from the world.

I can follow your thoughts from this patch and implement that GVRComponent.onOwnersParentChanged().

ragner:
I mean, when a SceneObject is added to the Scene we could  just àddToWorld()` function
and when it is removed just call `removeFromWorld()` function.

The attach/detach logic keeps just to onAttach/onDetach of the Component.

liaxim:
Should we close this one?

ragner:
Yes,
It already is in  #880

NolaDonato:
yes

PR #900, opened 2016-10-18
GVRConsole tweaks

liaxim:
- increase default text size
- update the fragment shader to mix so the overlay texture is
more of an overlay


liaxim:
Have to update couple of samples due to the different default text size. Left-only, right-only worked fine for me.

thomasflynn:
the stats look like they need their fudge factor adjusted (so they are visible while inside the headset), but otherwise looks good.

liaxim:
Adjusted the console x and y offsets used for the scene stats.

liaxim:
Using the offsets @thomasflynn suggested

thomasflynn:
ok, looks good.

PR #901, opened 2016-10-19
Introduce LOD component

liaxim:
LOD demo app: https://github.com/gearvrf/GearVRf-Tests/pull/41


liaxim:
@mwitchwilliams Please let me know what LOD tests do you have for x3d so I can double-check the changes I made to x3d. Thanks.

mwitchwilliams:
Can I get this to you tomorrow?
This sounds like an opportunity for me to assemble my LOD demo’s and include documentation and post it on the GVR-Tests repository.

Mitch

mwitchwilliams:
I can also send a quick file if you need one right away.

To post to the repository, I have to create some new, non-proprietary, textures.

Mitch

mwitchwilliams:
Here’s a good quick fast demo.

The text and color changes.  It animates between LOD boundaries, pauses, go past the boundary as it changes, pauses again, and then animates toward the next boundary.

There are 3 LOD’s here.
The Caltrain in the lower left is just a constant animation and not related to LOD.  I throw these in there to help detect any new errors.

One other issue of this demo – it uses X3D’s <Inline> command where it can specify an external file and embed it into this X3D scene.  So, it’s a good test of <inline> and why there are so many x3d files.

Mitch

liaxim:
@mwitchwilliams Tomorrow is fine. We will merge this after the release so there is plenty of time.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRLodGroup.java
```diff
@@ -0,0 +1,89 @@
...
+
+            mVector.sub(mCenter);
+            mVector.negate();
+
+            float distance = mVector.dot(mVector);
+
+            if (distance >= lodRange.getLODMinRange() && distance < lodRange.getLODMaxRange()) {
+                child.setEnable(true);
+            } else {
+                child.setEnable(false);
```
thomasflynn: so, question:  if there is a hierarchy under an LOD that is not in range, does that hierarchy still get traversed during scene traversal?

liaxim:
Found out that enabling the scene stats stops lod from working. Very odd.

mwitchwilliams:
Just did pull request in the GVR-Tests repository with a variety of X3D LOD (Level-of-Detail) tests.
Noticed the above conflict.  I don't have any outstanding pull requests for GearVR framework.  Is there something I need to do?

liaxim:
@mwitchwilliams Thank you for the tests. I rebased the pull request so there is no more conflict.

GVRf/Framework/framework/src/main/jni/objects/scene_object.cpp
```diff
@@ -445,31 +442,7 @@ int SceneObject::frustumCull(glm::vec3 camera_position, const float frustum[6][4
...
-    glm::vec4 transformed_sphere_center(bounding_volume_.center(), 1.0f);
-
-    // Calculate distance from camera
-  //  glm::vec3 camera_position = camera->owner_object()->transform()->position();
-    glm::vec4 position(camera_position, 1.0f);
-    glm::vec4 difference = transformed_sphere_center - position;
-    float distance = glm::dot(difference, difference);
-
-    // this distance will be used when sorting transparent objects
-    render_data()->set_camera_distance(distance);
```
liaxim: Just realized that I should not have removed setting the camera distance. It is used by the state_sort.

NolaDonato:
You could simplify this component by getting rid of GVRLodRange and just putting an array of floats into GVRLodGroup. That would avoid the issue of deciding what to display when the camera distance is between two LODs but the user did not make the max of one equal the min of the next.

If you just keep an array of floats in ascending order it makes the API simpler and makes it impossible to have a "dead zone" between two levels. If the distance from the camera is smaller than the smallest range, nothing should be displayed. If there are more ranges than children, nothing should be displayed outside the last range.

GVRLodRange.addRange(float v) adds a range, keeping the array sorted in ascending order
float[] GVRLodRange.getRanges() returns the ranges

In a future release we should consider blending between the levels to avoid popping. The implementation you have now will be good for foveated rendering where popping isn't as big an issue.

liaxim:
Thanks for the suggestions. Will look into implementing them.

liaxim:
@mwitchwilliams With or without this change I see the same behavior when using levelofdetail02.x3d. Behavior that doesn't match the description in Level-of-Detail demonstrations.docx. Could you please check whether everything works as expected in master when it comes to lod and x3d? Thanks.

mwitchwilliams:
I haven’t done a fetch in about two weeks, but my initial test shows this works as expected / as documented.
The geometry remains a flat panel but each level-of-detail has a different texture map.  There are also planned pauses in the animation as it passes marker 1 and marker 2.

There is another black and yellow diagonal animated in the scene but it has no relevance to level-of-detail.

Let me do a pull on my master branch and see f the results are the same.

Mitch

mwitchwilliams:
OK,

I did a “git fetch master” in my master branch, build and retried the demo and got the same correct and expected results.
Is the documentation unclear and thus needs some work?

Mitch

liaxim:
I probably should read more carefully the document.  The three screenshots do not match what I see. This is what I get:

screenshot_20161222-124859

mwitchwilliams:
The images didn’t send, or were just white.
It’s possible I didn’t fetch the latest.  I received from https://github.com/Samsung/GearVRf.git

liaxim:
I was referring to the screenshots in your document. There are three of them and the lod quad is missing from my screenshot.

mwitchwilliams:
I’m just wrapping up here, but maybe when I return after New Years, I can take a look at it.

But it all looked as expected on my end.

thomasflynn:
@mwitchwilliams, it's after New Years :).  Did you want to take another look?

mwitchwilliams:
OK, my mistake.  There were 3 texture maps - lod1.png, lod2.png and lod3.png - that were not included in the original pull requests.  This should correct the problem  Thanks for catching this.

liaxim:
Verified that the X3D tests match the test documentation. levelofdetailusedef03.x3d is missing from the tests repo so I didn't check that one.

Also added toggling of scene stats to the gvr-lodtest app. Having scene stats has no adverse effects (have mentioned an issue some time ago).

From my perspective this is it for this pull request.

mwitchwilliams:
Good catch and testing.

I did a pull request on the documentation to refer to “text-lod-demo.x3d” instead of “levelofdetailusedef03.x3d”, the last test listed in the documentation, a typo on my part.

I re-tested the file (after a git fetch) and it worked as documented.

Mitch

Mitch Williams
Samsung GearVR X3D development
310-809-4836 (outside line)

liaxim:
Will merge after lunch.

PR #903, opened 2016-10-21
Fix for gvr-transparencytest

liaxim:
Tests pull request:
https://github.com/gearvrf/GearVRf-Tests/pull/27
Demos pull request:
https://github.com/gearvrf/GearVRf-Demos/pull/357
https://github.com/gearvrf/GearVRf-Demos/pull/351

### DIFF
--- a/GVRf/Framework/framework/src/main/jni/objects/components/render_data.cpp
+++ b/GVRf/Framework/framework/src/main/jni/objects/components/render_data.cpp
@@ -69,6 +69,11 @@ void RenderData::set_material(Material* material, int pass) {
         renderdata_dirty_ = true;
     }
 }
+
+void RenderData::setCameraDistanceLambda(std::function<float()> func) {
+    cameraDistanceLambda_ = func;
+}
+
 bool compareRenderDataByShader(RenderData* i, RenderData* j) {
     // Compare renderData by their material's shader type
     // Note: multi-pass renderData is skipped for now and put to later position,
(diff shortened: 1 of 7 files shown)


liaxim:
- updates state_sort to respect transparency
- makes the camera distance calculation on demand meaning transparency works the same with culling enabled or disabled

deepakrawat22:
Reviewed the changes and I can see Hello_World! now.

GVRf/Framework/framework/src/main/jni/objects/components/render_data.cpp
```diff
@@ -84,45 +89,41 @@ bool compareRenderDataByShader(RenderData* i, RenderData* j) {
...
-                                           && i->rendering_order() < RenderData::Overlay) {
-                                        return i->camera_distance() > j->camera_distance();
-                                    }
-                                        // otherwise sort from front to back
-                                    return i->camera_distance() < j->camera_distance();
-                                }
-                                return i->getHashCode() < j->getHashCode();
-                            }
-                            return i->cull_face(pass) < j->cull_face(pass);
+                //@todo what about the other passes
```
liaxim: If unclear, I removed the loop over the passes since it wasn't actually looping and could be misleading.

liaxim:
Will have to re-test the samples since I made one more tweak to the sorting function.

The tests pull request: https://github.com/gearvrf/GearVRf-Tests/pull/27

I am confused by the material pointer comparison in the sorting function. Until I understand it or fix it there is a limit on the kinds of tests that can be written.

The problem with the pointer comparison:
- assumes batching is enabled
- do not see the same material being pointed to when I expect in my simple tests; will debug batching post-release
- in case the material pointers are not the same I changed the criteria to be camera distance

liaxim:
On a side note, for the sake of the exercise, the tests use CMake to build the native code. It is an improvement over Android.mk.

thomasflynn:
Here are the requirements for the comparison method that I can think of:

all objects:
 - need to sort by render order (BACKGROUND, GEOMETRY, TRANSPARENT, OVERLAY)
transparent objects: (>= TRANSPARENT and < OVERLAY)
 - for all objects of the same render order:
   - sort back to front
opaque objects:  (>= GEOMETRY and < TRANSPARENT)
 - for all objects of the same render order:
   - group by material (objects with the same material should be next to each other in the list)
   - sort front to back

am i missing anything?

liaxim:
@thomasflynn Added more tests; I am under no illusion they are exhaustive but are a sanity check at least. See https://github.com/gearvrf/GearVRf-Tests/pull/27/files#diff-de3f78d24ac189af16e60ac5388d81a1.

@roshanch should correct me if I am wrong anywhere below.

Since the fix and the tests are focused on state_sort and state_sort I do not believe sorts on material currently, there are no material sorting tests. The grouping by material actually happens later during batching. Batching comes up with a substitute material for the whole batch.

There is a pointer to pointer comparison for the material in state_sort but those pointers are always different.

Including material sorting in state_sort would require sensitive changes that I believe are better done after the release. There are no demos broken by this change.

--
On the lambda - it is not free. There is some memory overhead since the object and the camera_position are captured by copy.

roshanch:
@liaxim all render-data with same material goes to same batch so batching stores the same material it doesn't create a new one. 
When did you find pointers are different? if you create scene objects with same material then it would be same. So material comparison is just for grouping and this grouping occurs irrespective of batching.

liaxim:
The usual way you create scene objects yields different material instances. 

For example this code produces different material instances although logically the materials are the same. Are you saying this has to be rewritten to create one GVRMaterial and set it for each scene object?
```
        GVRTexture texture = context.getAssetLoader().loadTexture(new GVRAndroidResource(
                context, R.drawable.gearvr_logo));
        final GVRSceneObject geometryFar = new GVRSceneObject(context, 5.0f, 5.0f, texture);
        geometryFar.setName("geometryFar");
        geometryFar.getTransform().setPosition(1.0f, 0.0f, -15.0f);
        scene.addSceneObject(geometryFar);

        final GVRSceneObject geometryNear = new GVRSceneObject(context, 5.0f, 5.0f, texture);
        geometryNear.setName("geometryNear");
        geometryNear.getTransform().setPosition(-1.0f, 0.0f, -5.0f);
        scene.addSceneObject(geometryNear);

        final GVRSceneObject geometryMiddle = new GVRSceneObject(context, 5.0f, 5.0f, texture);
        geometryMiddle.setName("geometryMiddle");
        geometryMiddle.getTransform().setPosition(-1.5f, 0.0f, -10.0f);
        scene.addSceneObject(geometryMiddle);
```

roshanch:
Yes. We have to change the criteria of material equal-ness where we can check for texture and material colors which which we use in the shader.

liaxim:
Ok, part of the grouping is manual then. I will update the tests to check for that.

liaxim:
Reverted my last change to reflect the intended grouping by material pointer value. Added more tests and updated the existing ones to take into account the material criteria. This is it for this pull request I believe.

NolaDonato:
I think we should group renderdatas by material and transparency sort as part of the cull process. I also think we should use bounding spheres instead of boxes for culling - this would be faster (don't have to regenerate the bounding box, just transform center and radius) but it is less precise. This would be a merge sort during cull which is faster than a large sort after culling is done. The renderer would then visit the individual material groups - opaque first and then transparent. We can use the memory for these lists over and over again each frame.

I am in no way suggesting that you do this as part of this pull request. I want to record these ideas for reform of material processing for the future when we rework the renderer for Vulkan.

liaxim:
Someone please give thumbs up or let me know what needs further work. Thanks.

thomasflynn:
I did a bit of testing on this last week.  I'd still like a bit more time to continue to test and review.  I should be able to give comments or a thumbs up by tomorrow afternoon.

thomasflynn:
just for the record, i found something unusual in gvr-opacityanigallery that I don't see in master.  mihail is going to check it out and try to reproduce.
otherwise, this change looks good to me.

liaxim:
Did not quite reproduce it but I looked into how we could get two black quads on both sides of the tile in focus.

First, why the video quad is black when not in focus? The background and the video quad have the same rendering order so the order is determined by the shader type (OES/video is 2, Texture/background is 7) thus the video is always drawn first. Since depth testing is enabled the background pixels behind the video quad are never drawn so when there is no video playing we get to see the color the color buffer gets cleared to (the video quad when not in focus has opacity of 0).

Second, how to get two black quads on both sides of the focused one? All quads and the background have GEOMETRY rendering order. In that case we would be making pointer to pointer comparison to establish the rendering order (when the video quad is not involved). So the order is arbitrary. Almost always the background gets drawn before the quads because the memory allocator just so happens to work. If you move the creation of the background-related objects to the end of the onInit method you would get the black quads (when the opacity is 0) that @thomasflynn observed.

I will make updates to gvr-opacityanigallery to correct the issues; like making the background's rendering order BACKGROUND.

thomasflynn:
if the background quads and the picture/video quads are at the same Z, then yes the application would need to be changed to either have the background at a different Z, at a polygon offset, or a different rendering order (like BACKGROUND).

thomasflynn:
p.s. good detective work.

liaxim:
Change to the sample: https://github.com/gearvrf/GearVRf-Demos/pull/357

PR #905, opened 2016-10-25
Include the HEAD commit id in the native binaries

liaxim:
On library load this gets printed out:
```
FRAMEWORK HEAD: <commit-id>
BACKEND_<blah> HEAD: <commit-id>
```
You can also just search for prefix in the binary.

The task is not aware of local changes to the tree.
The Java classes will be marked in a subsequent pull request.
Works on Windows.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,5 @@ import-summary.txt
 !**/libgdx.so
 
 .externalNativeBuild
+buildinfo.h
+buildinfo.cpp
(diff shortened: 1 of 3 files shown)


liaxim:
@Vittalbabu would like the commit id to be printed out so I will look into it.

liaxim:
Updated according to @Vittalbabu's preferences.

liaxim:
Will merge in the afternoon if there are no objections

PR #907, opened 2016-10-27
Fix null collision list case

JulianaFigueira:
Bug fix to better handle empty updated collisions list and avoid conflict with the previousCollisions list.

GearVRf-DCO-1.0-Signed-off-by: JulianaFigueira juliana.f@samsung.com


thomasflynn:
+1, looks good to me.

PR #909, opened 2016-10-28
fixed direct and spot light shadow mapping issues

NolaDonato:
- modified math in shaders
- delete shadow framebuffer in proper thread
- fix problem with GVRAssetLoader not calling onAssetLoaded
- added GVRRenderData.setCastShadows for per-object shadow enable/disable

This does not fix the problem with VRBenchmark lighting not matching in Unity and GearVRF. There is still a problem with the asset importer and light direction.


NolaDonato:
We really should expose some of the shadow mapping parameters to Java to allow the user to tweak them. I exposed shadow_near and shadow_far already. This is a temporary hack. We should probably add more uniforms to the light to set the shadow bias, the shadow intensity, near and far planes, quality, etc.

liaxim:
This is how lightmap_test.fbx looks in gvr-modelviewer2 for me:

screenshot_20161028-105242


screenshot_20161028-105502

--

This is how gvr-shadows looks like (no shadows):

screenshot_20161028-111613

screenshot_20161028-111617

--

Am I missing some changes to the demos or something?

NolaDonato:
I added pull request #354 to GearVRf-Demos to update gvr-shadows. The light directions were wrong so instead I put code in to rotate the light in X and Y when you touch the screen. Right now it is working with a spot light but I put a direct light in and tested that and it works too. The trick is to set shadow_near and shadow_far in the material returned by GVRLightBase.getShadowMaterial appropriately for your scene.

Lightmaptest does not have any lights that cast shadows so you won't see any. Right now we have to explicitly call GVRLightBase.castShadows to make the light cast a shadow. The asset importer does not set this.

liaxim:
It is just that with this pull request the Lightmaptest model's textures flicker for the lack of better words.

NolaDonato:
I found and fixed the problem with the lightmap test textures.

liaxim:
Looks good. The shadows sample looks better. Planning to merge in the afternoon.

Having said that, there are some discrepancies I am sure you are aware of - some artifacts on the big soccer ball when the light is not rotating. The shadows "jump" when the light rotation is stopped.

GVRf/Framework/framework/src/main/jni/engine/renderer/gl_renderer.cpp
```diff
@@ -230,13 +231,16 @@ void GLRenderer::renderShadowMap(RenderState& rstate, Camera* camera, GLuint fra
...
-        GL(renderRenderData(rstate, *it));
+         it != render_data_vector.end(); ++it) {
+        RenderData* rdata = *it;
+        if (rdata->cast_shadows()) {
+            GL(renderRenderData(rstate, rdata));
+        }
     }
-
     glBindFramebuffer(GL_READ_FRAMEBUFFER, readFbo);
     glBindFramebuffer(GL_DRAW_FRAMEBUFFER, drawFbo);
```
roshanch: Are these 2 lines for debugging?

roshanch: can you add glinvalidateFramebuffer() call after glBindFramebuffer() . We need to discard previous attachments to avoid unnecessary copying from disk.

liaxim: @roshanch I added these some time ago. We can't leave the shadow fbo bound since daydream doesn't expect you are going to return control back to daydream with a different fbo bound.

liaxim:
No shadows are rendered if multiview is enabled. To enable multiview you need to set use_multiview in renderer.cpp to true *and* set useMultiview to true in the vr-app-settings section of gvr.xml. I thought changing gvr.xml should have been enough..

@NolaDonato We could merge the pull request the way it is and have the multiview case examined later. Your call.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRShaderTemplate.java
```diff
@@ -811,7 +810,8 @@ private String makeVertexCopy(String descriptor, String inBase, String outBase)
         }
         return desc;
     }
-    
+
+    protected boolean mWriteShadersToDisk = true;
```
liaxim: Since you are still working on this pull request, please set mWriteShadersToDisk to false.

NolaDonato:
The multiview issue is fixed - shadows show when multiview is enabled now

liaxim:
Looks good to me. Planning to merge in the afternoon.

PR #913, opened 2016-11-02
Have the experimental gradle plugin project use extra_settings_exp.gr…

liaxim:
…adle

to include modules outside of this repository.


NolaDonato:
I have tested this and it works if you use absolute paths in the extra settings file. You cannot use the same paths as in settings.gradle - these relative paths don't work.,

liaxim:
If using relative paths they are relative to the current directory for extra_settings_exp.gradle. For example this will include the experimental gradle project for gvr-eyepicking from extra_settings_exp.gradle:

```
include ':gvr-eyepicking'
project(':gvr-eyepicking').projectDir = file('GearVRf/GVRf/tools/Framework/demos/gvr-eyepicking')
```

PR #915, opened 2016-11-03
fixed errors in javadocs

NolaDonato:
This patch only changes comments, no coding changes at all. It fixes some (not all) of the errors emitted during javadoc construction.


liaxim:
Looks good to me

PR #920, opened 2016-11-05
Fix jassimp import of light intensities

NolaDonato:
GearVRF assumes these are 0-1, in jassimp they can be 0 to anything. Jassimp now scales the light intensity  by the maximum R,G,B so the light intensity won't be > 1.
This should improve the look of assets which use lights with very large intensities like the models in the VRBenchmark app.


GVRf/Framework/framework/src/main/java/org/gearvrf/GVRJassimpAdapter.java
```diff
@@ -626,17 +635,21 @@ private void importLights(List<AiLight> lights, Hashtable<String, GVRLightBase>
     }
 
     private void setLightProp(GVRLightBase gvrLight, AiLight assimpLight){
-        gvrLight.setFloat("attenuation_constant", assimpLight.getAttenuationConstant());
-        gvrLight.setFloat("attenuation_linear", assimpLight.getAttenuationLinear());
-        gvrLight.setFloat("attenuation_quadratic", assimpLight.getAttenuationQuadratic());
+        //gvrLight.setFloat("attenuation_constant", assimpLight.getAttenuationConstant());
```
liaxim: If those are truly not needed anymore why not remove the method altogether?

roshanch: where do we set these 3 constants from assimp?

liaxim:
Demos run fine. Code looks good except that for lights imported from assimp the attenuation_constant, attenuation_linear and attenuation_quadratic do not seem to be set anymore. Maybe this is intended but it is unclear from the code.

@roshanch any suggestions for models with lights we can test?

liaxim:
I also rebased the pull request. There was a merge conflict in GVRTexture.

roshanch:
@liaxim we have the cube models with lights in it. but i doubt it will be good test for the attenuation. Papiya has reported the issue with benchmark app with problem with light intensity for direct light, she will test this change with that app.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRJassimpAdapter.java
liaxim: jassimp.cpp:loadLights()?

liaxim:
Ok then. If you are ok with merging this then I am ok too.

roshanch:
let me talk with Papiya about VRBenchmark app Nola has mentioned. I will check whether it works with that

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRJassimpAdapter.java
NolaDonato: I screwed up and commented out the attenutation settings. Put them back in and you are good to go.

liaxim:
Per Nola, I uncommented the attenuation settings.

thomasflynn:
so, what's the latest status on this?

liaxim:
We can merge it.

PR #924, opened 2016-11-09
Switch to Oculus Mobile SDK 1.0.4

liaxim:

### DIFF
--- a/GVRf/Framework/backend_oculus/build.gradle
+++ b/GVRf/Framework/backend_oculus/build.gradle
@@ -17,28 +17,6 @@ android {
         targetSdkVersion 19
     }
 
-    task copyOculusFiles(type: Copy) {
-        if (rootProject.hasProperty("OVR_MOBILE_SDK")) {
-            println "copying oculus binaries"
-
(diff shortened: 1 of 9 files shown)


liaxim:
- the oculus backend is not going to be backwards compatible with 1.0.3
- less dependencies to oculus is better; if we are ok with including libvrapi.so then we can simplify the build further
- updated the experimental gradle plugin project too
- libvrapi's Android.mk uses non-standard "BUILDTYPE" subdirectory which forces us to set it explicitly to "Release" in our build files

Test scenarios:
- long back key press
- short back key press
- running app without oculus signature
- multiview gvr-shadows

thomasflynn:
so far, i have cherry-picked and tested with simplesample.  compiling against 1.0.4 will fail without this commit.  i should do a little more testing, but so far, this looks good.

thomasflynn:
p.s. let me know what i need to change in our build instructions:
https://resources.samsungdevelopers.com/Gear_VR/020_GearVR_Framework_Project/060_Gear_VRf_Developer_Guide/110_Build_Instructions

liaxim:
No need to copy SystemUtils.jar and VrApi.jar anymore.

PR #925, opened 2016-11-10
Fixes in 3DCursorLibrary and SensorManager

parthmehta209:
1. Added check for valid ray from CursorController before performing picking in SensorManager.
2. Added transferIoDevice() call to enableSettingsCursor() so IoDevice is not disabled and re-enabled when using the settings cursor


rahul27:
Will test this tomorrow.

rahul27:
Tested this and it works great. Good job.

This one is good to go.

PR #929, opened 2016-11-11
Motion Events in depth order

parthmehta209:
1. Adding motion events in cursor events. 
2. Set default scale and max scale for cursor
3. Send cursor events in depth order


liaxim:
Has this change been verified that works with the app that needs it? Clicks with the mouse or the HMD's trackpad do not work for me.

parthmehta209:
I have make significant changes and merged all commits into a single commit and put it in another pull request. So I will close this one.

PR #932, opened 2016-11-14
Fix for using framework thread

parthmehta209:
This fixes the setUseFrameworkThread call to use the one in the base class GVRContext. An empty implementation was preventing the framework thread from being created. 


liaxim:
We have discussed it already. There is a fix in one of the pending pull requests but it won't make this release. Will merge in the afternoon.

Without this fix even if an app uses GVRMain it will not execute onInit and onStep on the framework thread. Got broken during one of the backend refactors.

PR #933, opened 2016-11-15
Adding the GVRf Leap Motion integration library

rahul27:
This is a very early version of the integration library using GVRf
and LeapMotion that makes use of the leapC.

r.rudradevan@samsung.com


rahul27:
Here is a sample application that makes use of this library:

https://github.com/gearvrf/GearVRf-Demos/pull/368

PR #935, opened 2016-11-15
oculus backend: check for package availability to determine

liaxim:
whether to use vrapi or not

vrapi_Initialize in 1.0.4 doesn't return error code as it used to;
it crashes the app instead


thomasflynn:
@rahul27 , please double check.  thanks.

rahul27:
:+1:  Works. Tested with with the 6P and Pixel.

PR #937, opened 2016-11-16
Fix the splash screen dismissal logic

liaxim:
When in "automatic" mode we were dismissing the splash screen before onInit completes causing issues noticeable in gvr-meshanimation, gvr-accessibility and gvr-modelviewer.

Note that now the fw will wait correctly for onInit to complete and that may take time. For example, gvr-jassimp loads a tree model from the network so the splash screen stays on for some time.


rahul27:
:+1: This fixes the issues with meshanimation, jassimp and model viewer.

PR #951, opened 2016-11-21
Add a build task that produces a zip file for use by plugins

liaxim:


liaxim:
Also when running the clean task, remove build artifacts that get pushed to the various gearvrf-libs directories

liaxim:
Will merge after lunch

PR #952, opened 2016-11-22
On Note4 there is what looks like a JNI error leading to

liaxim:
an exception after GetStaticFieldID is called; even though
the call actually is successful.


liaxim:
This fixes Nola's recent problem where the framework-tests would crash on launch.

NolaDonato:
I have tested it on the Note4 and it fixes the crash on startup.

liaxim:
Will merge after lunch

PR #953, opened 2016-11-22
Fix crash when loading vrbenchmark asset bmw.fbx

NolaDonato:
There was a check for texture coordinates in Mesh::generateVAO that will
crash if a mesh does not have texture coordinates. We used to supply
them always but now we honor the FBX file and this asset had a mesh
without texcoords.

This does NOT fix the synchronization issue in vrappbenchmark. This
application loads assets simultaneously from two threads. It now hangs
when loading textures. I traced this to an issue in the event manager.
GVREventManager.findHandlerMethod caches the method based on the event
name only. This means you can only have ONE HANDLER in GVRContext for a
particular event. Of course, GVRAssetLoader needs two if it is going to
run on two threads simultaneously...

I am still working on fixing the threading problems with vrappbenchmark so we should not release yet. I might need Mihail's help because I am not confident in the event processing code.


liaxim:

liaxim:
I think I am starting to understand the problem you are talking about. By design the EventManager delivers events to all handlers of a certain type of event, so an onAssetLoaded event for one model would actually lead to onAssetLoaded events invoked for all asset handlers (all models). I am afraid we are misusing the EventManager. Not seeing a quick fix thus far. Will keep looking.

liaxim:
In the AssetLoader instead of ``mContext.getEventManager().sendEvent(mContext, IAssetEvents.class,`` try ``mContext.getEventManager().sendEvent(mUserHandler, IAssetEvents.class,`` to deliver directly to the handler..

PR #956, opened 2016-11-23
Fixed multi-threaded asset loader problem

NolaDonato:
GVRAssetLoader was adding a listener per thread to GVRContext for asset events. GearVRF event manager only permits one listener per target object for a particular event class. I modified the asset loader to not rely on the event manager for propagating internal asset load events. GVRAssetLoader no longer adds listeners for its internal event management, making direct calls whenever possible.

To do asset loading from multiple threads you have to add a user handler to the loadModel function. This API was already there, I just made it avoid the event manager.

I also gave embedded textures their own cache. I tried using the ResourceCache used for texture loading but this would not reliably cache stuff - it sometimes reported something as not cached when it actually was. Not sure what is wrong with this - we probably should look closely at texture caching to see if it actually still works.


GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
```diff
@@ -282,6 +288,13 @@ public GVRTexture loadEmbeddedTexture(TextureRequest request, AiTexture aitex, G
         public void onModelLoaded(GVRContext context, GVRSceneObject model, String modelFile) {
             mModel = model;
             Log.d(TAG, "ASSET: successfully loaded model %s %d", modelFile, mNumTextures);
+            if (mUserHandler != null)
+            {
+                mUserHandler.onModelLoaded(context, model, modelFile);
```
liaxim: You might want to wrap direct calls into user's code with try-catch to avoid their exceptions breaking the fw.

liaxim:
Yet another cache is not ideal but we can live with it to get the release out. But the fact that mEmbeddedCache only grows and never gets expunged, if I understand correctly, is a problem that has to be fixed.

The issue with ResourceCache could be down to the fact it keeps weak references. Depending on the usage pattern it could lead to what you are describing.

liaxim:
Ok, the cache is only for embedded textures. Not considered that big of a deal.

liaxim:
My tests checked out. Will prepare the release on Monday.

PR #959, opened 2016-11-23
Extract a setFovY method that can be used to dynamically

liaxim:
adjust the fovY by apps (via reflection)


liaxim:
Will merge tomorrow morning

PR #965, opened 2016-11-29
MAX_LIGHT count based on max uniforms supported by shaders

apjagdale:
Adding check for number of uniforms supported in VS and FS and assigning appropriate MAX_LIGHTS count

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRScene.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRScene.java
@@ -59,7 +59,7 @@ import java.util.Set;
 public class GVRScene extends GVRHybridObject implements PrettyPrint, IScriptable, IEventReceiver {
     @SuppressWarnings("unused")
     private static final String TAG = Log.tag(GVRScene.class);
-    public static final int MAX_LIGHTS = 16;
+    public static int MAX_LIGHTS = 0;
     private GVRCameraRig mMainCameraRig;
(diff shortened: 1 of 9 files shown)


GVRf/Framework/framework/src/main/java/org/gearvrf/GVRScene.java
```diff
@@ -73,6 +73,11 @@
      */
     public GVRScene(GVRContext gvrContext) {
         super(gvrContext, NativeScene.ctor());
+
+        if(MAX_LIGHTS == 0) {
```
liaxim: There is no guarantee the scene will be created on the GL thread.

NolaDonato:
Looks like I was wrong when I told Abhijit it would be OK. I looked into the initialization code and saw GL calls right before the first call to new GVRScene() and I assumed that code ran on the GL thread. Can you suggest another place where we can put this initialization that is on the GL thread?

liaxim:
- add configureRendering (or whatever feels more appropriately named) to GVRConfigurationManager
- call getActivity().getConfigurationManager().configureRendering() from GVRViewManager.onSurfaceCreated right before setMainScene
- determine the max_lights in configureRendering
- expose getter in GVRConfigurationManager to obtain max_lights
- use gvrContext.getActivity().getConfigurationManager() in GVRScene to get the max lights

I can think of few other things that we could centralize there eventually.

liaxim:

PR #964, opened 2016-11-29
Upgrade to joml 1.9.1

thomasflynn:
Referencing issue #930 (https://github.com/Samsung/GearVRf/issues/930),
this upgrades GearVRf to JOML 1.9.1-SNAPSHOT.  Also, we can reference
joml from gradle rather than include the sources in our build.

GearVRf-DCO-1.0-Signed-off-by: Tom Flynn
tom.flynn@samsung.com

PR #966, opened 2016-11-30
More changes required by the JOML upgrade

liaxim:


liaxim:
Depends on https://github.com/Samsung/GearVRf/pull/964

Samples crashing with the same JNI error out of all things:
gvr-jassimp
gvr-javascript
gvr-lua
gvr-meshanimation
gvr-modelviewer2
gvr-multilight

Will look into the crash.

liaxim:
So there is something wrong with ``new Vector3f(FloatBuffer.wrap(pos))``. But since we actually do not need to create all these new objects I eliminated the calls. Will not investigate further this time around.

liaxim:
The actual error is caused by the android gradle plugin versions used apparently. The demos were using 2+ and if built with this version (apparently something other than 2.2.2), gvr-jassimp would crash. If 2.2.2 is specified - it would not.. I will keep the removal of the superfluous objects change anyway.

liaxim:
It would be nice to run the automated tests. Since most of them are set up for Nola's Note4, @NolaDonato please help. Thanks.

liaxim:
BoundingVolumeTests, PickerTests, StateSortTests and MiscTests pass.

liaxim:
Will merge tomorrow morning

PR #969, opened 2016-12-01
Refactor how hybrid objects are finalized; replace the RenderData dirty listener with something that doesn't lead to invalid memory accesses

liaxim:
- Remove the custom finalizer thread
- Ensure no concurrent deletion of hybrid objects can occur while rendering
- Hybrid object dtors will always be executed on the GL thread paving the way
for the removal of the deleter hack

- Eliminate the RenderData dirty listener; was causing invalid memory accesses
due to the lack of explicit ownership of Mesh, Material, RenderPass and
RenderData pointers that were being passed around
- Introduce a shared pointer for the dirty flag that is shared across the
objects that care about it; it is a stop gap measure but it is an improvement
and as far as I can tell the crashes are gone

Test case: https://github.com/gearvrf/GearVRf-Tests/pull/38


liaxim:
I assumed 1 to 1 relationships when it comes to the dirty listeners. We have 1 to many so this needs more work.

liaxim:
The demos check out.

liaxim:
BoundingVolumeTests, MiscTests, PickerTests and StateSortTests pass.

liaxim:
I will need help with reviewing and testing this from someone.

Planned to create automated test but the app that causes the crash this pull requests fixes is very complicated and I could not induce the very same crash with my tests.

liaxim:
I have independent confirmation that the app that was crashing no longer is with this fix.

liaxim:
Added a test case: https://github.com/gearvrf/GearVRf-Tests/pull/38

Nothing is visibly rendered but few hundred iterations into the loop there will be a native crash without the fixes from this pull request.

rahul27:
Planning to review and test this today. Is there a specific app I need to run to recreate the issue?

liaxim:
None of our sample apps stresses the framework enough. I did add a test case to the framework-tests project. We can talk in private if you'd like to try the app that gets the framework to crash.

rahul27:
Ah I see. Will try a sample few apps and the test case.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRSceneObject.java
```diff
@@ -270,7 +270,13 @@ public GVRSceneObject(GVRContext gvrContext, float width, float height,
 
     protected void finalize() throws Throwable
     {
-        detachAllComponents();
+        try {
+            detachAllComponents();
+        } catch(final Exception exc) {
+            exc.printStackTrace();
```
liaxim: Any RuntimeExceptions. A finalizer should not throw exceptions and and should also ensure the super.finalize is invoked. I have seen this finalizer throw occasionally.

GVRf/tools/Framework/settings.gradle
```diff
@@ -1,19 +1,19 @@
 include ':framework'
-if (!hasProperty("ARM64")) {
-    include ':backend_oculus'
-}
+//if (!hasProperty("ARM64")) {
```
liaxim: Yes. This is for the experimental workspace and the 64bit support there is not a priority.

GVRf/Framework/framework/src/main/jni/objects/mesh.h
```diff
@@ -39,14 +40,24 @@
...
+            float_vectors_(),
+            vec2_vectors_(),
+            vec3_vectors_(),
+            vec4_vectors_(),
+            have_bounding_volume_(false),
+            vao_dirty_(true),
+            boneVboID_(GVR_INVALID),
+            vertexBoneData_(this),
+            bone_data_dirty_(true),
+            regenerate_vao_(true)
```
liaxim: regenerate_vao_ removed.

rahul27:
Tried a few samples and they ran fine, with the exception of a weird issue with model viewer (which I cannot reproduce consistently). Not sure if it is related, in any case @liaxim is having a look at it.

The test checks out.

liaxim:
The model viewer thing is not reproducible. It looks like some kind of a texture loading/usage problem specific to the texture shader (only objects rendered by it were affected). Numerous attempts to reproduce across various devices were not successful. Will tackle once it manifests again in a more consistent fashion.

Will merge the pull request tomorrow morning.

For reference this is how the modelviewer end up looking like for Rahul:

image

PR #970, opened 2016-12-02
v3.1.1: Upgrade to joml 1.9.1

liaxim:


rahul27:
Tested this and the samples worked fine. 

I did notice one unrelated issue with the 3d cursor sample that @parthmehta209 is currently looking into.

The is one minor followup commit that is needed to allow the extensions to be built individually. I will add that once this one is through.

liaxim:
Extensions can be built individually already: ``./gradlew 3DCursorLibrary:assembleDebug`` for example.

When importing in Android Studio use this option to enable selection of modules to import:
image

rahul27:
I usually open the 3DCursorLibrary in a separate window on Android Studio, it does not build that way.

PR #971, opened 2016-12-02
Added upload artifact task for release

parthmehta209:


GVRf/Framework/backend_daydream/build.gradle
```diff
@@ -124,3 +124,7 @@ assembleDebug {}.doLast {
         uploadToGitHub.execute();
     }
 }
+
+if(file("../release.gradle").exists()) {
```
liaxim: Please apply only in case artifactory_username is defined. Let's not force everyone define artifactory_username - dummy values or not.

PR #972, opened 2016-12-02
v3.1.1: MAX_LIGHT count based on max uniforms supported by shaders

liaxim:


GVRf/Framework/framework/src/main/java/org/gearvrf/GVRConfigurationManager.java
```diff
@@ -192,6 +192,17 @@ public boolean isHomeKeyPresent() {
         return false;
     }
 
+    public void configureRendering(){
+        NativeConfigurationManager.configureRendering(mActivity.get().getNative());
```
rahul27: The activity ptr is cast to a ptr type of configuration manager, @apjagdale was this intentional?

PR #976, opened 2016-12-07
Change compileSdkVersion to N

deepakrawat22:


GVRf/Framework/backend_daydream/build.gradle
```diff
@@ -9,12 +9,12 @@ repositories {
 }
 
 android {
-    compileSdkVersion 21
-    buildToolsVersion '23.0.3'
+    compileSdkVersion 24
```
rahul27: @deepakrawat22 does this need to be 24? I am seeing a weird crash with the daydream backend.

Can I revert it back to 21?

deepakrawat22: Yes you can revert back.

PR #981, opened 2016-12-09
Clean up of few AndroidManifest.xml files

liaxim:
No longer apps will have errors because of an icon being already specified


liaxim:
Will merge after lunch

PR #984, opened 2016-12-12
Add stencil buffer support

liaxim:
A simple test: https://github.com/gearvrf/GearVRf-Tests/pull/47

Works with all backends. Stencil attachment added unconditionally for Daydream since it is not yet using the values from gvr.xml. For other backends the depthFormat must be set to DEPTH_24_STENCIL_8.

Creates a special stencil bucket for stencil scene objects that get rendered very first with stencil test on.

GVRTransform, GVRRenderData - all the setters return "this" so multiple setXxx calls can be chained

### DIFF
--- a/GVRf/Framework/backend_oculus/src/main/jni/ovr_framebufferobject.cpp
+++ b/GVRf/Framework/backend_oculus/src/main/jni/ovr_framebufferobject.cpp
@@ -144,6 +144,9 @@ void FrameBufferObject::create(const ovrTextureFormat colorFormat, const int wid
         mResolveFrameBuffers = new GLuint[mTextureSwapChainLength];
     }
 
+    const GLenum depthStencilAttachment =
+            VRAPI_TEXTURE_FORMAT_DEPTH_24_STENCIL_8 == depthFormat ? GL_DEPTH_STENCIL_ATTACHMENT : GL_DEPTH_ATTACHMENT;
+
     for (int i = 0; i < mTextureSwapChainLength; ++i) {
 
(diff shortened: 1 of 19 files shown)


liaxim:
A certain app needs this for a new feature

liaxim:
Ready for review

GVRf/Framework/backend_oculus/src/main/jni/ovr_framebufferobject.cpp
```diff
@@ -277,8 +244,14 @@ bool FrameBufferObject::create(const ovrTextureFormat colorFormat, const int wid
                 if (depthFormat != VRAPI_TEXTURE_FORMAT_NONE) {
                     if (resolveDepth) {
                         GL( glFramebufferTexture2D(GL_FRAMEBUFFER, GL_DEPTH_ATTACHMENT, GL_TEXTURE_2D, depthTexture, 0) );
+                        if (VRAPI_TEXTURE_FORMAT_DEPTH_24_STENCIL_8 == depthFormat) {
+                            GL( glFramebufferTexture2D(GL_FRAMEBUFFER, GL_STENCIL_ATTACHMENT, GL_TEXTURE_2D, depthTexture, 0) );
+                        }
```
roshanch: when attaching depth and stencil we can use "GL_DEPTH_STENCIL_ATTACHMENT" with single call. Same applies to multiview.

liaxim:
Any other comments?

liaxim:
Assuming it is decent enough. Will merge Wednesday afternoon if there are no further comments.

PR #985, opened 2016-12-12
Zip up select headers and publish the artifact

liaxim:


liaxim:
Will have to add the dependencies

liaxim:
Keshav might have more input in the future when this becomes a priority.

PR #993, opened 2016-12-16
Fixed a bug where some components are disabled by default

rahul27:
Surprising that it took so long to discover this bug.

r.rudradevan@samsung.com


liaxim:
How did you find out?

rahul27:
Incorrect IPickEvents were being generated for colliders that were disabled. The sample app that I wrote for daydream enables and disables colliders inside IPickEvents.

Just to be sure I ran eye picker and the picking tests against this pull request and they ran fine.

PR #995, opened 2016-12-19
Use externalNativeBuild in the primary gradle files

liaxim:
- debugging, syntax highlighting should work now
- we can remove the experimental gradle workspace now


liaxim:
Will merge in the afternoon

liaxim:
@rahul27 You might want to have a quick look since I had to change how the native code for daydream is built. The generation of Android.mk doesn't seem to gel with externalNativeBuild so I added a cmake build file instead.

rahul27:
@liaxim I have a version of the daydream backend with Android.mk and Application.mk files. Will that be easier for you to maintain uniformity across backends?

liaxim:
From the look of things, cmake is android's preferred build system for native code. So we may very well start migrating bit by bit. It is a more concise than Android.mk for sure.

rahul27:
Alright.

This PR asks me to provide an OVR_MOBILE_SDK path:

````
Error:(22, 0) Could not get unknown property 'OVR_MOBILE_SDK' for root project 'Framework' of type org.gradle.api.Project.
````

Should we still consider a default path if none is provided?

rahul27:
I also get `IllegalArgumentException: new child is an ancestor` when I try to view files using the "Project Files" tab in Android Studio.

liaxim:
@rahul27 About the default path for OVR_MOBILE_SDK - I really prefer to have just one way of doing things. Especially for something that I end up having to maintain. How strongly do you feel about the hard-coded path? I guess we could remove the property instead.

liaxim:
@rahul27 About the IllegalArgumentException - doesn't happen for me in AS 2.2.3. But I've seen the problem quite often when switching between branches that are quite different. You should close AS, delete *.iml (or better ``git clean -fdx``) and re-import the project. Let me know what happens.

rahul27:
Reg OVR_MOBILE_SDK, it feels convenient to build the framework without having to modify the source files. But again if it is too much of a hassle then your way works.

Reg. the IllegalArgumentException, I still get the error after git clean -fdx and re importing the project. I am currently on AS 2.2.3 the Dec 2 build.

liaxim:
OVR_MOBILE_SDK - you should just set it in the global gradle.properties and pretty much forget about it.

liaxim:
I think this IllegalArgumentException is a bug in general. I just saw it when using the experimental workspace. Planning to merge tomorrow morning. If anybody feels strongly about keeping the support for the default OVR_MOBILE_SDK path should speak up.

PR #1001, opened 2016-12-28
Fix the path to the so files

rahul27:
The externalNativeBuild command no longer places the *.so files in
src/main/libs/

Use the build folder instead.

r.rudradevan@samsung.com


rahul27:
@deepakrawat22 could you confirm if this works?

deepakrawat22:
so files are getting copied.

PR #1002, opened 2017-01-03
Removed unwanted matrix mul calls from the daydream renderer

rahul27:
r.rudradevan@samsung.com


GVRf/Framework/backend_daydream/src/main/jni/daydream_renderer.cpp
```diff
@@ -155,15 +155,10 @@ void DaydreamRenderer::DrawFrame(JNIEnv &env) {
 
     gvr::Frame frame = swapchain_->AcquireFrame();
 
-
     gvr::ClockTimePoint target_time = gvr::GvrApi::GetTimePointNow();
     target_time.monotonic_system_time_nanos += kPredictionTimeWithoutVsyncNanos;
 
     head_view_ = gvr_api_->GetHeadSpaceFromStartSpaceRotation(target_time);
```
liaxim: The canonical Daydream sample code shows that the eye-specific matrices are taken into account. Do you know what might be the impact of not using them? Like what transformations are we missing?

I think you could apply the left/eye matrices to the left/right camera.

```
    cameraRig_->getHeadTransform()->setModelMatrix(MatrixToGLMMatrix(head_view_));
    cameraRig_->left_camera_->owner_object()->transform()->setModelMatrix(MatrixToGLMMatrix(left_eye_matrix));
    cameraRig_->right_camera_->owner_object()->transform()->setModelMatrix(MatrixToGLMMatrix(right_eye_matrix));
```

You will need an accessor for the left/right camera.

rahul27: I had tried it and found no difference, which is why I got rid of the calls to avoid these computations. It seems like the left/right matrices provided by daydream are just helpers that do the same thing as GVRf when it comes to calculating matrices for the left/right cameras.

liaxim:
Ok. Will merge in the afternoon.

PR #1003, opened 2017-01-03
Vk texture separation

roshanch:
Separated out GL and Vulkan versions of Material, Textures, RenderData and UniformBlock in the C++ layer. The Renderer now has routines to instantiate these in a device-specific way.

Moved updating bone matrices and view transforms into common code (moving the data to the GPU still device-specific).

Removed GVRPostEffect from Java - GVRShaderData can be used here instead. Also removed other post effect classes that are now meaningless.

GearVRF DCO signed off by: Nola Donato nola.donato@samsung.com and Roshan Chaudhari roshan.c@samsung.com


liaxim:
I think it is highly desirable to break out changes from this branch and create review-able and testable pull requests for the trunk. Things like the removal of the post effect, GVRImage, GVRJavaComponent, all shaders using GVRShaderTemplate, etc.

liaxim:
Should we merge?

liaxim:
AS is getting confused by this addition to Android.mk: ``$(call import-module, third_party/shaderc)``. I have no fix for it but after adding explictly the symbols path breakpoints in renderCamera and getInstance are hit. Add to Edit Configuration->Debugger->Symbol Directories this path:
``<your path to>/GearVRf/GVRf/Framework/framework/build/intermediates/ndkBuild/debug/obj/local/armeabi-v7a/``.

roshanch:
@liaxim Thanks a lot. Its working now.

PR #1004, opened 2017-01-03
Enable native debugging for the fw and the backends

liaxim:


liaxim:
- defined a gvr-simplesample project in extra_settings.gradle
- modified the demo's common.gradle to include just this dependency
```
dependencies {
    compile "org.joml:joml-android:${jomlVersion}"
}
```
- added these dependencies to the sample's build.gradle
```
dependencies {
    compile project(':framework')
    compile project(':backend_oculus')
}
```

Native breakpoints were hit and it all seemed to work normally. Will look into simplifying this.

PR #1006, opened 2017-01-04
Cleanup: remove GlDelete

liaxim:


liaxim:
All the samples check out. The complex proprietary app that we have checks out too.

liaxim:
Will merge Tuesday afternoon.

PR #1008, opened 2017-01-06
X3D, Scripting per Frame

mwitchwilliams:
Added code to the X3D SCRIPT node to be called per frame, enabling JavaScript code to be called / programmed per frame.  Features include procedural animation (transforms and color), with 3d objects that can be clicked or the cursor over the 3d object


thomasflynn:
Tested against the provided tests in GearVRf-Tests.  Works as expected.

liaxim:
Looks good.

PR #1011, opened 2017-01-09
Clean up of the head tracking processing for Oculus

liaxim:
There were redundant ApplyHeadModel calls and generally
streamlined the code.

I find a subtle improvement with this change - on Note5
which has no frame dropping issue like S7.


liaxim:
@rahul27 I would be curious to know whether you can spot a difference and how it compares to the Oculus 360 Video app.

rahul27:
:+1: Looks good.

The performance of the Oculus 360 video and the gvr-360video seem comparable to me.

liaxim:
Thanks. Will merge in the afternoon.

PR #1014, opened 2017-01-13
Simple fix for a problem when the transform hierarchy is concurrently

liaxim:
validated and invalidated.


liaxim:
@saurabhlalwani This addresses the problem in your app. Probably the best that can be done without redesigning the whole threading model.

saurabhlalwani:
@liaxim my app seems to be working fine with this change.

liaxim:
There a small overhead and it is more of a hack but will have to do for now.

liaxim:
Planning to merge Tuesday after lunch if there are no comments.

PR #1016, opened 2017-01-18
No longer require 'git' on the PATH to get the HEAD commit id

liaxim:


liaxim:
Directly reading files under the .git directory to determine the current head commit. No need to have the git executable in the PATH anymore. Verified it works on Windows and Linux.

liaxim:
Will merge tomorrow morning

PR #1017, opened 2017-01-18
Remove the getNextMainScene making the splash screen more transparent for end users

liaxim:
There will be one and only main scene from the user's perspective.

The use of the mainScene in the ViewManager has not been and still is not completely thread-safe. Instead of trying to fix that I rather completely remove the capability to change the main scene in the future.

Added a new method to the scene to set the background color. Deprecated the per-camera background color setter.

Demos update: https://github.com/gearvrf/GearVRf-Demos/pull/397

### DIFF
--- a/GVRf/Extensions/WidgetPlugin/widgetplugin/src/main/java/org/gearvrf/widgetplugin/GVRWidgetPlugin.java
+++ b/GVRf/Extensions/WidgetPlugin/widgetplugin/src/main/java/org/gearvrf/widgetplugin/GVRWidgetPlugin.java
@@ -166,7 +166,7 @@ public class GVRWidgetPlugin implements AndroidApplicationBase {
         @Override
         public void onEarlyInit(GVRContext context) {
             initWidget();
-        }  
+        }
     };
 
     public GVRWidgetPlugin(GVRActivity activity) {
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRContext.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRContext.java
@@ -2302,30 +2302,6 @@ public abstract class GVRContext implements IEventReceiver {
     public abstract void setMainScene(GVRScene scene);
 
     /**
-     * Returns a {@link GVRScene} that you can populate before passing to
-     * {@link #setMainScene(GVRScene)}.
-     * 
-     * Implementation maintains a single element buffer, initialized to
-     * {@code null}. When this method is called, creates a new scene if the
-     * buffer is {@code null}, then returns the buffered scene. If this buffered
-     * scene is passed to {@link #setMainScene(GVRScene)}, the buffer is reset
-     * to {@code null}.
-     * 
-     * <p>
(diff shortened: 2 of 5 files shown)


NolaDonato:
Yeah!

liaxim:
Having problems with some of the backends and the gvr-widgetviewer sample

liaxim:
Ready for review

NolaDonato:
I ran all of the unit tests. Everything works except the alpha to coverage test. The text wraps differently in the box than it used to.

The gvr-shadows demos has weird flashing when the light moves.

liaxim:
The text wrapping issue - did you change something? Like the test phone or maybe updated the OS? The text size can be unpredictable and has always been eye-balled. Maybe there is something more we can get out of the Android APIs to make it (more) consistent but so far it hasn't been high priority. I'd recommend something other than the textViewSceneObject when it comes to automated tests.

gvr-shadows: the flashing recently has become more apparent, I noticed that too.

gvr-simplephysics has been working reliably for me. What phone did you test it on? Does the problem happen every time you launch the app?

--
Planning to merge this pull request in the afternoon.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRCamera.java
```diff
@@ -94,6 +94,7 @@ public void setBackgroundColor(int color) {
      * Meaningful parameter values are from 0 to 1, inclusive: values
      * {@literal < 0} are clamped to 0; values {@literal > 1} are clamped to 1.
      */
+    @Deprecated
```
rahul27: Why is this getting deprecated?

If this one is deprecated there is also the `public void setBackgroundColor(int color)` function as well.

liaxim: Will deprecate the other one too. Thanks.

I just can't see what good it is to expose setting the clear color per camera. It is not like you will ever need to have different background color in the left eye from the right - that would be one odd thing to look at in VR. I planned to just hide the public APIs, not remove the capability which may be useful to us.

It is much more idiomatic to set the background color on the GVRScene. Easier to discover and code too.

rahul27:
Just tried the gvr-sceneobjects app on daydream and it crashes for me. Let me verify if it works on master.

liaxim:
Please use git clean -fdx before building, just in case.

rahul27:
I did a git clean -fdx. It seems to be happening on master as well.
 It may be a problem with gvr-sceneobjects. gvr-simplesample seems to work fine.

liaxim:
Could you please share the logcat output? Which phone are you using?

rahul27:
Am using the pixel. Will send you the logcat.

liaxim:
@rahul27 tracked down the crash to using the GVRCameraSceneObject on Pixel. Unrelated problem.

liaxim:
Removed the deprecation of setBackgroundColor in GVRCamera. There is a whole family of related APIs that I rather not touch right now. Let's see about this in the future.

I still think a helper in GVRScene makes sense.

PR #1018, opened 2017-01-19
Access for trackpad for other VR devices

sivapenke:
Signed-off-by: Siva Penke <siva.penke@samsung.com>


liaxim:
I am not sure this belongs to the mainline. @rahul27 ?

liaxim:
@rahul27 says it is ok. Will merge after lunch.

PR #1019, opened 2017-01-19
Call handleOnDock explicitly for devices that doesn't listen for dock …

sivapenke:
…events

Signed-off-by: Siva Penke <siva.penke@samsung.com>

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRActivity.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRActivity.java
@@ -273,17 +273,8 @@ public class GVRActivity extends Activity implements IEventReceiver, IScriptable
                     IActivityEvents.class,
                     "onSetMain", gvrMain);
             final GVRConfigurationManager localConfigurationManager = mConfigurationManager;
-            if (null != mDockEventReceiver && !isMonoscopicMode && localConfigurationManager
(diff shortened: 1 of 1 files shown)


liaxim:
I don't think we can call handleOnDock() unconditionally. What if the device is in developer mode?

liaxim:
Looked at it closer. Unfortunately the polling check I believe is still needed since we had problems with the listener not getting events on some devices. Based on your idea, I have come up with https://github.com/liaxim/GearVRf/tree/alwaysdock which should work both for your and the general case.

liaxim:
@sivapenke What do you think the commit in https://github.com/liaxim/GearVRf/tree/alwaysdock?

sivapenke:
Do you think this is more appropriate? Though it adds an extra 'if' statement it makes readable and does extra check of isMonoscopic Mode for SAVR. Rest looks good.

if(!isMonoscopicMode) {
	if (null != mDockEventReceiver && localConfigurationManager.isDockListenerRequired()) {
		//drop-in
	} else {
		//savr
		handleOnDock();
	}
}

liaxim:
@sivapenke Yes. Will make the change.

liaxim:
@sivapenke Should we close this?

sivapenke:
@liaxim Yes.

PR #1025, opened 2017-01-30
GVRScene: Make getRoot package private.

liaxim:
Exposing it is confusing since it is unclear what users may do with it.

Also add names for various objects to make debugging slightly easier.

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRCameraRig.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRCameraRig.java
@@ -105,6 +105,7 @@ public class GVRCameraRig extends GVRComponent implements PrettyPrint {
         getOwnerObject().attachCameraRig(this);
 
         headTransformObject = new GVRSceneObject(gvrContext);
+        headTransformObject.setName("_gvrf_head_transform");
         addHeadTransformObject();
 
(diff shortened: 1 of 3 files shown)


liaxim:
@NolaDonato Please let me know what you think since you added the getRoot method. Thanks.

NolaDonato:
If you make getRoot() inaccessible to users, a program will not be able to gather all the top level scene objects. I am confused about why having a method to get the root of the scene graph is a problem.

liaxim:
I guess then we could document it better. Like give examples what it is good for.

There was a specific case where a user expected that changing the transform of the root would have an effect on the scene rendered. Which it can't since this affects the whole world - camera rig, cameras and scene objects. What should happen if someone changes the position of the root?

If the true purpose is allowing to enumerate the top-level scene objects, it might be preferable exposing a dedicated method that does just that (and which returns only user-added objects). Returning a scene object implies many more possibilities.

NolaDonato:
Actually, changing the position of the root could have a meaning in augmented reality where the scene may be placed within a larger world. Changing the scale of the root could be used to map the virtual coordinate system onto a real world view with a different coordinate system.


-Nola



liaxim:
Ok

NolaDonato:
I will document it better. I am never opposed to making something clearer.

Issue #967, opened 2016-11-30
Texture goes missing if a new material is created and assigned to RenderData


liaxim:
Quite easy to reproduce; last known instance is the new physics sample app. There is this section there

```
        GVRMaterial material = new GVRMaterial(context);

        GVRSceneObject groundObject = new GVRCubeSceneObject(context, true, texture);

        groundObject.getRenderData().setMaterial(material);
        groundObject.getRenderData().getMaterial().setTexture("diffuseTexture", texture);
        groundObject.getRenderData().getMaterial().setMainTexture(texture);
        groundObject.getTransform().setScale(15.0f, 0.5f, 15.0f);
        groundObject.getTransform().setPosition(x, y, z);
```

If the setMainTexture call is removed the texture is not going to be applied. If the existing material associated with the scene object is used the texture is applied correctly. We have to look at this sooner rather than later.

NolaDonato:
We are already working on this but it will take a bit more time. I am rewriting all the old built-in shaders in terms of the custom shader. When this is complete, the new texture shader will let you omit the texture and it will use only the color like the phong shader does. You also will never have to call bindShaders again - the system will do it automatically.

NolaDonato:
This is not a bug. It is a "feature" of the texture shader (which is the default shader if you don't specify anything). If you remove the setMainTexture call, you have to create the cube with the Phong shader (which uses "diffuseTexture" instead of "mainTexture"). Right now the way to do this is via GVRRenderData.setShaderTemplate. Replace the setMainTexture call with this:

groundObject.getRenderData().setShaderTemplate(GVRPhongShader.class);
groundObject.getRenderData().bindShaders(scene);

This will force that render data to use the Phong shader. As a update: the bindShaders removal has been coded in the Vulkan branch and is working.

liaxim:
This is more complete excerpt from the simplephysics sample (MainHelper.java):

```
        Future<GVRTexture> texture = context.loadFutureTexture(new GVRAndroidResource(context, R.drawable.orange));
        GVRMaterial material = new GVRMaterial(context);

        GVRSceneObject groundObject = new GVRCubeSceneObject(context, true, texture);

        groundObject.getRenderData().setMaterial(material);
        groundObject.getRenderData().getMaterial().setTexture("diffuseTexture", texture);
        groundObject.getRenderData().getMaterial().setMainTexture(texture);
        groundObject.getTransform().setScale(15.0f, 0.5f, 15.0f);
[...]

PR #1028, opened 2017-02-02
Support rendering a scene to a texture

liaxim:
Demos pull request: https://github.com/gearvrf/GearVRf-Demos/pull/399

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRCameraRig.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRCameraRig.java
@@ -106,6 +106,7 @@ public class GVRCameraRig extends GVRComponent implements PrettyPrint {
 
         headTransformObject = new GVRSceneObject(gvrContext);
         addHeadTransformObject();
+        headTransformObject.setName("_gvrf_head_transform");
 
         leftCameraObject = new GVRSceneObject(gvrContext);
         rightCameraObject = new GVRSceneObject(gvrContext);
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRContext.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRContext.java
@@ -50,8 +50,10 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.lang.ref.PhantomReference;
 import java.lang.ref.ReferenceQueue;
+import java.lang.ref.WeakReference;
 import java.util.ArrayList;
 import java.util.EnumSet;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
@@ -2407,6 +2409,8 @@ public abstract class GVRContext implements IEventReceiver {
      */
     public abstract GVREventManager getEventManager();
 
+    abstract void renderTexture(GVRScene scene, GVRCamera camera, GVRRenderTexture renderTexture);
+
     /**
      * Returns the {@link GVRScriptManager}.
      *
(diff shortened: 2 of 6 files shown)


liaxim:
Ready for review

NolaDonato:
Is there some way we can implement this as a component that you attach to a scene object that will cause it to render from the camera component attached to that scene object? If we did this the feature could be used for shadow mapping and environment mapping too.

liaxim:
I was thinking of re-implementing this as a specialization of GVRRenderTexture. Say GVRSceneRenderTexture where you pass the scene to be rendered and you just use the thing as a regular texture otherwise. No new APIs, etc. I will think about your suggestions and come up with something more useful. Thank you.

NolaDonato:
I am excited about this feature because it will clean up the mess in that is shadow mapping. Right now this is hacked into the renderer in a very obtrustive way (mostly because I didn't want to take on reforming the render to texture stuff as you are doing now). There is a separate render texture for each shadow map but they all use the same framebuffer because they are rendered serially. Potentially this could change with Vulkan and we could render them in different threads. I know that many console and PC game engines do this, I am not sure if it is a win on mobile. We also have this thing in the C++ layer called a texture capturer. I am not sure how that fits into the scheme of things.

liaxim:
The texture capturer is a special facility only useful if you use the external renderer shader. In their current form both the external renderer shader and the texture capturer do not belong to this repo. They should be re-implemented differently. We probably should look into generalizing the texture capturer in the future. The external renderer shader is not something that we may even need.

In my opinion you should drop both from your vulkan development branches if they cause too much headache. @thomasflynn will correct me if I am wrong.

liaxim:
Minor change introducing GVRSceneRenderTexture and eliminating the proprietary apis.

NolaDonato:
I tried use this new feature to help konidk on issue #1060. I set up a scene which renders a blurred version of the camera input to a quad and then tried to use that resulting texture. Your example works beautifully. Mine displays nothing when I use the render to texture. I don't think it is a problem with my shader because it seems to work OK if I don't render to a texture first. I will try to include a ZIP file with my test app in this message if possible.
[test.zip](https://github.com/Samsung/GearVRf/files/822911/test.zip)

liaxim:
@NolaDonato I will investigate. Thanks.

liaxim:
Applying the rendered-to-texture to a quad just doesn't work regardless everything else. My sample doesn't work if I use a quad instead of a cube. Makes little sense but that is what it seems.

liaxim:
Rebased

liaxim:
@NolaDonato Got your test app working. No issues with render-to-texture and/or oes textures actually.


Detailed list of changes:

**gaussianblurhorz.fsh:1**
```
#extension GL_OES_EGL_image_external_essl3 : require
```
to
```
#extension GL_OES_EGL_image_external : require
```

**TestMain.java:61**
```
GVRSceneObject blurryCamera = new GVRSceneObject(context, 3.6f, 2.0f, null, GVRMaterial.GVRShaderType.OES.ID);
```
to
```
GVRTexture t = getGVRContext().getAssetLoader().loadTexture(new GVRAndroidResource(getGVRContext(), R.drawable.ic_launcher));
GVRSceneObject blurryCamera = new GVRSceneObject(context, 3.6f, 2.0f, t);
```
Do note that if the quad is created without a texture (any texture) passed to the ctor subsequent attempts to set the texture have no effect. Will be investigating further and I think it may have something to do with issue https://github.com/Samsung/GearVRf/issues/967.

**TestMain.java:113**
```
final GVRSceneObject blurryQuad = new GVRSceneObject(mContext, 1, 1, texture, GVRMaterial.GVRShaderType.OES.ID);
```
to
```
final GVRSceneObject blurryQuad = new GVRSceneObject(mContext, 3.6f, 2.0f, texture, GVRMaterial.GVRShaderType.OES.ID);
```

Since the render-to-texture is not focusing on a particular scene object the scene needs to be treated as if it were the regular main scene. Hence the quad size.

**TestMain.java:120**
```
blurryQuad.getTransform().setPositionZ(-0.01f);
```
to
```
blurryQuad.getTransform().setPositionZ(-2f);
```

Getting culled otherwise.

---

As expected the performance is bad even if the regular camera is shown since the blurry one is getting rendered anyway.

The blurry camera texture is flipped vertically because the GVRSceneRenderTexture is doing it intentionally. May be it shouldn't, will be checking that.

liaxim:
I will be looking into prototyping the component you suggested. It will still be Java-only but will be more in line with your idea.

NolaDonato:
Take a look at this branch to see what I did in C++: https://github.com/NolaDonato/GearVRf/tree/rendertotexture


Thanks so much for fixing the blur app! Now I can help the user with his problem.


-Nola

liaxim:
Obsoleted by https://github.com/Samsung/GearVRf/pull/1099

PR #1029, opened 2017-02-03
Centralized checkGLError

NolaDonato:
We had three different implementations of checkGLError, each slightly different. I picked the best of the lot and made all the GL code use that one. This makes it easier to set one breakpoint and trap all the GL errors.


GVRf/Framework/framework/src/main/jni/objects/components/bone_jni.cpp
```diff
@@ -130,8 +132,10 @@ Java_org_gearvrf_NativeBone_setFinalTransformMatrix(JNIEnv * env, jobject clz, j
 JNIEXPORT jfloatArray JNICALL
 Java_org_gearvrf_NativeBone_getFinalTransformMatrix(JNIEnv * env, jobject clz, jlong ptr) {
     Bone* bone = reinterpret_cast<Bone*>(ptr);
+    glm::mat4 matrix;
 
-    glm::mat4 matrix = bone->getFinalTransformMatrix();
+    if (bone)
+        glm::mat4 matrix = bone->getFinalTransformMatrix();
```
liaxim: The return value here will be lost. Assuming it is not intentional?

liaxim:
There is one more backend that needs to be updated. Otherwise it doesn't build.

GVRf/Framework/framework/src/main/jni/objects/components/bone_jni.cpp
NolaDonato: You are right. If there is a bone I want to set the matrix, Otherwise it should be identity. I pushed a fix for this.

NolaDonato:
I don't understand about updating another backend. The checkGLError function in daydream_backend is local to that module. Are you suggesting I use the gvr_util.h function in the Daydream backend too? If so, I need help with how to include the proper framework headers.

liaxim:
Will reply via email.

liaxim:
Will merge Friday morning unless there are comments.

PR #1031, opened 2017-02-06
Small glClear-related optimizations

liaxim:


liaxim:
Verified it is working on an Adreno Note 4 with KK. GL_EXT_texture_border_clamp is not available in this configuration. Any other comments?

I can see the setBackground api change (to disable clearing of the color buffer on every frame) being questionable but it is kind of hidden.

liaxim:

PR #1035, opened 2017-02-09
Remove releaseNative

liaxim:
- it is obsolete and can easily lead to crashes

Demos: https://github.com/gearvrf/GearVRf-Demos/pull/401


liaxim:
Will merge Monday morning if there are no objections.

PR #1043, opened 2017-02-14
Call handleOnDock immediately for devices that do not require a docking event listener

liaxim:
Supersedes https://github.com/Samsung/GearVRf/pull/1019


liaxim:
@sivapenke Please confirm it works as expected for you.

sivapenke:
@liaxim not working. Looks like the issue is with the sequence of calls. isDockListenerRequired() is not updated yet by the time it is called.

liaxim:
@sivapenke Should work now. Sending the event from onCreate requires bigger refactoring.

sivapenke:
@liaxim Ok, works now.

liaxim:
Will merge Thursday morning if there are no other comments.

sivaboggala:
isDockListenerRequired() depends on compile time isStandAloneVR(). There is no guarantee that Build.DEVICE.endsWith("_vr"). This is weird flag set by my team to distinguish regular builds from standalone build. Framework should query runtime for the capability.

PR #1054, opened 2017-02-23
Refactor screen capturing so more is shared between backends

liaxim:
And also fixes the non-working capturing when multiSamples>1


liaxim:
Easiest way to test is to launch GearVRf-Demos/old/gvr-sample and tap anywhere on the screen. Under /sdcard/ there should be a number of screenshot*.png files.

liaxim:
Planning to merge Thursday afternoon

PR #1058, opened 2017-02-24
3.1 NPE fix in GVRConfigurationManager

liaxim:


liaxim:
Tom, this is the most innocent fix I could think of. This pull request is for the 3.1 branch so it can't be committed. If it works, we need to decide on having a 3.1.1 release and what goes into it.

thomasflynn:
ok, this fixes the crash, but now i have a weird problem where it just goes to oculus home after i dock it.  i'm not seeing anything obvious in the logs.  i'll show you tomorrow.

liaxim:
@thomasflynn vr_only is missing from your AndroidManifest.xml. It is very likely your original problem had something to do with it too. Reference manifest: https://github.com/gearvrf/GearVRf-Demos/blob/master/template/GVRFApplication/app/src/main/AndroidManifest.xml

thomasflynn:
Do-oh!  that would likely explain it!  I'll verify later tonight...

thomasflynn:
OK, the fix is still needed.  The manifest entries are necessary for it to launch correctly and not go to occlus home (how long have I been doing this and missed that?), but I still get the crash on dock from developer mode without your fix.

liaxim:
Created https://github.com/Samsung/GearVRf/pull/1062 for the 3.1.1 release instead

PR #1061, opened 2017-02-27
Cleanup of the build files

liaxim:
- copy build artifacts only to GearVRf/GVRf/gearvrf-libs/
- demos, tests will be updated to use the one and only gearvrf-libs
- support for a release build

Demos update: https://github.com/gearvrf/GearVRf-Demos/pull/408


rahul27:
Looks good. I tested it with the samples and it works fine. Have not tested the plugins zip file. 

I see that you got rid of the upload to github script, are we done with the daily releases to github?

Also let me know if you need my help to update the Extensions accordingly?

liaxim:
Since we publish to the maven repo I don't see the value of those daily releases (and having to maintain the associated gradle tasks and shell scripts). We are not currently uploading the snapshots to maven daily but I do it manually every other week. @parthmehta209 has it all documented and it is very easy to do once you set up the credentials.

After we get this merged, it would be nice if you could update the Extensions. Thanks.

liaxim:
Will be double-checking the plugins zips.

liaxim:
While not perfect, this is much cleaner that what we had. "./gradlew clean" removes the published artifacts also.

The extensions will need a similar update.

Do note that there will be one and only place where the build artifacts will be copied: GearVRf/GVRf/gearvrf-libs/. Will update the Tests repo after this is merged.

Build artifacts are copied unconditionally on every run. Less than optimal but would eliminate some strange errors due to aars being outdated.

There is no more ``LIBS_DIRECTORY`` in gradle.properties.

liaxim:

PR #1064, opened 2017-03-02
Fix some Java-side leaks

liaxim:


liaxim:
- Oculus - simpification; eliminate unnecessary global reference to the view manager
- GVRActivity - changes to ensure references to the view manager are cleared; the activity can be kept around by the platform
- GVRAndroidResource.java - use application context since instances of GVRAndroidResource can be kept in static caches
- GVRAssetLoader - make DEFAULT_TEXTURE_PARAMETERS per-instance to avoid leaking a view manager
- GVRComponent - fix for the occasional crash when activities are frequently created and destroyed; might have been impacting the tests
- GVRScriptManager - Lua seems to leak various things most notably global variable bindings; use a proxy that weak-references the actual context instead to minimize the leak impact

We might have to update our Lua library to hopefully pick up bug fixes. The proxy thing can probably be implemented more elegantly that doesn't require updates whenever GVRContext changes, will evaluate in the future.

liaxim:
From the Java side things are better (no view manager and activity leaks) but all these changes amount to very little. The graphics memory (EGL mtrack) usage is more or less the same at the end of a full test run (1GB). Will be making another passes.

liaxim:
Ready for review. According to my tests all the demo apps work.

liaxim:
Can someone please help me verify these changes? Thanks.

rahul27:
@liaxim I can review this after Tuesday. If someone else wants to pick this up before then it is fine with me.

NolaDonato:
Testing results:
framework-tests:
    BoundingVolumeTests: always hangs on testScaledBV
    SceneObjectTests: always hangs on canDisplaySpheres
    PickerTests: always hangs on canPickFromObject
   The rest of the test categories ran to completion.
asset-tests:
    AssetImportTests: run to completion
    AssetLightTests: hung the first time, ran to completion the second time
    AssetTextureTests: ran to completion
   Running the entire suit crashes somewhere in AssetLightTests, not always in the same place

thomasflynn:
@Liahim  how were you looking to see this tested?  how would you like us to verify these leaks are indeed gone?  I can go run through the samples if you want, but not sure if that would tell you enough.

liaxim:
@NolaDonato I'd like to borrow your phone.

@thomasflynn For Java leaks we have http://www.eclipse.org/mat/. Run the automated tests so there are lots of activities created and destroyed by the same process; or comment out everything in GVRActivity.dispatchKeyEvent but ``return super.dispatchKeyEvent(event);`` - it would mean that on pressing the back key the activity will be destroyed and next time you launch the app - a new one created.

Then a heap dump has to be collected via Tools->Android->Android Device Monitor. The heap dump needs to be converted with hprof-conv (command line tool) and then loaded and analyzed in mat. I guess I can explain in person better if someone is willing to go through this.

Otherwise the changes could be reviewed and verified.

liaxim:
Since the samples apps show no problems and the automated tests issues are down to the testing framework itself, I plan to merge this Thursday afternoon.

rahul27:
@liaxim I might have some time today to review this.

GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrViewManager.java
```diff
@@ -417,5 +417,6 @@ public void onRotationSensor(long timeStamp, float rotationW, float rotationX, f
         }
     }
 
-    private static native void drawEyes(long ptr);
+    private native void drawEyes(long ptr);
+    private native void readRenderResultNative(long renderTexture, Object readbackBuffer);
```
rahul27: Was this added by mistake?

GVRf/Framework/framework/src/main/java/org/gearvrf/script/GVRContextProxy.java
```diff
@@ -0,0 +1,352 @@
...
+import org.gearvrf.scene_objects.GVRModelSceneObject;
+
+import java.io.IOException;
+import java.lang.ref.WeakReference;
+import java.util.EnumSet;
+import java.util.List;
+import java.util.concurrent.Future;
+
+public final class GVRContextProxy {
+    private final WeakReference<GVRContext> mContext;
```
rahul27: It is fine to wrap the context behind a weak reference, but for cases where the context is GC'd don't we have to check for a null after the get()?

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRActivity.java
```diff
@@ -66,11 +66,11 @@
     private volatile GVRConfigurationManager mConfigurationManager;
     private GVRMain mGVRMain;
     private VrAppSettings mAppSettings;
-    private static View mFullScreenView = null;
+    private static View mFullScreenView;
```
rahul27: Does this have to be static?

GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrViewManager.java
liaxim: No, intentional. Java's ``this`` is passed implicitly this way and it is needed by the native code.

GVRf/Framework/framework/src/main/java/org/gearvrf/script/GVRContextProxy.java
liaxim: Technically, we should. In this case though I consider it invalid state. If a script tries to use the "gvrf" global to access an activity that has been gc-ed then I rather have an exception thrown since it is not supposed to happen.

GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrViewManager.java
rahul27: Just to be sure, I was talking about the new `readRenderResultNative` native call you added.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRActivity.java
liaxim: Well, this is old code where I just removed the redundant initialization. Probably doesn't have to be static but have not looked at it.

GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrViewManager.java
liaxim: Ok, my bad. Let me see about readRenderResultNative.

liaxim: Removed readRenderResultNative. It was moved to the base class some time ago.

rahul27:
Looks good. Verified it works.

PR #1063, opened 2017-03-02
Add event handler to GVRExternalScene

NolaDonato:
Since GVRExternalScene loads the asset in another thread, an event handler is needed if the caller wants to know when the asset has loaded. This is to support X3D loading FBX, OBJ and other models.

GearVRF DCO signed off by: Nola Donato nola.donato@samsung.com

PR #1065, opened 2017-03-03
Fixed null pointer reference in mesh loader

NolaDonato:
The null pointer was caused by an effort to log a mesh load failure. The asset loader already logs this anyway so I removed the line.


NolaDonato:
rolled into PR #1063

Issue #1024, opened 2017-01-26
Problems when the useAndroidWearTouchpad is set to true


liaxim:
There is a report of ServiceConnection leak being flagged.

Also the sound control (volume up/down) doesn't work anymore.

liaxim:
Fix for the leak merged. The volume UI a separate issue.

PR #1071, opened 2017-03-07
Fix the ServiceConnection leak in GVRAndroidWearTouchpad

rahul27:
* bind to service only if installed
* fixed bugs with the GVRAndroidWearTouchpad code
* added a null check in WearInputService

r.rudradevan@samsung.com


rahul27:
Should address https://github.com/Samsung/GearVRf/issues/1024 and https://github.com/gearvrf/GearVRf-Demos/issues/414

liaxim:
Looks good. Fixes the problem. Will merge Thursday morning.

PR #1076, opened 2017-03-10
Use compileSdkVersion 19 for the framework and the Oculus backend

liaxim:
The app needs to have compileSdkVersion of **21** or remove the daydream backend to use 19.

Expecting further changes in other backends.


liaxim:
Proved to be ineffective

PR #1078, opened 2017-03-13
Support authoring post effect shaders using GVRShaderTemplate

NolaDonato:
Added support for authoring post effect shaders using GVRShaderTemplate the same way we use it to make material shaders. This will be the only way to author shaders eventually since we are deprecating the built-in shaders and the existing method of authoring them in Java.
1.  Derive GVRMaterialShaderId and GVRPostEffectShaderId from a new common class GVRShaderId. We will eventually only have one shader manager for both materials and post effects and one type of shader id.
2. Remove templatization from GVRShaderMaps and GVRShaders based on shader ID type
3. Add GVRShaderMaps.setShaderMapping to provide a way to set the material map for either a material or post effect shader. GVRMaterialShaderManager.setShaderMap and GVRPostEffectShaderManager.setShaderMap remain
3. GVRMaterial, GVRPostEffect getShaderType, setShaderType use GVRShaderId
4. Add GVRPostEffect.getTextureNames()
5. add GVRShaderTemplate.bindShader for post effect shaders


liaxim:
I assume we will be merging this after the 3.2 release.

NolaDonato:
This is for the user who is working on the blur shader. Their deadline is March 27 so I gave them instructions on how to build GearVRF with both of our patches. We don't have to merge into the latest release because I don't think our primary client can wait for that.

NolaDonato:
replaced by #1098

PR #1079, opened 2017-03-13
GVRAssetLoader support reading MTL files from res/raw 

NolaDonato:
OBJ files from res/raw cannot load MTL files at all without this fix -
the asset loader assumes the MTL files are in the "assets" directory.
This patch allows textures and MTL files to be loaded from res/raw if
the 3D model comes from res/raw.


liaxim:
gvr-3dcursor crashes with this pull request in.

NolaDonato:
Fixed the crash in gvr-3dcursor

liaxim:
gvr-accessibility crashes

NolaDonato:
I fixed the gvr-accessibility crash

liaxim:
Looks good. Will merge Friday afternoon.

PR #1080, opened 2017-03-13
IO Framework - Fix mouse to track the camera rig

rahul27:
r.rudradevan@samsung.com


rahul27:
Waiting for @salgubel to give this a :+1:

GVRf/Framework/framework/src/main/java/org/gearvrf/io/GVRMouseDeviceManager.java
```diff
@@ -75,9 +80,17 @@ void removeCursorController(GVRBaseController controller) {
...
             threadStarted = false;
+            if(gvrContext != null){
+                gvrContext.unregisterDrawFrameListener(this);
+            }
         }
     }
 
+    @Override
+    public void onDrawFrame(float frameTime) {
+        thread.invalidate();
```
liaxim: Does this all have to be on the GL thread? There might be a small portion that has to be but most probably doesn't have. The user callbacks ideally should not be on the GL thread. The so called framework main thread seems a good candidate.

liaxim: Or maybe I am not reading it right. There is GVRCursorController.invalidate and there is now one more (which maybe should have been called something like sendInvalidateMessage).

rahul27: Do we have per frame callbacks on the framework thread? Any idea on how to subscribe to it?

On the second comment - perhaps invalidate is not a good name. Will change it.

liaxim: While it would be easy enough to add aforementioned callbacks and also tempting, we need to revisit the whole listeners and events situation since we cannot keep adding stuff.

For now what you could do is:
1) use GVRNotifications on a background thread
2) use onDrawFrameListener as a notification and let the thread pool execute the job

rahul27: @liaxim the drawframelistener in my case does exactly (2).

It does very little and only notifies the thread to invalidate the position.

liaxim:
I take it we are still waiting for @salgubel's feedback.

rahul27:
He tested it and it worked fine.

rahul27:
Updated this PR. The crash is now fixed. 

Also rebased the changes.

liaxim:
Looks good. Will merge Tuesday after lunch.

rahul27:
Update: added a file I missed in the last check in.

salgubel:
+1 verified the changes for the mouse functionality

PR #1081, opened 2017-03-13
Fix for issue #967, "Texture goes missing ..."

liaxim:
Custom shaders can and do get reused. If any of the materials of the scene
objects using the same shader introduces a new texture key, the custom
shader instance will start checking for the key for all scene objects it
renders. If there is no texture for this key, the texture presence check
at the start of CustomShader::render will fail and not render the scene
object.

I am undecided whether in this case something needs to be done about sharing
shaders but am sure that the check by CustomShader:render is both
undesirable and obsolete. Hence I removed it.

Also disabled a bunch of GL error checking by default.

### DIFF
--- a/GVRf/Framework/framework/src/main/jni/shaders/material/custom_shader.cpp
+++ b/GVRf/Framework/framework/src/main/jni/shaders/material/custom_shader.cpp
@@ -116,9 +116,8 @@ void CustomShader::addTextureKey(const std::string& variable_name, const std::st
         glActiveTexture(GL_TEXTURE0 + textureIndex);
         Texture* texture = material.getTextureNoError(key);
         if (nullptr != texture) {
-            glBindTexture(texture->getTarget(), texture->getId());
-            glUniform1i(location, textureIndex++);
-            checkGLError("CustomShader::addTextureKey");
+            GL(glBindTexture(texture->getTarget(), texture->getId()));
+            GL(glUniform1i(location, textureIndex++));
         }
     };
 
@@ -179,6 +178,7 @@ void CustomShader::addAttributeVec4Key(const std::string& variable_name,
             };
     addAttributeKey(variable_name, key, f);
 }
+
 void CustomShader::addUniformKey(const std::string& variable_name,
         const std::string& key, UniformVariableBind f) {
     LOGV("CustomShader::uniform:add variable: %s key: %s", variable_name.c_str(), key.c_str());
@@ -199,7 +199,7 @@ void CustomShader::addUniformFloatKey(const std::string& variable_name,
         const std::string& key) {
     UniformVariableBind f =
             [key] (Material& material, GLuint location) {
-                glUniform1f(location, material.getFloat(key));
+                GL(glUniform1f(location, material.getFloat(key)));
             };
     addUniformKey(variable_name, key, f);
 }
@@ -210,7 +210,7 @@ void CustomShader::addUniformVec2Key(const std::string& variable_name,
     UniformVariableBind f =
             [key] (Material& material, GLuint location) {
                 glm::vec2 v = material.getVec2(key);
-                glUniform2f(location, v.x, v.y);
+                GL(glUniform2f(location, v.x, v.y));
             };
     addUniformKey(variable_name, key, f);
 }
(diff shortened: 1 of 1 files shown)


liaxim:
gvr-simplephysics explained

Scene objects are being added from couple of places using a ctor that is also setting the "main_texture". Later these scene objects switch to the phong shader and also supply a "diffuseTexture".

The platform is a cube. A cube that gets a brand new material by the user code. If that material gets only a "diffuseTexture" it would not render since all previously mentioned scene objects share the same custom shader and check for both "main_texture" and "diffuseTexture".

This is why the workaround of adding a main texture to the cube, even though it is not actually used by the shader, works.

NolaDonato:
I am fixing this monkey business in the renderer reform for Vulkan. The texture shader will dynamically adapt to the presence or absence of a texture, just like the phong shader does now.

liaxim:
This is happening *because* the phong shader is used. Without this change if you put any key without a texture in the material, the whole scene object would not be rendered if it were using the phong shader (or any custom shader for that matter).

liaxim:
@rahul27 Could you please help me verify this?

1) gvr-simplephysics:MainHelpers.java:createGround()
If you comment out ``groundObject.getRenderData().getMaterial().setMainTexture(texture);`` without this change the orange platform would not appear most of the times you launch the app

2) anywhere a phong shader is used, add a texture key but no texture (``setTexture("blah", null)``); the corresponding scene object would not be rendered without this change.

Thanks

rahul27:
@liaxim Verified that the gvr-simplephysics sample works with this pull request. I see the orange platform every time even with the the setMainTexture call commented out.

Reg. this pull request, I am assuming the texture variable "main_texture" belongs to the texture shader and "diffuseTexture" belongs to the phong shader. Switching shaders should essentially also replace the texture variables as well right?

liaxim:
The texture variables belong to the material too so while switching shaders means different texture variables may be used, the material retains whatever it already had. This is the tradeoff for the flexibility of the loose interface in place.

liaxim:

NolaDonato:
Actually what SHOULD happen is that the Phong shader should work like all the other builtin shaders. Unfortunately that would have seriously expanded the scope of the multiple lighting feature and I would still be working on it now :-) So instead of making a way through the existing code for the Phong shader, I grafter on the setShaderTemplate atrocity.
In the Vulkan reform branch I experimented with doing it right and Roshan and I got it working. The tradeoff we made in that branch is that you cannot change the shader associated with a material - GVRMaterial.setShaderType is gone. That way you don't have this inconsistency between the material properties and a different shader than the original. If you want a different shader, you have to make a new material that uses it.

rahul27:
Which is what I was getting at. This issue seems to be more of an improper use of the material/shader apis rather than a framework bug, which can be fixed on the app side by using a different material.

NolaDonato:
To avoid the strangeness with the PhongShader Mihail added a special material ID a while ago: GVRMaterial.GVRShaderType.BeingGenerated.ID. If you make a material with this ID and then use the Phong shader it might fix the problem.

NolaDonato:
If you take out the checks for null texture in CustomShader::render, it may crash if the user omits a texture. I have worked hard on the asset loader to make this case less likely but it is still possible. Right now with the checks in, the object will not render if the texture isn't there.

liaxim:
@NolaDonato I will double-check the case you mention. I do not believe it will crash since before binding there is a nullptr check.

liaxim:
There would be no crashes since all accesses to the texture instance are properly guarded.

In case you omit the texture (pass null or not even set a diffuseTexture) what happens is that you get some random texture that was previously bound for that texture unit.

It looks like the custom shader itself would be a good place to introduce a "missing texture" texture since it can be all covered in one place. I think we all agree that having the object render, with any texture, is better than not rendering it at all.

I still think this is a decent fix. @NolaDonato let me know what you think.

NolaDonato:
OK, I wrote a unit test which deliberately has a missing texture. This fix does not crash that case, even if I explicitly set the texture to null. We should merge it.

liaxim:
Thanks. Will merge Thursday afternoon.

Issue #801, opened 2016-08-11
Crash on loading a new texture to a GVRSphereSceneObject


parthmehta209:
This happens when I try to replace a texture using this code:
  Future<GVRTexture> futureSphereTexture = gvrContext.loadFutureTexture(resource);
gvrSphereSceneObject.getRenderData().getMaterial().setMainTexture(futureSphereTexture);

I do not see this crash if I initialize a new GVRSphereSceneObject every time I want to set a new texture. 

F/libc    (11351): Fatal signal 6 (SIGABRT), code -6 in tid 11400 (GLThread 1072)
I/DEBUG   ( 2939): **\* **\* **\* **\* **\* **\* **\* **\* **\* **\* **\* **\* **\* **\* **\* ***
I/DEBUG   ( 2939): Build fingerprint: 'samsung/zeroltexx/zerolte:5.0.2/LRX22G/G925FXXU1AOCV:user/release-keys'
I/DEBUG   ( 2939): Revision: '10'
I/DEBUG   ( 2939): ABI: 'arm'
I/DEBUG   ( 2939): pid: 11351, tid: 11400, name: GLThread 1072  >>> org.gearvrf.io.sceneeditor <<<
I/DEBUG   ( 2939): signal 6 (SIGABRT), code -6 (SI_TKILL), fault addr --------
I/DEBUG   ( 2939):     r0 00000000  r1 00002c88  r2 00000006  r3 00000000
I/DEBUG   ( 2939):     r4 df0ffdb8  r5 00000006  r6 0000000b  r7 0000010c
I/DEBUG   ( 2939):     r8 f4232bf0  r9 00000011  sl 00000011  fp 00000007
I/DEBUG   ( 2939):     ip 00002c88  sp df0fe100  lr f75b23cd  pc f75d5c7c  cpsr 600d0010
I/DEBUG   ( 2939): 
I/DEBUG   ( 2939): backtrace:
I/DEBUG   ( 2939):     #00 pc 0003ac7c  /system/lib/libc.so (tgkill+12)
I/DEBUG   ( 2939):     #01 pc 000173c9  /system/lib/libc.so (pthread_kill+52)
I/DEBUG   ( 2939):     #02 pc 00017fdb  /system/lib/libc.so (raise+10)
I/DEBUG   ( 2939):     #03 pc 00014925  /system/lib/libc.so (__libc_android_abort+36)
[...]

PR #1082, opened 2017-03-13
Possible fix for bug #801

thomasflynn:
In https://github.com/thomasflynn/GearVRf/tree/bug801, what is happening
is that a new future texture is asked to be loaded, meanwhile the
reference to the existing texture is released.  Since that reference is
released, it gets garbage collected, but the native side don't know
about it and is still trying to use the old texture.

This change holds onto the existing texture reference until the new texture is loaded.  I still haven't evaluated if there are any possibility for leaks with this change.  This change also now calls into question whether the setTexture(key, null) call is actually necessary at line 617.  Still evaluating this.

Putting this change up now to get comments.  Does this look like the
right fix?  Are there other / better ways?  Let me know.

tom.flynn@samsung.com

The bug: https://github.com/Samsung/GearVRf/issues/801


NolaDonato:
Yes the setTexture(key, null) is necessary. It tells GearVRF that this material will have a texture in the future even though it does not have one now. When GVRShaderTemplate.bindShader is called, it will compile a shader that needs a texture instead of one that does not.


thomasflynn:
@NolaDonato thanks.  that explains that mystery then.  it let my sample app run for two hours with no change in memory usage, so i'm feeling a bit safer about this change.

thomasflynn:
The more I think about this, the more I think this is probably the correct fix.

liaxim:
Would it be a more general fix if the null was getting propagated down to the native code and have it deal with it properly instead of trying to guard the native pointer from the Java side? That would be the diff:

```
diff --git a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRMaterial.java b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRMaterial.java
index e97a95a..1e72b93 100644
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRMaterial.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRMaterial.java
@@ -602,8 +602,9 @@ public class GVRMaterial extends GVRHybridObject implements
         {
             tinfo.Texture = texture;
         }
-        if (texture != null)
-            NativeMaterial.setTexture(getNative(), key, texture.getNative());
+
+        final long nativePtr = null != texture ? texture.getNative() : 0;
+        NativeMaterial.setTexture(getNative(), key, nativePtr);
     }
 
     public void setTexture(final String key, final Future<GVRTexture> texture) {
diff --git a/GVRf/Framework/framework/src/main/jni/objects/material.h b/GVRf/Framework/framework/src/main/jni/objects/material.h
index 0c56bce..d38b75c 100644
--- a/GVRf/Framework/framework/src/main/jni/objects/material.h
+++ b/GVRf/Framework/framework/src/main/jni/objects/material.h
@@ -113,7 +113,9 @@ public:
         textures_[key] = texture;
         //By the time the texture is being set to its attaching material, it is ready
         //This is guaranteed by upper java layer scheduling
-        texture->setReady(true);
+        if (nullptr != texture) {
+            texture->setReady(true);
+        }
         if (key == "main_texture") {
             main_texture = texture;
         }
```

thomasflynn:
I would think texture->setReady(false) would need to be called at some point, right?

liaxim:
If you have no texture instance, why would the ready flag matter?

liaxim:
With this change in, gvr-360video shows a black screen. Cleaned and rebuilt couple of times but it is consistent on my Note 5.

NolaDonato:
PR #1084 may fix this problem.

thomasflynn:
Ok, so, I tried Mihail's suggestion above.  it works.  Though the texture being replaced goes black while the new texture is loaded.  

I went back and found the issue with my earlier pull-request.  the call to setTexture() should not have been in an 'else' statement.  It should just be called anyway.  With that change, gvr-360video, simplesample, and others seem to work.

haven't tried Nola's pull-request yet.

thomasflynn:
ok, it looks like Nola's does the same thing plus a little extra.  If you wanna merge that go ahead and I'll close this.

thomasflynn:
we'll take Nola's fix.

PR #1084, opened 2017-03-15
 Fixes loading problem with future textures

NolaDonato:
Synchronized access to internal texture table in GVRMaterial
Do not set a texture key to null if it exists
This fixes issue #967


liaxim:
Doesn't fix #967. The platform disappears in gvr-physics if the ``MainHelper:createGround:groundObject.getRenderData().getMaterial().setMainTexture(texture)`` is commented out.

liaxim:
Does fix #801.

liaxim:
Looks good. Will merge Monday afternoon.

Issue #1077, opened 2017-03-11
Improvement: support models with vertex colors baked in.


mvasilkov:
**UPD.** I've noticed that while Assimp loads vertex colors, the GearVRf adapter doesn't copy this data to the mesh. Hence changing the title of the issue. I think this would be a neat feature for authoring simple models, for prototyping and such.

**Original issue:**

Hi dear all, I'm new here, and I'm trying to render a *.ply model from file. The model has vertex colors baked in.

I'm using `GVRAssetLoader::loadModel()`, which seems to work. When I add this model to the scene, it's light gray.

I can `.setDiffuseColor()` on a material and change the color of the model wholesale, but I can't seem to make it use vertex colors from the model. This seems to be unsupported by the built-in shaders, is this correct?

Can you please advise how to proceed? (My knowledge of GLSL can be summarized by an empty string: "".)

NolaDonato:
The built-in shaders do not currently support vertex colors. You could write your own shader that looks at vertex colors but it would not work because our asset importer might not preserve the vertex colors in import. I will look into adding support for "foreign" vertex components into the asset loader and write a sample of how to do vertex colors. I might not get to it until Weds though.

NolaDonato:
I worked on this issue and made some progress. I got the asset loader to load a mesh with vertex colors and wrote a shader which used them. Unfortunately GearVRf crashes in the middle of creating mesh attributes inside the renderer. I will debug further later this week.

NolaDonato:
My first crack at solving vertex colors is in PR #1085. I could only find one free PLY model with vertex colors. Assimp imports this as a 4 floats for each color. I have not tried it with other PLY models. Maybe  @mvasilkov  would like to test this pull request with other PLY models.

liaxim:
@mvasilkov Nola added support. She even added a test case for it; see https://github.com/gearvrf/GearVRf-Tests/pull/67. Assuming it is all good if we don't hear from you.

PR #1085, opened 2017-03-15
GVRAssetLoader import vertex colors

NolaDonato:
This patch adds support for importing vertex colors from assimp. It does not affect X3d models.
Addresses enhancement request #1077


mwitchwilliams:
There is an X3D Color node that GearVR's X3D implementation does not currently support.  The Color node includes a boolean colorPerVertex.  With this added support in GearVRf, it seems we could now support X3D's Color node when colorPerVertex is true.  When colorPerVertex is false, X3D expects each color for a polygon, which there is no plan to support in GearVR.

NolaDonato:
Great! Try to add the support in X3D and import some models with vertex colors. If that works, we can add X3D vertex color support to the pull request to make it complete.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRJassimpAdapter.java
```diff
@@ -134,6 +137,22 @@ public GVRMesh createMesh(GVRContext ctx, AiMesh aiMesh) {
             }
         }
 
+        // Vertex Colors
+        for(int c = 0; c < MAX_VERTEX_COLORS; c++) {
+            FloatBuffer fbuf = aiMesh.getColorBuffer(c);
+            if (fbuf != null) {
+                FloatBuffer coords = FloatBuffer.allocate(aiMesh.getNumVertices() * 4);
+                FloatBuffer source = aiMesh.getColorBuffer(c);
+                String name = "a_color";
```
thomasflynn: quick question on this.  if the model does have multiple 'vertex color' arrays (for whatever reason), it looks like they'll all get the same name.  that would be a problem, wouldn't it?

NolaDonato: The full source looks like this (c is the color set index):
                String name = "a_color";

                if (c > 0) {
                    name += c;
                }

thomasflynn: ah, somehow i missed that.  Ok, thanks.  Code looks good then.

thomasflynn:
ok, with @NolaDonato 's help, I was able to run the asset test and verify this loads the model as expected.  +1 Verified.  +1 Looks good to me.

PR #1088, opened 2017-03-16
Support Oculus Mobile SDK 1.5.0

liaxim:


rahul27:
Looks good. Tested a few samples and it worked fine.

My only comment would be to rename the commit from 1.0.5 to 1.5.0 to avoid confusion about the version of the SDK.

liaxim:
Fixed the incorrect version

thomasflynn:
keep in mind we need to update any READMEs / wikis that reference the oculus version.

GVRf/Framework/backend_oculus/src/main/jni/ovr_activity.cpp
```diff
@@ -196,14 +195,15 @@ void GVRActivity::onDrawFrame() {
 
     for ( int eye = 0; eye < VRAPI_FRAME_LAYER_EYE_MAX; eye++ )
     {
-        ovrFrameLayerTexture& eyeTexture = parms.Layers[VRAPI_FRAME_LAYER_TYPE_WORLD].Textures[eye];
+        ovrFrameLayerTexture& eyeTexture = parms.Layers[0].Textures[eye];
```
roshanch: we are always using 0th layer?

liaxim: We have one TimeWarp layer so 0 would be it.

thomasflynn:
also, this sdk supposedly supports 64bit.  so, perhaps, we no longer need to strip out 64bit libs in the gradle files?

liaxim:
Yes, I will test the 64bit support and update the gradle files.

rahul27:
Can we add the 64 bit support in a separate pull request. I need this one for the controller support.

liaxim:
Sure, I can have the 64bit related changes as a separate thing.

liaxim:
Planning to merge Tuesday morning.

PR #1089, opened 2017-03-16
Fix the packageForPlugins task execution

liaxim:
The file existance check it used to make is executed during the
so called configuration phase where nothing has been built yet.
The check is not strictly needed anyway.


liaxim:
Merging to get the jenkins build going

PR #1096, opened 2017-03-17
Backend's setMain delegate returns result; call onUndock from onDestroy if requested

liaxim:


liaxim:
@sukhwanyoon needs handleOnDock to be called from onCreate. It was a problem in the past. Will look into making it possible.

sukhwanyoon:
By the way, regardless of the handleOnDock issue, backend setMain having a returning value should be there when it comes to how landscape/portrait apps are handled appropriately by android platform.

sukhwanyoon:
Hi. Mihail,
It works for me now.
Thanks.
Sean,

liaxim:
@sukhwanyoon Since you've given thumbs up, I am planning to merge this pull request Thursday afternoon.

PR #1098, opened 2017-03-20
Support authoring post effect shaders using GVRShaderTemplate

NolaDonato:
Added support for authoring post effect shaders using GVRShaderTemplate the same way we use it to make material shaders. This will be the only way to author shaders eventually since we are deprecating the built-in shaders and the existing method of authoring them in Java.
1.  Derive GVRMaterialShaderId and GVRPostEffectShaderId from a new common class GVRShaderId. We will eventually only have one shader manager for both materials and post effects and one type of shader id.
2. Remove templatization from GVRShaderMaps and GVRShaders based on shader ID type
3. Add GVRShaderMaps.setShaderMapping to provide a way to set the material map for either a material or post effect shader. GVRMaterialShaderManager.setShaderMap and GVRPostEffectShaderManager.setShaderMap remain
3. GVRMaterial, GVRPostEffect getShaderType, setShaderType use GVRShaderId
4. Add GVRPostEffect.getTextureNames()
5. add GVRShaderTemplate.bindShader for post effect shaders
 This replaces PR #1078 which had extra files init.


thomasflynn:
i don't know how it was missed the first time, but could you add javadoc for the methods that are public (even if you didn't add them)?  Otherwise, even though it touches 13 files, it does look simpler than it first appears.  I'll do another pass of it tomorrow.  so far, it looks fine.

liaxim:
gvr-accessibility, gvr-inputsample do not compile. Probably it is not just these two.

NolaDonato:
I have fixed the compilation errors due to minor API changes which I reverted.

liaxim:
Question: should this be in the 3.2 milestone? I would guess no.

NolaDonato:
Yes I think it should. It actually improves the API a bit because it no longer relies on stuff in the scene object being updated by another thread.

liaxim:
The gvr-blurfilter has a problem that may or may not require updates to this pull request.

liaxim:
I guess the gvr-blurfilter problem will be dealt with in the app or in a future gvrf version. So this can be merged. Planning to do so Friday lunch-time.

Issue #940, opened 2016-11-16
The shadows in gvr-shadows "jump"


liaxim:
I haven't seen this before but today I am. The shadows kind of "jump" every once in a while. Maybe I am just doing something wrong. I picked up the binaries from the 3.1 prerelease and build with those. Tested on Note5. @NolaDonato Could you please double-check?

[video.zip](https://github.com/Samsung/GearVRf/files/595381/video.zip)

liaxim:
Confirmed as a bug. @NolaDonato and @deepakrawat22 are looking into it.

NolaDonato:
What I am seeing is that the entire scene appears to flicker to black periodically as the light moves around.

thomasflynn:
this can be closed now, right?

liaxim:
Yes. Fixed.

PR #1099, opened 2017-03-20
Refactored shadowmap code, exposed render to texture

NolaDonato:
This patch refactors the shadow mapping code from a hard-coded mess in
light.cpp into a set of general components that implements shadow
mapping. Now you can render the shadow maps because it is implemented on
top of RenderTexture.
1. Added new component GVRRenderTarget which will render to a texture
from the viewpoint of the scene object it is attached to. This component
renders to a GVRRenderTexture via a DrawFrameListener.
2. Added GVRContext.cullAndRender function to cull and render to a
GVRRenderTarget. This is what GVRRenderTarget calls each frame.
3. Added GVRShadowMap component which derives from GVRRenderTarget and
encapsulates the shadow mapping code which was formerly spewed all over
light.cpp
4. Added GVRRenderTextureArray to expose an array of render textures
(which is what shadow mapping uses).

Fixes issue #940 (jumpy shadows)


GVRf/Framework/framework/src/main/java/org/gearvrf/GVRDirectLight.java
```diff
@@ -184,10 +182,23 @@ public void setDiffuseIntensity(float r, float g, float b, float a) {
      * @param b blue component (0 to 1)
      * @param a alpha component (0 to 1)
      */
-    public void setSpecularIntensity(float r, float g, float b, float a) {
+    public void setSpecularIntensity(float r, float g, float b, float a)
+    {
         setVec4("specular_intensity", r, g, b, a);
     }
 
+    public void setCastShadow(boolean enableFlag)
```
liaxim: Javadoc

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRRenderTarget.java
```diff
@@ -0,0 +1,134 @@
...
+        return mCamera;
+    }
+
+    /**
+     * Sets the texture this render target will render to.
+     * If no texture is provided, the render target will
+     * not render anything.
+     * @param texture GVRRenderTexture to render to.
+     */
+    public void setTexture(GVRRenderTexture texture)
```
liaxim: Might be better if it is package private. Having this component should imply creating the render texture for the user. getTexture should probably return just a GVRTexture. Ideally end users should not be aware of GVRRenderTexture.

liaxim:
Nice pull request. Does fix #940. No problems found.

@roshanch @rahul27 Please take a look at the changes too.

roshanch:
looks good to me.

liaxim:
One of the commits is the java leaks change that we haven't yet merged. You either have to remove the commit or we can wait until after the java leaks change has been merged.

liaxim:
There are also couple of methods that need javadocs. Please either add that or make them non-public.

liaxim:
Other than that everything worked for me and we can merge after the minor non-code issues are addressed.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRSpotLight.java
```diff
@@ -68,26 +66,26 @@
     private static String fragmentShader = null;
     private static String vertexShader = null;
     private boolean useShadowShader = true;
-    private Matrix4f biasMatrix = null;
+    private boolean mChanged = true;
```
liaxim: Is this flag necessary?

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRCamera.java
```diff
@@ -309,6 +309,15 @@ public int getChildrenCount() {
         return 0;
     }
 
+    public void setViewMatrix(float[] matrix)
```
NolaDonato: I added a JavaDoc for this. It is public and is used to set up the view matrix for a camera that is not attached to a scene object so it can render to a  texture.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRDirectLight.java
NolaDonato: Technically these are overrides but I documented them anyway.

liaxim:
Needs rebase

NolaDonato:
I have rebased.

liaxim:
Planning to merge Friday morning

PR #1101, opened 2017-03-22
Adding Gear Controller support to GVRf

rahul27:
r.rudradevan@samsung.com

### DIFF
--- a/GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrGearController.java
+++ b/GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrGearController.java
@@ -1,0 +1,352 @@
+/*
+/* Copyright 2017 Samsung Electronics Co., LTD
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gearvrf;
+
+import android.graphics.PointF;
+import android.os.Handler;
+import android.os.HandlerThread;
+import android.os.Message;
+import android.view.KeyEvent;
+
+import org.gearvrf.io.CursorControllerListener;
+import org.gearvrf.io.GVRControllerType;
+
+import org.gearvrf.io.GVRInputManager;
+import org.joml.Quaternionf;
+import org.joml.Vector3f;
+
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.nio.FloatBuffer;
+
+/**
+ * This class represents the Gear Controller.
+ *
+ * The input manager notifies the application when the controller has connected successfully. Use
+ * the {@link GVRInputManager#addCursorControllerListener(CursorControllerListener)} to get notified
+ * when the controller is available to use. To query the device specific information from the
+ * Gear Controller make sure to type cast the returned {@link GVRCursorController} to
+ * {@link OvrGearController} like below:
+ *
+ * <code>
+ * OvrGearController controller = (OvrGearController) gvrCursorController;
+ * </code>
+ *
+ * Additionally register a {@link ControllerEventListener} using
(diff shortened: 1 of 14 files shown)


thomasflynn:
+1 verified with a controller with https://github.com/gearvrf/GearVRf-Demos/pull/427

GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrGearController.java
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    @Override
+    public void setSceneObject(GVRSceneObject object) {
+        context.getMainScene().addSceneObject(pivot);
+        pivot.addChildObject(object);
+    }
+
+    @Override
+    public void onDrawFrame(float frameTime) {
```
liaxim: I doubt it is a problem but isn't the controller data being consumed one frame later? It is produced by frame N on the native side by GVRActivity::onDrawFrame which executes after the drawFrameListeners for the current frame. So the frameListener on the Java side will run in frame N+1.

Also I think it might be simpler if the native code invoked the Java controller with the produced data, if any, immediately. The JNI overhead hasn't been a problem so far and also we could combine some of the JNI calls together.

GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrGearController.java
```diff
@@ -0,0 +1,352 @@
...
+        thread = new EventHandlerThread();
+        context.registerDrawFrameListener(this);
+    }
+
+    long getPtr() {
+        return mPtr;
+    }
+
+    @Override
+    public void setSceneObject(GVRSceneObject object) {
```
liaxim: The base class exposes resetSceneObject method that calls setSceneObject with null parameter. Would calling resetSO work as expected for OvrGearController?

GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrGearController.java
```diff
@@ -0,0 +1,352 @@
...
+            return null;
+        }
+        return thread.currentControllerEvent.handedness == 0.0f ? Handedness.LEFT : Handedness
+                .RIGHT;
+    }
+
+    @Override
+    protected void finalize() throws Throwable {
+        super.finalize();
+        OvrNativeGearController.delete(mPtr);
```
liaxim: Might also try to stop the event thread?

GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrGearController.java
rahul27: Valid point. I have refactored the code to piggyback on the onDrawFrame() called by the view manager.

GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrGearController.java
rahul27: Think I missed the resetSceneObject call the first time. Added it now. Thanks.

GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrGearController.java
rahul27: Done.

GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrViewManager.java
```diff
@@ -348,6 +352,10 @@ void onDrawEye(int eye) {
     protected void onDrawFrame() {
         beforeDrawEyes();
         drawEyes(mActivity.getActivityNative().getNative());
+
+        // update the gear controller
```
liaxim: Have you tried to measure the overhead in the case there is no controller? In my non-scientific test this takes ~0.06ms which is not that much but then it is not free either. Have seen it spike a lot too.

We can merge it like this but please examine it again in the future. You could call the controller's onDrawFrame from the native GearController::onFrame directly, if there is a controller. The check by findConnectedGearController is consistently fast.

Also the Java Buffers may not always be the universally best solution. I guess you need to have a large enough of payload to gain something out of them. For what we are passing around, copies may be simpler and faster.

liaxim:
Looks good. Will merge Friday afternoon if there are no comments.

GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrViewManager.java
rahul27: Sure. I can run a few tests with your suggested changes in place in the future.

PR #1103, opened 2017-03-22
Fix import of shininess in X3D

NolaDonato:
The X3D shininess value is between 0 and 1. The X3D spec indicates this should be scaled by 128 to match OpenGL shininess. X3D shininess does NOT correspond to the roughness parameter in physically based shading models.


liaxim:
@mwitchwilliams ?

mwitchwilliams:
Same implementation as I suggested.  Better to set this in the parser as this fix does than in the shader.  I support this change.  LGTM (Looks Good to Me).

PR #1109, opened 2017-03-24
Added GVRAndroidResource.setStream for custom I/O

NolaDonato:
This allows subclasses to provide an input stream they have previously opened.
Also removed unused arguments from GVRAssetLoader.loadX3dModel


GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAndroidResource.java
```diff
@@ -313,6 +313,23 @@ public synchronized final InputStream getStream() throws IOException{
...
     }
 
+    /**
+     * Sets the stream for a resource.
+     * This function allows you to provide a stream that is already open to
+     * an existing resource. It will throw an exception if that resource
+     * already has an open stream.
+     * @param s InputStream currently open stream to use for I/O.
+     */
+    protected synchronized void setStream(InputStream s)
```
rahul27: GVRAndroidResource already has a constructor that accepts an input stream - look at line 208. Might have to make it public to access it.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAndroidResource.java
```diff
@@ -313,6 +313,23 @@ public synchronized final InputStream getStream() throws IOException{
         return stream;
     }
 
+    /**
```
liaxim: Why not make ``GVRAndroidResource(String name, InputStream inputStream)`` public or a variant of it? This method circumvents all the existing ctors since you can use any of them and then override the stream with this method. It completely disables the encapsulation GVRAndroidResource is trying to do.

NolaDonato: Yes, it has a CONSTUCTOR that accepts an input stream. Problem is, the asset loader constructs the resource in one place and opens it in another. The method is protected so that only subclasses, like the one Srini is trying to write, can provide an open stream.

NolaDonato: Yes, it has a CONSTUCTOR that accepts an input stream. Problem is, the asset loader constructs the resource in one place and opens it in another. The method is protected so that only subclasses, like the one Srini is trying to write, can provide an open stream.

liaxim: To me it seems to be by design, not a problem. But since you don't see an issue then we'll merge after @rahul27 gives his ok.

rahul27: The API itself seems confusing since for one like @liaxim mentions, it breaks the clean encapsulation provided by GVRAndroidResource. What happens when I create a NETWORK type resource and then use the setStream call? Does it still remain a GVRAndroidResource of type NETWORK or does it become something else? 

I guess we might get a better idea if we see an example of this API in use. To me we could still work with the constructor where a GVRAndroidResource created using the input stream can be provided to the assetloader to be loaded.

NolaDonato:
The issue is that their project needs to intercept all I/O coming from the asset loader and ask Android for permission to do it. Overriding GVRResourceVolume allows you to get control over creating and opening GVRAndroidResources needed by the asset loader. The catch is that you asset loader first creates the resource, then checks if it is in the cache and then opens it. Their project doesn't have an open stream at the time the resource is created - they don't know the file name yet.

rahul27:
Seems to be a specialized use case. I guess it's fine to merge.

liaxim:
Will merge after lunch on Monday if there are no other comments.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAndroidResource.java
NolaDonato: You can't use that call unless you are a subclass of GVRAndroidResource - it is protected. If the subclass overrides GVRResourceVolume.openResource to make an application-specific subclass of GVRAndroidResource, that subclass needs a way to provide an open stream. The GVRAndroidResource is responsible for opening the stream so it needs to override openStream to do so. Once it opens the stream, it needs to set the private "stream" variable to the open InputStream object. Short of making a protected function to do so I cannot think of a better way. Making "stream" and "streamStates" protected exposes even more of the implementation to subclasses.

What I think we SHOULD do is to make openStream private and getStream public. That way to subclass doesn't have to provide a stream at all - it can just use its own internal stream. That is an API change that I would prefer not to make before release 3.2

liaxim: This brings the question whether GVRAndroidResource was ever intended to be derived from. Unless explicitly designed for that then you have to assume it is not. Let's see how it is used by Srini then we'll re-evaluate whether this is the only possible way of supporting what he needs. Having users derive from our classes can be a burden long term.

PR #1110, opened 2017-03-24
Unify screen capture code across backends

liaxim:
- provides 4 apis all backends use to perform capturing
- changed the fbo size for daydream to be what the rest of the framework expects when default values are used; it was too big anyway; should eventually be derived from the requested configuration
- all implementation moved to the common framework
- tweaked readRenderResultNative so it is aware of the current viewport

### DIFF
--- a/GVRf/Framework/backend_daydream/src/main/jni/daydream_renderer.cpp
+++ b/GVRf/Framework/backend_daydream/src/main/jni/daydream_renderer.cpp
@@ -15,10 +15,10 @@
 
 #include "daydream_renderer.h"
 #include "glm/gtc/matrix_inverse.hpp"
-#include <assert.h>
 
 namespace {
     static const uint64_t kPredictionTimeWithoutVsyncNanos = 50000000;
+    static const int kDefaultFboResolution = 1024;
 
     // Use the same default clipping distances as the perspective camera
     // TODO: Change this to read the values from the gvr.xml file
@@ -34,21 +34,6 @@ namespace {
         }
         return result;
     }
-
-    static gvr::Mat4f MatrixMul(const gvr::Mat4f &matrix1,
-                                const gvr::Mat4f &matrix2) {
-        gvr::Mat4f result;
-        for (int i = 0; i < 4; ++i) {
-            for (int j = 0; j < 4; ++j) {
-                result.m[i][j] = 0.0f;
-                for (int k = 0; k < 4; ++k) {
-                    result.m[i][j] += matrix1.m[i][k] * matrix2.m[k][j];
-                }
-            }
-        }
-        return result;
-    }
-
 
     static gvr::Rectf ModulateRect(const gvr::Rectf &rect, float width,
                                    float height) {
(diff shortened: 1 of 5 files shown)


liaxim:
This is it for this pull request. @rahul27 Please take a look at the Daydream tweaks, thanks.

There has always been a bug where the column 0 of the captured imaged is a copy of the very last column. All the golden images are tainted. Did not find the reason quickly and will examine it closer when I switch the capturing to use PBOs; and when we are ready to regenerate all the golden images too.

liaxim:
The aforementioned issue is tracked as https://github.com/Samsung/GearVRf/issues/1112

GVRf/Framework/backend_daydream/src/main/jni/daydream_renderer.cpp
```diff
@@ -104,9 +81,9 @@ DaydreamRenderer::~DaydreamRenderer() {
 void DaydreamRenderer::InitializeGl() {
     gvr_api_->InitializeGl();
 
-    // Because we are using 2X MSAA, we can render to half as many pixels and
-    // achieve similar quality.
-    render_size_ = HalfPixelCount(gvr_api_->GetMaximumEffectiveRenderTargetSize());
+    //@todo read gvr.xml and obtain the values from EyeBufferParams
+    render_size_.height = kDefaultFboResolution;
+    render_size_.width = 2*kDefaultFboResolution;
```
rahul27: aren't the default values 1024 x 1024?

liaxim: Yes. Daydream seems to use one FBO for both eyes though.

rahul27: The default daydream sample does, but in our case we explicitly create two separate FBOs for GVRf.

liaxim: We render though either to left or to the right half of these. So they have to be sized to accommodate two eyes.

rahul27: Ahh I see. I thought I had fixed that. In any case will submit a fix to repair these wasted half buffers.

NolaDonato:
Is it possible to get rid of the call to ImageUtils.generateBitmapFlipV? It is a really slow Java routine that flips the bitmap vertically? We can do the same thing by changing the camera view matrix before we capture the screen shot.

GVRf/Framework/backend_daydream/src/main/jni/daydream_renderer.cpp
rahul27: Fixed here https://github.com/Samsung/GearVRf/pull/1113/commits/482dc40ed48f2eb2b5130227f302177a701ce6f7

rahul27:
Verified that the screenshot APIs work with Daydream.

liaxim:
@NolaDonato Good idea. Will make the change. Should be possible for all but left/right eye capture. In that case the FBO that is handed off to the backend is being read.

liaxim:
Ready for review. No longer making the generateBitmapFlipV for the 3d screenshot and the center camera screenshot. Ended up using the HorizontalFlip post effect to achieve that. @NolaDonato It would be helpful if you could let me know the automated tests still work the same for you.

NolaDonato:
I will test this today.

liaxim:
I am under the impression that the slowest operation in the automated tests is the image comparison. So this change is not going to lead to a noticeable speed up.

NolaDonato:
I have tested this pull request with all of the unit tests. The tests seem to run a lot better now than they used to. I still see occasional hangs but they are much less frequent. All of the screen shots that compare correctly on master also compare correctly with this pull request.

PR #1119, opened 2017-03-31
fix pragma strings for shaders with GL_OES_EGL_image_external extension

roshanch:
It tries for other extension string if current one gives failure.


liaxim:
I would expect this to query for the presence of the extension and proceed from that. The compilation of a shader could fail for so for many reasons. Also GL_OES_EGL_image_external and GL_OES_EGL_image_external_essl3 may mean the use of different sampling functions from the looks of it. Are we going to patch this up too?

I personally do not think we should take the responsibility of making shaders "portable". Not unless we are planning to provide a whole abstraction layer on top of GLSL so shaders are written against that. And then have a good documentation and justification for making people go through this. Us parsing and patching shaders like this doesn't seem scalable to me.

Again, for this release the problem is most likely best dealt with on the app side (referring to gvr-blurfilter). It has all the tools at its disposal.

roshanch:
I tried declaring both strings with adreno and Mali s7 edge and it works. with s7 mali, it gives error for both the strings, seems to be driver issue. But same device works very well with gvr-video which uses glsl100.

liaxim:
@roshanch Are we going to close this one?

Issue #1100, opened 2017-03-20
SIGSEGV's reported by gvr-simplephysics


rahul27:
Here are the two stacktraces. Note that I am using Oculus SDK 1.5.0

````
03-21 06:47:02.066 18831 18831 F DEBUG   : backtrace:
03-21 06:47:02.066 18831 18831 F DEBUG   :     #00 pc 001b1d96  /data/app/org.gearvrf.gvrsimlephysics-1/lib/arm/libgvrf.so (_ZNSt8_Rb_treeIiSt4pairIKibESt10_Select1stIS2_ESt4lessIiESaIS2_EE8_M_beginEv+9)
03-21 06:47:02.066 18831 18831 F DEBUG   :     #01 pc 001b4885  /data/app/org.gearvrf.gvrsimlephysics-1/lib/arm/libgvrf.so (_ZNSt8_Rb_treeIiSt4pairIKibESt10_Select1stIS2_ESt4lessIiESaIS2_EE4findERS1_+24)
03-21 06:47:02.066 18831 18831 F DEBUG   :     #02 pc 001b4159  /data/app/org.gearvrf.gvrsimlephysics-1/lib/arm/libgvrf.so (_ZNSt3mapIibSt4lessIiESaISt4pairIKibEEE4findERS3_+16)
03-21 06:47:02.066 18831 18831 F DEBUG   :     #03 pc 001b31a9  /data/app/org.gearvrf.gvrsimlephysics-1/lib/arm/libgvrf.so (_ZN3gvr5Light6renderEii+44)
03-21 06:47:02.066 18831 18831 F DEBUG   :     #04 pc 001ea861  /data/app/org.gearvrf.gvrsimlephysics-1/lib/arm/libgvrf.so (_ZN3gvr12CustomShader6renderEPNS_11RenderStateEPNS_10RenderDataEPNS_8MaterialE+1204)
03-21 06:47:02.066 18831 18831 F DEBUG   :     #05 pc 00185545  /data/app/org.gearvrf.gvrsimlephysics-1/lib/arm/libgvrf.so (_ZN3gvr10GLRenderer20renderMaterialShaderERNS_11RenderStateEPNS_10RenderDataEPNS_8MaterialE+1468)
[...]

PR #1122, opened 2017-04-03
Possible fix for physics crashes

NolaDonato:
This patch synchronizes access to the native physics step() function and does not allow it to be called if it is still processing the last frame. This MAY fix some of the crashes in the physics apps. It needs stress testing before we can declare it a true fix.

Fixes for https://github.com/Samsung/GearVRf/issues/1100


GVRf/Framework/framework/src/main/jni/objects/components/render_data.cpp
```diff
@@ -36,9 +36,14 @@ const RenderPass* RenderData::pass(int pass) const {
...
-    mesh_ = mesh;
-    mesh->add_dirty_flag(dirty_flag_);
-    *dirty_flag_ = true;
+    if (mesh != nullptr) {
+        mesh_ = mesh;
+        mesh->add_dirty_flag(dirty_flag_);
+        *dirty_flag_ = true;
+    }
+    else {
+        LOGE("RenderData::set_mesh mesh cannot be NULL");
```
liaxim: It should probably std::terminate in addition to the log trace. Or just use the FAIL macro from gvr_log.h.

GVRf/Extensions/gvrf-physics/src/main/java/org/gearvrf/physics/GVRWorld.java
```diff
@@ -49,6 +51,8 @@ public GVRWorld(GVRContext gvrContext) {
     public GVRWorld(GVRContext gvrContext, GVRCollisionMatrix collisionMatrix) {
         super(gvrContext, NativePhysics3DWorld.ctor());
         mHasFrameCallback = false;
+        mIsProcessing = false;
```
liaxim: What is setting mIsProcessing to true? Are there changes missing from this pull request?

GVRf/Framework/framework/src/main/jni/objects/components/render_data.cpp
```diff
@@ -36,9 +36,14 @@ const RenderPass* RenderData::pass(int pass) const {
 }
 
 void RenderData::set_mesh(Mesh* mesh) {
-    mesh_ = mesh;
-    mesh->add_dirty_flag(dirty_flag_);
-    *dirty_flag_ = true;
+    if (mesh != nullptr) {
```
liaxim: This nullptr check prevents in the crash in NativePhysics3DWorld.step. Though I believe the better fix is to synchronize all access to GVRWorld.mRigidBodies.

liaxim:
This is much more than a crash fix by now and is not affecting just the physics extension. Some things could use few lines of explanation what problems are being fixed and how are those reproduced/verified as being fixed.

Since end users use the Java APIs to deal with Bullet there is still some locking needed on the Java side too and there will be overlap with the locking on the native side. This may very well fix all the problems with gvr-simplephysics but requires extensive testing better done after the release. I already have few comments and will have some more in the future.

liaxim:
gvr-simplephysics with the modifications to cause more crashes is not crashing with these changes in.

NolaDonato:
This pull request will accumulate everything necessary to make the demos work. I don't think we should check this in now. There are still more bugs in physics. This patch just fixes obviously bogus stuff. Neither physics nor GearVRf locks around updating transforms but they can both update them simultaneously. This will never be thread safe and we need to fix it. I  think we should hold the release to fix physics.

liaxim:
If we keep expanding the scope of the release then more things will crop up in the meantime and we are not going to have a release any time soon.

NolaDonato:
replaced by #1141 and #1140

PR #1124, opened 2017-04-03
Fix problem with picking test "canPickMeshWithObject" 

NolaDonato:
This test creates a picker, starts listening for pick events and then attaches the picker to a scene object. This can cause false events to be generated between the time the picker starts listening and the time it is attached. This constructor eliminates the problem by attaching the picker first and then starting to listen for events.


liaxim:
How come false events can be generated in first place?

liaxim:
Bundles the changes from the https://github.com/Samsung/GearVRf/pull/1122 pull request. Can't merge unless https://github.com/Samsung/GearVRf/pull/1122 is ready then.

NolaDonato:
#1122 doesn't actually fix all the  physics crashes - one is actually due to shadow mapping. But it does implement some thread safety that was lacking and guards against a null pointer. So it does no harm and does prevent re-entrant calls to the physics solver.

NolaDonato:
replaced by #1129

Issue #1104, opened 2017-03-22
Possible bug on 3.1.2-SNAPSHOT


felislynx-silae:
Ever since i've made update to 3.1.2 (for daydream support), i've random exceptions like this on gear
```
Exception java.util.ConcurrentModificationException:
java.util.ArrayList$ArrayListIterator.next (ArrayList.java:573)
org.gearvrf.GVRBaseSensor.processList (GVRBaseSensor.java:134)
org.gearvrf.SensorManager.processPick (SensorManager.java:88)
org.gearvrf.GVRCursorController.update (GVRCursorController.java:618)
org.gearvrf.GVRCursorController.setPosition (GVRCursorController.java:423)
org.gearvrf.io.GVRGazeCursorController.access$101 (GVRGazeCursorController.java:30)
org.gearvrf.io.GVRGazeCursorController$EventHandlerThread$1.handleMessage (GVRGazeCursorController.java:249)
android.os.Handler.dispatchMessage (Handler.java:98)
android.os.Looper.loop (Looper.java:158)
android.os.HandlerThread.run (HandlerThread.java:61)
```

liaxim:
@felislynx-silae Do you have any hints on how to reproduce? Does it happen with any of our demo apps? Thanks.

@rahul27 Let me know if you are too busy. I could look into it.

senseearth:
@felislynx-silae Hi there. Would be happy to help you with it. To help us reproduce, it would be great to share with us

1. A simplified demo project with source code
2. Steps to have this occur, and frequency, and if you are doing some multithreading as well
3. And other hardware factors such as is it gearvr only or daydream as well, and the particular device you use

Thanks

felislynx-silae:
1. I'll have to prepare demo project with sourcecode i could share
2. I'ts on startup of app (sometimes on going back to this scene) where i have 12 sceneobjects with selectable behavior (one on the middle is receiving event.isOver() event even if i'm not looking at it) only multithreading that is happening is retrofit call to our server which is loading data.
3. I didn't checked on daydream since this version is on early stage of development. Reference device is Galaxy S6, Android 6.0.1 most up to date oculus drivers. It happens on second S6 and on S7 with 7.0 Android

rahul27:
@liaxim I can take a look at it when @felislynx-silae shares the demo code.

rahul27:
Hi @felislynx-silae, do you have an update for us?

Also is this something you see with a stock sample as well?

[...]

PR #1128, opened 2017-04-04
Fix threading issues with the IO framework

rahul27:
* Moved the setScene and invalidate calls to the device threads
* Fixed an issue which caused a crash with the laser cursor
* Also fixed a few minor issues with the framework while
  using the 3dcursor sample

r.rudradevan@samsung.com

### DIFF
--- a/GVRf/Extensions/3DCursor/3DCursorLibrary/src/main/java/org/gearvrf/io/cursor3d/Cursor.java
+++ b/GVRf/Extensions/3DCursor/3DCursorLibrary/src/main/java/org/gearvrf/io/cursor3d/Cursor.java
@@ -649,7 +649,8 @@ public abstract class Cursor {
 
     protected void handleControllerEvent(GVRCursorController controller, boolean sentEvent) {
         lookAt();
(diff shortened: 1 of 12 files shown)


rahul27:
Should fix https://github.com/Samsung/GearVRf/issues/1104

liaxim:
Looks ok. Did not find problems with our demos and with Srini's app but it would be nice if we get his explicit approval. Then we will schedule it for merging.

The callParentSetEnable/callParentSetScene/callParentInvalidate vs setEnable/setScene/invalidate is confusing and I read it as a problem with the hierarchy. Not a major one though.

rahul27:
Update: added lines to ignore the gear controller in the 3D Cursor framework for now. Adding the controller to 3D cursor might take some work. Could you categorized as a future item.

@liaxim to address your point - might need to revisit the whole threading issue with the IO framework. It works at the moment but is not aesthetically pleasing. Could do it as a part of the <IO Framework,Picker Framework> refactoring.

liaxim:
Srini gave thumbs up. Will merge Thursday morning.

PR #1129, opened 2017-04-04
Fix problem with picking test "canPickMeshWithObject"

NolaDonato:
This test creates a picker, starts listening for pick events and then attaches the picker to a scene object. This can cause false events to be generated between the time the picker starts listening and the time it is attached. This constructor eliminates the problem by attaching the picker first and then starting to listen for events.
Replaces PR #1124

This is for https://github.com/gearvrf/GearVRf-Tests/pull/78


liaxim:
Pending verification with https://github.com/gearvrf/GearVRf-Tests/pull/78

liaxim:
The updated test still doesn't pass for me.

rahul27:
The test passes now.

liaxim:
The test passes for me too.

PR #1131, opened 2017-04-05
gvrf-physics: convert the build to cmake and externalNativeBuild

liaxim:


liaxim:
- use useLocalDependencies=true
- build fw and extensions with this change in
- open gvr-simplephysics separately (edit GearVRf-Demos/settings.gradle first line to be ``new File(".").eachDirMatch(~"gvr-simplephysics") {``)
- in the symbols section for gvr-simplephysics add these two: ``<your-path-to>/GearVRf/GVRf/Framework/framework/build/intermediates/bundles/debug/jni/armeabi-v7a`` and ``<your-path-to>/GearVRf/GVRf/Extensions/gvrf-physics/build/intermediates/bundles/debug/jni/armeabi-v7a``
- from within AS navigate to the physics source files and open whichever you need
- place breakpoints and debug launch

It can be flaky with buildGvrfAndDemos=true so this is why I recommend useLocalDependencies. The described procedure worked reliably for me. Your mileage can vary.

NolaDonato:
I tested this on Linux using AS 2.2.3. I was not able to get it to set breakpoints in the physics code. I built the Framework in one instance of AS and the demo in another. I opened the physics source files with the same AS instance I used to build the demo.

liaxim:
I assume you also built the physics extension, right?

liaxim:
Tried again after git clean -fdx and still worked for me. I am with AS 2.4 Preview3, should not matter but who knows. Can you hit breakpoints in the framework then (after adding the symbols)? If you hit a native breakpoint then in the Debugger tab there is a LLDB sub-tab. Enter there "image list" and let me know what the output is. It will show where lldb is getting the symbols for libgvrf-physics from.

liaxim:
Tried the procedure on my other desktop. Worked again. As long as you give AS the right symbols for the binary in the apk, it works quite reliably.

liaxim:
Planning to merge Friday morning.

PR #1132, opened 2017-04-06
Have GVRContextProxy derive from GVRContext

liaxim:
Fixes what was broken for the live scripting scenario

---

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRContext.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRContext.java
@@ -175,6 +175,15 @@ public abstract class GVRContext implements IEventReceiver {
     }
 
     /**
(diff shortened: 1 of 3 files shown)


liaxim:
Verified I can telnet into gvr-remote-scripting and execute the sample scripts it includes.

thomasflynn:
since i can now run the scripts, i did a little more testing.  I used to be able to do a print() on a sceneobject and it would print out info like its transform, who its parent(s) are, etc.  Now it just crashes.  To repro, execute the space_trex.js and then issue:  print(trex);

liaxim:
Crashes in GVRBone.getFinalTransformMatrix; joml expects a direct floatbuffer but this is not what we give it. I might just eliminate the buffer altogether and construct the matrix out of the float[].

liaxim:

PR #1136, opened 2017-04-07
gvr-remote-scripting: fix the crash on "print(trex)"

liaxim:
Fix for the crashing GVRBone.getFinalTransformMatrix

- joml expects a direct buffer; if not you get a strange crash in libart due to invalid buffer access
- joml has an option to enable checks for this (joml.debug)

---

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRBone.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRBone.java
@@ -1,12 +1,14 @@
 package org.gearvrf;
 
-import java.nio.FloatBuffer;
-import java.util.ArrayList;
-import java.util.List;
-
 import org.gearvrf.jassimp.AiWrapperProvider;
 import org.gearvrf.utility.Log;
 import org.joml.Matrix4f;
+
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.nio.FloatBuffer;
+import java.util.ArrayList;
+import java.util.List;
 
 /**
  * A single bone of a mesh.<p>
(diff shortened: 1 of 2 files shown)


liaxim:
This must have happened when we switched to the new joml.

liaxim:
Planning to merge Wednesday morning. @rahul27 I am using a direct buffer and you have experience with these so might be useful if you take a look.

httpdigest:
About JOML using direct NIO buffers: https://github.com/JOML-CI/JOML/wiki/Design#java-nio-buffers
The assumption is that NIO buffers are only used when interfacing with native code. Otherwise arrays should be used instead. Measurements showed that arrays are also faster than NIO buffers on Android 7.
I'm sorry that this became an issue here.
If you must work with non-direct NIO buffers then there is also `-Djoml.nounsafe` which also disables optimizations for direct buffers.

liaxim:
No worries. Now we know how to use JOML better. Thanks.

NolaDonato:
Do we need to say something about this in the release notes?


liaxim:
No

thomasflynn:
it was working in 3.1.  we broke it in development of 3.2.  now we're fixing it.  as far as going from 3.1 to 3.2 there is nothing to mention.

rahul27:
@liaxim will check it out.

rahul27:
@liaxim Looks good. `print(trex)` works fine.

thomasflynn:
ok, i did a bit more testing including doing the whole 'programming for VR, in VR, on a GearVR' thing and everything still works.  looks good to me.

rahul27:
I tried the editor and it worked fine. Also tested a few gvrf primitives and it worked well.

The only issue I found was  that the update button on the editor was not clearly visible. I don't know whether that is a regression or a known issue.

liaxim:
Will merge today in the afternoon.

PR #1140, opened 2017-04-11
Fix crash due to shadow map camera setup

NolaDonato:
Added GVRCamera argument to GVRShadowMap constructor allowing the camera to be set up before the shadow map. This fixes a crash in the renderer caused by having the camera not set up for a few frames.
Disabled logging in lights.

This fixes one of the crashes in gvr-simple physics.

GearVRF DCO signed off by Nola Donato nola.donato@samsung.com


GVRf/Framework/framework/src/main/java/org/gearvrf/GVRLightBase.java
```diff
@@ -116,29 +116,10 @@ static public long getComponentType() {
      */
     public void setCastShadow(boolean enableFlag)
     {
-        GVRSceneObject owner = getOwnerObject();
-
-        if (owner != null)
+        if (enableFlag)
```
liaxim: It looks like GVRLightBase could be an abstract class with no knowledge of shadows. Would make GVRLightBase and GVRPointLight slightly simpler and the hierarchy clearer.

liaxim:

PR #1141, opened 2017-04-11
Features and fixes for physics

NolaDonato:
1. RenderData::set_mesh now checks for null to avoid crash
2. MeshCollider::mesh() returns the mesh from an associated RenderData if the mesh was not specified (it used to return NULL). The physics extension assumes the new behavior.
3. Added GVRWorld.setGravity to allow app to change global gravity settings
4. Lock around updates to physics rigid bodies in C++ to prevent crashes. This is a global lock - not a per rigid-body lock. We may need to revisit this issue in the future to support more granular locking.
3. Added GVRRigidBody.setSimulationType to support dynamic, kinematic and static objects. THIS FEATURE IS NOT WORKING YET!
4. Removed confusing APIs in GVRRigidBody to set position and center. The correct way to move and orient the rigid body is by changing the transformation of the scene object which owns it. These APIs should not be exposed to the user.
5. Do not permit re-entering NativePhysics3DWorld.step by the draw frame listener, Instead wait until the next frame to run physics. This fixes some thread safety issues.

These changes fix some of the crashes in the gvr-simplephysics sample.

### DIFF
--- a/GVRf/Extensions/gvrf-physics/src/main/java/org/gearvrf/physics/GVRWorld.java
+++ b/GVRf/Extensions/gvrf-physics/src/main/java/org/gearvrf/physics/GVRWorld.java
@@ -34,6 +34,8 @@ import java.util.LinkedList;
  * {@link GVRWorld} is a component that must be attached to the scene's root object.
  */
 public class GVRWorld extends GVRBehavior implements ISceneObjectEvents, ComponentVisitor {
+    protected float mFrameTime;
+    private boolean mIsProcessing;
 
     static {
         System.loadLibrary("gvrf-physics");
@@ -60,6 +62,8 @@ public class GVRWorld extends GVRBehavior implements ISceneObjectEvents, Compone
     public GVRWorld(GVRContext gvrContext, GVRCollisionMatrix collisionMatrix) {
         super(gvrContext, NativePhysics3DWorld.ctor());
         mHasFrameCallback = false;
+        mIsProcessing = false;
+        mFrameTime = 0.0f;
         mCollisionMatrix = collisionMatrix;
     }
 
@@ -113,9 +117,15 @@ public class GVRWorld extends GVRBehavior implements ISceneObjectEvents, Compone
 
     @Override
     public void onDrawFrame(float frameTime) {
-        NativePhysics3DWorld.step(getNative(), frameTime);
+        mFrameTime += frameTime;
+        if (mIsProcessing) {
+            return;
+        }
+        NativePhysics3DWorld.step(getNative(), mFrameTime);
 
         generateCollisionEvents();
+        mFrameTime = 0.0f;
+        mIsProcessing = false;
     }
 
     private void generateCollisionEvents() {
@@ -211,6 +221,14 @@ public class GVRWorld extends GVRBehavior implements ISceneObjectEvents, Compone
 
     }
 
+    public void setGravity(float x, float y, float z) {
+        NativePhysics3DWorld.setGravity(getNative(), x, y, z);
+    }
+
+    public void getGravity(float[] gravity) {
+        NativePhysics3DWorld.getGravity(getNative(), gravity);
+    }
+
     @Override
     public boolean visit(GVRComponent gvrComponent) {
         if (!gvrComponent.isEnabled()) {
@@ -237,6 +255,10 @@ class NativePhysics3DWorld {
     static native void removeRigidBody(long jphysics_world, long jrigid_body);
 
     static native void step(long jphysics_world, float jtime_step);
+
+    static native void getGravity(long jworld, float[] array);
+
+    static native void setGravity(long jworld, float x, float y, float z);
 
     static native GVRCollisionInfo[] listCollisions(long jphysics_world);
 }
--- a/GVRf/Extensions/gvrf-physics/src/main/java/org/gearvrf/physics/GVRWorld.java
+++ b/GVRf/Extensions/gvrf-physics/src/main/java/org/gearvrf/physics/GVRWorld.java
@@ -121,8 +121,8 @@ public class GVRWorld extends GVRBehavior implements ISceneObjectEvents, Compone
         if (mIsProcessing) {
             return;
         }
+        mIsProcessing = true;
         NativePhysics3DWorld.step(getNative(), mFrameTime);
-
         generateCollisionEvents();
         mFrameTime = 0.0f;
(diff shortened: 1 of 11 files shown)


GVRf/Extensions/gvrf-physics/src/main/java/org/gearvrf/physics/GVRWorld.java
```diff
@@ -113,9 +117,15 @@ public void removeBody(GVRRigidBody gvrBody) {
 
     @Override
     public void onDrawFrame(float frameTime) {
-        NativePhysics3DWorld.step(getNative(), frameTime);
+        mFrameTime += frameTime;
+        if (mIsProcessing) {
```
liaxim: How does this work? mIsProcessing is never set to true.

NolaDonato:
Glad you noticed that. This must have occurred during merging. I added mIsProcessing = true back in and tested the gvr-simplephysics sample (but not with the controller).

rahul27:
I got this null pointer exception while testing this pull request with the gear controller:

````
04-13 13:59:55.817 10035-10073/org.gearvrf.gvrsimlephysics E/GVRViewManager: DrawFrameListener org.gearvrf.physics.GVRWorld@22bcc4ef threw java.lang.NullPointerException: Attempt to invoke virtual method 'org.gearvrf.GVRSceneObject org.gearvrf.GVRComponent.getOwnerObject()' on a null object reference
04-13 13:59:55.818 10035-10073/org.gearvrf.gvrsimlephysics W/System.err: java.lang.NullPointerException: Attempt to invoke virtual method 'org.gearvrf.GVRSceneObject org.gearvrf.GVRComponent.getOwnerObject()' on a null object reference
04-13 13:59:55.818 10035-10073/org.gearvrf.gvrsimlephysics W/System.err:     at org.gearvrf.physics.GVRWorld.sendCollisionEvent(GVRWorld.java:148)
04-13 13:59:55.818 10035-10073/org.gearvrf.gvrsimlephysics W/System.err:     at org.gearvrf.physics.GVRWorld.generateCollisionEvents(GVRWorld.java:142)
04-13 13:59:55.818 10035-10073/org.gearvrf.gvrsimlephysics W/System.err:     at org.gearvrf.physics.GVRWorld.onDrawFrame(GVRWorld.java:126)
04-13 13:59:55.818 10035-10073/org.gearvrf.gvrsimlephysics W/System.err:     at org.gearvrf.GVRViewManager.doMemoryManagementAndPerFrameCallbacks(GVRViewManager.java:258)
04-13 13:59:55.818 10035-10073/org.gearvrf.gvrsimlephysics W/System.err:     at org.gearvrf.GVRViewManager.access$600(GVRViewManager.java:46)
04-13 13:59:55.818 10035-10073/org.gearvrf.gvrsimlephysics W/System.err:     at org.gearvrf.GVRViewManager$3.beforeDrawEyes(GVRViewManager.java:433)
04-13 13:59:55.818 10035-10073/org.gearvrf.gvrsimlephysics W/System.err:     at org.gearvrf.GVRViewManager.beforeDrawEyes(GVRViewManager.java:490)
04-13 13:59:55.818 10035-10073/org.gearvrf.gvrsimlephysics W/System.err:     at org.gearvrf.OvrViewManager.beforeDrawEyes(OvrViewManager.java:214)
04-13 13:59:55.818 10035-10073/org.gearvrf.gvrsimlephysics W/System.err:     at org.gearvrf.OvrViewManager.onDrawFrame(OvrViewManager.java:353)
04-13 13:59:55.818 10035-10073/org.gearvrf.gvrsimlephysics W/System.err:     at org.gearvrf.OvrVrapiActivityHandler$7.onDrawFrame(OvrVrapiActivityHandler.java:424)
04-13 13:59:55.818 10035-10073/org.gearvrf.gvrsimlephysics W/System.err:     at android.opengl.GLSurfaceView$GLThread.guardedRun(GLSurfaceView.java:1562)
04-13 13:59:55.819 10035-10073/org.gearvrf.gvrsimlephysics W/System.err:     at android.opengl.GLSurfaceView$GLThread.run(GLSurfaceView.java:1262)

File: GVRf/Framework/framework/src/main/jni/objects/components/render_data.cpp [hunk 2]
```diff
@@ -36,9 +36,14 @@ const RenderPass* RenderData::pass(int pass) const {
 }
 
 void RenderData::set_mesh(Mesh* mesh) {
-    mesh_ = mesh;
-    mesh->add_dirty_flag(dirty_flag_);
-    *dirty_flag_ = true;
+    if (mesh != nullptr) {
```
liaxim: What is the case where the mesh parameter is nullptr? While this method technically accepts a pointer none of the uses imply passing nullptr is valid. We need to know where the nullptr is coming from first.

There are already some but we need to sprinkle nullptr checks for every gvrrenderdata::mesh() now since there is a new way for renderdata::mesh_ to become null.

---
I do know this nullptr check masks some of the crashes but they are down to thread-safety issues and we should be understanding and fixing those. In fact, we should abort the program if somebody passes a nullptr. Better yet, use something better than a raw pointer to indicate that the contract is to pass a valid mesh instance always.

GVRf/Extensions/gvrf-physics/src/main/java/org/gearvrf/physics/GVRWorld.java
```diff
@@ -113,9 +117,15 @@ public void removeBody(GVRRigidBody gvrBody) {
...
     @Override
     public void onDrawFrame(float frameTime) {
-        NativePhysics3DWorld.step(getNative(), frameTime);
-
+        mFrameTime += frameTime;
+        if (mIsProcessing) {
+            return;
+        }
+        mIsProcessing = true;
+        NativePhysics3DWorld.step(getNative(), mFrameTime);
```
liaxim: Is this getting called more than once per frame somehow? How is step exactly getting re-entered?

When I removed these changes but kept the mesh nullptr check there were no crashes.

NolaDonato: The draw frame listener is called pretty frequently. If you launch a lot of balls at the same time or the collisions are complex, the physics will not finish before the next call is issued. Also if you have more than one physics world there will be multiple draw frame listeners for step() and calls could happen at the same time.
This fix is not ideal and we need to revisit the physics thread safety issues.

PR #1145, opened 2017-04-11
Fix picking from object

NolaDonato:
Default the pick ray  to (0,0,0,0,0,1) when picking from an object.
This fixes the issue with picking a mesh from an object in the picking tests.


NolaDonato:
rolled into #1129

Issue #1139, opened 2017-04-10
S8: libgdx plugin problem with gvr-modelviewer2&gvr-widgetviewer


liaxim:
The app just freezes at the logo or even prior to that. Almost certainly it is down to the widget plugin extension. Works on Note 5. The plugin has some dodgy mechanics that can be highly sensitive to timing.

Doesn't work with Oculus and Daydream.

liaxim:
Unsurprisingly gvr-widgetviewer doesn't work either. I assume we will want this fixed for the release and will be looking into it.

liaxim:
Still working on it. Had trouble deciphering which thread needs to wait on what. Clearly something is deadlocking.

liaxim:
After getting the widget plugin working there is this exception in gvr-modelviewer2:

```
04-12 12:57:10.211 10725 10725 E AndroidRuntime: FATAL EXCEPTION: main
04-12 12:57:10.211 10725 10725 E AndroidRuntime: Process: org.gearvrf.modelviewer2, PID: 10725
04-12 12:57:10.211 10725 10725 E AndroidRuntime: java.lang.NullPointerException: Attempt to invoke virtual method 'org.gearvrf.GVRSceneObject org.gearvrf.modelviewer2.Model.getModel(org.gearvrf.GVRContext)' on a null object reference
04-12 12:57:10.211 10725 10725 E AndroidRuntime: 	at org.gearvrf.modelviewer2.Controller.onScrollOverModel(Controller.java:590)
04-12 12:57:10.211 10725 10725 E AndroidRuntime: 	at org.gearvrf.modelviewer2.ModelViewer2Manager.onScroll(ModelViewer2Manager.java:306)
04-12 12:57:10.211 10725 10725 E AndroidRuntime: 	at org.gearvrf.modelviewer2.ModelViewer2Activity.onScroll(ModelViewer2Activity.java:79)
04-12 12:57:10.211 10725 10725 E AndroidRuntime: 	at org.gearvrf.util.VRTouchPadGestureDetector.onScroll(VRTouchPadGestureDetector.java:227)
04-12 12:57:10.211 10725 10725 E AndroidRuntime: 	at android.view.GestureDetector.onTouchEvent(GestureDetector.java:622)
04-12 12:57:10.211 10725 10725 E AndroidRuntime: 	at org.gearvrf.util.VRTouchPadGestureDetector.onTouchEvent(VRTouchPadGestureDetector.java:252)
[...]

PR #1146, opened 2017-04-12
widgetplugin: reduce some of the locking, simplify and also fix the deadlock on s8

liaxim:
doResume must be called before a widget is initialized; the code though was written in such a way that doResume may end up waiting for widget initialization

Addresses the deadlock from https://github.com/Samsung/GearVRf/issues/1139

---


liaxim:
Verified with all backends on S8 and Note5.

liaxim:
Will merge Thursday after lunch

PR #1151, opened 2017-04-18
GVRAccessibilityPostEffectShader: use GLSL ES version 300

liaxim:


liaxim:
Will close this and push one for master instead

PR #1161, opened 2017-04-20
Fix for a UnknownFormatConversionException that leads to a crash

liaxim:
A certain app that has been known to be kind of a stress test has been crashing recently due to ``java.util.UnknownFormatConversionException: Conversion: p``.

Doesn't look like there is a good case that justifies the lookup returning nullptr. The app can detach colliders from some thread and runs GVRPicker.pickObjects on the GL thread. While there is locking on the C++ side, part of the collider implementation is on the Java side. If it becomes a bigger problem we will have to look into making the operations atomic from Java's perspective.

---


liaxim:

PR #1173, opened 2017-04-27
Add a texture that is updated from a Java NIO Buffer.

liaxim:
Will use for the YUV conversion.

---


NolaDonato:
How is this type of texture not just a different variant of GVRBitmapTexture? In the Vulkan branch we are distinguishing textures internally by the target format (TEXTURE_2D, TEXTURE_CUBEMAP, etc) Anything with a target of TEXTURE_2D is considered a bitmap texture.

My suggestion is to implement something like:
GVRBitmapTexture(GVRContext gvrContext, int format, int width, int height, int type, Buffer pixels)

I know our existing texture code is somewhat messy but I have cleaned it up in the Vulkan branch. We do all the texture updates on the native side in C++ directly from the Java data structures and then release the Java objects once the texture has been updated on the GPU. I do not know the tradeoffs of using NIO buffers vs arrays but perhaps later we should make NIO constructors for all the texture types.

Next week I will be making a pull request of just the Vulkan texture stuff to roll it back into master. In the meantime if you want to look at our work, Roshan and I have been using this branch: https://github.com/NolaDonato/GearVRf/tree/vk_shader_rebased

liaxim:
Just following the existing hierarchy where we have GVRCompressedTexture, GVRFloatTexture, etc. For now it is hard to make the argument this is part of GVRBitmapTexture while we have all the other types of textures. This deals with a ByteBuffer that doesn't contain android.graphics.Bitmap so making it part of GVRBitmapTexture goes against the existing style. The Bitmap moniker implies to the end users an Android Bitmap which this is not.

For one time operations I am not so sure it is really better to have everything on the C++ side since it may lead to extra code that doesn't really give you any advantage. To me the question is bigger - what is our vision for the Java part of the framework. If it is going to be just wrappers then we should be consistent. If not, then everything that can should be in Java unless proven by measurements it is better done in C++.

The Java NIO buffers (if "direct") are eliminating copies when using JNI. There are not always the best choice and it depends what you do with them. If you read them from Java apparently they are very slow.

---
You can ignore this texture in your pull request; I will make the necessary updates after your changes are merged.

NolaDonato:
Well I can't see how embedding OpenGL code in the Java side is going to help the Vulkan port.

liaxim:
Ok, will move the postBuffer implementation to C++.

liaxim:
@NolaDonato Made the changes you requested.

liaxim:
Planning to merge on Tuesday.

GVRf/Framework/framework/src/main/jni/objects/textures/base_texture.h
```diff
@@ -85,6 +85,12 @@ class BaseTexture: public Texture {
         }
     }
 
+    void updateFromBuffer(int internalFormat, int width, int height, int format, int type, void* data) {
+        glBindTexture(GL_TEXTURE_2D, getId());
+        glTexImage2D(GL_TEXTURE_2D, 0, internalFormat, width, height, 0, format, type, data);
```
roshanch: are we just updating texture here, is the memory preallocated? If we are updating texture data, then PBO would be best option and if not then glTexSubImage2D.

liaxim:
@roshanch Made the suggested change. Had to add few members to GVRTextureParameters though.

roshanch:
looks good.

liaxim:

PR #1174, opened 2017-04-27
Add an utility YUV-NV21 conversion shader

liaxim:
Depends on https://github.com/Samsung/GearVRf/pull/1173

---

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/utility/YuvNv21ToRgbShader.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/utility/YuvNv21ToRgbShader.java
@@ -28,8 +28,8 @@ import org.gearvrf.GVRShaderTemplate;
  * ByteBuffer mUvBuffer = ByteBuffer.allocateDirect(width*height/2);
  * myBuffer.order(ByteOrder.nativeOrder());
  * muvBuffer.order(ByteOrder.nativeOrder());
- * GVRTexture mYBufferTexture = new GVRNioBufferTexture(context);
- * GVRTexture mUVBufferTexture = new GVRNioBufferTexture(context);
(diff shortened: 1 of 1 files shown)


GVRf/Framework/framework/src/main/java/org/gearvrf/utility/YuvNv21ToRgbShader.java
```diff
@@ -0,0 +1,88 @@
...
+            "precision highp float;\n" +
+            "in vec2 v_texCoord;\n" +
+            "uniform sampler2D y_texture;\n" +
+            "uniform sampler2D uv_texture;\n" +
+            "out vec4 fragColor;\n" +
+            "  void main (void) {\n" +
+            "  float r, g, b, y, u, v;\n" +
+            "  y = texture(y_texture, v_texCoord).r;\n" +
+            "  u = texture(uv_texture, v_texCoord).a - 0.5;\n" +
+            "  v = texture(uv_texture, v_texCoord).r - 0.5;\n" +
```
roshanch: Lets avoid sampling the same fragment multiple times.
vec4 texColor = texture(ub_texture,v_texCoord);
u = texColor.a - 0.5;
v = texColor.r - 0.5;

liaxim: Sure

NolaDonato:
Do you think we should put all of our builtin shaders into a single package - like org.gearvrf.shaders? Right now they are sprinkled throughout our code. In the Vulkan branch we are removing all the C++ builtin shaders so we will have a lot more of them.

liaxim:
I think core shaders we should keep in the gearvrf package. Especially if it means making some internal apis public just to be able to move them to a separate package. This one is just a helper for a very special YUV format so should not be polluting the main package.

liaxim:
Planning to merge on Tuesday.

liaxim:

PR #1175, opened 2017-05-01
Add the ColorShader to the utilities

liaxim:
- serves a common need and helps simplify few samples that have their own copies


liaxim:

PR #1180, opened 2017-05-03
Add a bit of logging to the io library

liaxim:
debug-level logging to try track down an issue with an app where input stops working

for the next release will add a proguard rule to strip out all Log.d calls in release builds


liaxim:
Planning to merge Friday afternoon.

PR #1176, opened 2017-05-02
Enable building the platform sdk support library always

liaxim:
GearVRf-DCO-1.0-Signed-off-by: Mihail Marinov <m.marinov@samsung.com>

PR #1184, opened 2017-05-04
Removing GVRContext loading APIs

NolaDonato:
This pull request moves GVRContext.loadMesh and loadFutureMesh functions
to GVRAssetLoader. It also removes the deprecated GVRContext functions
to load models, meshes and textures leaving only the asset loader
versions.

This patch removes all variants of loadModel, loadTexture, loadFutureTexture, loadMesh and loadFutureMesh from GVRContext. It changes the API.

I made patch requests for API changes to the demos and tests for the loadMesh call.


liaxim:
3DCursorLibrary needs updating too.

NolaDonato:
Extensions project no longer builds out of the box. It references "buildGvrfAndDemos" and OVR_PLATFORM_SDK which should be OVR_MOBILE_SDK. We need to look at the build scripts in the extensions to make them compatible with the others.

liaxim:
References to buildGvrfAndDemos are obsolete but will not break the build.

We enabled building the Oculus Platform SDK integration always in https://github.com/Samsung/GearVRf/pull/1176. Hence the need to download this SDK (https://developer.oculus.com/downloads/package/oculus-platform-sdk/1.8.0/) too and point OVR_PLATFORM_SDK to it.

NolaDonato:
I have updated this pull request and pull request #448 from GearVRf-Demos to not use deprecated asset APIs. In particular, loading a Future<GVRMesh> makes absolutely no sense. You save nothing since it has to parse the entire 3D asset file to create the mesh in the first place. The function did not even return the future mesh until the asset file was parsed...

liaxim:
This and the corresponding pull request in the Demos are still waiting on the GVRShaderTemplate changes pull request for the Demos.

PR #1186, opened 2017-05-05
Fix for cases where the Oculus backend may end up rendering to a portrait-oriented surface

liaxim:
- fixes leaked egl surfaces
- gradle: publishNonDefault to ensure apps use the correct build variant
of the libraries (not always release like it is by default)

---


liaxim:
Calling setFixedSize always stopped the creation of surfaces in portrait orientation. Something that I saw a lot of when running the automated tests and something that usually leads to the tests hanging.

liaxim:
Ran the framework-tests few times and they seem to be a lot more stable than in the past. Also at the end of the whole run the memory used is a lot less (~150MBs) in total than prior to this change.

liaxim:
The leak specifically was caused by not making no surface current prior to their destruction.

liaxim:
Planning to merge Thursday afternoon.

PR #1192, opened 2017-05-11
Fix x3d emission, external scene loading

NolaDonato:
This patch fixes two X3D issues.
1. fixes emissive color by modifying the phong shader
2. fixes external scene loading by modifying the asset loader


liaxim:
Looks good. @mwitchwilliams Do you have any comments? Planning to merge Tuesday morning.

Related tests pull request: https://github.com/gearvrf/GearVRf-Tests/pull/86

mwitchwilliams:
I haven't tested either of these, but I can in the next day or two.  Let's postpone merging until I test these.

liaxim:
Sure

mwitchwilliams:
Emissive color is working.
When calling GVRExtenralScene, the old scene isn't being removed, even when I set 'replaceScene' true.  Here's the code I'm running

                                GVRExternalScene gvrExternalScene = new GVRExternalScene(gvrContext, url, true);
                                root.attachComponent( gvrExternalScene );
                                boolean load = gvrExternalScene.load(gvrContext.getMainScene());

I also have a DirectionalLight and animation in the new scene that aren't running.  The DirectionalLight does work when using the debugger.  I am investigating both.

NolaDonato:
The directional light problem does not have anything to do with this pull request. I think there is something wrong with lighting when you replace the entire scene. I will write a unit test for it to see if I can reproduce it independently of X3D.
As far as the animation is concerned, we need to add a constructor for GVRExternalScene that takes import settings. Then you can tell it to start the animations.

liaxim:
Ok then. If I understand correctly this pull request is fine as-is. Planning to merge Wednesday morning.

mwitchwilliams:
Just re-tested and emissive color is working for me.
GVRExternalScene also works - there is a lighting problem, but there is a separate issue posted for that.
There is a Pull Request in the Testing repository under x3d -> anchor that shows it works.

I attached my demo X3D file with a red, green, blue and white cone in the upper left, upper right, lower left and lower right corners respectively.  It also has a box with only white diffuse color in the center.  With a dim light ( 10% intensity), the white box is correctly very dark gray.
[emissivecolor.zip](https://github.com/Samsung/GearVRf/files/1032504/emissivecolor.zip)

PR #1193, opened 2017-05-11
Added START_ANIMATIONS import setting

NolaDonato:
The START_ANIMATIONS setting will cause animations from a loaded model to automatically start when the asset is added to the scene. (Currently X3D does not honor this flag).

Also made the "cacheEnabled" flag actually do something. It was not correctly implemented. Now if you clear this flag when loading the model it will NOT cache textures.


liaxim:
Looks good. Will merge Thursday morning.

liaxim:
@mwitchwilliams Please look into supporting the start animations flag in X3D.

Tests pull request: https://github.com/gearvrf/GearVRf-Tests/pull/87

PR #1201, opened 2017-05-17
Fixing shader template changes for mat4

apjagdale:
Fixes GearVRf-Demos pull request 436


GVRf/Framework/framework/src/main/java/org/gearvrf/GVRShaderTemplate.java
```diff
@@ -673,6 +673,11 @@ protected void makeMaterialMap(GVRShaders material, GVRShaderMaps map)
                     nfloats = Integer.parseInt(size);
                 }
             }
+
+            if(name.equals("mat4")){
```
liaxim: That should be type.equals("mat4")

liaxim:
Will merge Friday morning and fix the type vs. the name problem. The demos work fine, including vuforia with this small fix.

liaxim:
Thanks

Issue #1203, opened 2017-05-19
GVRRenderData.setAlphaToCoverage() blending background over opaque objects


mwitchwilliams:
GVRRenderData.setAlphaToCoverage(true) is blending the background color over opaque object.
In the X3D file, if the meshes with png texture maps with transparencies are listed before the opaque object, the background color is blended with the transparency when it is over an opaque object.

The demo contains 4 animated planes containing texture maps with transparency.  Two of the objects come before the opaque object and these seem to be wrong.  The screen shots show what happens when there is a red, green and blue background.  The two objects following the opaque object in the X3D file seems ok.

To run this demo, in X3Dobject.java, add:
 _gvrRenderData.setAlphaToCoverage(true);_
after the GVRRenderData() is constructed in the &lt;Shape&gt; node.

**Two transparent texture maps:**
![round_icon_shadow](https://cloud.githubusercontent.com/assets/18128830/26257989/9a050c62-3c77-11e7-8261-8f294645d0c5.png)
![round_half_icon_shadow](https://cloud.githubusercontent.com/assets/18128830/26257993/9cb2a2d0-3c77-11e7-9ebd-2d95a4711e8f.png)

**X3D zip file, contains the X3D file plus texture maps.**
[x3d_files.zip](https://github.com/Samsung/GearVRf/files/1014983/x3d_files.zip)

**Screen Shots with red, green and blue backgrounds.**
![screenshot_20170519-090159](https://cloud.githubusercontent.com/assets/18128830/26258483/a76c8e1e-3c79-11e7-864a-279ef1752568.png)

![screenshot_20170519-090408](https://cloud.githubusercontent.com/assets/18128830/26257976/8b7ded8a-3c77-11e7-99cd-40ba0df41a6d.png)
![screenshot_20170519-092950](https://cloud.githubusercontent.com/assets/18128830/26257979/8daa0a12-3c77-11e7-98ce-59a7ee7bfd95.png)

mwitchwilliams:
The issue seems related to the position of the camera.  New (as yet not merged) code enables the camera to be at other positions besides (0, 0, 0).  At (0, 0, 0), it seems to work, but at (0, 0, 8) as in the demo, the problems occur.

mwitchwilliams:
Still an active bug.
Investigating renderdata.cpp for algorithm of sorting object based on z-distance from camera versus actual distance [ sqrt(x*x + y*y + z*z) from camera ].

[...]

PR #1206, opened 2017-05-22
fixing camera_position for culling

roshanch:
this fixes issue #1203 .


mwitchwilliams:
Verified this fixed Issue #1203.
This scene now renders (with animation) correctly.

thomasflynn:
good find!

liaxim:
Needs ``GearVRf-DCO-1.0-Signed-off-by``

liaxim:
gvr-shadows crashes due to nullptr de-reference.

liaxim:
Is the fix complicated?

roshanch:
waiting for @NolaDonato

NolaDonato:
The fix will fail if the camera does not have an owner object. If you are rendering a shadow map or render target, this will be the case.

liaxim:
Should we close this?

PR #1211, opened 2017-05-25
ttf Font and Style implementation

mwitchwilliams:
New GVRTextViewSceneObject constructor and methods to enable loading .ttf font files with styles (bold, italic) and justification (left, middle, right). 
Calls made from X3D implementing the &lt;Text&gt; and &lt;FontStyle&gt; nodes.


GVRf/Framework/framework/src/main/java/org/gearvrf/x3d/X3Dobject.java
```diff
@@ -3273,7 +3305,15 @@ else if (qName.equalsIgnoreCase("Appearance")) {
             } else if (qName.equalsIgnoreCase("ProximitySensor")) {
                 currentSensor = null;
             } else if (qName.equalsIgnoreCase("Text")) {
-                ;
+                // since the text node is complete, now calculate
```
liaxim: Looks like this whole block belongs to GVRTextViewSO. Something like updateTextPlaneSize() which doesn't even need arguments from what I can tell.

GVRf/Framework/framework/src/main/java/org/gearvrf/scene_objects/GVRTextViewSceneObject.java
```diff
@@ -93,6 +112,11 @@
     private int mCount;
     private final GVRDrawFrameListenerImpl mFrameListener;
 
+    // individual lines of text.  Assists in determining lenghts of
+    // strings once the fonts are set.
+    private String linesOfText[] = null;
+    private GVRMaterial gvrMaterialX3D = null;
```
liaxim: Why do we need a special material for just one case? It doesn't seem necessary to me.

GVRf/Framework/framework/src/main/java/org/gearvrf/scene_objects/GVRTextViewSceneObject.java
```diff
@@ -181,6 +205,87 @@ public GVRTextViewSceneObject(GVRContext gvrContext) {
...
+        mTextView.setLayoutParams(new LayoutParams( canvasWidthInt,  canvasHeightInt) );
+
+        float width = canvasWidth / FACTOR_IMAGE_SIZE;
+        float height = canvasHeight / FACTOR_IMAGE_SIZE;
+
+        GVRContext gvrContext = getGVRContext();
+        GVRMesh gvrMesh = gvrContext.createQuad(width, height);
+
+        GVRRenderData gvrRenderData = new GVRRenderData(gvrContext);
+        gvrRenderData.setRenderingOrder(GVRRenderData.GVRRenderingOrder.TRANSPARENT);
```
liaxim: If X3D needs the object being transparent assumption, it should just make the call from the parser instead of embedding it in what could be generic otherwise method.

GVRf/Framework/framework/src/main/java/org/gearvrf/scene_objects/GVRTextViewSceneObject.java
```diff
@@ -258,6 +378,168 @@ public void setBackgroundColor(int color) {
...
+    public int getMaxLines() {
+        return mTextView.getMaxLines();
+    }
+
+    /**
+     * Saves the lines of text and used later to calculate the
+     * length in pixels based on font-family, size, etc.
+     * @param textString
+     */
+    public void setTextStrings(String[] textString) {
```
liaxim: The array and the related methods shouldn't be necessary. The text view can just accept one string, detect if it is multiline (look for '\n') and in case it is, handle all the measurements behind the scenes.

GVRf/Framework/framework/src/main/java/org/gearvrf/x3d/X3Dobject.java
```diff
@@ -2962,16 +2996,14 @@ else if (qName.equalsIgnoreCase("Background")) {
                         root.addChildObject(mCubeEvironment);
                     } else {
                         // Not cubemapping, then set default skyColor
-                        cameraRigAtRoot.getLeftCamera()
-                                .setBackgroundColor(skycolor[0], skycolor[1], skycolor[2], 1);
-                        cameraRigAtRoot.getRightCamera()
-                                .setBackgroundColor(skycolor[0], skycolor[1], skycolor[2], 1);
+                        gvrContext.getMainScene().getMainCameraRig().getLeftCamera().setBackgroundColor(skycolor[0], skycolor[1], skycolor[2], 1);
```
liaxim: Fyi: there is a GVRScene.setBackgroundColor method nowadays.

liaxim:
This is rather large and can use another pair of eyes. Few tests show no side-effects on the rest of the framework. Have not yet gone through the x3d tests. We should generalize the changes to the TextViewSO though.

williamsma:
One thought was for me to create a Font class inside X3Dobject.java, and then once I get all the properties, create a single Constructor with all the properties such as font family, size of font, style (bold, italic) and other parameters, instead of breaking it up the way I did.

I'm discussing this with these 2 kangaroos right now to get their helpful input.
feeding kangaroos

Will discuss more when I return on Monday, June 12.

NolaDonato:
What do the kangaroos think we should do? Do they like the single constructor idea better?


williamsma:
Turns out kangaroos like inner classes, based on their being marsupials.  By adding a new Font inner class to X3Dobject.java, I can save all the properties from X3D's &lt;Text&gt; and &lt;FontStyle&gt; nodes and come up with a new constructor in GVRTextViewSceneObject which will set everything.
The koalas like it too.
me with koala

NolaDonato:
Yes I also used inner classes in the asset loader. Saber thought it was a better approach. I guess dogs and roos think alike.



mwitchwilliams:
Request this pull request be closed as its issues are now superseded by pull request 'ttf Font implementation' #1238

PR #1212, opened 2017-05-25
Try to automatically detect transparency

thomasflynn:
Right now, if your SceneObject has a texture with transparency, you must explicitly mark that SceneObject's RenderData RenderingOrder as TRANSPARENT in order for it to get sorted and blended properly.  This commit is an attempt to detect this condition and set the RenderingOrder appropriately.  To do this, the following steps are taken:

1)      Add flag in GVRTexture for transparency
    a.      If a Bitmap and texture hasAlpha()
        i.     Call through jni to check alpha values.  Will be similar to extractAlpha() in SkBitmap in Skia, but no need for allocating or copying anything.  Will early-out once a non-opaque pixel is detected.
        ii.    If it hasAlpha(), but is fully opaque, warn the user in logcat that it would be faster if it was an RGB (no alpha) texture.
    b.      If compressed (ASTC, ETC2, etc) and has an alpha component, set for transparency.
2)      Add flag inside GVRRenderData for transparency
    a.      When setMaterial() is called, check for transparency flag in the "MAIN_TEXTURE", set flag and set rendering order to TRANSPARENT if appropriate.


thomasflynn:
This is still a work in progress.  I still need to do some testing before I'll consider this ready.  Welcome any comments in the meantime.

thomasflynn:
This currently does not handle the case where one calls material.setTexture() after calling renderData.setMaterial().  Handling that will require some surgery that we haven't figured out how to do yet.  Otherwise, this should work now.

NolaDonato:
Don't worry too much about it. We will revisit it again in the Vulcan branch.



GVRf/Framework/framework/src/main/jni/objects/textures/bitmap_transparency.cpp
```diff
@@ -0,0 +1,97 @@
...
+                done = true;
+            }
+            ptr += stride;
+            height--;
+        }
+    } else if(format == ANDROID_BITMAP_FORMAT_RGBA_8888) {
+        const uint32_t *ptr = (uint32_t *)addrPtr; 
+        while(height >= 0 && !done) {
+            for(int x = 0; x < width && !done; x++) {
+                uint8_t alpha = ptr[x] << 24;
```
liaxim: The test with png_3_opaque always fails for me because the code determines that this particular png is non-opaque when it is supposed to opaque.

1. 0xFF808080 << 24 truncated to uint8_t yields a 0 for me; I think the result is implementation-defined. Shouldn't the shift be in the other direction actually?
2. When height becomes 0, ptr points to a 0 so even with 1. fixed, the png is marked as transparent. Are we getting out of bounds or is this png not quite as opaque as it should be?

liaxim:
Tested on S8+, user build.

thomasflynn:
interesting.  it passes on my S6, but not on my S8+.  interesting.  maybe an OS difference (the S6 is running 5.1.1, the S8+ 7.0).

thomasflynn:
ok, it boiled down to an off-by-one error in my loops.

liaxim:
Tests pass. Planning to merge Friday morning.

PR #1219, opened 2017-05-31
X3D: use better minification filter for textures

liaxim:
It used to be GL_LINEAR.

---

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/x3d/X3Dobject.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/x3d/X3Dobject.java
@@ -1036,6 +1036,7 @@ public class X3Dobject {
                     gvrTextureParameters = new GVRTextureParameters(gvrContext);
                     gvrTextureParameters.setWrapSType(TextureWrapType.GL_REPEAT);
                     gvrTextureParameters.setWrapTType(TextureWrapType.GL_REPEAT);
+                    gvrTextureParameters.setMinFilterType(GVRTextureParameters.TextureFilterType.GL_LINEAR_MIPMAP_NEAREST);
 
                     String urlAttribute = attributes.getValue("url");
                     if (urlAttribute != null) {


GVRf/Framework/framework/src/main/java/org/gearvrf/x3d/X3Dobject.java
```diff
@@ -1036,6 +1036,7 @@ else if (qName.equalsIgnoreCase("ImageTexture")) {
                     gvrTextureParameters = new GVRTextureParameters(gvrContext);
                     gvrTextureParameters.setWrapSType(TextureWrapType.GL_REPEAT);
                     gvrTextureParameters.setWrapTType(TextureWrapType.GL_REPEAT);
+                    gvrTextureParameters.setMinFilterType(GVRTextureParameters.TextureFilterType.GL_LINEAR_MIPMAP_NEAREST);
```
roshanch: did we test this with compressed textures? filtering says we are using mip-mapping.

liaxim:
Nothing requires X3D's ImageTexture to support compressed textures (http://www.web3d.org/documents/specifications/19775-1/V3.2/Part01/components/texturing.html#ImageTexture). It is probably nice to have but we have some work to do.

- our X3D implementation still doesn't support the TextureProperties where you could explicitly set the minification filter; the GL_LINEAR default doesn't work too well for some textures. So one way would be to explicitly have the filter set but we need to add support for TextureProperties then.

- we have this magic GVRBitmapTexture.updateCall that just forces GL_LINEAR_MIPMAP_NEAREST on all non-compressed texture; it needs to go.

- GVRAssetLoader.TextureRequest.loaded updates the texture parameters but has no concept of defaults for compressed vs non-compressed textures; in any case it is fed certain texture parameters by X3D and it does just what it is asked to do. The GVRAssetLoader.TextureRequest.loaded method happens to be called after GVRBitmapTexture.updateCall.

So if we are ok with explicitly requesting things then it is simple - add support for TextureProperties to X3D, have all models use them accordingly.

If we want to do sensible things automatically then things become fuzzier and require some refactoring.

liaxim:
@roshanch If you don't mind I plan to merge on Friday morning and open ticket for adding compressed textures support. Let me know..

roshanch:
yeah, opening a separate issue sounds good. we can merge this for now.

PR #1220, opened 2017-05-31
Removing GVRContext loading APIs

NolaDonato:
Revival of Pull Request #1184 
This pull request moves GVRContext.loadMesh and loadFutureMesh functions
to GVRAssetLoader. It also removes the deprecated GVRContext functions
to load models, meshes and textures leaving only the asset loader
versions.

This patch removes all variants of loadModel, loadTexture, loadFutureTexture, loadMesh and loadFutureMesh from GVRContext. It changes the API.

I made patch requests for API changes to the demos and tests for the loadMesh call.


liaxim:
Looks good. Planning to merge Tuesday morning.

PR #1226, opened 2017-06-02
Extensions: Removing GVRContext loading APIs

liaxim:
Depends on: #1220

Just make small changes so the Extensions also compile.


liaxim:
Planning to merge Tuesday morning together with #1220.

PR #1230, opened 2017-06-08
Set the minimum support API level to 21

liaxim:


liaxim:
Thanks. Will merge Monday morning.

PR #1232, opened 2017-06-09
Provide UV and Barycentric coordinate information when picking objects with MeshColliders

J0Nreynolds:
This change modifies the picking classes and functions to calculate Barycentric and UV coordinates based on the HitPosition and the face which the picking ray passed through. Using an explicit formula for finding the Barycentric coordinates of our HitPosition, we can interpolate the UV coordinates of the three face vertices in order to find the UV coordinates of the HitPosition on the mesh. 

Also note that UV coordinates, Barycentric coordinates, and the collision face index are standard outputs of other raycasters such as Unity's.


liaxim:
Please revert unnecessary javadoc changes. Like the removal of the "<p/>" tags. Thanks.

J0Nreynolds:
I returned the HTML tags to their proper places (although it seems that many of these tags are unclosed or misaligned), but I can't keep the spaces at the end of comment lines, as Android Studio automatically strips those whitespace characters whenever I save. I can probably turn off this formatting-at-save, but I'm not sure how necessary it is. I would assume that most developers will be using Android Studio and these changes will happen eventually anyways. Let me know what you think.

NolaDonato:
What does this do if I am not using a mesh collider? Does it give me texture coordinates for the sphere or box colliders? I think this pull request is incomplete unless does something useful for all of the basic collider types. Maybe we can't do barycentric coordinates for all of them but perhaps we can pass something else that would be equally useful.

J0Nreynolds:
As of now, it populates the corresponding GVRPickedObject instance with Barycentric coordinates of `{ -1.0f, -1.0f, -1.0f }` and texture coordinates of `{ -1.0f, -1.0f }` if a GVRMeshCollider isn't used. This is supposed to indicate that Barycentric /Texture coordinates weren't calculated for the object since the coordinates are normally on a range of [0, 1].  This convention was based loosely off of [Unity's raycaster behavior](https://docs.unity3d.com/ScriptReference/RaycastHit-textureCoord.html):

> RaycastHit._textureCoord is a texture coordinate when a hit occurs. A Vector2 zero is returned if no mesh collider is present in the GameObject.

However, I do agree that supplying information about other colliders could prove useful (think spherical coordinates for SphereColliders, front/back/left/right/bottom/top face indication for BoxColliders).

J0Nreynolds:
@NolaDonato All texture and Barycentric coordinate information is now populated based on a boolean argument passed into GVRMeshCollider. The coordinate calculation functions have been moved from picker.cpp to mesh_collider.cpp in accordance with this change.

NolaDonato:
Good job. Now we need to visit the samples that use GUI views to see if this will help clean them up.

J0Nreynolds:
Yes, anything that features interaction with an Android View can be greatly simplified by using my new [GVRGUISceneObject](https://github.com/J0Nreynolds/GearVRf/blob/feature/GVRGUISceneObject/GVRf/Framework/framework/src/main/java/org/gearvrf/scene_objects/GVRGUISceneObject.java), which incorporates these picking changes and will have its own pull request once this one is merged.

NolaDonato:
Can we have some other opinions on whether removing the collider group functionality is OK? The GVREyePointee class allows you to group together multiple colliders to describe the collision geometry. I am thinking we should keep it and rename it GVRColliderGroup. What does everybody else think?

liaxim:
It is high time to get rid of eyePointee references. GVRColliderGroup sounds good to me too.

On the other hand this pull is changing too many things. It would be nice if the eyePointee stuff can be done separately and this kept to the topic as stated by the title.

J0Nreynolds:
@liaxim I agree that this change should probably be in a different PR. I will make the necessary changes and prepare a branch to remove the references to EyePointee and create a GVRColliderGroup object.

GVRf/Framework/framework/src/main/jni/engine/picker/picker_jni.cpp
```diff
@@ -104,24 +109,43 @@ Java_org_gearvrf_NativePicker_pickObjects(JNIEnv * env,
...
     SceneObject* scene_object =
             reinterpret_cast<SceneObject*>(jscene_object);
-    CameraRig* camera_rig = reinterpret_cast<CameraRig*>(jcamera_rig);
-    return Picker::pickSceneObject(scene_object, camera_rig);
+
+    ColliderData data;
+    Picker::pickSceneObject(scene_object, ox, oy, oz, dx, dy, dz, data);
+
+    jclass pickerClass = env->FindClass("org/gearvrf/GVRPicker");
+    jclass hitClass = env->FindClass("org/gearvrf/GVRPicker$GVRPickedObject");
```
liaxim: I know there is already code doing it but FindClass and GetMethod are not free. Please change the code to cache the results of these lookups.

liaxim:
Looks good to me and what I tested worked as expected. Can be merged even with the extra JNI calls.
@NolaDonato let us know whether you are ok with merging this. Thanks.

NolaDonato:
Yes, it looks good.

PR #1233, opened 2017-06-12
3DCursorLibrary: eliminate a singleton causing the input to not work on some S8s

liaxim:
On Tom's S8 two activities could get created in a quick succession on launching Erik's app. The second activity adds cursor controller listeners to the wrong instances of the cursor controllers due to the caching by the singleton.

I could reproduce the problem by first locking the screen then launching the app from the launcher. Was happening for me 50% of times.

I think it is possible to end up with two cursors (gaze+mouse) but that would be a different problem. The input would work though.

---


liaxim:

PR #1234, opened 2017-06-12
framework: stop the fw thread from onDestroy

liaxim:
leaking a GVRContext is too easy and common so just stop it from onDestroy

---


liaxim:

PR #1236, opened 2017-06-13
[DockListener] separated addDockListener from the constructor

sisimma:
- changed it so that constructor won't call it. Instead it's called separately.

Signed-off-by: Sean Yoon <sukhwan.y@samsung.com>


liaxim:
@sisimma Sean, do you still need this change?

sisimma:
Yes I do.
Thanks.

liaxim:
Ok. Planning to merge Thursday morning.

PR #1237, opened 2017-06-14
framework, backend_oculus: refactor OvrGearController

liaxim:


b1foster:
Probably workable. Will have to explore in more detail but the reader class is definitely much cleaner.

liaxim:
Verified as working with gvr-controller. With or without this change, there seems to be a picking problem. It is independent of this change. Not all scene objects are getting picked as I would expect.

liaxim:
Planning to merge Thursday morning.

PR #1238, opened 2017-06-14
ttf Font implementation

mwitchwilliams:

Added DEF/USE capability to X3D &lt;FontStyle&gt; node, so Font Styles can be shared by text within an X3D file.
Addressed issues presented in 'ww_font_PullReq' pull request by implementing a new Constructor in GVRTextViewSceneObject and an inner class 'Text_FontParams' in X3DObject.java.
Also implemented the gvrScene.setBackground() command per request from previous pull request.

This pull request supersedes 'ww_font_PullReq' and request that one be closed.


mwitchwilliams:
X3D test cases with documentation and ,ttf font files can be found in the GVRf-Tests repository under x3d -> fonts

liaxim:
This looks much better. Thanks.

Looking at the new ctor, the Builder pattern comes to mind: https://en.wikipedia.org/wiki/Builder_pattern.

GVRf/Framework/framework/src/main/java/org/gearvrf/scene_objects/GVRTextViewSceneObject.java
```diff
@@ -224,6 +365,21 @@ public void setText(CharSequence text) {
         mIsChanged = true;
     }
 
+    public void setText(String text) {
```
liaxim: This method is unnecessary. String implements CharSequence. We can remove later.

liaxim:
I think the justification should have been kept inside x3d and translated by x3d to/from the Android Gravity flags instead of pushing this foreign concept into the TextViewSceneObject. We can live with it.

Planning to merge Thursday morning.

liaxim:
Tests match the expected output but the test pull request need rebase.

PR #1239, opened 2017-06-15
framework: set a web client in GVRWebView

liaxim:
otherwise gvr-renderableview opens a regular full-screen browser window on invoking loadUrl


liaxim:
Happens on my S8-s and S7.

PR #1240, opened 2017-06-16
Create GVRGUISceneObject

J0Nreynolds:
This PR creates a scene object that not only displays an Android View but also enables interaction with the View using the new picking information supplied by #1232. In order to accomplish this, the SensorEvents API has been updated to supply all picking information (in the form of a GVRPickedObject) rather than only supplying a HitPoint.


J0Nreynolds:
Will reopen once #1232 has been reviewed and any changes are merged into the necessary branches.

PR #1247, opened 2017-06-22
Physics constraints

JulianaFigueira:
This adds the following constraints features to the physics-extension:
Fixed
Point2Point
Hinge
Slider
ConeTwist
Generic (w/ 6 degrees of freedom)

GearVRf-DCO-1.0-Signed-off-by: Cristiano Bozzetto c.bozzetto@samsung.com 
GearVRf-DCO-1.0-Signed-off-by: Juliana Figueira juliana.f@samsung.com


GVRf/Framework/framework/src/main/jni/objects/components/mesh_collider.h
```diff
@@ -40,7 +42,18 @@ class MeshCollider: public Collider {
     }
 
     Mesh* mesh() const {
-        return mesh_;
+        if (mesh_) {
```
liaxim: I would like to understand the reason for all these nullptr checks. Here and in render_data.cpp.

liaxim:
I can see this includes some fixes from the past when we tried to patch up gvr-simplephysics. I believe the locking on the native side is not enough since there is state on the Java side too. In any case, we should separate out the thread-safety changes from the feature changes and have two pull requests. @NolaDonato Please review.

Also some of the nullptr checks just cover up problems, possibly deeper, in my opinion.

NolaDonato:
I suspect it is because some of our built in scene objects create renderdatas without meshes. We saw that some cylinder variants do this.



NolaDonato:
We need to lock around changes to the transforms at the C++ level to fix it for real.



GVRf/Extensions/gvrf-physics/src/main/java/org/gearvrf/physics/GVRWorld.java
```diff
@@ -113,9 +133,15 @@ public void removeBody(GVRRigidBody gvrBody) {
 
     @Override
     public void onDrawFrame(float frameTime) {
-        NativePhysics3DWorld.step(getNative(), frameTime);
-
+        mFrameTime += frameTime;
+        if (mIsProcessing) {
```
liaxim: @NolaDonato Why do you think onDrawFrame needs to make provisions to be reentrant? It is only called once per frame by the view manager. While technically it is a public method and the user could call it any time, they shouldn't and we mention that somewhere. And the method doesn't have to be public at all to do what it does.

GVRf/Extensions/gvrf-physics/src/main/jni/engine/bullet/bullet_conetwistconstraint.cpp
```diff
@@ -0,0 +1,140 @@
...
+    BulletConeTwistConstraint::~BulletConeTwistConstraint() {
+        if (0 != mConeTwistConstraint) {
+            delete mConeTwistConstraint;
+        }
+    }
+
+    void BulletConeTwistConstraint::setSwingLimit(float limit) {
+        if (0 != mConeTwistConstraint) {
+            mConeTwistConstraint->setLimit(4, limit);
+            mConeTwistConstraint->setLimit(5, limit);
```
liaxim: Is setLimit for limitIndex of 5 necessary?

GVRf/Extensions/gvrf-physics/src/main/jni/engine/bullet/bullet_conetwistconstraint.h
```diff
@@ -0,0 +1,76 @@
...
+
+        void setSwingLimit(float limit);
+
+        float getSwingLimit() const;
+
+        void setTwistLimit(float limit);
+
+        float getTwistLimit() const;
+
+        void set_owner_object(SceneObject* obj);
```
liaxim: It would be nice if you could use the override keyword for overridden methods.

GVRf/Extensions/gvrf-physics/src/main/jni/engine/bullet/bullet_conetwistconstraint.h
```diff
@@ -0,0 +1,76 @@
...
+    class PhysicsRigidBody;
+    class BulletRigidBody;
+
+    class BulletConeTwistConstraint : public PhysicsConeTwistConstraint, BulletObject {
+    public:
+        BulletConeTwistConstraint(PhysicsRigidBody *rigidBodyB, PhysicsVec3 pivot,
+                                  PhysicsMat3x3 const &bodyRotation,
+                                  PhysicsMat3x3 const &coneRotation);
+
+        ~BulletConeTwistConstraint();
```
liaxim: Since this class has virtual methods the dtor must be virtual too. Please check the rest of the classes.

GVRf/Extensions/gvrf-physics/src/main/jni/engine/bullet/bullet_rigidbody.cpp
```diff
@@ -111,13 +168,31 @@ void BulletRigidBody::setCenterOfMass(const Transform *t) {
...
+    Transform* trans = owner_object()->transform();
+    btTransform aux; getWorldTransform(aux);
+
+    if(std::abs(aux.getOrigin().getX() - prevPos.getOrigin().getX()) >= 0.1f ||
+       std::abs(aux.getOrigin().getY() - prevPos.getOrigin().getY()) >= 0.1f ||
+       std::abs(aux.getOrigin().getZ() - prevPos.getOrigin().getZ()) >= 0.1f)
+    {
+        mRigidBody->setWorldTransform(aux);
+        prevPos = aux;
+        //TODO: incomplete solution
```
liaxim: Incomplete solution to what? What should it be been doing instead?

liaxim:
Looks ok to me. A lot more stuff from Bullet is wrapped now. Some minor things to do. @NolaDonato ?

GVRf/Extensions/gvrf-physics/src/main/jni/engine/bullet/bullet_conetwistconstraint.cpp
crbozz: Bullet has two swing spam limits which allows one to set a "rollercoaster like" translation trajectory when these limits have different values. I assume this was not a valuable feature then I decided to set the same value to both swing spam limits.

NolaDonato:
The reason I put provisions in onDrawFrame is that I think it is possible for it to be called again before the previous one has finished. Physics can easily take more than one frame.

GVRf/Extensions/gvrf-physics/src/main/jni/engine/bullet/bullet_rigidbody.cpp
JulianaFigueira: This part is about updating rigid bodies whose scene object transform were changed outside the physics engine, currently we trade one physics step for the updating and static object are not updated as the physics ignores them. The comment is about looking for a more elegant solution. It will be deleted on the next commit.

liaxim:
Planning to merge next week

PR #1249, opened 2017-06-22
framework: add an app theme that disables the so called preview window

liaxim:
- the white screen with a titlebar that flashes when launching gvr-immersivepedia will be gone; the app will not launch faster but the transition to it can be smoother in case you happen to launch it while wearing the headset
- windowBackground might not be necessary but just in case

Dependent pull request: https://github.com/gearvrf/GearVRf-Demos/pull/462


liaxim:

PR #1253, opened 2017-06-26
framework,gradle: allow supplying build configuration properties from a file that is not under source control nor global

liaxim:

### DIFF
--- a/GVRf/Framework/build.gradle
+++ b/GVRf/Framework/build.gradle
@@ -1,3 +1,7 @@
+if(file("../../../extra_properties.gradle").exists()) {
+    apply from: '../../../extra_properties.gradle'
+}
+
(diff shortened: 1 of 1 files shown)


liaxim:
To avoid modifying either the project-specific gradle.properties or the global one create an extra_properties.gradle file in the same location you would create extra_settings.gradle. Unfortunately these two can't be merged.

Sample extra_properties.gradle:
```
allprojects {
    project.ext {
		useLocalDependencies = true
		OCULUS_SIGS_DIRECTORY = "~/oculussigs"
		OVR_MOBILE_SDK = "~/ovr_mobile_sdk/1.5.0"
		OVR_PLATFORM_SDK = "~/ovr_mobile_sdk/platform_1.7.0"
    }
}
```

liaxim:
If you prefer to use gradle.properties you can still do that instead.

PR #1254, opened 2017-06-26
framework, controller: add a new controller event

liaxim:
Add a new event that is broadcast every frame with the position and orientation

---

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVREventListeners.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVREventListeners.java
@@ -1,6 +1,8 @@
 package org.gearvrf;
 
 import org.gearvrf.script.IScriptable;
+import org.joml.Quaternionf;
+import org.joml.Vector3f;
 
 import android.content.Intent;
 import android.content.res.Configuration;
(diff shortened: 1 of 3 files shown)


liaxim:
The event instance may be recycled but the Vector3f and Quaternionf instances will live on as long as some keeps references to them.

liaxim:
Will pass the pointF too.

liaxim:
Touchpad point added

NolaDonato:
Unfortunately you are correct :-(



liaxim:
Sending the event directly from the controller thread. If the user needs to keep the event data around, he or she should copy it.

SteveGreatApe:
Thanks, I've copied the changes into my framework build and it's working okay.

I'm wondering if it would make more sense for this callback to be integrated into GVRCursorController.ControllerEventListener as that's where I'm receiving KeyEvent's from the controller and it seems a bit odd that I'm using two different mechanisms for the Key and Touch/Orientation events.

Also, is there an official define for the maximum x & y co-ordinates the touchpad can deliver? The min/max I'm picking up at the moment seem to be 7 & 314, I'm guessing the official limits are 0 & 320 and a hardware issues mean I don't quite get those values.

liaxim:
If you want to continuously receive the touch&orientation then the on-key event is not going to work. This mechanism, while not terribly nice, gives a bit more flexibility. Posting android.view.KeyEvent limits us to what it can represent. We may have to switch to custom events. Also, in general, we will have to look at the design of the whole thing.

liaxim:
For the maximum coordinates, the information is in ovrInputTrackedRemoteCapabilities but we are not exposing it yet. Will keep this issue open until we do.

liaxim:
Planning to merge this on Friday.

PR #1255, opened 2017-06-26
X3D Touch Sensor controlling animation in FBX file

mwitchwilliams:
Enables x3d files with objects containing Touch Sensors, to start fbx files that have key frame animations using the &lt;ROUTE&gt; node in X3D from the Touch Sensor to the inlined fbx file.
To invoke the entire animation, set the 'toField' to 'set_enabled':

&lt;ROUTE fromField="isOver" fromNode="_myTouchObj_"
	toField="**set_enabled**"  toNode="_someObject_"/&gt;

for an animation from one frame to another, use:
&lt;ROUTE fromField="isOver" fromNode="_myTouchObj_"
	toField="**2, 5**"  toNode="_someObject_"/&gt;
which will run the animation from key frame 2 to key frame 5.

_myTouchObj_ is the DEFined name of the X3D &lt;TouchSensor&gt; node.
_someObject_ is the DEFined name of the X3D &lt;Inline&gt; node with the url of the fbx file.

There is a pull request of test cases and documentation in the GearVRf-Tests repo under:
 x3d->control_fbx_key_frame_animations


liaxim:
Sooner rather than later we should move x3d to the org.gearvrf package. It seems you need too many public methods for no other reason but because x3d is in a separate package. Looks like x3d needs to be tightly integrated with the rest so it belongs to the same package then.

mwitchwilliams:
Most of the files in the X3D folder support X3D activities and don't interact with GearVR framework.  Perhaps only 3 files that interface with GearVR framework are: X3Dobject.java (the parser), AnimationInteractivityManager.java, which acts on sensors, and AnchorImplementation.java, the X3D version of HTML's Anchor tags going to other web files including x3d pages, or cameras.

So, perhaps just these 3 files..

NolaDonato:
Why does it need more than one public method. Isn't it only referenced from the asset loader?



mwitchwilliams:
I think the reference was to having to return some protected values in one of the GVRAnimations that had a reference to GVRKeyFrameAnimation.  I had to access the saved key frame animations in order to build new key frame animations from the specified key frames.

GVRf/Framework/framework/src/main/java/org/gearvrf/animation/GVRAnimation.java
```diff
@@ -471,4 +471,9 @@ public float getElapsedTime() {
      *            The start state is 0; the stop state is 1.
      */
     protected abstract void animate(GVRHybridObject target, float ratio);
+
+    public GVRHybridObject getTargetObj () {
```
liaxim: Public API needs javadoc.

GVRf/Framework/framework/src/main/java/org/gearvrf/x3d/AnimationInteractivityManager.java
```diff
@@ -628,17 +663,122 @@ public void onSensorEvent(SensorEvent event) {
...
+                                            GVRSceneObject inlineSceneObject = inlineObject.getInlineGVRSceneObject();
+                                            GVRAnimator animator = (GVRAnimator) inlineSceneObject.getComponent(GVRAnimator.getComponentType());
+                                            if ( animator != null ) {
+                                                List<GVRAnimation> gvrAnimations = animator.getAnimationList();
+                                                GVRAnimation gvrAnimation = null;
+
+                                                if  ( (inlineObject.getBeginKeyFrame() == InlineObject.UNSET_KEY_FRAME) || (inlineObject.getEndKeyFrame() == InlineObject.UNSET_KEY_FRAME)) {
+                                                    if ( (inlineObject.getBeginKeyFrame() == InlineObject.UNSET_KEY_FRAME) && (inlineObject.getEndKeyFrame() == InlineObject.UNSET_KEY_FRAME)) {
+                                                        // use the original FBX animation.
+                                                        gvrAnimation = gvrAnimations.get(0);
```
liaxim: What if there are no animations in the list? What are there guarantees there will be at least one?

GVRf/Framework/framework/src/main/java/org/gearvrf/x3d/AnimationInteractivityManager.java
```diff
@@ -628,17 +663,122 @@ public void onSensorEvent(SensorEvent event) {
...
+                                                        Log.e(TAG, "Error: ROUTE specifying the keyFrame range - only 1 may be an integers");
+                                                    }
+                                                }
+                                                else {
+                                                    // FBX animation specified in ROUTE as a set of keyFrames
+                                                    // play only a subset of the animation based on keyframes
+                                                    gvrAnimation = interactiveObjectFinal.getFBXAnimation();
+                                                    // check if we already copied over the animation
+                                                    if (gvrAnimation == null) {
+                                                        GVRAnimation gvrOriginalAnimation = gvrAnimations.get(0);
```
liaxim: What if there are no animations in the list? What are there guarantees there will be at least one?

GVRf/Framework/framework/src/main/java/org/gearvrf/x3d/AnimationInteractivityManager.java
```diff
@@ -628,17 +663,122 @@ public void onSensorEvent(SensorEvent event) {
...
+                                                        if (gvrOriginalAnimation instanceof GVRKeyFrameAnimation) {
+                                                            GVRKeyFrameAnimation gvrOriginalKeyFrameAnimation = (GVRKeyFrameAnimation) gvrOriginalAnimation;
+                                                            GVRAnimationChannel gvrOriginalAnimationChannel = gvrOriginalKeyFrameAnimation.getChannel(0);
+
+                                                            try {
+                                                                // copy the original animation, then save just the requested frames
+                                                                // and their translation, scale and rotate
```
liaxim: What if you could modify the existing animation instead of creating a new one? Could you please explain what is going in the block and why?

liaxim:
The X3D parser as-is is unusable without GVRf. It should be all be in the same package as GVRf since it requires, and will probably continue to, access to non-public APIs and state.

NolaDonato:
X3dObject.java is almost 4000 lines long and indents so deeply the code is not readable. Please make some internal classes or functions or something other than a with 20 levels of indentation.

GVRf/Framework/framework/src/main/java/org/gearvrf/x3d/AnimationInteractivityManager.java
NolaDonato: I don't think FBX guarantees there will be position, scale and rotation keys always. I think you can sometimes get rotation keys without scale and position. In this case, your code accessing these keys will fail.

mwitchwilliams: My tests have animations that have translation and scale, but no rotation.  Or rotation with translation, and no scale, and they all passed.  Tests cases in the tests repo.

I agree this needs to be tested in case they use an FBX file with no animations, yet have an X3D file that tries to begin an animation with none there.

GVRf/Framework/framework/src/main/java/org/gearvrf/animation/GVRAnimation.java
mwitchwilliams: Got it, thanks.

GVRf/Framework/framework/src/main/java/org/gearvrf/x3d/AnimationInteractivityManager.java
mwitchwilliams: We don't have a function to call GVRAnimation or GVRAnimator to begin and end at specific key frames.  Currently, we can only begin at key frame 0 and ends at the end of the animation.
However, selecting from one key frame to another would be a great feature for GearVR.  3d artists often build character animations on a single mesh, and to go from specific key frames on the same mesh.  For example, 'running' would be key frame 0 to 1, 'crawling' could be keyframe 1 to 2, etc.  Currently, GearVR doesn't allow selecting from one specific key frame to another.
The feature I wrote was specific to control FBX animations from X3D, a non-standard, non-documented feature requested by our art team.
We should discuss this to see if we want to instead modify GVRAnimation to see if we can implement the ability to select an animation from specific key frames.

mwitchwilliams:
I know Nola has changes to X3DObject parser relating to GVRMesh, and Tom and I scheduled some issues that I need to improve X3D parser.  I agree the X3D parser needs some clean-up and this will happen after SIGGRAPH.
Since this X3D feature calling an FBX animation is an undocumented, non-spec, work-around, what about a re-work of GVRAnimation to enable selecting animations from specific key frames?
It really seems like invoking an animation for specific key frames would be a nice feature for animators working in FBX.
Is there anything I should check into regarding moving X3DObject into GVRf same package?

NolaDonato:
We are planning an overhaul of the entire animation system later this year. If you want to rework GVRAnimation to supply the features you need that is a good first step. I won't be able to work on animation until after Siggraph.

mwitchwilliams:
Upon Tom's approval (or Mihail or Nola), my recommendation is to withdraw this Pull Request, and modify GVRAnimation to enable selecting animations by a range of key frames (from key frame 4 to 9 for example), that would be useful to FBX or X3D animations.  My modifications were based on a request from the art team, which can still use this code in the mean time for their own development and testing.

liaxim:
Modifying GVRAnimation sounds good. It seems like a useful capability in general.

NolaDonato:
Go for it!

liaxim:
@mwitchwilliams So we close this one?

mwitchwilliams:
Unless you think we should wait for Tom, then go ahead and close it.

liaxim:
We can always reopen unless you delete the branch.

PR #1256, opened 2017-06-27
Update SensorEvents API to supply new picking information

J0Nreynolds:
This update modifies the SensorEvents API to

1.  Use an object's respective collider for picking, rather than defaulting to picking against a bounding box
2.  Supply GVRPickedObject in SensorEvent instances instead of only a float[3] HitLocation

The modification to use a SceneObject's collider reflects the fact that a developer chooses a collider according to how accurately they wish to raycast at an object. For less accurate, but also less expensive raycasting, a developer may attach a box or sphere collider. For an object with complex geometry (holes, curves, or oblong appearance), a mesh collider may be more appropriate.  Furthermore, a mesh collider can be attached to an object to supply more picking information (such as that introduced by #1232) , while still maintaining low cost picking (think about picking the texture coordinates of a quad, which only has two faces to process). 

Since all picking information is already passed around in the form of a GVRPickedObject, I believe it makes more sense to supply this in SensorEvents rather than only storing the HitLocation.  Also, I removed the storage of the GVRSceneObject since the picked GVRSceneObject is supplied in GVRPickedObject instances.

### DIFF
--- a/GVRf/Extensions/3DCursor/3DCursorLibrary/src/main/java/org/gearvrf/io/cursor3d/LaserCursor.java
+++ b/GVRf/Extensions/3DCursor/3DCursorLibrary/src/main/java/org/gearvrf/io/cursor3d/LaserCursor.java
@@ -19,6 +19,7 @@ package org.gearvrf.io.cursor3d;
 import org.gearvrf.GVRContext;
 import org.gearvrf.GVRCursorController;
 import org.gearvrf.GVRCursorController.ControllerEventListener;
+import org.gearvrf.GVRPicker;
 import org.gearvrf.GVRSceneObject;
 import org.gearvrf.SensorEvent;
 import org.gearvrf.io.cursor3d.CursorAsset.Action;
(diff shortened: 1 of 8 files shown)


NolaDonato:
Since you are working in the controller API, please take a look at this pull request made by rahul:
https://github.com/rahul27/GearVRf/commit/c334da6873199aa3310e7e88180b987d625267ab
It attempt to fuse the picker and the cursor API in a sensible way. We actually want these to work together better than they do now so maybe Rahul's code will help. I was supposed to look at this but I haven't had the time...

NolaDonato:
Storing the hit location in the scene object is not thread-safe. Your fix to use the picked object information is much better.

liaxim:
The 3DCursorLibrary needs updates too. Go to GVRf/Extensions/ and run ./gradlew assembleDebug to build the extensions.

J0Nreynolds:
@liaxim The 3DCursor library has been updated and the Extensions project directory builds without errors.

liaxim:
Looks good. Will merge on Wednesday.

liaxim:
Needs DCO per https://resources.samsungdevelopers.com/Gear_VR_and_Gear_360/GearVR_Framework_Project/040_Get_Involved/GearVRf_Developer_Certificate_of_Origin

J0Nreynolds:

PR #1257, opened 2017-06-27
Create GVRGUISceneObject

J0Nreynolds:
This pull request introduces a SceneObject that extends the GVRViewSceneObject to handle user interaction through GVRCursorControllers. With the modifications to SensorEvents introduced by  #1256, this class can forward motion events to its respective GVRView without any need for user setup, which makes it a desireable for creating menus and other user interfaces using only Android Views, which is easy to do using Android's Studio's Layout editor.

Additionally, the class features the option for a mesh to be generated based on the Android View passed to the constructor. Using the width and height of the View, the class can generate an appropriate planar or curved surface for the View to be displayed on.

This is marked as in progress because #1256 is yet to be reviewed/merged, and some work needs to be done in order to support interaction with the GearCursorController.

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/scene_objects/GVRGUISceneObject.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/scene_objects/GVRGUISceneObject.java
@@ -20,7 +20,23 @@ import java.util.List;
 
 
 /**
- * Created by j.reynolds on 6/12/2017.
+ * This class extends the {@link GVRViewSceneObject} to make interaction with Android
+ * Views simple. This class achieves the following:
+ *
+ * 1.  Allows any {@link org.gearvrf.GVRCursorController} to be used for interactions
+ * with the {@link GVRView} displayed. This is achieved with an {@link ISensorEvents} listener
+ * that uses the motion events set in {@link org.gearvrf.GVRCursorController}s to dispatch
+ * appropriate touch events to the Android View contained in the {@link GVRView}.
+ *
+ * 2. Generates an appropriate mesh for displaying the {@link GVRView} passed to the
+ * constructor. The mesh will always have the correct aspect ratio, and the user can
+ * specify the curvature of the mesh if desired.
+ *
+ * To use this class, simply instantiate a new {@link GVRGUISceneObject} with the appropriate
+ * arguments, and add the instance to the current {@link org.gearvrf.GVRScene} using
+ * {@link org.gearvrf.GVRScene#addSceneObject}. Then use the application's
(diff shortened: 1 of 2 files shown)


NolaDonato:
I suggest exposing createCurvedMesh as a static function on GVRMesh so users can call it directly. In the Vulkan branch I moved GVRContext.createQuad to GVRMesh.createQuad so we already have static mesh constructions functions.

J0Nreynolds:
@NolaDonato Thank you for the suggestion! I'll make that change now with additional parameters such as a central angle argument so users can choose the curvature of the mesh themselves.

liaxim:
Is this still in progress? Do you have an app where we could try it out?

J0Nreynolds:
Yes, this is essentially finished. I have an application that I've created for testing on [this branch](https://github.com/J0Nreynolds/GearVRf-Demos/tree/gui-demo). The application is in the gvr-gui folder.

Please note that only GVRCursorControllers that dispatch motion events will be able to interact with the GVRGUISceneObjects, which means the GearCursorController can't be used at the moment. I've attempted to modify the GearCursorController myself, dispatching ACTION_DOWN, ACTION_UP and ACTION_MOVE motion events at appropriate times, but there seems to remain some problem I'm unable to debug. I believe that this can be addressed separately though, as no further modifications are needed for this feature.

J0Nreynolds:

liaxim:
Please look into adding this to the vulkan branch.

liaxim:
This could be rather useful for the users that prefer Android views. I would argue it all belongs to GVRViewSceneObject but we will look into it in the future.

GVRf/Framework/framework/src/main/java/org/gearvrf/scene_objects/GVRGUISceneObject.java
```diff
@@ -0,0 +1,175 @@
...
+import org.gearvrf.GVRMeshCollider;
+import org.gearvrf.GVRPicker;
+import org.gearvrf.ISensorEvents;
+import org.gearvrf.SensorEvent;
+import org.gearvrf.scene_objects.view.GVRView;
+
+import java.util.List;
+
+
+/**
```
liaxim: It would be real nice if you could give a high-level overview what is this all about; what it does for users; short explanation how to use. Thanks.

liaxim:
Unfortunately can't be just cherry-picked onto vulkan since it depends on few prior commits. @dnvasilev Sorry.

liaxim:
Looks good to me. Planning to merge on Monday.

J0Nreynolds:
Thanks for checking it out. I'll add an description/overview at the top of the class. Also, I believe this can be cherry picked after #1364 is merged into the vulkan branch.

PR #1262, opened 2017-06-28
Adding Push Constants support for Material UBO. 

apjagdale:
Note: The size of Material UBO cannot be more than 128 bytes


GVRf/Framework/framework/src/main/jni/vulkan/vulkanCore.cpp
```diff
@@ -548,9 +549,14 @@ namespace gvr {
                                           &descriptorLayout);
         GVR_VK_CHECK(!ret);
 
+        VkPushConstantRange pushConstantRange = {};
+        pushConstantRange.offset                        = 0;
+        pushConstantRange.size                          = (uint32_t) vkMtl.uniforms().getTotalSize();
+        pushConstantRange.stageFlags                    = VK_SHADER_STAGE_FRAGMENT_BIT;
```
roshanch: Lets enable it for both the shaders, custom shaders might use them in vertex shaders as well.

PR #1264, opened 2017-06-28
Fix 360 video

dnvasilev:
Different phone models need different extensions for samplerExternalOES:
Adreno-based, GL_OES_EGL_image_external_essl3
Mali-based, GL_OES_EGL_image_external
Instead of 'require' a particular extension in the fragment shader, 'enable' both.


roshanch:
So the devices not supporting one of the extension strings still work fine with this change?

dnvasilev:
TL;DR: Yes, it works.
All well-behaved shader compiler must conform to Khronos GLSL ES specification, more formally known as "The OpenGL ES® Shading Language" specification. I am looking at Version 3.00. Chapter3.5, "Preprocessor" of the Section 3 "Basics", explains the semantics of "enable" behavior in the following way:
  "Behave as specified by the extension extension_name.
Warn on the #extension if the extension extension_name is not supported.
Give an error on the #extension if all is specified."
Compiler writers usually interpret this clause in the following way: if a device is not supporting an extension, compiler issues a warning, and ignores the declaration.

NolaDonato:
Did we test this with the demos that use external textures on both Mali and Adreeno? If so then it is good to go.

PR #1265, opened 2017-06-28
framework: correct missing return statements, fix include guard

liaxim:


liaxim:
Planning to merge on Friday.

Issue #755, opened 2016-07-29
Extension 'GL_OES_EGL_image_external' not supported


rahul27:
gvr-camera2renderscript and gvr-simplegallery crash with the following log. This happens after https://github.com/Samsung/GearVRf/pull/622

Note that both apps use the OES shader.
````
07-29 10:00:13.840 24430-24430/com.samsung.android.gvr.sample I/GVRViewManager: camera rig yaw reset
                                                                                
                                                                                [ 07-29 10:00:13.860 24430:24472 D/         ]
                                                                                MALI GL_COMPILE_STATUS is not GL_TRUE, 0
07-29 10:00:13.860 24430-24472/com.samsung.android.gvr.sample E/gvrf: Could not compile shader 35632:
                                                                      2:1: P0003: Extension 'GL_OES_EGL_image_external' not supported
                                                                      2:3: L0001: Expected token '{', found 'identifier'
07-29 10:00:13.860 24430-24472/com.samsung.android.gvr.sample A/libc: Fatal signal 11 (SIGSEGV), code 1, fault addr 0x3c in tid 24472 (GLThread 520)
````

rahul27:
@roshanch any idea what might be wrong here?

roshanch:
https://community.arm.com/thread/5204 just found that this extension is not supported by GLSL 300. Khronos has other extension for this but not sure whether this is present in current driver, we can give it a  shot. Replace the string in shader with "GL_OES_EGL_image_external_essl3". Workaround is we have to fall-back to GLSL 100 for this extension which does not support multiview.

rahul27:
See the same error with GL_OES_EGL_image_external_essl3

````
07-29 10:30:38.040 30154-30418/com.samsung.android.gvr.sample I/OVR: DeviceManagerThread - poll(fds,1,1000) = 0 (Tid=30418)
07-29 10:30:38.120 30154-30154/com.samsung.android.gvr.sample I/GVRViewManager: camera rig yaw reset
07-29 10:30:38.140 30154-30353/com.samsung.android.gvr.sample E/gvrf: Could not compile shader 35632:
                                                                      2:1: P0003: Extension 'GL_OES_EGL_image_external_essl3' not supported
                                                                      2:3: L0001: Expected token '{', found 'identifier'
[...]

PR #1267, opened 2017-06-28
Fix EGL_image_external

dnvasilev:
1. Different phone models need different extensions for samplerExternalOES:
Adreno-based, GL_OES_EGL_image_external_essl3
Mali-based, GL_OES_EGL_image_external
Instead of 'require' a particular extension in the fragment shader, 'enable' both.

2. Uniform locations are always GLint
Uniform location -1 is reserved for situations when a particular uniform was optimized out by compiler.

### DIFF
--- a/GVRf/Framework/framework/src/main/jni/shaders/material/oes_horizontal_stereo_shader.cpp
+++ b/GVRf/Framework/framework/src/main/jni/shaders/material/oes_horizontal_stereo_shader.cpp
@@ -38,7 +38,8 @@ static const char VERTEX_SHADER[] = "attribute vec3 a_position;\n"
         "}\n";
 
 static const char FRAGMENT_SHADER[] =
-        "#extension GL_OES_EGL_image_external : require\n"
+        "#extension GL_OES_EGL_image_external : enable\n"
+        "#extension GL_OES_EGL_image_external_essl3 : enable\n"
                 "precision highp float;\n"
                 "uniform samplerExternalOES u_texture;\n"
                 "uniform float u_opacity;\n"
--- a/GVRf/Framework/framework/src/main/jni/shaders/material/oes_shader.cpp
+++ b/GVRf/Framework/framework/src/main/jni/shaders/material/oes_shader.cpp
@@ -29,30 +29,14 @@
 namespace gvr {
 static const char VERTEX_SHADER[] =
         "#version 300 es\n"
-        "attribute vec4 a_position;\n"
-        "attribute vec2 a_texcoord;\n"
+        "in vec4 a_position;\n"
+        "in vec2 a_texcoord;\n"
         "uniform mat4 u_mvp;\n"
-        "varying vec2 v_tex_coord;\n"
+        "out vec2 v_texcoord;\n"
         "void main() {\n"
-        "  v_tex_coord = a_texcoord.xy;\n"
+        "  v_texcoord = a_texcoord.xy;\n"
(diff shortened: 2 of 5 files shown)


dnvasilev:
We should keep '#version 300 es' in the shaders for the following reasons:
(1) MULTIVIEW shader matches the unchanged non-multiview shader that already has this directive.
(2) Omitting version directive is equivalent to declaring '#version 100 es' according to [https://www.khronos.org/registry/OpenGL/specs/es/2.0/GLSL_ES_Specification_1.00.pdf](url) Section 3.4. The shader is not conformant to GLSL ES 1.00: it uses 'in' and 'out' storage specifiers instead of 'varying'.
GLProgram class passes shader source unchanged to glShaderSource() and specifically does not add any version directives to the source.

NolaDonato:
GVRShaderTemplate and GVRShaderTemplate adds the directive for you.



dnvasilev:
This is master branch, and I am modifying old-style native-side oes_shader.cpp
We use GLProgram(const char*, const char*) constructor that calls GLProgram::createProgram(int, const char**, const GLint*, const char**, const GLint*), that calls GLProgram::loadShader(GLenum, int, const char**, const GLint*), that calls glShaderSource(). Nothing adds the directive on the way. No GVRShaderTemplate -- this is in the native code.

roshanch:
I don't remember exactly why I did keep glsl 100 es shaders, but we should check multiview enabled with oes_shader once.

dnvasilev:
Code before this change was using GLSL ES 1.00 shaders with multiview disabled, and GLSL ES 3.00 shaders with multiview enabled.
@roshanch , what tests do you propose to run with multiview enabled and with multiview disabled?

roshanch:
Try 360-video with multiview enabled on Mali s6 and s7.

liaxim:
I think we still have to keep using GLSL 1.0 shaders for the non-multiview case. This is the issue Rahul encountered, unfortunately we don't know the device: https://github.com/Samsung/GearVRf/issues/755. Roshan's fix https://github.com/Samsung/GearVRf/pull/759.

dnvasilev:
@liaxim No. My fix fixes #755 properly.

liaxim:
Why is it fixed properly? Rahul said he was getting errors for both:
``2:1: P0003: Extension 'GL_OES_EGL_image_external' not supported ``
``2:1: P0003: Extension 'GL_OES_EGL_image_external_essl3' not supported``

NolaDonato:
As I recall, we are using version 100 in some of the demos because when we coded them with version 300 they did not work on all devices. This was especially apparent in the cinema demo gvr-video. We should try recoding the shaders in this demo as level 300 instead of 100 and see if it is really fixed. When I did that, the colors in the external texture for the video were wrong.

dnvasilev:
Rahul's error P0003 is triggered by 'require'. This is fixed by 'enable'.
I need a positive proof that fix doesn't work.

liaxim:
Ok, there is missing information from Rahul's issue so it is hard to argue.

I do expect this to work but it seemed like there was some weird device/sw combination where neither worked for 300. Can't tell from Rahul's issue whether he was testing on the same device with 300 shader or not.

dnvasilev:
I plan to merge vertex shaders (#ifdef MULTIVIEW) and test on a multitude of phones

liaxim:
Planning to merge on Monday.

PR #1270, opened 2017-06-30
Set glDepthMask from application.

sushantojal:
Api to enable/disable writing to depth buffer for a scene object.


GVRf/Framework/framework/src/main/jni/engine/renderer/gl_renderer.cpp
```diff
@@ -306,6 +311,11 @@ namespace gvr
             GL(glEnable(GL_DEPTH_TEST));
         }
 
+        if (!render_data->depth_mask())
+        {
+            GL(glDepthMask(GL_FALSE));
```
liaxim: Assuming the default state is that the depth mask is enabled, shouldn't we enable it from restoreRenderStates?

sushantojal: yes, will do!

liaxim:
Looks good. Planning to merge on Monday.

PR #1281, opened 2017-07-11
Two culling calls when using monoscopic rendering mode.

sushantojal:
When using monoscopic rendering, "OVRMonoscopicViewManager" calls "beforeDrawEyes()" from "onDrawFrame()". Since this class extends from "OVRViewManager"  and it is not overriding this method , the "beforeDrawEyes()" of the "OVRViewManager" gets called, where culling takes place.  It appears as if the removed culling call in this commit is redundant because is actually causing a second cull pass to happen.


NolaDonato:
Looks correct to me. I traced through the code and it calls GVRViewManager.beforeDrawEyes which calls cull() - the native function Java_org_gearvrf_GVRViewManager_cull. The monoscopic version of cull() calls Java_org_gearvrf_OvrNativeMonoscopicRenderer_cull. Both of these native functions do exactly the same thing.

liaxim:
Then the OvrMonoscopicRenderer.cull and Java_org_gearvrf_OvrNativeMonoscopicRenderer_cull should be removed too.

PR #1288, opened 2017-07-13
Create GVRGearControllerSceneObject, GVRLineSceneObject (in progress)

J0Nreynolds:
This pull request creates a scene object to better represent interaction with Gear VR Controllers. The main features of this scene object are:

- Default controller scene object
  - Users may also specify a GVRSceneObject to take the place of the default one.
- Enable/disable ray attached to controller object
   - A GVRLineSceneObject was created to represent the ray (could be nested if desired?).
   - Also able to modify the distance of the ray, which is **extremely important for interaction** since the cursor is not straight ahead of the camera like it is in the GazeCursorController. 
- Ability to attach an object to represent the cursor
- Ability to have "projective objects" which the ray/cursor will project itself on
   - Optionally can project onto the surface of objects with MeshColliders. Otherwise, only the hit distance will be used to adjust the ray-depth 

controller4-cropped

Marked in progress for a few reasons:
- Dependent on #1256 
- Dependent on ability to pick surface norms #1290
- Dependent on GearCursorController origin correction in #1291


NolaDonato:
We have a scene editor in GearVRfGVRf/tools/gvr-scene-editor. You can test the usability of your new objects by adding them to this demo.

GVRf/Framework/framework/src/main/java/org/gearvrf/scene_objects/GVRGearControllerSceneObject.java
```diff
@@ -0,0 +1,297 @@
...
+import org.gearvrf.ISensorEvents;
+import org.gearvrf.R;
+import org.gearvrf.SensorEvent;
+import org.gearvrf.io.GVRControllerType;
+
+import org.joml.Vector3f;
+import java.util.Arrays;
+
+/**
+ * Created by j.reynolds on 6/30/2017.
```
liaxim: Some documentation would be great.

liaxim:
Where did you get the assets from? Are we ok using them?

J0Nreynolds:
The controller and it's corresponding texture came from the Assets folder of the Unity Project provided here: https://resources.samsungdevelopers.com/SDP_Webinars/Take_Control%3A_Developing_for_the_Gear_VR_Controller

liaxim:
Is this done? Would you consider updating the gvr-controller demo to use it?

J0Nreynolds:
Yes, I believe this is done for now. I may update it in the future if changes are ever made to SensorEvents/Picking. As of now there is no simple way of ensuring only the nearest picked-object is the only one to receive an ISensorEvents or IPickEvents callback, which would be useful.

Also I updated gvr-controller demo in [this PR](https://github.com/gearvrf/GearVRf-Demos/pull/493).

liaxim:
Hi, is the ray supposed to be visible?

J0Nreynolds:
Yes it should be. It was visible for me when testing yesterday. Do you have
any idea what might cause it not to appear?

liaxim:
Missing bindShader but not the usual case. I am surprised anything is rendered because the controller scene object is not explicitly added to the scene graph. Not sure is this a bug or a feature.

Diff: 
[diff.txt](https://github.com/Samsung/GearVRf/files/1208961/diff.txt)

Let's make GVRLineSceneObject a bit more self-contained and have it handle the bindShader call.

liaxim:
Needs explicit bindShader call

J0Nreynolds:
@liaxim Tested your changes and included them. Also rebased to update calls to GVRPickedObject.getHitX/Y/Z and GVRPickedObject.getNormalX/Y/Z.

liaxim:

PR #1292, opened 2017-07-14
Screenshot feature for multiview extension

roshanch:
This adds screenshot feature for multiview extension. 
1. It modifies rendercamera rendering to support multiview/non-multiview rendering.
2. create separate FBO for screenshot capture if multiview is enabled.

GearVRf-DCO-1.0-Signed-off-by: Roshan Chaudhari, roshan.c@samsung.com


liaxim:
Needs rebase

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRRenderBundle.java
```diff
@@ -38,13 +40,16 @@
                 sampleCount = maxSampleCount;
             }
         }
-
         if (sampleCount <= 1) {
-            mPostEffectRenderTextureA = new GVRRenderTexture(mGVRContext, width, height);
-            mPostEffectRenderTextureB = new GVRRenderTexture(mGVRContext, width, height);
+            if(appSettings.isMultiviewSet())
+                mEyeCaptureRenderTexture  = new GVRRenderTexture(mGVRContext, width, height, 1);
```
liaxim: Suggest to create the capture render texture on demand. The capture is rarely used so we shouldn't be using up the memory all the time.

GVRf/Framework/framework/src/main/jni/engine/renderer/renderer.cpp
```diff
@@ -30,7 +30,7 @@
 #include "vulkan_renderer.h"
 #define MAX_INDICES 500
 #define BATCH_SIZE 60
-bool do_batching = true;
+bool do_batching = false;
```
liaxim: Is batching intentionally disabled?

GVRf/Framework/framework/src/main/jni/objects/textures/render_texture.cpp
```diff
@@ -23,215 +23,438 @@
 #include "eglextension/msaa/msaa.h"
 
 namespace gvr {
-RenderTexture::RenderTexture(int width, int height, GLTexture* tex) :
+GLFrameBuffer *RenderTexture::renderTexture_gl_read_buffer = nullptr;
```
liaxim: Please give removing this static member a try.

liaxim:
Suggest adding a useMultiview flag to the base ViewManager that is initialized by the ctor. Then some of the parameters passing can be replaced by using it.

GVRf/Framework/framework/src/main/jni/objects/textures/render_texture.cpp
roshanch: we need to store readFBO somewhere, we don't have render-texture object in "readRenderResultNative()" as textures are created by oculus and it doesn't make sense to create our own render-texture as we just need texture.

liaxim: texId is being passed to readRenderResultNative. And texId is obtained from the multiview-specific RenderTexture eyeCaptureRenderTexture. So if you pass eyeCaptureRenderTexture to readRenderResultNative instead you will have the instance you need. Right?

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRRenderTexture.java
```diff
@@ -48,9 +48,9 @@ public GVRRenderTexture(GVRContext gvrContext, int width, int height) {
      *            MSAA sample count.
      */
     public GVRRenderTexture(GVRContext gvrContext, int width, int height,
-            int sampleCount) {
+            int sampleCount, int number_views) {
```
liaxim: Please update the javadoc. Also please add a new ctor that uses 1 view by default. Like:
```
    public GVRRenderTexture(GVRContext gvrContext, int width, int height) {
        this(gvrContext, width, height, 1);
    }
```

Seems more sensible than making code changes to the demos.

GVRf/Framework/framework/src/main/jni/objects/textures/render_texture.cpp
roshanch: we are using eyeCaptureRenderTexture only in case of multiview, with oculus backend, in non-multiview case, we are using the texture coming from oculus swap chain in which we don't have render-texture instance.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRRenderTexture.java
roshanch: will update it.

GVRf/Framework/framework/src/main/jni/objects/textures/render_texture.cpp
liaxim: I clearly misunderstand something. You will explain tomorrow in person. Last question, isn't the renderTexture_gl_read_buffer used only for the multiview case?

roshanch: yes, I will move the generation code.

NolaDonato:
We have a thing called a render target now. It has a camera and a texture and is exposed in Java. What I think we should do is have the concept of a multi-view render target and use that abstraction when communicating with the back end. Eventually the renderer will have to take over scheduling the rendering of all displays so we can use multiple threads in Vulkan for it. Encapsulating these rendering tasks in an object makes it easier to do that work. It might also make it cleaner internally.

roshanch:
@NolaDonato for changing parameter list, it requires major changes in renderer, which we can do it 4.0. I talked with Tom and he said we can wait till 4.0 for this.

liaxim:
@roshanch My understanding is that you should look into creating a similar pull request for Vulkan also.

roshanch:
@liaxim yes, I will start working on it this week. Its not same as that of master, vulkan branch would need whole architectural change.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRViewManager.java
```diff
@@ -694,14 +695,22 @@ protected void captureCenterEye() {
         if (mScreenshotCenterCallback == null) {
             return;
         }
+        GVRRenderTexture eyeCaptureRenderTexture = mRenderBundle.getEyeCaptureRenderTexture();
```
liaxim: Getting (and creating) this render texture needs to be done only if multiview is in use. Please move to the if's for begin and end.

GVRf/Framework/framework/src/main/jni/objects/textures/render_texture.cpp
liaxim: Java_org_gearvrf_GVRViewManager_readRenderResultNative is a special case and modifying RenderTexture to support something in Java_org_gearvrf_GVRViewManager_readRenderResultNative in not desirable.

Attaching a proposal diff: [diff.txt](https://github.com/Samsung/GearVRf/files/1234802/diff.txt).

renderTexture_gl_read_buffer becomes instance member and should be properly deleted also. Java_org_gearvrf_GVRViewManager_readRenderResultNative deals with the problem independently and also takes care of the case where the process lingers but new activities (and egl contexts) get created. Based on the glIsFramebuffer docs it looks like glBindFramebuffer needs to be inserted in GLFrameBuffer also.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRViewManager.java
roshanch: check for multiview is in getEyeCaptureRenderTexture() which returns null if its not multiview.

liaxim:
Looks good to me. There is a dependency that needs to be ready before merging this though.

liaxim:
Planning to merge on Friday.

PR #1147, opened 2017-04-13
Fix problem with some of the demos that have custom shaders that are using GLSL ES version 100.

liaxim:
For example gvr-inputsample has a problem without this fix.

---
GearVRf-DCO-1.0-Signed-off-by: Mihail Marinov <m.marinov@samsung.com>

PR #1295, opened 2017-07-15
Vulkan port of GLSLESVersion

dnvasilev:
Vulkan port of GLSLESVersion
    
    Port of:
    Fix problem with some of the demos that have custom shaders that are using GLSL ES version 100. #1147
    
    - add useMultiView to _gvr.xml
    - add GLSLESVersion enum
          creatively redone as compared to Master. Uses strings (for example "300 es") instead of integers.
          Has an explicit Vulkan enum with value "400" . This is a hack because:
            for OpenGL ES, we will detect this version and rewrite the shader, removing layout directives. Replace version with "300 es"
            for Vulkan, shaderc compiler is perfectly OK with "#version 400" and we don't need to touch the shader source
    - always use #version directive, even for version "100" where this directive is optional

This PR shall be 'merged'

### DIFF
--- a/GVRf/Framework/framework/src/main/assets/_gvr.xml
+++ b/GVRf/Framework/framework/src/main/assets/_gvr.xml
@@ -28,6 +28,7 @@
         framebufferPixelsHigh="DEFAULT"
         framebufferPixelsWide="DEFAULT"
         showLoadingIcon="true"
+        useMultiview="false"
         useProtectedFramebuffer="false"
         useSrgbFramebuffer="false" >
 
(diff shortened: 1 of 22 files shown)


NolaDonato:
Looks good to me. Is it required to use this new method? If so, most of the demos that do custom shaders don't do it this way yet. We may need to change them.

dnvasilev:
@NolaDonato We have 6 demos using GVRShader or GVRShaderTemplate. All are broken before the change and require more extensive fix than just the shader version:
4 fail in Gradle:
old/gvr-modelviewer fails because Gradle files do not recognize old subdirectory
gvr-remote-scripting needs debugwebserver
gvr-widgetviewer needs widgetplugin
gvr-simplephysics needs gvrf-physics and google-vr
2 fail because of GVRMaterialMap change:
gvr-outline
gvr-video

PR #1297, opened 2017-07-17
Fixed skinning, optimized vertex weights & indices

NolaDonato:
Fixed the problem with skinning in the vulkan branch.

GVRVertexBoneData used to contain the bone weights and indices. Now computation of bone indices and weights is done directly by the asset loader.
Removed GVRBoneWeight API, GVRVertexBoneData now only contains bone
matrices. This makes bone weights and indices behave exactly like all the other vertex components.

Removed hardcoding of 60 as maximum bones.

Added GVRVertexBuffer.dump to print a component of a vertex buffer to the log for debugging.

Use uniform arrays to limit bone matrices.

Removed VertexBuffer::parseDescriptor


dnvasilev:
In the original merge, adding 'es' to the shader #version broke Vulkan. I fixed this problem by merging my GLSLESVersion changes first and then resolving.

NolaDonato:
I put the fix in because some of the unit tests broke without it. I am OK with removing the fix but then we should do the proper thing in the demos and tests. I did not know what "the proper thing" was. Sorry I broke something you fixed :-(

PR #1299, opened 2017-07-18
backend_oculus: elevate the render and main thread's priority

liaxim:
This dates back to Oculus 0.6.0. Ever since then we have not been
passing the thread ids. The TimeWarp thread itself has always been
elevated.

---


liaxim:
Planning to merge on Tuesday.

PR #1302, opened 2017-07-18
Make physics work on Vulkan branch

NolaDonato:
1. fix build
2. fix vertex accessors
3. fix mesh access for mesh collider


NolaDonato:
made a mess of it

PR #1304, opened 2017-07-18
change the distance metric on which sceneobjects are sorted

sushantojal:
This PR changes the distance metric on which the scene objects are sorted.
The current distance metric used to sort two objects is the 
euclidean distance between their centers. This might be wrong for cases such as 
this:

```
--------------------**---------------------------  1
				--------**-------  2





		C (camera center)
```

The distance from the camera to object 2 is greater than object 1. But object 1
should be rendered before object 2 in case of back to front rendering.

A method to solve this problem would be to sort the objects based on 
the projection of their distances on the lookat vector of the camera as shown:

```
--------------------**---------------------------  1
		     |~~~~~~~~~	--------**-------  2
		     |			/
		     |		    /
		     |		/
		     |	  /
		     |/	
		     C
```
			
The projection of distance on the lookat vector will be smaller for 
object 2 as it should be. The above case fails too if the lookat vector is at a sufficiently
large angle to both the object centers. The advantage of this distance metric is
that it suffices for a reasonable fov before coming up with a wrong distance ordering.
The disadvantage is that the ordering changes at a specific angle to the lookat 
vector of the camera, and therefore there might be a change in the effects which
depend on sorting of the objects. ( In the case of sorting based 
on distance between centers, these effects might be always wrong, but they wouldnt
change with respect to fov).


liaxim:
For this pull request we definitely will need tests. That prove the old code doesn't work and confirm this does. Edge cases where even this doesn't would be nice. Please evaluate the overhead here compared to the old method. This will be only merged after Tom has approved it.

dnvasilev:
I think being consistent is important to the users of GVRF: it's easier to work around bugs that are always there. The new method does not always fix problem of transparency ordering: you need BSP or similar to order your objects correctly.

sushantojal:
I took a look at OSG's codebase. They are using the distance between centers as a metric, and they have their fair share of issues on their forums regarding this. Workarounds suggested on forums include using separate render bins for different transparent objects , create bounding boxes of sceneobjects of comparable sizes, alpha to coverage etc.  That being said, I did find an implementation of depth peeling in their code, but order independant algorithms are expensive.
I came across similar issues on the jmonkeyengine forums too, but didnt go through their codebase.
Should probably close this PR for now, because of the reasons mentioned by @dnvasilev.

PR #1307, opened 2017-07-19
gvrf-physics: relax the mesh presence requirement in gvr-rigidbody

liaxim:

### DIFF
--- a/GVRf/Extensions/gvrf-physics/src/main/java/org/gearvrf/physics/GVRRigidBody.java
+++ b/GVRf/Extensions/gvrf-physics/src/main/java/org/gearvrf/physics/GVRRigidBody.java
@@ -17,6 +17,7 @@ package org.gearvrf.physics;
 
 import org.gearvrf.GVRComponent;
 import org.gearvrf.GVRContext;
+import org.gearvrf.GVRRenderData;
 import org.gearvrf.GVRSceneObject;
 
 /**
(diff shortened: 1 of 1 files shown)


liaxim:
@JulianaFigueira Please advice. gvr-bullet loads a model; the real mesh is under two transform-only nodes without render data (and mesh). The new restriction means that UnsupportedOperationException is thrown on attach.

NolaDonato:
If we add the code below to bullet_gvr_utils.cpp we don't need to check for a mesh at all on the Java side. The real issue is how often do we check? It is possible to attach a mesh collider to a scene object and then attach the mesh. If physics is run in between these two actions, there will be a mesh collider without a mesh. Next time physics is run, the mesh will probably be there. I suggest that, if the physics code encounters a rigid body that has a collider but does not have a collision shape, it could create the collision shape at that time if a mesh is present. That way, it would work in all cases.

btCollisionShape *convertMeshCollider2CollisionShape(MeshCollider *collider) {
    btCollisionShape *shape = NULL;

    if (collider != NULL) {
        Mesh* mesh = collider->mesh();
        if (mesh == NULL) {
            SceneObject* owner = collider->owner_object();
            if (owner == NULL) {
                return NULL;
            }
            RenderData* rdata = owner->render_data();
            if (rdata == NULL) {
                return NULL;
            }
            mesh = rdata->mesh();
            if (mesh == NULL) {
                return NULL;
            }
        }
        shape = createConvexHullShapeFromMesh(mesh);
    }

JulianaFigueira:
Cool, the new convertMeshCollider2CollisionShape would work to avoid checking for a mesh on the Java side, but if it returns null we can not add the rigidBody to the world. Then, instead of the throwing the error should we wait for the RenderData.setMesh and then try to add the rigidBody again when it loads?

NolaDonato:
Yes, you need to account for the possibility that physics will run before the collider gets a mesh depending on how the programmer wrote the code. This is because Jon is doing work to make GearVRF thread-safe and run physics in its own thread.

liaxim:
The node will never have a mesh since it is a transform-only node. You can work on a better fix but in the meantime the demo is broken. Hence I am planning to merge this on Monday.

liaxim:
Can't be cherry-picked onto the vulkan branch. Other changes to gvrf-physics are missing.

PR #1310, opened 2017-07-20
Run GVRWorld on own thread and use thread-safe transforms

J0Nreynolds:
This change allows the gvrf-physics extension to calculate collisions in a thread of its own. Furthermore, the transform class has been refactored to allow for thread-safe reads and writes.

### DIFF
--- a/GVRf/Extensions/gvrf-physics/src/main/java/org/gearvrf/physics/GVRPhysicsContext.java
+++ b/GVRf/Extensions/gvrf-physics/src/main/java/org/gearvrf/physics/GVRPhysicsContext.java
@@ -1,0 +1,65 @@
+package org.gearvrf.physics;
+
+/* Copyright 2015 Samsung Electronics Co., LTD
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import android.os.Handler;
+import android.os.HandlerThread;
+
+/**
+ * This class represents the Physics context
+ * with its own main loop.
+ */
+public class GVRPhysicsContext {
+    private static GVRPhysicsContext mInstance;
+    private final HandlerThread mHandlerThread;
+    private final Handler mHandler;
+
+    public static GVRPhysicsContext getInstance() {
+        if (mInstance == null) {
+            mInstance = new GVRPhysicsContext();
+        }
+
+        return mInstance;
+    }
+
+    private GVRPhysicsContext() {
+        mHandlerThread = new HandlerThread("gvrf-physics");
+        mHandlerThread.start();
+        mHandler = new Handler(mHandlerThread.getLooper());
+    }
+
+    public boolean runOnPhysicsThread(Runnable r) {
+        return mHandler.post(r);
+    }
+
+    public boolean runDelayedOnPhysicsThread(Runnable r, long delayMillis) {
+        return mHandler.postDelayed(r, delayMillis);
+    }
+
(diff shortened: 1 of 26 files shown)


NolaDonato:
Juliana Figueroa from the physics team should review this PR. I will ask her to. I don't know how to add her to the reviewers list.

liaxim:
It would be nice to have few test cases that exercise this change well.

JulianaFigueira:
Hi Jon. Were you able to test it with all the physics unit tests? Would be nice to compare the results. I tried it really quickly here, but it failed disabling the physics world, maybe we missed something.

J0Nreynolds:
Hi Juliana. Thanks for catching that. I've committed a one-line fix that should take into account whether or not the GVRWorld is enabled. It passes all of the unit tests now. Please let me know if there's anything else that needs my attention.

ragner:
Hi @J0Nreynolds,
your patch look good to me. 
I just think that GVRWord doesn't need extend GVRBehavior
because it doesn't implement onDrawFrame anymore.
So I think you could change GVRWord to extend GVRComponent and just check
if GVRWorld already was initialized before call the start of the thread.

But this comment is just to improve the class hierarchy. Would not improve the performance :)

Thank you,

JulianaFigueira:
I was able to test with that fix, superficially its performs 2% faster than the older version. I think the patch is good :)

liaxim:
We need a variant of this of pull request for vulkan.

liaxim:
Run gvr-bullet and tap on the screen to pause the simulation. Usually leads to a NPE:

```
08-11 01:11:35.220 19144 19192 E AndroidRuntime: FATAL EXCEPTION: Timer-0
08-11 01:11:35.220 19144 19192 E AndroidRuntime: Process: org.gearvrf.gvrbullet, PID: 19144
08-11 01:11:35.220 19144 19192 E AndroidRuntime: java.lang.NullPointerException: Attempt to invoke virtual method 'org.gearvrf.GVRSceneObject org.gearvrf.GVRComponent.getOwnerObject()' on a null object reference
08-11 01:11:35.220 19144 19192 E AndroidRuntime: 	at org.gearvrf.physics.GVRWorld.sendCollisionEvent(GVRWorld.java:191)
08-11 01:11:35.220 19144 19192 E AndroidRuntime: 	at org.gearvrf.physics.GVRWorld.generateCollisionEvents(GVRWorld.java:184)
08-11 01:11:35.220 19144 19192 E AndroidRuntime: 	at org.gearvrf.physics.GVRWorld.access$100(GVRWorld.java:36)
08-11 01:11:35.220 19144 19192 E AndroidRuntime: 	at org.gearvrf.physics.GVRWorld$GVRWorldTask.run(GVRWorld.java:302)
08-11 01:11:35.220 19144 19192 E AndroidRuntime: 	at java.util.TimerThread.mainLoop(Timer.java:555)
08-11 01:11:35.220 19144 19192 E AndroidRuntime: 	at java.util.TimerThread.run(Timer.java:505)
```

For gvr-bullet to run you will need https://github.com/Samsung/GearVRf/pull/1307.

NolaDonato:
I have run gvr-bullet on the Note 8 and paused it repeatedly and it does not crash. Which phone did it crash on?

ragner:
Hi @NolaDonato,
I will take a look at this issue. I hope have a patch soon.

ragner:
Ok, 
I am fixing this issue but I need reimplement the current logic. 

All physics call should be running at the same physics thread/looper.

The touch event(UI Thread) is removing the rigidbody at same time that the collision notification
occur. The collision occurred at native side but the rigibody it not in `mRigidBodies` list anymore.

ragner:
Hi @J0Nreynolds,
could you update this pull request with my changes at https://github.com/ragner/GearVRf/tree/gvrf-physics ? Our should I make a new pull request?

@NolaDonato, I implemented a own main loop to the physics.
So our physics api that needs be thread-safe should be changed to run
on Physics loop. 

This issue:
 * We are removing the ball during its collision with the ground that causes "collision exit" event. When the physics tries notify this event the null pointer occurs.

My fix to this issue of bullet demo:
 * The physics world does not notify "collision exit" to SceneObject removed of the scene.

J0Nreynolds:
Hi @ragner, I will be sure to update the PR with your requested changes today.

J0Nreynolds:
Added

NolaDonato:
I tested this again on S7 and Note 8 and it no longer seems to crash when you stop and start the action.

PR #1317, opened 2017-07-22
fix screen shot shader, texturing bug

NolaDonato:
1. screen shots were using horizontal flip shader instead of vertical flip shader
2. iterate over textures in proper order
3. make phong layered shader match phong shader

Fixes two failures in multi-texturing unit tests


NolaDonato:
wrong branch

PR #1320, opened 2017-07-24
take the vertex-attribute values from mesh

roshanch:


GVRf/Framework/framework/src/main/jni/objects/shader_data.cpp
```diff
@@ -240,7 +240,7 @@ namespace gvr {
                 }
             }
         }
-        return (uniforms().updateGPU(renderer) ? 1 : 0);
+        return 1;
```
dnvasilev: Looks like a debugging change to me. Revert or justify,

roshanch: the call is made for updating UBO for material which we are not using right now.

PR #1331, opened 2017-07-26
Vulkan: Added START_ANIMATIONS import setting

dnvasilev:


dnvasilev:
Vulkan port of  #1193

Issue #1214, opened 2017-05-25
Picking not aligned correctly when camera's z not at origin


mwitchwilliams:
X3D allows cameras to be in locations other than the origin.
Prior to his leaving, Rahul suggested I set GVRCursorController's setOrigin() to the location of the camera.
However, the picking is not aligning correctly.

I have a code demo at https://github.com/mwitchwilliams/GearVRf/tree/pickCursorCtllrIssue.
I zipped an X3D file where the camera is at (-3, 0, -3).  The cylinder should rotate when over the orange sphere.  However, it rotates when over the area between the sphere and the red plane.

To run the demo, besides the code above, change X3DparserScript.java to set the file name to 'pickTest_CameraPos.x3d'.

[pickTest_CameraPos.zip](https://github.com/Samsung/GearVRf/files/1030304/pickTest_CameraPos.zip)

![screenshot_20170525-141941](https://cloud.githubusercontent.com/assets/18128830/26471994/0bb21680-4159-11e7-945d-b58141857ca4.png)

mwitchwilliams:
Without the bringing in the code, it can be simulated by changing GVRCursorController.setOrigin(-3, 0, -3)
and then set:
mainCameraRig.getTransform().setPosition(-3.0f, 0.0f, -3.0f);

liaxim:
@mwitchwilliams https://github.com/Samsung/GearVRf/pull/1332 has been merged. Could you please confirm this issue has been fixed. Thanks.

mwitchwilliams:
test3d, looks good.

PR #1332, opened 2017-07-26
Allow GVRGazeCursorController to work from any camera position

J0Nreynolds:
A long requested bug-fix from @mwitchwilliams, this PR addresses the problem outlined in #1214. It turns out that Rahul was correct in suggesting that setOrigin must be used on the GVRGazeCursorController when the camera is moved, but unfortunately, the gaze-direction was being calculated incorrectly, which is why problems still persisted for Mitch.

This fix corrects the calculation of the gaze-direction and overrides the getOrigin function to always return the main camera rig position. I believe this is a reasonable change since the gaze direction is always calculated with the main camera's head transform, which implies the GVRGazeController is always intended to be used from the point of view of the main camera.

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRCursorController.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRCursorController.java
@@ -646,7 +646,7 @@ public abstract class GVRCursorController {
         origin.set(x,y,z);
     }
 
-    Vector3f getOrigin(){
+    protected Vector3f getOrigin(){
         return origin;
     }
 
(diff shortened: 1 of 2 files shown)


J0Nreynolds:

liaxim:
@mwitchwilliams I assume you verified this fix works for you?

mwitchwilliams:
I haven't tested it, but I provided Jonathan the test case with the problem.  If it fixed that, then only thing left would be to try different camera positions, which can be set by changing _x, y, z_ inside the x3d file &lt;Viewpoint position-'_x, y, z_'/&gt; node.

PR #1337, opened 2017-07-27
Fix null pointer crash in GVRRenderData.setMaterial

dnvasilev:
This PR makes gvr-controls work on OpenGL

_Recreated on the correct branch from #1333_

GearVRF DCO signed of by: Nola Donato nola.donato@samsung.com


dnvasilev:
Looks good to me (this is Nola's change)

Issue #1338, opened 2017-07-28
Rendering order set TRANSPARENT when creating scene object with future texture


SteveGreatApe:
Hi, I've just pulled the latest GearVRf code and found a bug in my App as a result of the rendering order being automatically changed to TRANSPARENT where I'm loading a GVRSceneObject with a future texture and mesh.

The problem is caused by the newly added GVRRenderData:adjustRenderingOrderForTransparency function. I can see the sense it doing this when it does detect a texture with transparency, but in my case both the main and diffuse textures are null. In this case the code behaves as if the textures have transparency and automatically switches the rendering order to TRANSPARENT, which is not what I was expecting.

bohuang3d:
As a side note: maybe you know this already, it's been recommended to not use future texture

thomasflynn:
he doesn't have a lot of easy choices in the current top of trunk.  once we merge the vulkan branch into master, then futureTextures will be unnecessary and be deprecated.

thomasflynn:
and thank you @sushantojal for the quick diagnosis and fix!

liaxim:
https://github.com/Samsung/GearVRf/pull/1341 has been merged. @SteveGreatApe Can we close?

SteveGreatApe:
Yes, the fix looks good thanks.

PR #1341, opened 2017-07-28
set rendering order to transparent only if texture is not null

sushantojal:
In response to issue #1338. Automatically set rendering order to transparent only if a texture is present.


liaxim:

liaxim:
Impossible to cherry-pick to vulkan since Tom's change is missing from there.

PR #1352, opened 2017-08-01
framework: propagate all exceptions openStream might throw

liaxim:
Since users can extend from GVRAndroidResource and override openStream their implementations could be throwing all sorts of exceptions, like in a particular case a SecurityException. Propagate them all to the callback.


liaxim:
Planning to merge Friday.

PR #1357, opened 2017-08-02
Set MotionEvents in GearCursorController

J0Nreynolds:
This PR sets MotionEvents for the GearCursorController. It also enables one to query if the touchpad of the GearVR controller is being touched (not pressed down).


J0Nreynolds:

NolaDonato:
Upon inspection this pull request seems like it will add touchpad support. I cannot find a test program for it. I think we should go ahead and merge. It does not break the gvr-controller demo.

liaxim:
I can see the motion events being sent to the ControllerEventListener.

liaxim:
gvr-controller, prior to this pull request but must have happened in the last month or so:
- the controller model can just freeze somewhere
- the object selected by the controller can flash at the origin for a frame or two

PR #1370, opened 2017-08-08
Vulkan transparency

dnvasilev:
Creatively reworked "Try to automatically detect transparency #1212"

- Change state_sort() compare function: check transparency depth right after rendering order
- Implement all transparency bookkeeping in the native code
- Adjust render order during every cull pass, and when Java asks Native for render order (that may happen before the cull pass). 

We can make it more efficient in two ways: 

- optimize adjust RenderOrderForTransparency() by reducing by-name lookups: either get the first texture and assume it's either main texture or diffuse texture; or, more robustly, look for main texture, and if it is null, then look for the diffuse texture.
- check dirty state before adjusting: this may need a new dirty flag.

@NolaDonato, @thomasflynn, @roshanch Please let me know if we want to implement efficiency improvements right now.
 
_This PR shall be 'merged'_
- - -

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRBitmapTexture.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRBitmapTexture.java
@@ -116,7 +116,7 @@ public class GVRBitmapTexture extends GVRImage
      */
     public void setBitmap(Bitmap bmap)
     {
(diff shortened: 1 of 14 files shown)


NolaDonato:
The transparency adjustment code in addRenderData is ugly but we can refactor it as part of a later effort. I think we will ultimately want to call ShaderData::updateGPU as part of culling and do the adjustment in that function (which already examines all of the textures for readiness).

PR #1395, opened 2017-08-16
Vulkan ubo padding

apjagdale:
OpenGL change:
- Added support for passing array of vecs or ints to Material

Vulkan change:
- Added support for passing array of vecs or ints to Material
- Done respective padding for the same since Vulkan uses UBO unlike GL


NolaDonato:
It would be easier to modify the offsets in VulkanUniformBlock constructor the descriptor has been parsed. DataDescriptor::parseDescriptor is called in the UniformBlock constructor and it set up mLayout. You can change the offsets immediately afterward and avoid having to add padding later.

NolaDonato:
Please retest

NolaDonato:
OK I guess my code doesn't work. Looks OK now, we can merge.

PR #1406, opened 2017-08-22
Make GearCursorController revolve around main camera position, not origin

J0Nreynolds:
Translates the pivot and ray-origin positions by the main camera rig position.


J0Nreynolds:

NolaDonato:
This fix looks good to me. I tested it with the gvr-controller demo and it does no harm.

liaxim:
The controller stops being tracked quite often for me on S8. Anybody else seeing problems?

What I do: gvr-controller, keep selecting, moving and rotating objects. Within 20 seconds it freezes pretty much always.

liaxim:
Again, the problem was already there prior to this pull request.

NolaDonato:
We should test again on S8 with the physics check-in merged. It locks around transform updates. This will help the controller code when it changes the matrix from a separate thread.

PR #1407, opened 2017-08-22
Allow for variable size and offset subimage texture uploads

J0Nreynolds:
Allows users to update any rectangular subarea of a texture.


dnvasilev:
Can be tested with gvr-painting from J0Nreynolds/GearVRf-Demos

PR #1408, opened 2017-08-23
Multiview support

roshanch:
this adds multiview support for vulkan branch. It creates render-target for each eye rendering. Each render-target is having separate render-list. Existing renderCamera() not removed yet, as other backends still using those functions. Tested with few of the samples which includes shadows, render-texture, tutorial-6, screenshot feature.


NolaDonato:
Really good job at RenderTarget reform. I tried to deal with the post effect situation. In the vk_multiview2 branch in my repo I merged this PR with my change to represent post effects as a single RenderData. I moved the checks in addRenderData to RenderData::isValid. If this function returns >=0 all of the shaders in all the render data passes have been successfully generated. Take a look at it.

I would like to get rid of the old renderCamera and cullFromCamera functions in the renderer. I remember you said they might be used somewhere so I left them alone.

GVRf/Framework/framework/src/main/java/org/gearvrf/IRenderBundle.java
```diff
@@ -23,4 +23,9 @@
     GVRRenderTexture getPostEffectRenderTextureA();
 
     GVRRenderTexture getPostEffectRenderTextureB();
+    GVRRenderTarget getEyeCaptureRenderTarget();
+    GVRRenderTarget getDaydreamRenderTarget();
```
liaxim: Backend-specifics do not belong to the framework. Please move this to the daydream backend.

GVRf/Framework/settings.gradle
```diff
@@ -1,6 +1,12 @@
 include ':framework'
 include ':backend_oculus'
 include ':backend_daydream'
+include ':gvr-simplesample'
```
liaxim: Please do not commit changes to settings.gradle. extra_settings.gradle exists for this very reason - so you can add whatever other projects you want to the framework workspace outside of the source controlled directories.

roshanch: oops.. i forgot to remove this one.

roshanch:
@NolaDonato Right now only pure backend is using those functions.. we will eventually remote them.

NolaDonato:
vk_multiview2 now has post effect rendering fixed. You can merge with it.
I notice that you put GL-specific code into shadowmap.cpp. This is a platform-independent file, it should not have GL code in it. Maybe we need a GL-specific shadow map, not sure

NolaDonato:
I think we should go ahead and merge this PR asis. I will make my updates in a future PR. I think that will be cleaner.

dnvasilev:
We will have to fix tests after merging.
GearVRF-Tests RenderTests compile fails after this change.

roshanch:
I have tested with gvr-shadows, gvr-opacitygallery, gvr-sample, gvr-simplesample.

dnvasilev:
Will merge Tuesday morning. Please test and review now.

PR #1334, opened 2017-07-27
BoxCollider and MeshCollider fixes

J0Nreynolds:
This PR implements an isHit function for BoxColliders, which is necessary in order for it to be used. It uses the same algorithm as the BoundingVolume, and as such, the intersection method for bounding volumes has been made static for both classes to use it.

I also made a small adjustment to mesh_collider.cpp, which removes an unnecessary matrix operation, as no accuracy is gained from calculating the hit distance in world coordinates - a matrix and its inverse produce the same error.

PR #1411, opened 2017-08-24
framework, backend_oculus: fixes for PickerTests

liaxim:
Strictly necessary are the ExceptionClear (it is a Oculus 1.0.7 thing) and the extra check in mesh_collider.cpp.

---

### DIFF
--- a/GVRf/Framework/backend_oculus/src/main/jni/ovr_activity.cpp
+++ b/GVRf/Framework/backend_oculus/src/main/jni/ovr_activity.cpp
@@ -126,6 +126,7 @@ void GVRActivity::onSurfaceChanged(JNIEnv& env) {
         }
 
         oculusPerformanceParms_ = vrapi_DefaultPerformanceParms();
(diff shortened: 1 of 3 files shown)


liaxim:
@J0Nreynolds Please let me know what you think. Especially about the change to mesh_collider.cpp.

J0Nreynolds:
@liaxim The change to mesh_collider.cpp should fix the problem with the picker tests. I actually made the same change in #1334.

liaxim:
Sorry. I missed that change.

NolaDonato:
These changes look like an improvement.  After testing on the S7 European  several times, I notice that some of the picking tests still fail randomly. Usually it is only one test out of the lot and not the same one. In some runs, all of the tests succeed. I think we should go ahead and merge these changes.

liaxim:
@dnvasilev Can't cherry-pick cleanly. Looks like a lot of white space difference in ovr_activity.cpp.

dnvasilev:
@liaxim this solves it:

git merge -s recursive -Xignore-space-change

PR #1415, opened 2017-08-25
Fix: make GVRLineSceneObject render properly

J0Nreynolds:
Fixes GVRLineSceneObject so it can render properly in scene.


roshanch:
its good to go, we have verified and tested it..

PR #1416, opened 2017-08-28
Integrate sensors and picking

NolaDonato:
The original sensor implementation came before picking. I added a GVRPicker to the GVRCursorController class so the cursor automatically picks whenever it gets an event. The buttons are passed to the picker as touchFlags so all pick events now contain information on what buttons are pressed by the controller.

1. Needed to add GVRContext to all GVRCursorController subclass constructors
2. Made GVRBaseSensor derive from GVRComponent, now it is attached to the scene object instead of being a member. GVRSceneObject.setSensor is removed, GVRSceneObject.attachComponent can be used instead
3. Allowed GVRPicker to have event listeners for pick events (added GVRPicker.getEventReceiver, GVRPicker.addListener)
4. Got rid of the CursorEvent.EventGroup, was not being used heavily

This PR is the first step in I/O reform to better integrate controller support into GearVRF.


dnvasilev:
Per discussion with Nola and Tom, remove from Release 3.3. This will be released later on 3.x branch (maybe 3.4)

mwitchwilliams:
Tested with 3 X3D demos that did not work:

- touchSensor3.x3d:  the camera is at (0, 0, 0), and the blue cone at (2, 0, -6) has the Touch Sensor. It looks like the picking / bound box is to the left side of the blue cone.  Testing against master, the picking wascorrect. The blue cone invokes a rotation of a yellow cylinder, green cone (different lengths of time) and a pink box.
- touchensorChildren.x3d: In this scene, the camera is at (0, 0, 0). The are 3 green cones.  The largest has the Touch Sensor.  The middle green code is a child of the largest, and the smallest green code is a child of the middle cone. Rolling over any of the green cones should invoke the animation of the yellow Cylinder - it does not. Tested against master and it worked properly. The animated red cone is just there to random the test case.  It is not related to any Touch Sensors.
- JavaScript_PerFrame_02_modified.x3d: This demo uses JavaScript when the items are clicked or over. The sphere is dragged.  When clicked (anywhere since the cursor is over the sphere dragging it), the sphere rolls away. When rolling over either plane, the diffuse color is changed.  And clicking on either changes the color-animating sphere, or bring the sphere back. It all works on Master, but not in this new picking code. Only dragging the sphere works correctly, though that is done via JavaScript, not any picking code..

The demo's are in the attached zip file.

[x3d demos.zip](https://github.com/Samsung/GearVRf/files/1306876/x3d.demos.zip)

NolaDonato:
replaced by #1455

PR #1426, opened 2017-09-01
controller support for daydream backend

roshanch:
this adds controller support for daydream backend. Tested with gvr-controller app

### DIFF
--- a/GVRf/Framework/backend_daydream/build.gradle
+++ b/GVRf/Framework/backend_daydream/build.gradle
@@ -64,6 +64,7 @@ dependencies {
     compile fileTree(include: ['*.jar'], dir: 'src/main/libs')
     compile 'com.google.vr:sdk-base:1.30.0'
     compile project(':framework')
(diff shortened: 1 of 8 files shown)


liaxim:
Looks good. Will borrow a controller and a device to try it out unless somebody else verified it already.

liaxim:
The controller tends to freeze but I've seen it with other controller types and I am pretty sure it is in the framework and not in this pull request. It is likely a bug we have to fix for 3.3.

liaxim:
Ignoring the freeze problem that we will probably fix in a future release please consider these two:
- apply an additional rotation around the Y axis of about -45 degrees in updateRotation
- add to the position vector in updatePosition this additional position vector - (0.341f, -0.486f, -0.383f).

This makes the model look kind of similar to what the other backends do. Otherwise it is completely off. The position vector I got from the initial position in the Oculus backend. The -45 degrees rotation I came up with.

You can just run the gvr-controller app with Oculus and with Daydream and you'll see what I am talking about.

liaxim:

PR #1433, opened 2017-09-05
Do not change rendering order to 'geometry' for opaque objects.

dnvasilev:
Fixes gearvrf/GearVRf-Tests#126


liaxim:
The test does pass but I don't think the test is correct anymore. The texture has no alpha and rightly so the rendering order gets reverted. It seems like a useful feature to me that if possible we should keep.

I updated the test here: https://github.com/gearvrf/GearVRf-Tests/pull/136. A framework change is necessary since there is a path through the asset loader where the bitmap will not be tested for alpha otherwise. Here: https://github.com/Samsung/GearVRf/pull/1437.

liaxim:
Will merge on Friday.

PR #1435, opened 2017-09-05
Fix billboard resetting scale to 1

ragner:
* Current implementation is always resetting the owner's scale to 1
this patch fix this issue.


liaxim:
Looks good. Will merge on Monday. Added a test case for the scale: https://github.com/gearvrf/GearVRf-Tests/pull/138.

PR #1442, opened 2017-09-08
Vulkan no lighting

apjagdale:
Selecting appropriate shader base on settings.
Selecting texture shader only when NO_ANIMATION is applied.


NolaDonato:
Good catch. Merge this one.

PR #1444, opened 2017-09-11
fix multiview rendering for render-mask

roshanch:
this fixes multiview rendering if one of the eye rendering is disabled. Changed uniform locations from GLuint to GLint. 
For testing change the rendermask of scene-object with multiview enabled.

### DIFF
--- a/GVRf/Framework/framework/src/main/res/raw/vertex_template_multitex.vsh
+++ b/GVRf/Framework/framework/src/main/res/raw/vertex_template_multitex.vsh
@@ -5,6 +5,7 @@ uniform mat4 u_view_[2];
 uniform mat4 u_mvp_[2];
 uniform mat4 u_mv_[2];
 uniform mat4 u_mv_it_[2];
+uniform uint u_render_mask;
 #else
 uniform mat4 u_view;
 uniform mat4 u_mvp;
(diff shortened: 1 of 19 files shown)


GVRf/Framework/framework/src/main/res/raw/vertex_template_multitex.vsh
```diff
@@ -5,6 +5,7 @@ uniform mat4 u_view_[2];
 uniform mat4 u_mvp_[2];
```
liaxim: Why not update vertex_template.vsh also?

liaxim:
Looks good and works as expected. Tested with the texture shader.

---

Branching in the shaders has been brought up again and again as a performance concern. I upped the number of bunnies in gvr-complexscene and tested with and without your change. There is no difference as far as I can tell (on Adreno). Is it something specific about the branching in this pull request that makes it cheaper or branching in general is not that big of a concern anymore?

GVRf/Framework/framework/src/main/res/raw/vertex_template_multitex.vsh
roshanch: we are not using vertex_template anymore. all phong shaders are using multitex ones by default.

roshanch:
@liaxim cannt' say that branching wont hurt the performance at all. GPUs are getting smart about the branching. However, using fps as metric for the branching performance may not be good idea if the shaders are not bottleneck. If we really want to check the performance of each stage, we should be using tools. Right now our bottleneck is number of draw calls and not the shaders.

liaxim:

PR #1445, opened 2017-09-12
framework: tweak so the stats can show when multiview is enabled

liaxim:


liaxim:
Needs work

liaxim:
Works now with and without multiview

liaxim:

PR #1446, opened 2017-09-12
X3D: Changed GVRRenderingOrder to GEOMETRY from TRANSPARENT

mwitchwilliams:


liaxim:
Will test with Erik's app and merge on Wednesday if it is all good.

PR #1447, opened 2017-09-14
gvrf-physics: stop/start the simulation on pause/resume

liaxim:
Doesn't quite immediately stop on pause. Few simulation steps do happen away from user's eyes. Doing something about it would require rearranging some potentially sensitive things in onPause.

---


ragner:
Hi @liaxim,
Looks good to me
but I think we should consider the time spend on pause and sum it to the mLastSimulTime
otherwise mTimeStep may get a wrong value. 

` mTimeStep = mSimulationTime - mLastSimulTime;`

liaxim:
@ragner On resume, mLastSimulTime gets set to the current uptime. Some time later mSimulationTime is set to the current uptime at that point. The delta should be reasonable. How exactly do you see mTimeStep being wrong?

ragner:
Hi @liaxim,
Your patch is correct. 
I did a mistake, sorry.

Looks good to me.

liaxim:
Thanks. Planning to merge on Friday.

PR #1454, opened 2017-09-18
framework, backend_oculus: do not regenerate the buildinfo sources unconditionally

liaxim:


liaxim:

Issue #516, opened 2016-04-26
Weird artifacts on SM-R322 with Galaxy S7


cemkod:
Hello,

I'm using the new Gear VR (SM-R322) with Galaxy S7 Phone. There are some pixels repeat artifacts on the edges of views, and viewport sizes seem to be different than the native sdk as can be seen in the image:
![gearvrf](https://cloud.githubusercontent.com/assets/4003996/14835588/232c0d74-0c13-11e6-960e-c886ee25c9c5.jpg)
Here is a screenshot from a software running with NDK:
![native](https://cloud.githubusercontent.com/assets/4003996/14835589/23314aaa-0c13-11e6-8e14-f2c359a70ce8.jpg)

I am using the standard [gvr.xml](https://github.com/gearvrf/GearVRf-Demos/blob/3e71ed8815ff157a24e6eec4f4af470a4ee55272/gvr-360photo/gvr360photo/src/main/assets/gvr.xml) file supplied in the samples. Is this happening because the lens distortion parameters do not match my device?

liaxim:
Hi,
This was fixed by pull request https://github.com/Samsung/GearVRf/pull/506. We don't have a binary release yet that includes this fix but we will soon have one. If you can't wait you can always build and use the latest framework code. Let us know if you need any help.

cemkod:
Thank you @liaxim 
I'll just wait. New binary will probably be ready until I'm ready to release what I'm developing.

PR #1455, opened 2017-09-19
Controller reform - integrate picking, fix sensors

NolaDonato:
1. Added a picker to each controller which picks from the viewpoint and direction of the controller: GVRCursorController.getPicker() and GVRCursorController.setPicker()
2. Added a callback for when a cursor controller becomes active or inactive: CursorControllerListener.onCursorActive(), CursorControllerListener.onCursorInactive
Call GVRCursorController.isConnected() to explicitly check connection
3. GVRBaseSensor has a pick handler which listens for pick events and converts them to sensor events. The old logic is gone. The new logic scans upwards from the collider until it finds a scene object with a sensor and sends the sensor event there.
4. Modified X3D parser to use the sensor as a component and to add mesh colliders
to all the children of the sensor
5. Added GVRCursorControllerSceneObject which will move the cursor for any controller
6. Added ITouchEvents interface to describe touch events sent by the picker.
7. Changed GVRViewSceneObject to use touch events from the picker
fixed bug in meshcollider
8. Changed GVRKeyboardSceneObject to use touch events from the picker
9. removed redundant IHoverEvents and ITouchEvents interfaces
10. remove GVRPickerInput, use GVRPicker instead
11. Added GVRPicker.propagateEventsToTarget() to send touch events to the owner of the collider
12. Added GVRPicker.runOnUiThread() to process pick events on the UI thread

These changes break key-up events in GVRKeyboardSceneObject (keyboard disappears)
The 3D cursor library samples are also broken


liaxim:
What is the relation to the https://github.com/Samsung/GearVRf/pull/1416 pull request?

mwitchwilliams:
Works with X3D _TouchSensor.isOver_ including children (when the cursor is over a mesh), but does not work with _TouchSensor.isActive_ (when clicking on an object).
Appears to fix the issue of LOD (level-of-detail) with TouchSensors (as long as it is _isOver_ and not _isActive_).

This fix also requires a change to X3Dobject.java, deleting `Sensor currentSensor`

thomasflynn:
ok, though isActive does need to work though.

NolaDonato:
I will look at it. Which test is broken?



mwitchwilliams:
There is touchSensor5.x3d in the X3D Tests repo.  The green cones (with children) uses _isOver_ and works fine.  The blue cone (also with children) uses _isActive_ but does not work.

I also built some simplified versions which has a single blue cone. 

- touchSensor_isOver.x3d runs animatiions of a box, cone and cylinder when over the blue cone.
- touchSensor_isActive.x3d runs the same animations, except it needs to be clicked.

I tested against master and it worked.  But the code in this Pull Request, the _isActive_ X3D file does not work.

[touchSensorTests.zip](https://github.com/Samsung/GearVRf/files/1318996/touchSensorTests.zip)

NolaDonato:
​I tried touchSensor_isActive.x3d. When I click on the blue cone the other objects animate. Is it supposed to do something else?

roshanch:
gvr-controller sample doesn't work. Controller looks freezed also ray coming out from controller is not at proper location. tested with private backend.

NolaDonato:
There is an updated gvr-controller pull request in GearVRf-Demos #516

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRPicker.java
```diff
@@ -56,21 +58,51 @@
...
+ * information. You can register as a listener for pick events from a specific cursor.
  * @see IPickEvents
  * @see GVRSceneObject#attachCollider(GVRCollider)
  * @see GVRCollider
  * @see GVRCollider#setEnable(boolean)
  * @see GVRPickedObject
+ * @see GVRCursorController#addPickEventListener(IEvents)
  */
-public class GVRPicker extends GVRBehavior {
+public class GVRPicker extends GVRBehavior implements IEventReceiver {
```
liaxim: Why should the base picker do so many new things now? Can't the base picker be used by or derived from a class that deals the controller stuff?

liaxim:
gvr-controller with non-Oculus backend; the model and the ray move but there is no picking happening. Using the updated gvr-controller.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRPicker.java
NolaDonato: The reason for IEventReceiver is that lots of objects need to observe pick and touch events. The original picker sent the events to the scene only but you can't have multiple listeners for the same event on the same target. Making the picker an event receiver allows it to have many objects listening to the same events. This change is independent of the controller although it makes working with the controller easier.
The picker now has the ability to generate touch events (which it could not before). I should have done this originally but at that time IO was owned by a different team. I think we should unify pick and touch events in 4.0 but I did not want to change the API in 3.3.
Each GVRCursorController now has its own picker and the logic to interact with that picker is in the controller.  One of our internal apps needs to be able to drive the picker independently from their own motion events which is why GVRPicker.processPick is exposed.

nitosan:
Dose this implementation supports adding EventListener on the SceneObjects?
```java
            GVRMeshCollider collider = new GVRMeshCollider(getGVRContext(), true);
            GVRCubeSceneObject scene_obj = new GVRCubeSceneObject(getGVRContext());
            scene_obj.getRenderData().setShaderTemplate(GVRPhongShader.class);
            scene_obj.attachComponent(collider);
            scene_obj.getEventReceiver().addListener(new IPickEvents() {
                @Override
                public void onPick(GVRPicker picker) {}
                @Override
                public void onNoPick(GVRPicker picker) {}
                @Override
                public void onEnter(GVRSceneObject sceneObj, GVRPicker.GVRPickedObject collision) {}
                @Override
                public void onExit(GVRSceneObject sceneObj) {}
                @Override
                public void onInside(GVRSceneObject sceneObj, GVRPicker.GVRPickedObject collision) {}
            });
```

Currently only the picking events on the `MainScene` will trigger.

mwitchwilliams:
Tested this code against the test repo x3d-> Scripting_Per_Frame
There is still one more change required, but all five tests work.  These include a variety of Touch Sensors, with scripting, attribute changes to lights/transforms/materials, and per-frame procedural animations.
I will be testing more from the test repo x3d -> Scripting and TouchSensor, anchor, boolean_toggle and possibly other directories Tuesday (9/26)

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRPicker.java
liaxim: The controller is in the picker for the sole reason events can be send. Users should not have to create a GVRPicker at all - it should be created for them automatically somewhere in the controller framework.

If the CursorController has its own picker why the Picker has a CursorController too?

Unify touch and pick events as in have them all in once place or provide higher level, more useful, events?

Still not convinced picking and android MotionEvents belong to the same unit. The very existence of processPick is a good sign it belongs to a more specialized entity.

liaxim:
@nitosan I think you need a sensor for the scene object and then you would be getting ISensorEvents.

liaxim:
@nitosan and a collider. For example this worked for me:

```
            GVRCubeSceneObject cube_obj = new GVRCubeSceneObject(gvrContext);
            cube_obj.getRenderData().getMaterial().setVec4("diffuse_color", 1, 0, 0, 1);
            cube_obj.getRenderData().setShaderTemplate(GVRPhongShader.class);
            cube_obj.getTransform().setPosition(0, -1, -2);
            gvrContext.getMainScene().addSceneObject(cube_obj);

            cube_obj.attachCollider(new GVRMeshCollider(gvrContext, true));
            cube_obj.attachComponent(new GVRBaseSensor(gvrContext));
            cube_obj.getEventReceiver().addListener(new ISensorEvents() {
                @Override
                public void onSensorEvent(SensorEvent event) {
                }
            });
```

liaxim:
I honestly think we need strongly-typed addListener methods, at least at the public API level. It will be easier to document and to large extent it will be self-documenting.

We shouldn't force the users to dig through the docs unnecessarily and/or expose ambiguous APIs like some generic addListener to which you can pass anything that derives from IEvents. I would expect any of the event interfaces deriving from IEvents to work. Why wouldn't I?

mwitchwilliams:
Performed through regression test on the Tests repo -> x3d -> 

- anchor
- boolean_toggle
- Scripting
- Scripting_Per_Frame
- Switch_node
- TouchSensors
- plus Level-of-Detail (lod) plus touch sensors 

[lod_TouchSensor.zip](https://github.com/Samsung/GearVRf/files/1338097/lod_TouchSensor.zip)
(roll over _level1_ invokes an animated translation of the green cylinder; roll over _level2_ invokes an animated clockwise rotation of the orange cone).

In all, about 40 tests of touch sensors involving camera location changes, launching web pages, and new X3D scenes, turning lights on/off, moving lights, changing light directions, modifying outer/inner spotlight cones, material colors, procedural animations, grabbing and moving meshes, changing animations, and switching 3d meshes in the scene, I can say this Pull Request...
**Looks Good To Me (LGTM)**

liaxim:
With the Oculus backend the gvr-controller app works fine now for me. With the other backend there is still no picking. After this gets merged I will exclude the 3DCursorLibrary and the related demos from the build.

liaxim:
PickerTests do not seem to be affected by this change.

GVRf/Framework/framework/src/main/java/org/gearvrf/scene_objects/GVRGearControllerSceneObject.java
```diff
@@ -29,18 +31,9 @@
...
-
-public class GVRGearControllerSceneObject extends GVRSceneObject {
-    private final String TAG = GVRGearControllerSceneObject.class.getSimpleName();
-    private float rayDepth = 1.0f;
-    private boolean projectToSurface = false;
-    private GVRBaseSensor disabledSensor;
-    private GVRSceneObject cursor = null;
-    private GVRCursorController gearCursorController = null;
-    private SensorListener sensorListener = new SensorListener();
+public class GVRGearControllerSceneObject extends GVRCursorControllerSceneObject
```
liaxim: The GVRGearControllerSceneObject is so thin now that in my opinion it should be rolled into the GVRCursorControllerSceneObject as something you could optionally enable.

liaxim:
I am ok with merging this.

roshanch:
works with daydream controller.

ragner:
Looks good to me.

PR #1469, opened 2017-09-21
 Vulkan: Allow for variable size and offset subimage texture uploads

dnvasilev:
eimplement #1407 on Vulkan branch

Implemens plumbing for xoffset/yoffset parameters from setBuffer to glTexSubImage2D
Vulkan API buffer loads are still a TODO

_This PR shall be 'merged'._
- - -

### DIFF
--- a/GVRf/Framework/framework/src/main/jni/objects/textures/image.h
+++ b/GVRf/Framework/framework/src/main/jni/objects/textures/image.h
@@ -160,6 +160,8 @@ protected:
     std::mutex mUpdateLock;
     short   mType;
     short   mLevels;
+    int   mXOffset;
+    int   mYOffset;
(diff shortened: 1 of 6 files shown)


GVRf/Framework/framework/src/main/jni/objects/textures/image.h
```diff
@@ -160,6 +160,8 @@ class Image : public HybridObject
     std::mutex mUpdateLock;
     short   mType;
     short   mLevels;
+    int   mXOffset;
```
liaxim: I think explicit initialization to 0 is necessary.

dnvasilev: OK

PR #1480, opened 2017-09-28
Render target rearchitecture

roshanch:
restructure render-targets for vulkan. vulkan is running now.


roshanch:
this branch is working with daydream backend now. Added all the pending commits from master.

GVRf/Framework/framework/src/main/jni/engine/renderer/renderer.h
```diff
@@ -31,6 +31,7 @@
 #include "objects/bounding_volume.h"
 #include "shaders/shader_manager.h"
 #include "batch_manager.h"
+#include "../../../../../../backend_oculus/src/main/jni/ovr_activity.h"
```
liaxim: @dnvasilev This has to be it. The framework must not reference a backend. @roshanch is aware and will take a look.

PR #1492, opened 2017-10-04
framework: hide the unnecessarily public getViewManager method

liaxim:

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRActivity.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRActivity.java
@@ -160,7 +160,7 @@ public class GVRActivity extends Activity implements IEventReceiver, IScriptable
         return mAppSettings;
     }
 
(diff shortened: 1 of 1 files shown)


NolaDonato:
VRTop needs to get the view manager in onCreate in order to listen for controller events (it needs a context). This is why this method was made public. The controller events saying what devices there are happen BEFORE onInit, before setMainScene, right after the context is made. Oculus starts sending events that indicate whether the controller is connected or not on onDrawEyes as soon as there is a scene. When it determines a valid connection, it sends a controller activation event. If the listeners for this event are not hooked up BEFORE the event occurs, VRTop misses the controller activation event and uses the wrong controller.
Instead of making the method public, perhaps setMain could return the context instead of a boolean. That changes the API a lot which is why I made the method public.

liaxim:
I do not agree

NolaDonato:
What should we do about it then?

thomasflynn:
why don't you guys get on the phone and discuss it.

PR #1502, opened 2017-10-05
Merging changes from PR #1310 for physics extensions

roshanch:
tested with gvr-simplephysics and bulletsample


liaxim:
For whatever it is worth I tested the apps too and they work. Except pause and resume handling which was a subsequent change.

PR #1508, opened 2017-10-09
framework: jline: update so it works on S8

liaxim:
catching an exception that is only thrown on s8 that breaks the normal flow for jline


liaxim:
JLine repo: https://github.com/danke-sra/jline-android

PR #1512, opened 2017-10-09
framework: disable some of the verbose log statements

liaxim:


GVRf/Framework/framework/src/main/jni/gl/gl_index_buffer.cpp
```diff
@@ -7,6 +7,10 @@
 #include "gl_index_buffer.h"
 #include "gl_shader.h"
 
+//comment out to enable verbose logging
+#undef LOGV
+#define LOGV
```
dnvasilev: Clever, but:
LOGV(x,y,z) becomes (x,y,z) -- meaning all garbage strings, side effects, etc. are still in the code.
I would rather do varargs macro. Does anyone remember how varargs macros work in C preprocessor?

dnvasilev: #define LOGV(...) /* nothing */

liaxim: The compiler takes care of it. Strings are nowhere to be found.

Now if someone has side-effects in a trace statement, it is theirs problem.

liaxim: But if changing ``#define LOGV`` to ``#define LOGV(...)`` will make you happier, I will make the change.

liaxim:
If there are no other comments, planning to merge on Wednesday

GVRf/Framework/framework/src/main/jni/gl/gl_index_buffer.cpp
dnvasilev: Better idea. Kill LOGV in gvr_log.h

liaxim: Can be quite useful to turn on and off verbose logging for the individual units. As long as those are well organized of course.

liaxim: And also LOGV should be used appropriately for what truly is verbose logging. I am not convinced it is throughout the codebase so I rather deal with individual cases for now. I am even seeing LOGE for cases where LOGI should've been used so..

But if you insist, I will turn it off globally and be done with it.

dnvasilev: This is new functionality and I would organize it a bit differently, without undef

dnvasilev: Makes sense for now to turn it off globally.

dnvasilev: If we allow verbose logging to impact performance, it must be turned off for release

PR #1513, opened 2017-10-09
fix multiview rendering for posteffect

roshanch:
It creates mesh on java side for posteffect renderdata which is latter updated on native side. Modified shaders for posteffect multiview rendering. It adds #define strings in shaders for multiview


liaxim:
Planning to merge on Thursday

PR #1514, opened 2017-10-09
backend_oculus: fix the crash on hitting the back key

liaxim:


liaxim:
Will merge on Wednesday

PR #1520, opened 2017-10-10
Fixing GVRPhongShader for multiview

apjagdale:
Fragment shader requires only the extension to be enabled.


liaxim:
Planning to merge on Thursday

PR #1523, opened 2017-10-11
Fixing particle system shaders for multiview

apjagdale:
Converted shaders to multiview.
Now gvr-particles would work with multi-view


liaxim:
Planning to merge on Thursday

PR #1526, opened 2017-10-11
fix render-to-texture demo for the vulkan, adding copyrights

roshanch:
this adds missing copyright in some of the files. Separating copy vulkan texture code to oculus code.

PR #1530, opened 2017-10-12
framework, backend_oculus: ensure there is a renderer instance always

liaxim:
- fixes the gvr-blurfilter crash after pause/resume

+ fix formatting
+ some cleanup

The fix is adding gRenderer = Renderer::getInstance() to GVRActivity::onSurfaceChanged

---


liaxim:
Backend-specific fix so I will be coming back to this in the future.

A very good example why the renderer should not be a singleton. It is not even a true singleton since it is not the same instance during the process lifetime.

NolaDonato:
I agree it should not b a singleton. We will address this in the next release.

dnvasilev:
Conflicts with #1526

liaxim:
Now includes #1526

liaxim:
No longer includes #1526 since it introduces some problems. Merging this. #1526 needs updating anyways.

PR #1539, opened 2017-10-13
fix texparameters for mipmapping

roshanch:
defaulting min filter to GL_LINEAR_MIPMAP_NEAREST. and disable it internally if not applicable.
tested with 3dcursor, cubemap, shadows, render-to-texture, solarsystem, multilight, performance.

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRTextureParameters.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRTextureParameters.java
@@ -42,7 +42,7 @@ public class GVRTextureParameters {
      */
     public GVRTextureParameters(GVRContext gvrContext) {
         mGVRContext = gvrContext;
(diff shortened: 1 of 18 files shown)


roshanch:
@NolaDonato please review it.

liaxim:
Breaks gvr-modelviewer:


---

```
        "3.7.14 External Textures

        External textures cannot be used with TexImage2D, TexSubImage2D,
        CompressedTexImage2D, CompressedTexSubImage2D, CopyTexImage2D, or
        CopyTexSubImage2D, and an INVALID_ENUM error will be generated if this
        is attempted.  Their default min filter is LINEAR.  It is an
        INVALID_ENUM error to set the min filter value to anything other than
        LINEAR or NEAREST.
```

roshanch:
tested with all the demos. External textures are using two different types of images. I had missed one of it.

PR #1540, opened 2017-10-14
Fixing crash issue of vulkan on pause or resume of apps

apjagdale:
App crashed since Vulkan Instance was freed on pause/resume of app, the RenderTarget which had 
Command Buffer made from the Command Pool associated with Vulkan Instance was no more valid.
Hence freeing vulkan instance only when the app is closed.


NolaDonato:
I have another potential fix in PR #1546. Please test on Vulkan.

PR #1541, opened 2017-10-14
fix for multiple texcoord sets in jassimp

NolaDonato:
GVRMaterial.setTexCoord would silently fail if you did not set a texture
beforehand. Really we should get rid of this API and just use
GVRTexture.setTexCoord directly - this does the real work. I changed it
to throw an exception if it doesn't have a texture so the user at least
knows about it. I also changed jassimp to call GVRTexture.setTexCoord.

This fix is highly localized to jassimp.

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRJassimpAdapter.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRJassimpAdapter.java
@@ -732,13 +732,13 @@ class GVRJassimpAdapter {
             shaderKey += texIndex;
             gvrmtl.setInt(textureKey + "_blendop", blendop);
         }
-        gvrmtl.setTexCoord(textureKey, texCoordKey, shaderKey);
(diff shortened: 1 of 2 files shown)


NolaDonato:
This PR makes the lightmap appear in AssetTextureTests.jassimpLightmap. The overall test still fails because there is a slight difference in hue between the golden master and the 4.0 screenshot. The 4.0 version seems to be brighter. The other jassimp lightmap tests also show some specular highlights not present in the golden masters. 
testspecularlightmapunlit
testspecularlightmapunlit

NolaDonato:
It is even more confusing when you look at the other tests. AssetTextureTests.jassimpSpecularLightmap passes but AssetTextureTests.testSpecularLightmap fails. These tests both use exactly the same model, they just display it from a different point of view. The same is true for AssetTextureTests.jassimpNormaLightmap and AssetTextureTests.testNormalLightmap. The first one passes and the other one fails. We need to rerun the tests which fail against maint_v3.x branch to see if they match.

PR #1546, opened 2017-10-16
Fix crash on pause or resume of Vulkan apps

NolaDonato:
I moved the reset into GVRViewManager.onDestroy. This doesn't mess up
the picking tests. Maybe this instead of PR #1540?


liaxim:
Doesn't compile for me

NolaDonato:
Closed in favor of PR #1551

PR #1548, opened 2017-10-16
framework: disable stats until the posteffect bug is fixed

liaxim:


liaxim:
This probably should be merged at the very last moment in case the post effect bug gets fixed. It is so low risk that merging late is not a problem.

liaxim:
@roshanch fixed it properly in https://github.com/Samsung/GearVRf/pull/1553

PR #1550, opened 2017-10-16
Precision fix for s7

apjagdale:
Fixes issues on Mihail's phone:
Now gvr-accessibility, gvr-immersivepedia, gvr-javascript, gvr-outline, gvr-particles and gvr-shadows work now.


thomasflynn:
quick question.  why is this a problem on s7 now with 4.0 when it wasn't a problem before?

apjagdale:
Earlier u_render_mask  was not present for multi-view.

roshanch:
@thomasflynn we added rendermask fix for multiview recently. I am not sure whether we tested it on that particular device.

PR #1553, opened 2017-10-17
fix post effect pause resume freeze issue

roshanch:
Fbo and tex ids are recreated by oculus after pause resume which are not populated in rendertexture stored on java side. now retrieving currently bound fbo by oculus every frame. tested with opacitysample, widdgetviewer, modelviewer2, tutorial-6, vuforia.

### DIFF
--- a/GVRf/Framework/framework/src/main/jni/engine/renderer/gl_renderer.cpp
+++ b/GVRf/Framework/framework/src/main/jni/engine/renderer/gl_renderer.cpp
@@ -211,8 +211,6 @@ namespace gvr
         rstate.uniforms.u_view = camera->getViewMatrix();
         rstate.uniforms.u_proj = camera->getProjectionMatrix();
 
-
-        RenderTexture* saveRenderTexture = renderTarget->getTexture();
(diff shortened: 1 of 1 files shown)


GVRf/Framework/framework/src/main/jni/engine/renderer/gl_renderer.cpp
```diff
@@ -282,8 +284,8 @@ namespace gvr
                 renderPostEffectData(rstate, input_texture, post_effects, i);
                 input_texture = renderTexture;
             }
-            GL(glBindFramebuffer(GL_FRAMEBUFFER, saveRenderTexture->getFrameBufferId()));
-            GL(glViewport(0, 0, saveRenderTexture->width(), saveRenderTexture->height()));
+            GL(glBindFramebuffer(GL_FRAMEBUFFER, drawFboId));
+            GL(glViewport(0, 0, viewport[2], viewport[3]));
```
dnvasilev: If we are restoring a viewport, how about restoring all 4 values? Don't assume old viewport always starts at (0,0)

apjagdale:
Works on S7. Tested using gvr-accessibility demo.

PR #1555, opened 2017-10-20
fix controller model visibility problem

NolaDonato:
1. call controller activation from framework thread
2. call bindShaders after controller model is added
3. remove unused assets

This fixes the problem with the controller model not becoming visible when the controller is activated after the app starts. gvr-controller exhibited this problem.


roshanch:
not sure if i messed up something with my repo, but it shows black screen with gvr-controller.

liaxim:
Worked fine for me with Erik's app.

roshanch:
@liaxim did you see the texture on the controller? I dont see it, but it fixes the issue mentioned.

GVRf/Framework/framework/src/main/java/org/gearvrf/io/GearCursorController.java
```diff
@@ -206,7 +203,14 @@ public void onDrawFrame()
...
                 initialized = true;
-                context.getInputManager().activateCursorController(GearCursorController.this);
+                Runnable task = new Runnable()
+                {
+                    public void run()
+                    {
+                        context.getInputManager().activateCursorController(GearCursorController.this);
+                    }
+                };
+                context.runOnTheFrameworkThread(task);
```
liaxim: Why the framework thread? It seems to me that all this adds is some latency. There is no other occasion the cursor controllers are using the framework thread. They probably should be for all callbacks but they don't. Also the callbacks right now are on a number of possible threads. Is this change mandatory for the fix to work?

NolaDonato:
Doesn't onInit run on the framework thread? That is where the cursor controllers are activated except for the Gear which might not be ready then. The reason I made this change is that the cursor controller loads an asset which should probably not occurs on the GL thread.

liaxim:
The cursor controllers are quite frivolous when it comes to threading. onInit does run on the framework thread, yes.

liaxim:
Planning to merge on Tuesday. @roshanch ?

liaxim:
Erik's app has been verified by QA.

roshanch:
@liaxim again something wrong with controller, i don't see the controller in app. @NolaDonato is going to check it on her end.

NolaDonato:
Will include this in upcoming controller reform

PR #1557, opened 2017-10-20
framework: fix disabling the gear cursor controller

liaxim:
- the gear cursor controller was impossible to disable since the flag was being passed in the wrong field (obj instead of arg1)
- fix formatting in GVRPicker
- add a proper hashCode() to GVRPicker

Required by https://github.com/gearvrf/GearVRf-Demos/pull/542

---


liaxim:
I don't think Erik's app cares about this change. I see no attempts to disable this specific type of controller.

liaxim:
Planning to merge on Tuesday.

PR #1563, opened 2017-10-25
framework: fix gvrmesh.getvertices

liaxim:

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRMesh.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRMesh.java
@@ -77,7 +77,7 @@ public class GVRMesh extends GVRHybridObject implements PrettyPrint {
      * @param gvrContext GVRContext to associate mesh with
      */
     public GVRMesh(GVRContext gvrContext) {
-        this(gvrContext, "float3 a_position float2 a_texcoord float3 a_normal ");
+        this(gvrContext, "float3 " + KEY_POSITION + " float2 " + KEY_TEXCOORD + " float3 " + KEY_NORMAL);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


NolaDonato:
If we are going to add getVerticesDirect, what about indices, texcoords, normals? We have accessors for those too. Personally I think these one-liners make the API kinda messy. Maybe just documenting in getVertices how you would access them as a buffer might cover the issue? In any case, if we are going to add a Direct option for vertices, we should add it for the others as well. I think the same bug also exists for getNormals and getTexCoords. I obviously did not understand the nuances of Java Direct buffers when I wrote that code :-)

liaxim:
It is not just a one liner. It allows you avoiding yet another copy, hence direct. Not that there is no copying already on the native side. To be honest there is no need for the direct buffers since the consumers are almost guaranteed to be Java consumers.

The other needs that too. When I get to them after creating a test and exercising them all.

NolaDonato:
The direct buffers on input are for assimp. If we take in direct buffers we can avoid a copy because we can directly use the C++ arrays.  If you don't think we need them on output then we can remove these methods. I figured if we took them in we should output them as well but I am OK if we remove them. Simple is better.

liaxim:
For output we can let the users decide. Hence two methods for user convenience. It is likely they need Java friendly output but we can't assume. If you prefer to fix it, let me know and I will close. The user has already figured out a workaround.

NolaDonato:
Are you going to fix normals and texcoords too?

liaxim:
Yes, but I want to add a simple test case together with those changes.

liaxim:
Update few more APIs, fixed a problem with getIndices. Added a simple test in https://github.com/gearvrf/GearVRf-Tests/pull/214.

liaxim:
@NolaDonato Please review

NolaDonato:
You can just call GVRVertexBuffer.getFloatArray in GVRMesh.getVertices. It is a native function which returns a Java array directly.

liaxim:
Using getFloatArray instead

liaxim:

PR #1564, opened 2017-10-25
changing anisotropic filter settings

roshanch:
if the user input for anisotropic is greater than supported value by driver, take the maximum supported value.

### DIFF
--- a/GVRf/Framework/framework/src/main/jni/gl/gl_texture.h
+++ b/GVRf/Framework/framework/src/main/jni/gl/gl_texture.h
@@ -103,17 +103,19 @@ public:
             glGenTextures(1, &id_);
             glBindTexture(target_, id_);
 
-            GLfloat anisotropic_level;
+            const char* extension_string = (const char*)glGetString(GL_EXTENSIONS);
+            if(strstr(extension_string, "GL_EXT_texture_filter_anisotropic")) {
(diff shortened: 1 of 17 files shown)


thomasflynn:
commit message needs DCO and some sort of description ;-)

GVRf/Framework/framework/src/main/jni/gl/gl_texture.h
```diff
@@ -103,10 +103,15 @@ class GLTexture : public GLPendingTask {
             glGenTextures(1, &id_);
             glBindTexture(target_, id_);
 
+            GLfloat anisotropic_level;
+
+            glGetFloatv(GL_MAX_TEXTURE_MAX_ANISOTROPY_EXT, &anisotropic_level);
```
liaxim: I believe before making this call you need to be certain EXT_texture_filter_anisotropic is supported.

liaxim:
@roshanch Do you agree the presence of EXT_texture_filter_anisotropic needs to be checked or should we merge as-is?

roshanch:
@liaxim yes, I will add the check for that.

roshanch:
@liaxim added extension check

roshanch:
uhh.!! too many commits from the merge.

PR #1569, opened 2017-10-27
framework: add undertaker thread

liaxim:
framework: fix shadowmap dtor crash

---

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRContext.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRContext.java
@@ -797,18 +797,47 @@ public abstract class GVRContext implements IEventReceiver {
      * Our {@linkplain GVRReference references} are placed on this queue, once
      * they've been finalized
      */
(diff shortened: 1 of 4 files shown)


liaxim:
- framework-tests runs show hybrid objects closed after done rendering

- the real problem with the tests is a ton of 4MB byte[] that are held around by native code, assuming platform's Bitmap impl
- study shows there is no Java-side leak of Bitmap instances
- the memory problem arises from the bitmap capture and comparing code in the tests; if eliminated footprint is much more reasonable
- possible platform bug or a very subtle bug in the bitmap usage; should be studied separately, as the time permits
- s8 shows 100MBs less of a footprint at the end of a complete test run compared to s7
- s7 finishes a complete test run but it is close; few more tests with image comparison and it will be outofmemory-ing
- clearly there are other leaks, Java and native side; this just plugs one hole

dnvasilev:
To summarize, the fix is: GVRContext.onDestroy() launches a new thread that takes over the reference queue and guarantees all C++ chunks of hybrid objects are destroyed.
I like the solution! There are some corner cases I'd like to discuss for 4.1; no questions about 3.x.

liaxim:
Planning to merge on Thursday

Issue #1567, opened 2017-10-26
Don't call virtual functions from destructors and constructors


dnvasilev:
**"Don't call virtual functions from destructors"** is a rarely-thought-of C++ (mis)feature. When a C++ object goes away, destructors are called in sequence, first the leaf class (most derived) destructor (destroying the leaf-specific parts of the object) in turn all the way to the top-of-the-inheritance-tree base object. Destructor destroys the underlying _slice_ of the object, and memory occupied by this slice becomes a graveyard of non-functioning bits. However if a base destructor calls into a virtual function, implementation will pick ~~the most-derived implementation of the virtual function and pass _this_ pointer to a corpse of the corresponding derived object.~~ the current class implementation of the virtual function and derived-class-specific actions will not be run. Bad. 

And yes, it happens in our code.

Example:
Component class has virtual destructor

      virtual ~Component();

that calls virtual function Component::set_owner_object(). There is only one implementation of this function, and it is in Component itself, and _we dodged one bullet_. Here's the implementation:

    inline Component::~Component() {
        set_owner_object(NULL);
    }

Component::set_owner_object() is another virtual function (_second bullet dodged_), and it calls other virtual functions, including, for example, Component::onDetach(). 

    virtual void onDetach(SceneObject* owner) { }

    inline void Component::set_owner_object(SceneObject *owner_object) {
        if (owner_object_) {
            onDetach(owner_object_);
        }
        . . .

EDIT: _during Component::~Component, onDetach() call selects Component::onDetach() implementation and never calls Transform::onDetach(). We don't call onTransformChanged(). This is a real bug._
~~onDetach() has implementation in Transform object, which will be destroyed and decompositing at the time Transform::onDetach() is called indirectly from Component::\~Component(). Luckily, Transform::onDetach() does not use any data members (why is it a member function then?) and _we dodged the third bullet_, but why are we **pushing our luck**?~~

    void Transform::onDetach(SceneObject *owner_object) {
        owner_object->onTransformChanged();
    }

NolaDonato:
?Do you have an suggestion on what we should do about it?


________________________________

dnvasilev:
But yes of course I have a suggestion!
**Don't call virtual functions from destructors**
Meaning: all polymorphism must be handled in the destructor itself. If you need to do something specific to Transform during destruction time, do it in Transform::~Transform(), and don't even think about calling a virtual function from Component::~Component()

NolaDonato:
[...]

PR #1570, opened 2017-10-27
Don't use virtual functions in destructors I

dnvasilev:
set_owner_object() is a virtual function. Don't call it from destructor please.

Part of the #1567 series

-----

### DIFF
--- a/GVRf/Framework/framework/src/main/jni/objects/components/transform.cpp
+++ b/GVRf/Framework/framework/src/main/jni/objects/components/transform.cpp
@@ -34,7 +34,9 @@ Transform::Transform() :
 }
 
 Transform::~Transform() {
-    owner_object_->onTransformChanged();
+    if(owner_object_) {
+        owner_object_->onTransformChanged();
+    }
 }
 
 void Transform::invalidate()
--- a/GVRf/Framework/framework/src/main/jni/objects/components/transform.cpp
+++ b/GVRf/Framework/framework/src/main/jni/objects/components/transform.cpp
@@ -34,6 +34,7 @@ Transform::Transform() :
 }
 
 Transform::~Transform() {
+    owner_object_->onTransformChanged();
 }
 
 void Transform::invalidate()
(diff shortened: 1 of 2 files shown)


dnvasilev:
This is the only place we deal with set_owner_object() and onDetach().
According to Android Studio, there are two implementations of onDetach():
Component::onDetach() /* empty */
Transform::onDetach()
and only one implementation of set_owner_object()

NolaDonato:
Check the physics extension just in case.

dnvasilev:
Physics extension: no grep matches for onDetach and set_owner_object.

dnvasilev:
@NolaDonato , any other comments?

GVRf/Framework/framework/src/main/jni/objects/components/transform.cpp
```diff
@@ -34,6 +34,7 @@ Transform::Transform() :
 }
 
 Transform::~Transform() {
+    owner_object_->onTransformChanged();
```
ragner: Looks that owner_object_ is not obligatory by Transform's constructor.
I think could be a good practice check if owner_object_ is not NULL before reference it just to avoid some future issue.

dnvasilev: Good catch!

GVRf/Framework/framework/src/main/jni/objects/components/transform.cpp
```diff
@@ -34,6 +34,9 @@ Transform::Transform() :
 }
 
 Transform::~Transform() {
+    if(owner_object_) {
+        owner_object_->onTransformChanged();
```
liaxim: @NolaDonato What is the purpose of calling onTransformChanged from the dtor?

On a side note the onTransformChanged method violates the component abstraction in SceneObject. There is a need for a general Component-changed mechanism?

liaxim:
Planning to merge on Thursday

PR #1572, opened 2017-10-28
vulkan code restructure

roshanch:
It fixes pipeline catching for vulkan objects. fixes gvr-controls and gvr-performance for vulkan.
Part of Vulkan renderer code restructured, more on the way.

GearVRf-DCO-1.0-Signed-off-by: Roshan Chaudhari roshan.c@samsung.com

PR #1575, opened 2017-10-30
framework: fix memory leaks in scene and javacomponents

liaxim:

### DIFF
--- a/GVRf/Framework/framework/src/main/jni/objects/components/render_target.cpp
+++ b/GVRf/Framework/framework/src/main/jni/objects/components/render_target.cpp
@@ -92,10 +92,6 @@ RenderTarget::RenderTarget()
 
 RenderTarget::~RenderTarget()
 {
-    if (mRenderTexture)
-    {
-        delete mRenderTexture;
-    }
 }
 
 /**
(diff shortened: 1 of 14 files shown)


NolaDonato:
Should we give all hybrid objects a reference to the Java object? Would this help with cleanup? I know it would get rid of the table of colliders in the picker.

liaxim:
Most don't need that. Could be an optional trait of a hybrid object. Something like that is needed to avoid the current split between Scene and JavaComponents.

GVRf/Framework/framework/src/main/jni/objects/components/render_target.cpp
```diff
@@ -92,10 +92,6 @@ RenderTarget::RenderTarget()
 
 RenderTarget::~RenderTarget()
 {
-    if (mRenderTexture)
-    {
-        delete mRenderTexture;
```
dnvasilev: Good catch

liaxim:
Planning to merge on Thursday. https://github.com/Samsung/GearVRf/pull/1572 makes a bindShader call that needs to be updated to look like the ugly section I added to render_data.cpp:283-289 here.

PR #1589, opened 2017-11-01
Fold RenderData::copy() function back into RenderData copy constructor

### LINKED ISSUE #1582: RenderData::copy needs redesign
Smells:
Java delegate constructor syntax in C++ code is a no-no. 
RenderData::copy() is at odds with Component::type. Component is a base class of RenderData. Component::type is protected (so it can be set from RenderData), but is not set ever after Component was constructed (what is the intent?)
RenderData is a kind of HybridObject (so it is owned by a Java object), and is a kind of JavaComponent (so it can maintain a pointer to the Java object), but it also is constructed by Batch::add() directly (and then there will be no Java object)

dnvasilev:
Now code is written in C++ and not Java

Partially fixes #1582 

As a bonus, I made Component type private member. I believe this is the intent.

- - -


liaxim:
Planning to merge on Tuesday

Issue #1579, opened 2017-10-31
reinterpret_cast considered harmful


dnvasilev:
We have 77 occurrences of reinterpret_cast in Framework/framework, outside of places where reinterpret_cast is justified (JNI interface, OpenGL extensions PFNs, contribs). Need to examine and justify each. Consider replacing with static_cast or const_cast.
~~We have 787 occurrences of reinterpret_cast in Framework alone, and I bet dollars for donuts that all of them should not be there. Use static_cast or const_cast.~~

dnvasilev:
fixed.

PR #1590, opened 2017-11-01
Downcast using static_cast

dnvasilev:
Downcast using static_cast
No explicit cast necessary for upcast
Remove unused ugly header file
Smell detected: VulkanShader::bindTextures casts texture to both VkTexture and VkRenderTexture that are not related.


dnvasilev:
Fixes #1579

PR #1592, opened 2017-11-02
[handleOnUndock] handleOnUndock is called by onDestroy

sisimma:
- undock event didn't make Activity finish properly.
     :activity didn't get called onDestroy,
     :unregister dock event receiver was not called,
     :since activity was not finished, some services were connected for some time until garbages got collected.

   - Changes are
     :undock handler calls Activity finish,
     :onDestroy calls handleOnUndock so that activity finishes after doing things to be done before.

Signed-off-by: Sean Yoon <sukhwan.y@samsung.com>


liaxim:
Will merge on Wednesday

PR #1593, opened 2017-11-02
framework: add a missing return statement

liaxim:


liaxim:
@roshanch @apjagdale This has to be the problem.

gcc usually does what you want here, which can be very confusing. clang crashes, so much easier to catch this way.

roshanch:
Good catch :+1:

PR #1596, opened 2017-11-03
v8 JavaScript engine for GearVR

mwitchwilliams:
Extends to v8 JavaScript engine to all of GearVR.  Previously, v8 had only been used with X3D.


mwitchwilliams:
Currently does not work for gvr-JavaScript example in the Dem's repository (renamed "disabled-gvr-javascript" in the maint_v3.x branch https://github.com/gearvrf/GearVRf-Demos/tree/maint_v3.x ).

liaxim:
What is the problem with gvr-javascript and what is the plan for fixing it? The disabled demos will be enabled again after Nola's controller reform.

liaxim:
Also are we really dropping support for lua? Its interpreter is a source of memory leaks for us and it would be great not to deal with it anymore. Or having to update it for which the time is due.

thomasflynn:
there doesn't seem to be the expected interest in Lua.  I think we can deprecate it and work towards removing it.

mwitchwilliams:
The issue with gvr-javascript demo is that it it reads the javaScript (.js) file from the assets directory, unlike X3D or Tom's demo, which has JavaScript embedded in &lt;Script&gt; nodes within the X3D file.

When the .js file is read from the assets directory, GVRScriptManager.loadScript() method is called.  GVRScriptManager.loadScript() calls GVRScriptManager.initalizeEngines() to invoke the V8 JavaScript engine, where the program hangs waiting on the GLThread.
Unlike the previous Rhino JavaScript engine, the V8 JavaScript engine must be run on a specific thread.  

The proposed solution is to create a new thread (proposed name "JSThread") within GVRScriptManager and use that instead of GLThread.

liaxim:
Since this pull request is for the 3.x branch it can't be in the 4.1 release milestone. @mwitchwilliams What should we do for 4.1?

GVRf/Framework/framework/src/main/java/org/gearvrf/script/GVRScriptManager.java
```diff
@@ -217,22 +250,94 @@ public GVRScriptFile getScriptFile(IScriptable target) {
      * @throws GVRScriptException if script processing error occurs.
      */
     public GVRScriptFile loadScript(GVRAndroidResource resource, String language) throws IOException, GVRScriptException {
-        if (getEngine(language) == null) {
-            mGvrContext.logError("Script language " + language + " unsupported", this);
-            throw new GVRScriptException(String.format("The language is unknown: %s", language));
-        }
+        //{
```
liaxim: This method could use a cleanup

mwitchwilliams: Yes, this code is still a work-in-progress and the Pull Request was only done to allow Tom to test his application.

liaxim: Thanks for clarifying. Updated the status so everybody knows.

mwitchwilliams:
The issue introduced by having the v8 JavaScript engine run on its own thread (called _JSthread_) instead of _GLthread_ is that both the _GLthread_ and _JSthread_ are writing/reading values for example to the same GVRTransform or GVRMaterial.
In addition, to enable procedural animations using JavaScript (for example, having a 3d mesh animating based on a sine wave), X3D's AnimationInteracitivtyManager.java has an onDrawFrame() listener.  Running on the _GLthread_, we seemed to be insync with each draw frames, but now running on the _GLthread_, the _JSthread_ might be falling behind the frame rate.
Currently investigating adding synchronization, but this may be some code rewrite.

NolaDonato:
I believe that writing the same transform is OK because we added locking around that so physics could run in its own thread.

mwitchwilliams:
This Pull Request is superseded by https://github.com/Samsung/GearVRf/pull/1623 for the portion that enables X3D's &lt;SCRIPT&gt; node to have JavaScript call Java methods.  
The ability to have the v8 engine run JavaScript (for example, gvr-JavaScript from the Demo's repo) in all GearVR will be developed for master branch (v4.x) instead.

PR #1606, opened 2017-11-07
Use synchronized in GVRSceneObject.detachAllComponents

NolaDonato:
Fix thread-safety issue in GVRSceneObject.detachAllComponents.
This was causing a crash in startup in gvr-3dcursor-simple.


NolaDonato:
too many commits, needs squash

PR #1607, opened 2017-11-07
Use synchronized in GVRSceneObject.detachAllComponents

NolaDonato:
Fix thread-safety issue in GVRSceneObject.detachAllComponents.
This was causing a crash in startup in gvr-3dcursor-simple.


liaxim:
Just curious, what was the crash (callstack)? Was it a Java exception or native? detachAllComponents is called by the finalizer thread presumably because the instance is being gc-ed. Which means there would be no strong references to the object and no opportunity to concurrently modify the collection.

liaxim:
Will merge on Wednesday

PR #1609, opened 2017-11-08
Bitfields for render data flags

apjagdale:
Assigning single bit to each boolean variables and 3 bits for the draw mode.
Also changed the render data hash code generation.

### DIFF
--- a/GVRf/Framework/framework/src/main/jni/objects/components/render_data.h
+++ b/GVRf/Framework/framework/src/main/jni/objects/components/render_data.h
@@ -79,20 +79,21 @@ public:
             texture_capturer(0),
             bones_ubo_(nullptr)
     {
-        renderFlags.use_light_ = false;
-        renderFlags.use_lightmap_ = false;
-        renderFlags.batching_ = true;
-        renderFlags.offset_ = false;
-        renderFlags.depth_test_ = true;
-        renderFlags.depth_mask_ = true;
-        renderFlags.alpha_blend_ = true;
-        renderFlags.alpha_to_coverage_ = false;
-        renderFlags.hash_code_dirty_ = true;
-        renderFlags.cast_shadows_ = true;
-        renderFlags.dirty_ = false;
-        renderFlags.invert_coverage_mask_ = GL_FALSE;
-        renderFlags.stencilTestFlag_ = false;
-        renderFlags.draw_mode_ = GL_TRIANGLES;
+        render_data_flags.HashCode = 0;
+        render_data_flags.BitFields.use_light_ = false;
+        render_data_flags.BitFields.use_lightmap_ = false;
+        render_data_flags.BitFields.batching_ = true;
+        render_data_flags.BitFields.offset_ = false;
+        render_data_flags.BitFields.depth_test_ = true;
+        render_data_flags.BitFields.depth_mask_ = true;
+        render_data_flags.BitFields.alpha_blend_ = true;
+        render_data_flags.BitFields.alpha_to_coverage_ = false;
+        hash_code_dirty_ = true;
+        render_data_flags.BitFields.cast_shadows_ = true;
+        render_data_flags.BitFields.dirty_ = false;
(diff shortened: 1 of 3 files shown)


GVRf/Framework/framework/src/main/jni/objects/components/render_data.h
```diff
@@ -433,7 +433,26 @@ class RenderData: public JavaComponent {
...
+        bool alpha_to_coverage_:1;
+
+        bool batching_:1;
+        bool hash_code_dirty_:1;
+        bool dirty_:1;
+        bool cast_shadows_:1;
+        GLboolean invert_coverage_mask_:1;
+        bool stencilTestFlag_:1;
+        GLenum draw_mode_:3;
+    };
```
roshanch: can you make it union, to get the hashcode..

NolaDonato:
I suggest you make a union of the bit fields and an integer. This integer is your hash code.

GVRf/Framework/framework/src/main/jni/objects/components/render_data.h
```diff
@@ -433,7 +440,27 @@ class RenderData: public JavaComponent {
     int stencilOpDpfail_ = 0;
     int stencilOpDppass_ = 0;
     unsigned int stencilMaskMask_ = 0;
-    bool stencilTestFlag_ = false;
+    bool hash_code_dirty_;
+
+    union {
```
liaxim: Strictly speaking it is undefined behavior to access the HashCode member if it was not the last written to. Implementations usually support it but we should be obeying the rules.

Excerpt from http://en.cppreference.com/w/cpp/language/union
The union is only as big as necessary to hold its largest data member. The other data members are allocated in the same bytes as part of that largest member. The details of that allocation are implementation-defined, and it's undefined behavior to read from the member of the union that wasn't most recently written. Many compilers implement, as a non-standard language extension, the ability to read inactive members of a union.

NolaDonato: While I agree that type punning using unions is implementation dependent, GNU compiler extensions allow it and define the behavior. However, if we switch to Clang, all bets are off. Maybe we can use reinterpret_cast instead. It will do the same thing but avoid the union.

dnvasilev: Language layer here. TL/DR: it's safe for us to overlap int and bitfields in the same union.

C++ standard is designed to be truly universal. Strictly speaking, someone can implement compiler that compiles C++ to Java (or to Bash shell.) In this case, you don't have 'real' pointers (machine addresses); the only pointers you can use are pointers that are within the same array, that is, array indices. You don't have 'real' pointers and hence you can't put two different objects at the same address; ok, I can implement union as struct (members at different locations); wasting some memory here, but it will work. That explains the standard language about undefined behavior: in this case, implementation writes data to completely separate location, and the other location remains uninitialized and can throw exception if accessed (think about some interpreted language that can check all uninitialized variables.)

Real life is different from C++ standard, and in real life there are some common practices; some of them were called in the old days 'BSD conventions' because BSD-derived code will not compile if these conventions are not true for a particular C compiler. For example, byte must be 8 bits; signed integers must be two's complement (C does not mandate two's comlement ints; in fact, Cray supercomputers had compliant C implementation with hardware one's complement integers). Overlapping unions are part of BSD convention. Combining bitfields and ints will work; however combining ints and floats does not always work as intended -- I remember that there are some nasty corner cases on i386 architecture.

liaxim: We will switch to clang. There were too many crashes for me to fix them all at the time I tried. Other project requires it and gcc has been deprecated anyway. Theoretically it all should be portable too.

liaxim:
Planning to merge on Friday.

Really wanted to try this with the 64bit build but it turns out it is crashing much earlier. Will be dealing with the non-working 64bit master build separately.

PR #1610, opened 2017-11-08
GVRMaterial.setMainTexture now sets first texture

NolaDonato:
This used to always set "u_texture" which is wrong for any shader but the texture shader.
Now it sets the first texture in the descriptor which should work for all shaders.


NolaDonato:
needs squah

PR #1611, opened 2017-11-08
GVRMaterial.setMainTexture now sets first texture  

NolaDonato:
This used to always set "u_texture" which is wrong for any shader but
the texture shader.
Now it sets the first texture in the descriptor which should work for
all shaders.

GearVRF DCO signed off by Nola Donato nola.donato@samsung.com


GVRf/Framework/framework/src/main/java/org/gearvrf/GVRMaterial.java
```diff
@@ -536,5 +552,19 @@ public void setLineWidth(float lineWidth) {
         setFloat("line_width", lineWidth);
     }
 
+    private String findFirstTexture()
+    {
+        if (mMainTextureName == null)
+        {
+            Pattern pattern = Pattern.compile("([a-zA-Z0-9]+) ([a-zA-Z0-9_]+)");
```
liaxim: The pattern ideally should be compile just once. Static should be ok.

liaxim:
Looks ok but unlikely to be thread-safe and it is a method that gets called from everywhere. @NolaDonato do you agree or should we merge as-is?

NolaDonato:
I will just put it in the constructor.

PR #1612, opened 2017-11-08
rebuild shader if texture added later

NolaDonato:
This patch propagates the dirty bits which indicate a new texture has
been added to the RenderPass and RenderData so the shader can be
recreated in this case. It fixes a bug in the 3D cursor sample with the
animated cursor.


liaxim:
gvr-controls: some of the textures are missing
gvr-3d-cursorsimple: some of the other textures of the red cube are missing

NolaDonato:
What textures are wrong in gvr-controls? I have compared it with the maint_v3.x version and it looks the same.

liaxim:
In my tests the sky and the fence were missing - all black. Will git clean -fdx and retest.

liaxim:
Not reproducing anymore after git clean -fdx. Will merge on Friday.

PR #1613, opened 2017-11-08
framework: prevent double-delete of a renderdata not owned by the camera

liaxim:
Before adding the undertaker thread problems like this one have to be fixed.

The post effect renderData is its own hybridObject that is taken care of by Java as any other hybridObject.


liaxim:
Planning to merge on Tuesday

PR #1614, opened 2017-11-08
framework: add undertaker thread to master

liaxim:
The tests do not crash due to native memory problems but because of the screen capturing not working it is not good enough of a verification

---


liaxim:
The test ran to completion without crashes on S8. @dnvasilev has informed me Note4 is a problem in the 3.3 branch and most likely it is a problem for master too. Will get to it eventually. Planning to merge this on Tuesday.

On a side note, testAlphaToCoverage failed with what seems a genuine problem.

PR #1615, opened 2017-11-09
Vulkan line width and renderPass hashCode

apjagdale:
Enabling line width for vulkan. (Won't affect the line width since devices are not supporting it yet)
Moved VertexBuffer descriptor string to RenderData hash code.
Added HashCode in renderPass to differentiate between cull faces and shader ID's

GearVRf-DCO-1.0-Signed-off-by: Abhijit Jagdale a.jagdale@samsung.com


liaxim:
Please resolve conflict in vulkan_renderer.cpp

liaxim:
@apjagdale Could you please resolve the conflicts?

GVRf/Framework/framework/src/main/jni/objects/components/render_data.cpp
```diff
@@ -219,7 +219,7 @@ std::string RenderData::getHashCode()
         render_data_string.append(to_string(stencilOpSfail_));
         render_data_string.append(to_string(stencilOpDpfail_));
         render_data_string.append(to_string(stencilOpDppass_));
-
+        render_data_string.append(to_string(mesh_->getVertexBuffer()->getDescriptor()));
```
liaxim: Can mesh_ or the result of getVertexBuffer() be nullptr?

GVRf/Framework/framework/src/main/jni/vulkan/vulkanCore.cpp
```diff
@@ -955,6 +956,11 @@ void VulkanCore::InitPipelineForRenderData(const GVR_VK_Vertices* m_vertices, Vu
                 VulkanShader *shader = static_cast<VulkanShader *>(shader_manager->getShader(
                         rdata->get_shader(false,curr_pass)));
 
+                float line_width;
+                if(!rdata->pass(curr_pass)->material()->getFloat("line_width", line_width))
```
liaxim: material() could theoretically return nullptr, right?

liaxim:
Not part of this change, but please add a return to RenderData::pass for the "else" case.

liaxim:
Just curious: why RenderPass needs the hash code?

liaxim:
@apjagdale Any comments?

GVRf/Framework/framework/src/main/jni/objects/components/render_data.cpp
apjagdale: No since Vertex buffer is created in constructor of mesh.

GVRf/Framework/framework/src/main/jni/vulkan/vulkanCore.cpp
apjagdale: Theoretically yes, adding the check.
About RenderPass hashcode. In vulkan we are caching vulkan pipeline on that basis.
So if two Objects have different cull_types or shaderID should have different pipelines.

GVRf/Framework/framework/src/main/jni/objects/render_pass.cpp
```diff
@@ -74,5 +74,15 @@ int RenderPass::isValid(Renderer* renderer, const RenderState& rstate, RenderDat
     return !dirty;
 }
 
+std::string RenderPass::getHashCode(bool is_multiview){
```
liaxim: This should be returning a ``const string&`` instead.

liaxim:
Planning to merge on Thursday

PR #1617, opened 2017-11-09
backend_oculus: minimum changes to compile and work with Oculus SDK 1.9.0

liaxim:
headModel params completely removed

---


liaxim:
gvr.xml and associated code has been in a need of revamp for some time; now there is no head model for Oculus. There is this tracking transform now that doesn't map to what we have. Maybe we need separate backend-specific config files for some things.

liaxim:
Tested all demos with a bunch of other pending pull requests. Have not seen problems caused by this particular one. Planning to merge on Tuesday.

PR #1619, opened 2017-11-10
fix bug for screenshot capture

roshanch:
retrieving shader instead of returning. Screenshot capturer gets called only once. 
tested with accessibility, modelviewer2, gvr-sample(for screenshot), cubemap.

GearVRf-DCO-1.0-Signed-off-by: Roshan Chaudhari, roshan.c@samsung.com


liaxim:
Works. Planning to merge on Tuesday.

PR #1624, opened 2017-11-13
framework: fix crashes in gvr-shadows and gvr-simplephysics

liaxim:
there is a case where a rendertarget passes null as a rendertexture;
added a check for it


liaxim:
@apjagdale Please review

apjagdale:
@liaxim Both demos now don't crash. But I noticed that in simple-physics the ball is not animated/thrown, it just sticks at the start position.

liaxim:
Planning to merge on Tuesday.

NolaDonato:
I tested with gvr-simplephysics and it works great. Thanks.

PR #1625, opened 2017-11-14
Restore scene editor

dnvasilev:
Restore ruined scene editor. Update API to 4.0. 
Compiles, runs, not tested.


liaxim:
Doesn't work that well for me. There is some work to be done to restore the functionality. Had hard time bringing up the file choose dialog, cursor disappeared then. Dialog looks weird. But it is a start.

PR #1626, opened 2017-11-14
maint_v3.x: Fix controller model visibility issue

NolaDonato:
This patch fixes the problem with the controller model not always showing up. It calls bindShaders after the model is loaded. It fixes the problem with the controller not being visible in VRTop.

This patch also adds GVREventReceivers.AssetEvents to make handling of asset loading events easier.


liaxim:
Planning to merge on Thursday

liaxim:
Will trigger the build, verify and create the gerrit change

PR #1627, opened 2017-11-15
[maint_v3.x] Fix controller stops working after reset it

ragner:
Picker's touch events stop working if you restart the controller
by removing and putting its battery back.


liaxim:
Setting the scene to null from removeCursorController seem desirable. Please take a look #1636 which fixes this problem and another one.

ragner:
@liaxim #1636 works for me. It looks good.

PR #1629, opened 2017-11-16
Fix logic to detect controllers

ragner:
Some controllers do not work without this fix.


GVRf/Framework/framework/src/main/java/org/gearvrf/io/GVRInputManager.java
```diff
@@ -212,9 +212,11 @@ private GVRControllerType getGVRInputDeviceType(InputDevice device) {
         int vendorId = device.getVendorId();
         int productId = device.getProductId();
 
-        if (sources == (InputDevice.SOURCE_TOUCHSCREEN | InputDevice.SOURCE_KEYBOARD)) {
+        if ((sources & InputDevice.SOURCE_TOUCHSCREEN) == InputDevice.SOURCE_TOUCHSCREEN
```
liaxim: Isn't it odd that this is the criteria for detecting a controller? A device with touchscreen and keyboard?

liaxim:
Will merge on Friday. @ragner agrees the criteria is rather strange. We'll look into it separately.

PR #1630, opened 2017-11-16
Fix NO_LIGHTING import setting

NolaDonato:
Disable lighting on the RenderData and use the Texture shader if lighting is disable on import. This fixes the controller showing up white in gvr-simplephysics.

The test for this patch is in PR #218 of GearVRf-Tests


liaxim:
Planning to merge on Thursday

PR #1633, opened 2017-11-16
Restore tiled rendering enhancer

dnvasilev:
Turns out other repo uses TiledRenderingEnhancer.

Restore from ea2b3ec77b38aaf408a9afe22a86194079974f43
Convert to a C++ namespace


liaxim:
I already added the header to the other repo. Doesn't the file belong there anyway?

PR #1635, opened 2017-11-17
Eliminate GVRScene.bindShaders

NolaDonato:
t is no longer necessary to call bindShaders when objects are added to the scene. This function is called individually for each RenderData if the shader is not valid. It is better to defer binding the shader until it is known what type of RenderTarget it will be used with.


liaxim:
Needs rebase

NolaDonato:
I rebased this one

liaxim:
Please include only the necessary files in the pull request. I can't quite guess what the actual change is. @dnvasilev was kind enough to create a git guide for us - please peruse it.

NolaDonato:
Replaced by #1654. I tried to clean this one up in place but it didn't go as I would have liked so I am opening a new PR.

PR #1636, opened 2017-11-20
[maint_v3.x] input manager refactor, fix npe in gaze, fix #1627

liaxim:
- merge inputmanager and inputmanagerimpl, expose interface instead
- fix npe in gaze cursor controller due to the recent scanDevices/scanControllers change
- fix #1627 

---


NolaDonato:
The problem with setting the scene to NULL is that, if that cursor controller is reactivated, there is no longer a scene and it doesn't do picking anymore. This case can be caused by removing the batteries from the Gear controller. It will switch to the Gaze controller. Putting the batteries back in causes it to switch back to the Gear controller which no longer has a scene.

liaxim:
Fixed the controller activation after reconnect by setting the main scene; commit 
8b19081. Seems more appropriate fix and also precautionary. Immersivepedia had problems because an unused controller instance was doing its own picking on the scene - one specific reason it is better to null out the scene, just in case.

ragner:
It looks good to me.

liaxim:
Thank you @ragner for reviewing. Planning to merge on Tuesday.

PR #1640, opened 2017-11-27
updated libjav8.so binaries.

thomasflynn:
updated libjav8.so binaries.  These should eliminate a crash when trying
to resolve double vs. float arguments to methods.

tom.flynn@samsung.com


thomasflynn:
this is for the 3.x maint branch

PR #1642, opened 2017-11-28
AssetLoader: error checking

NolaDonato:
If a texture is loaded early during initialization and cannot be found, a default image is used. If the error occurs early enough, the default image isn't there yet.

GVRPhongLayeredShader is only capable of layering two textures. Now we check if the FBX file layers more than two and no longer crash the shader.
This patch checks for these cases.


liaxim:
Conflicts with https://github.com/Samsung/GearVRf/pull/1630

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
```diff
@@ -208,7 +208,10 @@ public void loadTexture(TextureRequest request)
                                                               request, r, DEFAULT_PRIORITY, GVRCompressedTexture.BALANCED);
 
                     GVRImage whiteTex = getDefaultImage(mContext);
-                    request.loaded(whiteTex, null);
+                    if (whiteTex != null)
```
liaxim: When can whiteTex be null and what is the exception raised in getDefaultImage?

liaxim:
Planning to merge on Tuesday. Not very fond of the null check for the white texture.

NolaDonato:
Do you have a suggestion of how to avoid that null check? I don't like it either. It is caused by loading a texture that is not there, trying to get the default image which returns NULL because it can't create the GVRBitmapTexture.

liaxim:
Why the creation of GVRBitmapTexture fails?

PR #1646, opened 2017-11-30
cursor controller bug fixes, 3D cursor reassignment, widget plugin & scene editor

NolaDonato:
**Cursor controller bug fixes**
1. Fix cursor depth scaling in Gear controller
2. Set controller active state in Gaze controller
3. Combined GVRInputManager and GVRInputManagerImpl into one class
4. GVRPicker enables SEND_TO_HIT_OBJECT when using controller
5. GVRInputManager.selectController remove context argument
6. Fix Gear controller duplicate motion events
Picking with the controller was wonky because the controller was
automatically generating motion events. GVRViewSceneObject and
GVRWidgetPlugin also convert controller events into motion events within
the quad representing the Android view. This patch suppresses the motion
events automatically generated by the controller which fixes picking
inside Android views. GVRGearController.sendEventsToActivity is added to
enable routing the events to the activity if this is desired.

**3DCursor: handle cursor reassignment**
The Gear controller comes on line later than the other controllers. This
patch makes the 3D cursor library recognize the controller and reassign
it to one of the active cursors. Previously an active cursor would never
get assigned to a different controller even if it was higher priority.

**Support controller in scene editor and widget plugin**
Modified widget plugin to handle controller events. Clicking works but
scrolling inside the widget does not.

Demos pull request: https://github.com/gearvrf/GearVRf-Demos/pull/556


GVRf/Extensions/3DCursor/3DCursorLibrary/src/main/java/org/gearvrf/io/cursor3d/Cursor.java
```diff
@@ -461,8 +461,46 @@ public IoDevice getIoDevice() {
     }
 
     /**
-     * Returns a list of {@link IoDevice}s compatible with the {@link Cursor}.
+     * Returns the priority for the {@link IoDevice} currently attached to the {@link Cursor}.
+     *
+     * @return -1 if no device attached, otherwise returns device priority
+     */
+    public int getCurrentDevicePriority() {
```
liaxim: Should getCurrentDevicePriority, getDevicePriority, isDeviceCompatible be public? What use can they be to end users?

NolaDonato: I agree, these are only used by the cursor manager.

liaxim:
gvr-controller - I think the bunnies could be scaled by dragging left and right, right? Whatever it is, if you attempt that the bunnies disappear.

liaxim:
gvr-keyboardview: the left-most keys can't be hovered above nor selected

liaxim:
gvr-modelviewer2: pause/resume the app, can't select anything anymore; even without pause/resume can't select alternative camera

liaxim:
gvr-widgetviewer crashes after pause/resume

liaxim:
There are way too many problems. I am quite confident we need to include few more bug fixes to this pull request before merging it. Otherwise the situation can quickly get out of hand.

NolaDonato:
In gvr-controller, the bunnies move but they do not scale. I think I have fixed the issue with dragging objects using the controller by making the drag thread-safe. Previously the app was doing the drag setup and the cursor controller could interrupt in the middle and change the cursor position.

NolaDonato:
I fixed the issue with gvr-keyboardview key picking.

liaxim:
I guess we could merge this round of fixes. But there is still good amount left. The widgetplugin-based demos have problems after pause/resume. I still see offset when picking and more. @NolaDonato any preferences?

liaxim:
The cursor scaling feels somewhat erratic.

liaxim:
In the next round of fixes please fix some regression issues when not using the controller. For example, no clicks go through in gvr-accessibility if the controller is not present.

GVRf/Framework/framework/src/main/java/org/gearvrf/io/GVRTouchPadGestureDetector.java
```diff
@@ -0,0 +1,280 @@
+/************************************************************************************
```
liaxim: I am under the impression we could not include this class in the Framework for some licensing issue. @thomasflynn ?

NolaDonato:
I did not know that. I will make our own version of this class. I can make the API more closely match the Android Gesture detector. I thought it was OK since we are including it in all of our demos.

liaxim:
The widget plugin problem where events do not work after pause/resume is present without this change in both master and maint_v3.x. So it is completely separate issue.

liaxim:
Every 5-10 minutes of using the demo apps, I get a NPE like these:
```
12-14 08:06:54.688  7894  7894 E AndroidRuntime: FATAL EXCEPTION: main
12-14 08:06:54.688  7894  7894 E AndroidRuntime: Process: org.gearvrf.GVRWidgetViewer, PID: 7894
12-14 08:06:54.688  7894  7894 E AndroidRuntime: java.lang.NullPointerException: Attempt to invoke virtual method 'float android.view.MotionEvent.getX()' on a null object reference
12-14 08:06:54.688  7894  7894 E AndroidRuntime: 	at org.gearvrf.io.GVRTouchPadGestureDetector.onFling(GVRTouchPadGestureDetector.java:171)
12-14 08:06:54.688  7894  7894 E AndroidRuntime: 	at android.view.GestureDetector.onTouchEvent(GestureDetector.java:667)
12-14 08:06:54.688  7894  7894 E AndroidRuntime: 	at org.gearvrf.io.GVRTouchPadGestureDetector.onTouchEvent(GVRTouchPadGestureDetector.java:251)
12-14 08:06:54.688  7894  7894 E AndroidRuntime: 	at org.gearvrf.widgetViewer.GVRWidgetViewer.onTouchEvent(GVRWidgetViewer.java:83)
12-14 08:06:54.688  7894  7894 E AndroidRuntime: 	at org.gearvrf.widgetplugin.GVRWidgetPlugin$7$1.run(GVRWidgetPlugin.java:750)
12-14 08:06:54.688  7894  7894 E AndroidRuntime: 	at android.os.Handler.handleCallback(Handler.java:751)
12-14 08:06:54.688  7894  7894 E AndroidRuntime: 	at android.os.Handler.dispatchMessage(Handler.java:95)
12-14 08:06:54.688  7894  7894 E AndroidRuntime: 	at android.os.Looper.loop(Looper.java:154)
12-14 08:06:54.688  7894  7894 E AndroidRuntime: 	at android.app.ActivityThread.main(ActivityThread.java:6776)
12-14 08:06:54.688  7894  7894 E AndroidRuntime: 	at java.lang.reflect.Method.invoke(Native Method)
--
12-14 08:10:44.726  1266  1877 D ConnectivityService: sending notification CAP_CHANGED for NetworkRequest [ id=17, legacyType=-1, [ Capabilities: INTERNET&NOT_RESTRICTED&TRUSTED] ]
12-14 08:10:44.726  1266  1826 D InputReader: Input event(10): value=0 when=5428421684000
12-14 08:10:44.726  1266  1826 I InputReader: Touch event's action is 0x1 (deviceType=0) [pCnt=1, s=] when=5428421684000
12-14 08:10:44.727  1266  1825 I InputDispatcher: Delivering touch to (9415): action: 0x1, toolType: 1
12-14 08:10:44.727  2037  2213 D NetworkController: onReceive: intent=Intent { act=android.net.wifi.RSSI_CHANGED flg=0x4000010 launchParam=MultiScreenLaunchParams { mDisplayId=0 mBaseDisplayId=0 mFlags=0 } bqHint=1 (has extras) }
12-14 08:10:44.728  9415  9415 D ViewRootImpl@31e4041[ModelViewer2Activity]: ViewPostImeInputStage processPointer 1
12-14 08:10:44.730  9415  9415 D TOUCH   : dispatchPickerActivity action = 1 2684.667969, 739.335938
12-14 08:10:44.747  1266  5147 D SSRM:SmoothScrollBooster: execute:: type = SMOOTH_SCROLL, value = TRUE, pid = 1266, uid = 1000
12-14 08:10:44.747  1266  5147 D SSRM:SmoothScrollBooster: Request from background package is ignored.
12-14 08:10:44.747  9415  9415 D AndroidRuntime: Shutting down VM
12-14 08:10:44.753  1266  5147 D SSRM:SmoothScrollBooster: execute:: type = SMOOTH_SCROLL, value = FALSE, pid = 1266, uid = 1000
12-14 08:10:44.755  1266  5147 D GameManagerService: identifyGamePackage. org.gearvrf.modelviewer2
12-14 08:10:44.757  9415  9415 E AndroidRuntime: FATAL EXCEPTION: main
12-14 08:10:44.757  9415  9415 E AndroidRuntime: Process: org.gearvrf.modelviewer2, PID: 9415
12-14 08:10:44.757  9415  9415 E AndroidRuntime: java.lang.NullPointerException: Attempt to invoke virtual method 'float android.view.MotionEvent.getX()' on a null object reference
12-14 08:10:44.757  9415  9415 E AndroidRuntime: 	at org.gearvrf.io.GVRTouchPadGestureDetector.onFling(GVRTouchPadGestureDetector.java:171)
12-14 08:10:44.757  9415  9415 E AndroidRuntime: 	at android.view.GestureDetector.onTouchEvent(GestureDetector.java:667)
12-14 08:10:44.757  9415  9415 E AndroidRuntime: 	at org.gearvrf.io.GVRTouchPadGestureDetector.onTouchEvent(GVRTouchPadGestureDetector.java:251)
12-14 08:10:44.757  9415  9415 E AndroidRuntime: 	at org.gearvrf.modelviewer2.ModelViewer2Activity.onTouchEvent(ModelViewer2Activity.java:90)
12-14 08:10:44.757  9415  9415 E AndroidRuntime: 	at org.gearvrf.widgetplugin.GVRWidgetPlugin$7$1.run(GVRWidgetPlugin.java:750)
12-14 08:10:44.757  9415  9415 E AndroidRuntime: 	at android.os.Handler.handleCallback(Handler.java:751)
12-14 08:10:44.757  9415  9415 E AndroidRuntime: 	at android.os.Handler.dispatchMessage(Handler.java:95)
12-14 08:10:44.757  9415  9415 E AndroidRuntime: 	at android.os.Looper.loop(Looper.java:154)
12-14 08:10:44.757  9415  9415 E AndroidRuntime: 	at android.app.ActivityThread.main(ActivityThread.java:6776)
12-14 08:10:44.757  9415  9415 E AndroidRuntime: 	at java.lang.reflect.Method.invoke(Native Method)
```
While checking for null in GVRTouchPadGestureDetector can be effective the root cause must be identified for this one, I believe.

gvr-simplephysics - still can't launch balls without the controller

Whomever is going to review and finally merge this - the changes must be independently verified by going through all the demos with and without the controller.

liaxim:
gvr-3dcursor - no controller; after resetting the scene (the big red button) the reticle disappears.

NolaDonato:
I have fixed the issues with gvr-3dcursor.

NolaDonato:
I fixed the issues with gvr-simplephysics in GearVRf-Demops #556

NolaDonato:
I have replaced the GVRTouchPadGestureDetector class that we cannot distribute. Now it is calls GVRTouchPadGestureListener and it is implemented as a subclass of OnGestureListener instead of a GestureDetector. This is a completely different implementation and a different API (matches OnGestureListener) so we should be OK.

mwitchwilliams:
Tested on S7 with all the files involving interactivity and procedural animations from the Tests->X3D repo -> Scripting, Scripting_Per_Frame, Switch_node, TouchSensors, anchor folders.
25 total test invoking animations or using JavaScript to modify GVRTransform, GVRMaterial, light properties, change viewpoint location and open web pages.  All passed!
**LGTM - Looks Good To Me**

thomasflynn:
has conflicts and needs rebase.

NolaDonato:
Results of testing on Pixel:
1. apps which used GestureDetector do not get events from controller. The onDown method of the listener is never called so it is not getting the events even though we are routing them thru GVRActivity.dispatchTouchEvent. These demos include gvr-accessibility, gvr-immersivepedia, gvr-controls
2. gvr-immersivepedia displays the wrong stuff on startup - objects are wrong size and possibly duplicated.
3. gvr-keyboardview does not display keypresses in the View even though they are being sent.

NolaDonato:
Until you brought it up I had no idea anyone was relying on 0,0 to tell if something was touched. I am trying to get everything working with the Daydream controller as well. It gives much different behaviour than Oculus so I doubt the 0,0 thing works on Daydream so we need this API change.

In case you want to run your app on the Pixel, here is the update. So far Daydream controller works in GearVRf except that the gesture detector ignores the events GVRActivity sends to it (never calls onDown). Until recently very few of our demos used the controller. I stuck the controller into lots of the demos and got the Oculus one to kinda work with most of them, even the ones that use the Android gesture detector. Daydream doesn't work with that yet and I haven't figured out why.

liaxim:
I don't know about this ICameraEvents. Why is the locking in transform.cpp not enough?

I will look into reproducing the flicker and root-causing. @SteveGreatApe If you have a sample you could share with us it will be appreciated.

NolaDonato:
Because his application uses a draw frame listener to update the camera rig. This update is not synchronized with the headset update so any object attached to the camera in between updates will be in the wrong position. ICameraEvents ensures that the head tracker update and the user update happen at the same time.

liaxim:
gvr-controller: with gaze cursor close to 60fps; with the controller - dips into the 40s. Why is that? Should it be that expensive? Tested on s8.

GVRf/Framework/framework/src/main/java/org/gearvrf/io/GearCursorController.java
```diff
@@ -533,16 +571,24 @@ void handleControllerEvent(final ControllerEvent event) {
...
-            if (mSendEventsToActivity)
+            if (mSendEventsToActivity && ((keyEvent.size() > 0) || (motionEvent.size() > 0)))
             {
-                GVRActivity activity = getGVRContext().getActivity();
-                for (KeyEvent e : keyEvent)
+                mPropagateEvents.init(keyEvent, motionEvent);
+                if (Looper.getMainLooper().equals(Looper.myLooper()))
                 {
-                    activity.dispatchKeyEvent(e);
+                    mPropagateEvents.run();
```
liaxim: What is mPropagateEvents for? It seems to me like both branches of this if effectively do the same thing on the same thread.

NolaDonato: It crashes in some applications when I call GVRActivity.dispatchMotionEvent. I will go back and figure out which one crashed. Maybe there is a better way to handle it. I was getting "you are doing stuff on the wrong thread" messages from Android.

NolaDonato: I checked into it and this code can go. The crash happens in gvr-events. GVRActivity.dispatchKeyEvent is called from the controller thread and it causes this exception: 
android.view.ViewRootImpl$CalledFromWrongThreadException: Only the original thread that created a view hierarchy can touch its views.

liaxim: See Activity.runOnUiThread: https://developer.android.com/reference/android/app/Activity.html#runOnUiThread(java.lang.Runnable). Probably the sample itself needs to place the call on the right thread; not GVRf's responsibility what happens in an event handler.

NolaDonato: I took a look at the logic and you are right - it sucks. I recoded it to always propagate the events to the UI thread and now gvr-events works.

NolaDonato: The Gear controller thread will NEVER be the UI thread and usually you want to respond to these events on that thread. I had the comparison backwards.

liaxim:
Isn't it time to consider breaking this pull request down? There are some generic changes you mentioned, like collider-related that could be separated out and merged. It seems to me it will be easier to progress with this if we could merge the stuff that works well. We could even merge controller stuff if it works well with Oculus and separately work on the other backends.

liaxim:
Copying from the demos pull request:
Double-checked against master. On master with/without controller gvr-controller is at ~60fps. With this change in, after I enable the controller the fps drops by ~10. It doesn't recover after the controller is disconnected. I believe this needs to be understood before merging.

NolaDonato:
I believe I have fixed this performance problem. I now do picking on the GL thread and it works a lot better. Was originally doing it on the controller thread.

liaxim:
Had a chat with Nola, the reticle has been fixed. Planning to merge on Tuesday.

GVRf/Framework/framework/src/main/java/org/gearvrf/io/GearCursorController.java
```diff
@@ -520,7 +516,10 @@ public void run()
                 for (MotionEvent e : mMotionEvents)
                 {
                     activity.dispatchTouchEvent(MotionEvent.obtain(e));
+                    e.recycle();
```
liaxim: This recycle is causing problems with gvr-events and I doubt only with it.
```
01-23 08:26:39.176 12053 12053 E AndroidRuntime: java.lang.RuntimeException: MotionEvent { action=ACTION_MOVE, actionButton=0, id[0]=0, x[0]=187.0, y[0]=67.0, toolType[0]=TOOL_TYPE_FINGER, buttonState=BUTTON_PRIMARY, metaState=0, flags=0x0, edgeFlags=0x0, pointerCount=1, historySize=0, eventTime=176380, downTime=176346, deviceId=0, source=0x100008 } recycled twice!
01-23 08:26:39.176 12053 12053 E AndroidRuntime: 	at android.view.InputEvent.recycle(InputEvent.java:120)
01-23 08:26:39.176 12053 12053 E AndroidRuntime: 	at android.view.MotionEvent.recycle(MotionEvent.java:2053)
01-23 08:26:39.176 12053 12053 E AndroidRuntime: 	at org.gearvrf.io.GearCursorController$EventHandlerThread$SendEvents.run(GearCursorController.java:519)
01-23 08:26:39.176 12053 12053 E AndroidRuntime: 	at android.os.Handler.handleCallback(Handler.java:751)
01-23 08:26:39.176 12053 12053 E AndroidRuntime: 	at android.os.Handler.dispatchMessage(Handler.java:95)
01-23 08:26:39.176 12053 12053 E AndroidRuntime: 	at android.os.Looper.loop(Looper.java:154)
01-23 08:26:39.176 12053 12053 E AndroidRuntime: 	at android.app.ActivityThread.main(ActivityThread.java:6776)
01-23 08:26:39.176 12053 12053 E AndroidRuntime: 	at java.lang.reflect.Method.invoke(Native Method)
01-23 08:26:39.176 12053 12053 E AndroidRuntime: 	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1520)
01-23 08:26:39.176 12053 12053 E AndroidRuntime: 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1410)
```

Probably removing it is enough. That is what I did locally.

liaxim:
gvr-events + daydream - native crash; just clicking around, using the touchpad.

```
01-23 11:41:34.988 11974 11974 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
01-23 11:41:34.988 11974 11974 F DEBUG   : Build fingerprint: 'google/sailfish/sailfish:8.1.0/OPP5.170921.005/4373449:user/release-keys'
01-23 11:41:34.988 11974 11974 F DEBUG   : Revision: '0'
01-23 11:41:34.988 11974 11974 F DEBUG   : ABI: 'arm'
01-23 11:41:34.988 11974 11974 F DEBUG   : pid: 11506, tid: 11506, name: .gearvrf.events  >>> org.gearvrf.events <<<
01-23 11:41:34.988 11974 11974 F DEBUG   : signal 6 (SIGABRT), code -6 (SI_TKILL), fault addr --------
01-23 11:41:34.988 11974 11974 F DEBUG   :     r0 00000000  r1 00002cf2  r2 00000006  r3 00000008
01-23 11:41:34.988 11974 11974 F DEBUG   :     r4 00002cf2  r5 00002cf2  r6 ffb49c64  r7 0000010c
01-23 11:41:34.988 11974 11974 F DEBUG   :     r8 00000000  r9 ea05a000  sl 00000000  fp 00000000
01-23 11:41:34.988 11974 11974 F DEBUG   :     ip 00000000  sp ffb49c50  lr eb802bdf  pc eb7fc5ec  cpsr 200b0030
01-23 11:41:35.349 11974 11974 F DEBUG   : 
01-23 11:41:35.349 11974 11974 F DEBUG   : backtrace:
01-23 11:41:35.352 11974 11974 F DEBUG   :     #00 pc 0001a5ec  /system/lib/libc.so (abort+63)
01-23 11:41:35.352 11974 11974 F DEBUG   :     #01 pc 00010737  /system/lib/libinput.so (android::MotionEvent::getRawPointerCoords(unsigned int) const+50)
01-23 11:41:35.352 11974 11974 F DEBUG   :     #02 pc 0001081f  /system/lib/libinput.so (android::MotionEvent::getAxisValue(int, unsigned int) const+10)
01-23 11:41:35.352 11974 11974 F DEBUG   :     #03 pc 010219a9  /system/framework/arm/boot-framework.oat (offset 0x5c5000) (android.view.MotionEvent.nativeGetAxisValue [DEDUPED]+128)
01-23 11:41:35.352 11974 11974 F DEBUG   :     #04 pc 000283c1  /dev/ashmem/dalvik-jit-code-cache (deleted)
```

liaxim:
The daydream controller is generating Android events in addition to the events our fw synthesizes. I think some of the problems are down to the mix of controller-generated and our events flowing through the system.

GVRf/Framework/framework/src/main/java/org/gearvrf/io/GearCursorController.java
liaxim: I guess the duped motion event needs to be recycled instead.

liaxim:
Also saw a ConcurrentModificationException

```
01-23 13:21:06.189 25724 25724 E AndroidRuntime: FATAL EXCEPTION: main
01-23 13:21:06.189 25724 25724 E AndroidRuntime: Process: org.gearvrf.events, PID: 25724
01-23 13:21:06.189 25724 25724 E AndroidRuntime: java.util.ConcurrentModificationException
01-23 13:21:06.189 25724 25724 E AndroidRuntime: 	at java.util.ArrayList$Itr.next(ArrayList.java:860)
01-23 13:21:06.189 25724 25724 E AndroidRuntime: 	at org.gearvrf.io.GearCursorController$EventHandlerThread$SendEvents.run(GearCursorController.java:518)
01-23 13:21:06.189 25724 25724 E AndroidRuntime: 	at android.os.Handler.handleCallback(Handler.java:790)
01-23 13:21:06.189 25724 25724 E AndroidRuntime: 	at android.os.Handler.dispatchMessage(Handler.java:99)
01-23 13:21:06.189 25724 25724 E AndroidRuntime: 	at android.os.Looper.loop(Looper.java:164)
01-23 13:21:06.189 25724 25724 E AndroidRuntime: 	at android.app.ActivityThread.main(ActivityThread.java:6494)
01-23 13:21:06.189 25724 25724 E AndroidRuntime: 	at java.lang.reflect.Method.invoke(Native Method)
01-23 13:21:06.189 25724 25724 E AndroidRuntime: 	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:438)
01-23 13:21:06.189 25724 25724 E AndroidRuntime: 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:807)
```

liaxim:
I am seeing these:

1. No reaction at all to multiple clicks on the checkbox: there are no onTouch events being delivered to the ITouchEvents listener in GVRViewSceneObject. It looks like kind of a picking problem. Moving the cursor around I can see the reticle jumping sometimes - I guess surface projection is getting involved. Eventually ITouchEvents start coming through and then the view reacts.

2. Even if ITouchEvents work, quick succession of clicks not always works. You may need to pause for a half a second before clicking again.

3. Native crashes on the UI thread somewhere in MotionEvent::getRawPointerCoords.

PR #1649, opened 2017-12-02
Vulkan monoscopic renderer

apjagdale:
Initial pull request for vulkan monoscopic.
Vulkan monoscopic now works, also with multi-sampling and post effects.
Created two new classes for onScreen and offScreen VkRenderTextures.
Fixed culling with vulkan.


GVRf/Framework/framework/src/main/java/org/gearvrf/GVRViewManager.java
```diff
@@ -124,7 +124,7 @@ public void setMainScene(GVRScene scene) {
         }
     }
 
-    private void setMainSceneImpl(GVRScene scene) {
+    protected void setMainSceneImpl(GVRScene scene) {
```
liaxim: Please make it private again and call GVRViewManager.onSurfaceCreated instead. There is a good reason this method ends with **Impl**.

liaxim: Reason being it is sensitive and subject to the occasional tweaks. It should be all confined to the base class.

GVRf/Framework/framework/src/main/jni/objects/components/render_data.cpp
```diff
@@ -57,7 +57,7 @@ void RenderData::set_mesh(Mesh* mesh)
     }
 }
 
-bool RenderData::cull_face(int pass) const {
+int RenderData::cull_face(int pass) const {
```
liaxim: Please also add the missing return to this method; in case of a fall-through.

GVRf/Framework/framework/src/main/jni/objects/textures/render_texture_jni.cpp
```diff
@@ -72,6 +76,22 @@ Java_org_gearvrf_NativeRenderTexture_ctorMSAA(JNIEnv* env, jobject obj,
...
+Java_org_gearvrf_NativeRenderTexture_ctorWithMonoscopic(JNIEnv* env, jobject obj,
+                                              jint width, jint height, jint sample_count, jint number_views, jboolean monoscopic)
+{
+    int depth_format = DepthFormat::DEPTH_24_STENCIL_8;
+
+    if(number_views > 1) // multiview doesn't work with stencil attachment
+        depth_format = DepthFormat::DEPTH_24;
+    RenderTexture* tex;
+    if(Renderer::getInstance()->isVulkanInstance())
+        tex = static_cast<VulkanRenderer*>(Renderer::getInstance())->createRenderTexture(width, height, sample_count, ColorFormat::COLOR_8888, depth_format , 0, NULL, number_views, monoscopic);
```
liaxim: How is the monoscopic flag being used? Quick search showed no use. If it is not necessary let's remove the flag and this new method.

GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrMonoscopicViewManager.java
```diff
@@ -126,32 +136,111 @@
...
-        // Log.v(TAG, "onDrawFrame");
         beforeDrawEyes();
         drawEyes();
         afterDrawEyes();
     }
 
+    public class Worker extends Thread {
+
+        @Override
+        public void run() {
```
liaxim: How does this thread get stopped?

liaxim:
General question: how did you test? Which demo apps did you verify with Vulkan and which with OpenGL?

apjagdale:
@Liahim I tested with complexscene, meshanimation and eyepicking.
For post effect I modified meshanimation to horizontally flip the scene.

liaxim:
@apjagdale Did you test pause and resume?

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRRenderTexture.java
```diff
@@ -53,10 +53,18 @@ public GVRRenderTexture(GVRContext gvrContext, int width, int height, int sample
         this(gvrContext,width,height,sampleCount,1);
     }
 
+    public GVRRenderTexture(GVRContext gvrContext, int width, int height, int sampleCount, boolean monoscopic)
```
liaxim: Can't push backend specifics into the framework. In a future change the monoscopic backend will become its own separate backend.

Please create a more abstract way of requesting and producing on and off screen targets.

GVRf/Framework/framework/src/main/jni/engine/renderer/vulkan_renderer.h
```diff
@@ -119,6 +112,9 @@ class VulkanRenderer: public Renderer {
                                                const TextureParameters* texture_parameters, int number_views);
     virtual RenderTexture* createRenderTexture(int width, int height, int sample_count, int layers) { }
     virtual RenderTexture* createRenderTexture(const RenderTextureInfo&);
+    RenderTexture* createMonoscopicRenderTexture(int width, int height, int sample_count,
```
liaxim: createRenderTexture should be enough. Can't have monoscopic-specifics in this interface.

GVRf/Framework/framework/src/main/jni/objects/textures/render_texture_jni.cpp
```diff
@@ -56,21 +57,41 @@ JNIEXPORT jlong JNICALL
...
+                                              jint width, jint height, jint sample_count, jint number_views, jboolean monoscopic)
 {
     int depth_format = DepthFormat::DEPTH_24_STENCIL_8;
 
     if(number_views > 1) // multiview doesn't work with stencil attachment
         depth_format = DepthFormat::DEPTH_24;
-    RenderTexture* tex = Renderer::getInstance()->createRenderTexture(width, height, sample_count, ColorFormat::COLOR_8888, depth_format , 0, NULL, number_views);
+    RenderTexture* tex;
+    if(Renderer::getInstance()->isVulkanInstance() && monoscopic)
+        tex = static_cast<VulkanRenderer*>(Renderer::getInstance())->createMonoscopicRenderTexture(width, height, sample_count, ColorFormat::COLOR_8888, depth_format , 0, NULL, number_views);
```
liaxim: No is-Vulkan-instance or is-monoscopic. Should go through the Renderer interface.

liaxim:
There is some work to be done. I understand there might be need for more specific suggestions but lets have you give it a try first. One basic rule we can't break - no if-backend-A-do-this in the framework. Another: we have the renderer interface - use that; no explicit if-vulkan-do-B. There is some code already doing that, let's not add more.

Also please put somewhere in the wiki a definitive list of Vulkan samples that are supposed to work. Sometimes I am scratching my head wondering should gvr-cubemap work or not for example. Eventually we need to do something about gracefully failing when something unsupported is used.

liaxim:
complexscene shows some visual artifacts on head rotation.

GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrMonoscopicViewManager.java
```diff
@@ -126,32 +140,129 @@
...
+                        else{
+                            NativeVulkanCore.resetTheInstance();
+                            NativeVulkanCore.getInstance(vulkanSurfaceView.getHolder().getSurface());
+                        }
+                    }
+
+                    @Override
+                    public void surfaceChanged(SurfaceHolder surfaceHolder, int i, int i1, int i2) {
+                        Log.d("Vulkan", "surfaceChanged");
+                        vulkanDrawThread.start();
```
liaxim: surfaceChanged can be called multiple times. Thread.start() can't be called more than once per instance.

GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrMonoscopicViewManager.java
```diff
@@ -126,32 +140,129 @@
...
     }
 
+    public class Worker implements Runnable {
+        public void run() {
+            while(activeFlag){
+                beforeDrawEyes();
+                drawEyes();
+                afterDrawEyes();
+            }
+            mRenderTarget[0] = null;
```
liaxim: Why not null out the other elements of the array while at it?

liaxim:
Used complexscene and accessibility - although there is no fps in the logs, you can tell they are rendering faster.

apjagdale:
@liaxim I have done the interface modifications. Do let me know if thats the expected change. 
Regarding artifacts for complexscene, I will see why its happening. For me it appears at the starting frames and latter its gone.

apjagdale:
@liaxim The artifacts issue for me appears on S7. I tried on Note 8 and it worked just fine.

GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrMonoscopicViewManager.java
```diff
@@ -126,32 +140,131 @@
             mViewportY = (screenHeightPixels / 2) - (fboHeight / 2);
         }
 
+        sampleCount = gvrActivity.getAppSettings().getEyeBufferParams().getMultiSamples();
+
+        try {
+            Process proc = Runtime.getRuntime().exec(new String[]{"/system/bin/getprop", "debug.gearvrf.vulkan"});
```
dnvasilev: I see
1. Vulkan-specific code in backend-oculus. I'd like to understand the design in detail, explaining why it is necessary and can't be hidden somewhere else.
2. Reading a property directly, instead of using our numerous property-reading routines (that need to be consolidated, and using yet another way to read properties is wrong)
3. Something that requires at least storage-read permission and may not work at all on user build phone

GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrMonoscopicViewManager.java
```diff
@@ -15,10 +15,17 @@
...
 import android.util.DisplayMetrics;
+import android.view.SurfaceHolder;
+import android.view.SurfaceView;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
 
 /*
  * This is the most important part of gvrf.
```
dnvasilev: Need to add Vulkan-specific comments. 

_Maybe_ we need to rethink Vulkan design because this comment block says that there's nothing GL-specific in the constructor, but we do Vulkan init in the constructor.

**Possibilities**:
1. the original comment is wrong (there _is_ some GL-specific code in constructor and adding Vulkan code is OK): **need to rewrite the comment**
2. the oringinal _design_ is wrong (works with GL, but Vulkan is _so_ different that we need to rethink the whole design): **rethink the design, write new documentation, redo GL**
3. Vulkan implementation is not following the original design (design holds, but Vulkan as-is is built right across it): **redo Vulkan**

GVRf/Framework/framework/src/main/jni/vulkan/vulkanCore.cpp
```diff
@@ -1144,28 +1143,29 @@ void VulkanCore::InitPipelineForRenderData(const GVR_VK_Vertices* m_vertices, Vu
...
 
-       // LOGE("Abhijit fence %d",vkQueueWaitIdle(m_queue));
-
-        //LOGE("Abhijit swapchain flags %d", swapChainFlag);
-        if(!fence)
-        PresentBackBuffer();
+        if(swapChainFlag) {
+            VkResult success = VK_INCOMPLETE;
+            while(success){
+                success = vkGetFenceStatus(m_device, fence);
```
liaxim: Why not use vkWaitForFences? If vkGetFenceStatus must be used, VK_ERROR_DEVICE_LOST probably should be handled somehow.

GVRf/Framework/framework/src/main/jni/vulkan/vk_render_texture_onscreen.cpp
```diff
@@ -0,0 +1,42 @@
...
+        if(fbo == nullptr){
+            fbo = new VKFramebuffer(mWidth,mHeight);
+            createRenderPass();
+            VulkanRenderer* vk_renderer= static_cast<VulkanRenderer*>(Renderer::getInstance());
+
+            fbo->createFrameBuffer(vk_renderer->getDevice(), DEPTH_IMAGE | COLOR_IMAGE, mSamples, true);
+        }
+    }
+
+    void VkRenderTextureOnScreen::initVkData(){
```
liaxim: I don't see the reason why initVkData is virtual and has identical impl in both Off and OnScreen.

liaxim:
Thought gvr-accessibility works with Vulkan but it crashes for me on S8.

liaxim:
Looks ok to me. Probably can be merged. We will have to come back for things like the monoscopic flag in the frontend, creating a separate monoscopic backend. Maybe even two backends - one for GL and for Vulkan.

apjagdale:
@liaxim gvr-accessibility won't work since it has text in it and we don't support external textures as of now.

PR #1654, opened 2017-12-04
Eliminate GVRScene.bindShaders

NolaDonato:
It is no longer necessary to call bindShaders when objects are added to the scene. This function is called individually for each RenderData if the shader is not valid. It is better to defer binding the shader until it is known what type of RenderTarget it will be used with.


liaxim:

liaxim:
@NolaDonato I came to realize that the render data's dirty flag is never cleared (gl renderer). So bindShader is getting called continuously for each and every render data. If a light is added but render data's dirty flag is correctly cleared prior to that, how would the system work without the removed Scene.bindShaders (or something similar)?

I have a small fix for the render data being dirty all the time but found problems with shadows missing.

PR #1650, opened 2017-12-04
[maint_v3.x] Revert "[handleOnUndock] handleOnUndock is called by onDestroy (#1592)"

liaxim:
Reverts pull request https://github.com/Samsung/GearVRf/pull/1592.

It is causing problems for a certain application.

---

GearVRf-DCO-1.0-Signed-off-by: Mihail Marinov <m.marinov@samsung.com>

PR #1655, opened 2017-12-05
Revert "[handleOnUndock] handleOnUndock is called by onDestroy (#1592)"

sisimma:
The patch caused an issue of black screen on undocking HMT. SO revert it.
This reverts commit c0062e3b33193c50102df6aaf57574b0736294d7.


liaxim:
Was reverted by https://github.com/Samsung/GearVRf/pull/1650

PR #1659, opened 2017-12-06
Reset static V8 engine to prevent reuse after onDestroy

xcaostagit:
The V8 engine declared in GVRJavascriptV8File is static, making it possible
to be reused in a different thread and cause a crash. This change ensures
such static reference gets reset after activity's onDestroy is called.


thomasflynn:
ok, yeah, this is similar to what I did in my own tree.  this will work.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRActivity.java
```diff
@@ -236,6 +237,7 @@ protected void onDestroy() {
         mRenderableViewGroup = null;
         mConfigurationManager = null;
 
+        GVRJavascriptV8File.destroy();
```
liaxim: If a second activity is created for this process, the old activity's onDestroy is usually called after onCreate for the new one. Meaning there is theoretically a window of time where the incorrect instance could be used by the new activity.

Ideally there should be no static writable data but if you must, you could use GVRContext.addResetOnRestartHandler to set mEngine to null in a safer manner.

xcaostagit: OK will try this approach.

xcaostagit: Done

thomasflynn: you no longer need these changes to GVRActivity.  the ResetOnRestart handler is all that is needed.  in fact, since that destroy() method no longer exists, it now breaks the build.

xcaostagit: That's right. GVRActivity is not touched at all in my 2nd commit. However I am confused by the 2 commits appearing in this pull request. I used 'git commit --amend -a' before 'git push'. Did I miss something?

xcaostagit: Only the 2nd commit shall be merged. The 1st one shall be discarded. I may have used git commands incorrectly.

PR #1663, opened 2017-12-12
[maint_v3.x] Assimp returns one scene object if asset contains single mesh

NolaDonato:
Assimp adds nodes UNDER the root model. Internal projects have requested
to add the mesh directly to the root model if there is only one mesh to
make it easier to access. This patch eliminates the extra root node if
the asset has a single mesh.


liaxim:

PR #1670, opened 2017-12-15
X3D parse exponent

mwitchwilliams:
Creates a fix in X3D parser.  Previously SAX parser splits 3e-2 (which is .03) into two number, 3 and -2.
New code pre-processes this value so that it will be a single number.

Test Case has many examples of both negative and positive exponents
[exponent demo.zip](https://github.com/Samsung/GearVRf/files/1563755/exponent.demo.zip)

screenshot_20171215-105518

GearVRF DCO signed off by: Mitch Williams


thomasflynn:
looks good.  will merge tomorrow if there are no further comments.

liaxim:
Seems like it could've been simpler to use String.split to tokenize and then BigDecimal.toPlainString to get rid of the exponent field.

PR #1674, opened 2017-12-22
Optimize texture parameter setting in GL

NolaDonato:
Do not update the GL texture parameters unless they have changed.


thomasflynn:
immersivepedia seems to have a bunch of black textures with this.  maybe wanna double check?

PR #1676, opened 2017-12-22
API changes for 4.1

NolaDonato:
These API changes are up for review. Corresponding pull requests in GearVRf-Demos and GearVRf-Tests incorporate these api changes.

Removed GVRMaterialShaderManager, GVRPostEffectShaderManager (use GVRShaderManager now)
Removed GVRGUISceneObject (GVRViewSceneObject has this functionality now)
GVRBaseSensor -> GVRSensor
GVRBitmapTexture -> GVRBitmapImage
GVRCubemapTexture -> GVRCubemapImae
GVRCompressedTexture -> GVRCompressedImage
GVRCompressedCubemapTexture -> GVRCompressedCubemapImage
GVRFloatTexture -> GVRFloatImage
Removed GVRMesh.setTriangles, GVRMesh.getTriangles
Added GVRMesh.setCharIndices, GVRMesh.setIntIndices, GVRMesh.getCharIndices, GVRMesh.getIntIndices
GVRLightBase -> GVRLight
GearCursorController -> GVRGearCursorController
Moved GVRxxxShader to org.gearvrf.shaders
Move GVRCursorController to org.gearvrf.io


liaxim:
Have not examined the change but in general: new packages are only ok if the dependencies are one-way (to org.gearvrf), not too many (hard to set exact number) and there is a clean API boundary (no make this and that public just to get to it).

liaxim:
Needs rebase

NolaDonato:
It needs internal discussion more than rebase :-)

GVRf/Framework/framework/src/main/java/org/gearvrf/io/GVRCursorController.java
```diff
@@ -988,7 +1002,7 @@ public boolean isEventHandledBySensorManager() {
         return eventHandledBySensor;
     }
 
-    void setEventHandledBySensor()
+    public void setEventHandledBySensor()
```
liaxim: setEventHandledBySensor is an example why I am against having the io package. As long as we need such methods to be public, we should keep the io stuff in the main package.

liaxim:
It is a nice improvement. You should rebase so we can get it in. Please consider the need for the io package but I am not going to get hung up on that.

thomasflynn:
@liaxim are there any other API changes you would like to see?  (other than the io package that you mentioned)

liaxim:
Have nothing specific. But it would be nice to remove at least some of the deprecated methods.

liaxim:
Well, I find ``public void addChildObject(GVRComponent childComponent)`` not particularly good. Should be removed in my opinion and its use appropriately replaced with ``addChildObject(GVRSceneObject child)``.

NolaDonato:
I agree about GVRComponent.addChildObject. I left that in for compatibility with old stuff but it can be removed.

NolaDonato:
If there are no complaints I will merge this tomorrow. Tests won't run without it.

liaxim:
Needs rebase

PR #1677, opened 2017-12-27
Detect transparency in ASTC.

thomasflynn:
Detect transparency in ASTC.  ASTC files will scan their Color Endpoint
Modes to tell if the file has transparency or not.

tom.flynn@samsung.com

### DIFF
--- a/GVRf/Framework/framework/src/main/jni/objects/textures/astc_transparency.cpp
+++ b/GVRf/Framework/framework/src/main/jni/objects/textures/astc_transparency.cpp
@@ -1,0 +1,691 @@
+/* Copyright 2015 Samsung Electronics Co., LTD
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


thomasflynn:
need to write some tests before it's ready for review.

thomasflynn:
i should also probably tone down the logging :)

NolaDonato:
Transparency tests now have their own class and each test is a separate function.  This is different than your original test.

mwitchwilliams:
I can build some quick tests in X3D with animations of texture maps with transparency moving in front of and behind opaque objects. What do we expect to see different after implementing this change?
Is this an additional fix to https://github.com/Samsung/GearVRf/issues/1203 ? (see screen shots captured in the PR)

thomasflynn:
prior to this change anything with an ASTC texture was treated as transparent since there was no way to tell in the header information.  Now we parse through and read the block endpoint information so we can tell if even an individual block has transparency.

in other words, prior to this change, even opaque ASTC textures were reported as transparent.  with this change, they are recognized properly.

I'm looking to add some tests to GearVRf-Tests -> framework -> TextureTransparencyTests.java.

thomasflynn:
Added tests in:  https://github.com/gearvrf/GearVRf-Tests/pull/224
I think this is ready for review now.

liaxim:
Works as expected. Tested by using the automated tests and supplying few astc textures I encoded myself. Planning to merge on Wednesday.

PR #1679, opened 2017-12-29
Avoid adding shader multiple times

NolaDonato:
Check shader signature on Java side before requesting a new shader from C++.
This removes the DataDescriptor: ... message every frame from the OESHorizontalShader

Fix handling of u_right in renderer. It was not getting set correctly causing some shaders that rely on it (OESHorizontalStereoShader, OESVerticalStereoShader) to produce incorrect visual results for the right camera.

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRMaterial.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRMaterial.java
@@ -135,7 +135,7 @@ public class GVRMaterial extends  GVRShaderData
     };
 
     /**
-     * A new holder for a shader's uniforms.
+     * Create a new holder for a shader's uniforms.
      * 
      * @param gvrContext
      *            Current {@link GVRContext}
@@ -162,6 +162,21 @@ public class GVRMaterial extends  GVRShaderData
      */
     public GVRMaterial(GVRContext gvrContext) {
         this(gvrContext, GVRShaderType.Texture.ID);
+    }
+
+    /**
+     * Create a new holder for a shader's uniforms from an existing material.
+     *
+     * @param src   GVRMaterial to copy from
+     * @param shaderId  shader ID
+      */
+    public GVRMaterial(GVRMaterial src, GVRShaderId shaderId)
+    {
+        super(src, shaderId);
+        if (hasTexture(mMainTextureName))
+        {
+            mMainTextureName = src.mMainTextureName;
+        }
     }
 
     /**
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRShader.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRShader.java
@@ -349,39 +349,9 @@ public class GVRShader
         synchronized (shaderManager)
         {
             int nativeShader = shaderManager.getShader(signature);
-            if (nativeShader > 0)
-            {
-                nativeShader = addShader(shaderManager, signature, material);
-            }
-            return nativeShader;
-        }
-    }
-
-    /**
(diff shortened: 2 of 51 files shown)


GVRf/Framework/framework/src/main/java/org/gearvrf/GVRShader.java
```diff
@@ -343,7 +348,11 @@ public int bindShader(GVRContext context, GVRShaderData material, String vertexD
 
         synchronized (shaderManager)
         {
-            int nativeShader = addShader(shaderManager, signature, material);
+            int nativeShader = shaderManager.getShader(signature);
+            if (nativeShader > 0)
```
ragner: Hi @NolaDonato , at line 319  you are checking "if (nativeShader == 0)" after get getShader(...) but at line 352 it has  "if (nativeShader > 0)". 
Just to confirm if it is right because that seems to be some mistake.

NolaDonato: int nativeShader = shaderManager.getShader(signature);
// getShader will return 0 if the shader has not been added
if (nativeShader == 0)
{
      // addShader will a native shader ID return > 0 if a shader is successfully added.
      nativeShader = addShader(shaderManager, signature, mtl);
}
// If it returns 0, the shader was not created so we don't want to set it on the render data.
if (nativeShader > 0)
{
      rdata.setShader(nativeShader, isMultiview);
}

ragner: Hi @NolaDonato, the line 319 looks good to me
but I mean about the line 352.

NolaDonato: You are 100% correct. This looks like a merge problem. It should be == instead of >.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRMaterial.java
```diff
@@ -164,6 +164,21 @@ public GVRMaterial(GVRContext gvrContext) {
...
+    /**
+     * Create a new holder for a shader's uniforms from an existing material.
+     *
+     * @param src   GVRMaterial to copy from
+     * @param shaderId  shader ID
+      */
+    public GVRMaterial(GVRMaterial src, GVRShaderId shaderId)
+    {
+        super(src, shaderId);
+        if (hasTexture(mMainTextureName))
```
liaxim: This if looks to be redundant. mMainTextureName is going to be null at this time.

liaxim:
Looks good. Tested the demos, gl, s8. Planning to merge on Thursday.

NolaDonato:
Does immersivepedia run at 60 fps now?

liaxim:
There is 4-5 fps improvement. The fps fluctuates though. Tried with enabling the various options in the museum scene. It is closer to 60 most of the time than without this change.

NolaDonato:
I think 4.0 has a lot more logging in the rendering loop than previous releases. I am going to put this logging on a #define to see if it is the problem.

NolaDonato:
I commented out logging completely. I get solid 60fps on S8. I get only 45fps on S7. What did you test with?

thomasflynn:
we used to get 60fps on a Note4 when that sample was first submitted.  Just a point of reference.

liaxim:
Tested on S8.

liaxim:
Tested on S7. Master is the mid-40s, maint_v3.x is in high 50s. It also depends what you do in the museum scene. I touch all four tombstones and look around.

liaxim:
Merging this; the performance problem is a different topic.

PR #1680, opened 2017-12-29
V8 Thread

xcaostagit:
To minimize the impact of script execution on GL thread, we create
a dedecated thread called V8 thread, aka 'gvrf-v8' in framework. All
script execution will happen on this thread from now on.

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRContext.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRContext.java
@@ -179,6 +179,10 @@ public abstract class GVRContext implements IEventReceiver {
         mHandlerThread = new HandlerThread("gvrf-main");
         mHandlerThread.start();
         mHandler = new Handler(mHandlerThread.getLooper());
+
+        mV8HandlerThread = new HandlerThread("gvrf-v8");
+        mV8HandlerThread.start();
+        mV8Handler = new Handler(mV8HandlerThread.getLooper());
     }
 
     /**
@@ -188,6 +192,8 @@ public abstract class GVRContext implements IEventReceiver {
         mContext = null;
         mHandler = null;
         mHandlerThread = null;
+        mV8Handler = null;
+        mV8HandlerThread = null;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


mwitchwilliams:
Similar to the Master v4.x version, I believe this code will have the same issue in Pull Req https://github.com/Samsung/GearVRf/pull/1690.
Should be "work in progress". 3 tests on the repo: Tests > X3D > Scripting_Per_Frame that involve using the camera direction to drag objects are having a delayed response from 10ths of a second to what seems like a few seconds. Was not an issue when JavaScript ran on GLthread. The camera direction is read on a Call back from GVRDrawFrameListener but of course the JavaScript is on GLthread. Will need more investigation.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRContext.java
```diff
@@ -709,6 +725,13 @@ public void runOnTheFrameworkThread(final Runnable runnable) {
         mHandler.post(runnable);
     }
 
+    /**
+     * Execute a runnable on the dedicated V8 thread
+     */
+    public void runOnV8Thread(final Runnable runnable) {
```
liaxim: Why would we need a public api for that?

liaxim: And answering to it myself - because of the fact x3d is in its own package. I am pretty sure because of incidents like this that x3d should be moved to the gearvrf package. Now we are exposing an internal detail in the most prominent gvrf class just so x3d can get to it.

liaxim:
I get compilation errors. There are differences from 4.0 that require tweaks to this change.

liaxim:
My bad. I have too many branches. Builds fine.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRContext.java
mwitchwilliams: How might this change if this V8thread were moved into Extensions, just like gvr-physics?
If you are suggesting every X3D file be moved from it's own 'X3D' package into GearVR, then I suppose every X3D Java file could be prefixed with "X3D" in its name  just because these files are exclusive to X3D implementation.

liaxim: Since scripting in its own extension is not happening right now, in the meantime we could hide unnecessary public methods if X3D was in the gearvrf package. If one day we clean it all up and everything can happen through real public API then X3D does belong to its own package.

thomasflynn:
i don't think we need this on the 3.3 branch.  i'd like to close out any development there as soon as practical.

PR #1681, opened 2017-12-31
gradle version

pnemonic78:
Android gradle build tools plugin version 3.0.1 and build tools 26.0.3

### DIFF
--- a/GVRf/Extensions/3DCursor/3DCursorLibrary/build.gradle
+++ b/GVRf/Extensions/3DCursor/3DCursorLibrary/build.gradle
@@ -27,8 +27,8 @@ repositories {
 }
 
 android {
-    compileSdkVersion 21
-    buildToolsVersion '25.0.2'
+    compileSdkVersion 26
+    buildToolsVersion '26.0.3'
 
     defaultConfig {
         minSdkVersion 21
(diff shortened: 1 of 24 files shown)


thomasflynn:
@NolaDonato can you check if this works on your machine?

NolaDonato:
It does not work on my machine. I did "Invalidate Caches & Restart" in Android Studio. I did git clean -fdx in the Framework directory. I installed build tools 26.0.2. I still get:

Gradle 'Framework' project refresh failed
Error:Cannot set the value of read-only property 'outputFile' for ApkVariantOutputImpl_Decorated{apkData=Main{type=MAIN, fullName=debug, filters=[]}} of type com.android.build.gradle.internal.api.ApkVariantOutputImpl. 

I am not sure which build.gradle it comes from because it doesn't say.

pnemonic78:
@NolaDonato I think you compiled the _GearVRf-Demos_ project instead.

There is no outputFile in any GearVRf gradle file, but there is in GearVRf-Demos "common.gradle":
`output.outputFileName = System.getProperty("appName") + ".apk"`
and VRHackathons:
```
            output.outputFile = new File(
                    output.outputFile.parent,
                     rootProject.name + ".apk")

```

NolaDonato:
If I don't build any demos, just the Framework, it builds for me. I have no way to test since demos don't build.

thomasflynn:
ok, i was mostly concerned with whether the framework built for you.  as pointed out, the Demos repo is a different issue.  Thanks for checking.

PR #1683, opened 2017-12-31
joml version 1.9.3

pnemonic78:
joml is already version 1.9.3-SNAPSHOT


thomasflynn:
@pnemonic78 after merging the gradle and gson upgrades this pull-request has conflicts.  could you resolve them real quick?  thanks.

PR #1690, opened 2018-01-04
V8 Thread

xcaostagit:
To minimize the impact of script execution on GL thread, we create
a dedecated thread called V8 thread, aka 'gvrf-v8' in framework. All
script execution will happen on this thread from now on.


mwitchwilliams:
In X3D, all but 3 tests that did not run properly: _JavaScript_PerFrame_02_LaunchSphere.x3d_, _JavaScript_PerFrame_04_GrabAnObject.x3d_ (simplest of the 3) and _JavaScript_PerFrame_05_LightControls.x3d_. These are located in the Tests repo at: https://github.com/gearvrf/GearVRf-Tests/tree/master/x3d/Scripting_Per_Frame
In each x3d test, a mesh is grabbed, stays in front of the camera and moves with the user until the mesh is released using the camera lookAt. However now the mesh movement is delayed from a few 10ths of second to what seems like many seconds.  
In my earlier attempt at coding a new thread for the JavaScript, I had a similar issue.

In the X3D file, a  &lt;ROUTE&gt; sends the camera orientation to the &lt;SCRIPT&gt; node:
`<ROUTE fromField='orientation' fromNode='VP1' toField='vpRot' toNode='AnimateScript'/>`
The camera lookAt (orientation) is set in AnimationInteractivityManager.java, SetJavaScriptArguments() method 
```
else if (scriptObject.getFromDefinedItemField(field).equalsIgnoreCase("orientation")) {
   if ( definedItem.getViewpoint() != null ) {
      // have a Viewpoint which for the time-being means get the current direction of the camera
      float[] lookAt = gvrContext.getMainScene().getMainCameraRig().getLookAt();
      Vector3f cameraDir = new Vector3f(lookAt[0], lookAt[1], lookAt[2]);
      Quaternionf q = ConvertDirectionalVectorToQuaternion(cameraDir);
      AxisAngle4f cameraAxisAngle = new AxisAngle4f();
      q.get(cameraAxisAngle);
      scriptParameters.add( cameraAxisAngle.x );
      scriptParameters.add( cameraAxisAngle.y );
      scriptParameters.add( cameraAxisAngle.z );
      scriptParameters.add( cameraAxisAngle.angle );
   }
}  // orientation parameter
```
This method is called by onDrawFrame() which implements GVRDrawFrameListerner.

  gvr-javascript demo also crashed. There was a Pull Request for https://github.com/Samsung/GearVRf/pull/1623https://github.com/Samsung/GearVRf/pull/1623 for the maint_v3.x which will be merged soon, but has not be merged into v4.x

Also tested with Tom's gvr-x3d-360video but that had code to work in maint_v3.x that had not been merged into v4.x yet.

xcaostagit:
Out of the 3 failed tests, one thing to keep in mind is RunScript, hence SetResultsFromScript now runs outside of the GL thread. SetResultsFromScript() seems complex and it triggers a lot of sub-function calls. Need to analyze which one might cause this delay.

mwitchwilliams:
I do have other demo's which are procedural animations - transforms, colors, etc. - and those run ok.  So, the issue seems to involve the camera transform.
And of course these were ok on the GLthread. But agreed, we need to investigate, as I had similar results when I created my own thread.

liaxim:
@mwitchwilliams JavaScript_PerFrame_02_LaunchSphere works with master. Couple of things:

1. Calling into JS is pretty slow. Calling into an empty JS function on S8 takes 20-25ms. I guess it is what it is.

2. The slow animation. I think it is something you need to deal with. On the gl thread your RunScript runs every frame bringing down the fps to 27-28 on S8. The clock is correct at the time of RunScript execution though and RunScript is effectively throttled. On the v8 thread the demo runs at 60fps. You enqueue RunScript calls every 16ms with a snapshot of the clock that is correct for the time of enqueue. But the JS is very slow to execute (~30ms) and it has to run every single enqueued RunScript. So everything gets "replayed" at a much slower pace than the real clock. In my opinion you need to run a "tick" in the scripting thread and manage the whole thing from there.

mwitchwilliams:
Some serious performance issues. We need to investigate.
Re:  run a "tick" in the scripting thread
For a while, I was using the gvr-physics code in GearVR's Extension as a template, which has its own looping function. That effort was suspended so I could return to work on other X3D features.  Perhaps we should review it again
I wonder if there is an optimization we need to do for the v8 engine that may has been overlooked.
Thanks for the performance timing measurements.  Good guidance and I will hae to review X3D's AnimationInteractivityManager.java for performance opportunities.

NolaDonato:
The approach taken by the physics thread is not necessarily the best. It is OK for physics to run slower or faster than the GL thread. If it runs faster, you get a better simulation. If it runs slower, the simulation is slower but it doesn't affect the performance on the GL thread. In the case of physics, the only information exchanged between the two threads is matrices.
In the case of V8, it seems it runs slower than the GL thread by far and you don't ever want it to run faster. (That will not result in a better animation.) I agree you need a timer in your V8 thread to control the rate at which it runs.

liaxim:
Should we close for now?

liaxim:
Ok then. Planning to close on Thursday.

liaxim:
Closing until we have an update for it.

PR #1692, opened 2018-01-05
Spatial audio support

ragner:
Add new ResonanceAudio extension with:
 * GVRAudioSource component - A representation of audio sources in 3D.
 * GVRAudioListener component - Representation of a listener in 3D space.

Known issues:
The same scene object cannot support multiple sources.
The listener must be attached to scene's root.

```
GearVRf-DCO-1.0-Signed-off-by: Cristiano Bozzetto <c.bozzetto@samsung.com>
```


GVRf/Extensions/ResonanceAudio/resonanceaudio/src/main/java/org/gearvrf/resonanceaudio/GVRAudioListener.java
```diff
@@ -0,0 +1,294 @@
...
+
+    private static class AudioEngineEventsHandler extends GVREventListeners.ActivityEvents
+            implements GVRDrawFrameListener {
+        private static WeakReference<AudioEngineEventsHandler> mInstance = new WeakReference<AudioEngineEventsHandler>(null);
+
+        private GVRContext mContext;
+        private GVRAudioListener mCurrentListener;
+        private GVRScene mCurrentScene;
+
+        public static synchronized AudioEngineEventsHandler getInstance() {
```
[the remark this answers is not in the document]
ragner: Hi @liaxim,
we have considered that issues during our development but we didn't figure out a way to get notifications when the main scene is changing.

When we change from SceneA to SceneB we need pause the audio sources from SceneA and resume the audio sources from SceneB. But to do that we need listen to main scene's changes.

Because of that we implemented the singleton. :(

But I agreen that we need a beter logic to fix that issue. Do you have any suggestion?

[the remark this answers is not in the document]
ragner: @liaxim You can see at line 245 the code to check scene's changes.

NolaDonato:
When the scene becomes the main scene, it's onInit function is called. This is done via an event listener for ISceneEvents. I think that if you add a listener for ISceneEvents to the new scene's event receiver ISceneEvents.onInit will be called when that scene becomes the main scene. This event is issued on the framework thread.

NolaDonato:
It seems both physics and audio can benefit from an event indicating when a scene object or component is added to the scene. The C++ layer knows when this happens as of 4.0. It could notify the Java side when a component/scene object is added or removed from the scene.

GVRf/Extensions/ResonanceAudio/resonanceaudio/src/main/java/org/gearvrf/resonanceaudio/GVRAudioListener.java
```diff
@@ -0,0 +1,294 @@
...
+
+import java.lang.ref.WeakReference;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Represents a listener in the scene. It records the sound around it and plays through the
+ * speakers. It is allowed to have only one listener in the scene.
+ */
+public class GVRAudioListener extends GVRComponent {
```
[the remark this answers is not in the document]
ragner: Hi @liaxim, I know that makes the most sense to attach the listener to the Main Camera but we could make some improvements to allow attach the listener to other scene objects like Unity3D does (https://docs.unity3d.com/Manual/class-AudioListener.html) so every scene has its own listener.

We didn't choose by some GVRAudioSession because we was going handle the states(paused/stopped/playing ...) of all audio sources of all scenes in the single GVRAudioSession adding some complexity to our logic.

But we can talk more about it and make improvements :)

ragner:
Hi @NolaDonato, I have tried ISceneEvents.onInit and it has worked for me only for the first main scene. When I change from SceneA to SceneB the ISceneEvents.onInit event does not happen anymore.

ragner:
@NolaDonato  and ISceneObjectEvents.onInit behaves like  ISceneEvents.onInit
Works just for the first main scene.

ragner:
I have tried a fix calling  ISceneEvents.onInit every time the main scene changes.
But there are some internal logics of GVRf that should happens just at initialization of the application listening to ISceneEvents.onInit

NolaDonato:
I am confused as to how your changes failed to work. What is it that your changes broke?

ragner:
GVRViewManager.this.notifyMainSceneReady() is called at onFinish event of splash screen or
when there is no splash screen at first frame handler.
I moved that notifyMainSceneReady() calls to GVRViewManager.setMainSceneImpl(..) but the main issue that I have found was related to GVRScriptBehavior component attached to the scene.
How there is no ISceneEvents.onFinish(...) event to release resources allocated at onInit()
if I change from SceneA to SceneB and change back SceneB->SceneA the onInit of GVRScriptBehavior will be called twice.

NolaDonato:
I have written some code to try and fix some of the scene switching issues by making the audio listener NOT depend on the scene. I have made a pull request (repo NolaDonato branch audiopr) on top of this PR #1692. Please take a look at it and give it a review. Maybe this is what you are looking for?

ragner:
Hi @NolaDonato,
that looks nice but have just one question about the change between scenes.
Considering my pull request,
that singleton was pausing the audio sources from SceneA and resuming the audio sources from SceneB when the application changes the main scene from SceneA->SceneB.
At your pull request I have not found any logic about that.
Are you considering that the application should handle the audio sources on changes between scenes.
the cardboard-audio demos is too simple but I have tested my code with a application with multiple scenes. Maybe we should change the cardboard-demo to have two scenes so we
could test changes between scenes.

crbozz:
We considered the scenario when there are more than one scene for the same application and those scenes will be switched several times. This was why we attached GVRAudioListener to a scene and also wrote AudioEngineEventsHandler class. This approach would save some code lines for the application developer since s/he would not be responsible for calling pauses and resumes when switching scenes.

NolaDonato:
When you switch scenes, all of the scene objects in the old scene, including the components, are removed. When the owner of the audio source component goes away, onDetach is called and that component is removed from the audio manager.

When you attach audio sources to the objects in the objects in the new scene, you also explicitly add them to the audio manager so the app can control when this actually occurs. I content the audio manager does not need to know anything about when scenes switch. Do you have a sample app that tests switching of scenes? I would like to try it with my pull request against your branch.

crbozz:
I uploaded a sample at my github: https://github.com/crbozz/gvrf_multiscene
This sample contains four scenes, three of them with audio sources, that are continously switched as main scene. It uses ResonanceAudio binaries built from our original pull request, also included in the repository.

NolaDonato:
I modified the multiscene demo to use my "audiopr" branch and it seems like it handles scene switching properly. I made a pull request against gvrf_multiscene with my changes.

ragner:
Hi @NolaDonato,
looks good to me.
We can make further improvements as needed.

ragner:
Hi @NolaDonato,
I "rebased" your commits  on my branch.
Thank you.

NolaDonato:
All I did was mimic the way the other "managers" wor (like GVREventManager, CursorManager, etc.) They are global for the most part. The tricky bit was to emit the "onInit" event on every scene change. Previously, we just did it on the first one making it impossible to tell when the scene changes.

I am trying to figure out a good way to emit an event on the Java side when a scene object becomes part of a scene. Currently we are not keeping references to the Java GVRSceneObject on the C++ side. That would be necessary to implement this feature.

liaxim:
The semantics of onInit is different. It is about app initialization and not scene initialization.

liaxim:
I am tempted to suggest we remove the support for scene changing from the next release instead of struggling how to support it. I see no value in supporting multiple scenes but it sure does raise the complexity for us. We've been thinking about this for some time..

NolaDonato:
Are you saying that I should not call onInit on every scene change? Does this break stuff? I think we need this capability but the way I did it is probably wrong.

liaxim:
It possibly could break stuff since now it has different semantics. If we do want to keep supporting the scene switching, we could emit a different event.

ragner:
I understand @liaxim  worries about that changes but
In my opinion make initialization stuff of the application at onInit event of the scene does not look good to me :( because any simple application can have multiple scene like "menu scene", "configuration scene", "level scene" for games, etc. So that onInit event of the scene looks like it is a event for scene's initialization, not for application. 

In my opinion we could create something new to application events like 
IContextEvents {
     onInit(context);
    onMainSceneChanged(oldScene, newScene);
}

So, every logic that needs handle scenes changes could listen to onMainSceneChanged(..).

liaxim:
onMainSceneChanged sounds good to me

NolaDonato:
I pushed a change to replace the onInit logic with IContextEvents. Please test with the application that uses multiple scenes. I could not find this application. The one that is in gvr-cardboardaudio does not switch scenes.

GVRf/Extensions/ResonanceAudio/resonanceaudio/src/main/java/org/gearvrf/resonanceaudio/GVRAudioManager.java
```diff
@@ -0,0 +1,240 @@
...
+ * Manages audio sources for the applications.
+ * Handles pausing and resuming sound and maintains
+ * a list of {@link GVRAudioSource} objects which are attached
+ * to scene objects. The position of each audio source is
+ * obtained from the {@link GVRTransform} of the {@GVRSceneObject}
+ * the audio source is attached to. The audio manager also
+ * tracks the head transform so the sound will be spatially
+ * correct.
+ */
+public class GVRAudioManager extends GVREventListeners.ActivityEvents
```
liaxim: Extending ActivityEvents and implementing GVRDrawFrameListener pollutes the api space with unnecessary methods. Could as easily be implemented as inner classes.

liaxim:
Looks pretty good. Thank you. Planning to merge on Wednesday. There is one bug that I will fix afterwards if it is not dealt with here.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRViewManager.java
```diff
@@ -355,9 +361,10 @@ public void run() {
                     }
 
                     // Trigger event "onAfterInit" for post-processing of scene
-                    // graph after initialization.
+                    // graph after initialization. Also trigger "onInit" for the GVRContext.
                     getEventManager().sendEvent(mMain, IScriptEvents.class,
                             "onAfterInit");
+                    getEventManager().sendEvent(this, IContextEvents.class, "onInit", this);
```
liaxim: Should be replaced with ``getEventManager().sendEvent(GVRViewManager.this, IContextEvents.class, "onInit", GVRViewManager.this)``. ``this`` refers to the Runnable instance.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRScene.java
```diff
@@ -145,6 +145,7 @@ public void removeAllSceneObjects() {
 
         rig.removeAllChildren();
         for (GVRSceneObject child : mSceneRoot.getChildren()) {
+            child.detachAllComponents();
```
liaxim: Why did you need to add this line?

I believe because of it the tests started crashing left and right. It is a framework problem and not an issue with your change. Currently nothing can be done other than sprinkling nullptr checks everywhere.

NolaDonato:
I added it so that onDetach would be called for all the components. I can take it out and I believe ResonanceAudio will still work. The application will be responsible for clearing the GVRAudioManager sources because it won't happen automatically when the scene is cleared. I will remove the line and test again.

liaxim:
It is not a problem by itself. Just exposes a bunch of other problems in the framework. I say we keep the line and fix the real issues. The tests clear the scene during tear down and there are components without parents suddenly. A lot of code doesn't expect that.

NolaDonato:
I removed it for now. We can easily put it back and investigate. GVRScene.clear() has always been a problem. Maybe the tests shouldn't clear during tear down?

liaxim:
They don't have to but it is a good test. We may never come back to this if we work around the problem.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRScene.java
liaxim: @ragner I am about to remove the detachAllComponents from removeAllSceneObjects. Let me know if you have concerns. If you are to reuse scene objects after removeAllSceneObject detachAllComponents would break the reused scene objects.

crbozz: I believe audio will continue playing after removing scene objects. We can fix that after you finish this change.

NolaDonato:
I think lights might also remain. I will check that.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRScene.java
liaxim: Unfortunately there is no good general and lightweight mechanism to know when a scene object starts and stops being rendered. While something could be added to removeSceneObject on the Java side to let scene objects know they are no longer parented to the main scene root it won't be very lightweight. Nor complete - since addSceneObject then should be producing similar notification. And then the main scene could change too. On the renderer side there is a bit more flexibility but it has its own complexity. And making a ton of JNI callbacks to Java might not be the best idea from the render thread. Nola may have better ideas.

I think for now GVRAudioSource's javadoc needs to mention that it is up to the user to stop the audio playback whenever appropriate.

NolaDonato: There is a notification on the C++ side. Component::onAddedToScene and Component::onRemovedFromScene are called. I did not know how to notify Java. I think we need a way to communicate events from C++ to Java asynchronously.

PR #1693, opened 2018-01-07
latest daydream

pnemonic78:
daydream version 1.120.0

### DIFF
--- a/GVRf/Framework/backend_daydream/build.gradle
+++ b/GVRf/Framework/backend_daydream/build.gradle
@@ -58,7 +58,7 @@ android {
 }
 
 project.ext.jomlVersion = "1.9.3-SNAPSHOT"
-project.ext.daydreamVersion = '1.101.0'
+project.ext.daydreamVersion = '1.120.0'
 
 dependencies {
     compile fileTree(include: ['*.jar'], dir: 'libs')


liaxim:
Looks good. Tested with few demos. Apps that could use a controller if configured with Daydream fail with
```
01-18 07:02:37.418 12764 12806 W System.err: java.lang.UnsupportedOperationException: Cannot select controllers if none are enabled in gvr.xml settings
01-18 07:02:37.419 12764 12806 W System.err: 	at org.gearvrf.io.GVRInputManager.selectController(GVRInputManager.java:175)
01-18 07:02:37.419 12764 12806 W System.err: 	at org.gearvrf.immersivepedia.Main.onInit(Main.java:84)
01-18 07:02:37.419 12764 12806 W System.err: 	at java.lang.reflect.Method.invoke(Native Method)
01-18 07:02:37.419 12764 12806 W System.err: 	at org.gearvrf.GVREventManager.invokeMethod(GVREventManager.java:307)
01-18 07:02:37.420 12764 12806 W System.err: 	at org.gearvrf.GVREventManager.sendEventWithMaskParamArray(GVREventManager.java:126)
01-18 07:02:37.420 12764 12806 W System.err: 	at org.gearvrf.GVREventManager.sendEventWithMask(GVREventManager.java:113)
01-18 07:02:37.422 12764 12806 W System.err: 	at org.gearvrf.GVREventManager.sendEvent(GVREventManager.java:108)
01-18 07:02:37.423 12764 12806 W System.err: 	at org.gearvrf.GVRViewManager$1$1.run(GVRViewManager.java:333)
01-18 07:02:37.423 12764 12806 W System.err: 	at android.os.Handler.handleCallback(Handler.java:751)
01-18 07:02:37.423 12764 12806 W System.err: 	at android.os.Handler.dispatchMessage(Handler.java:95)
01-18 07:02:37.423 12764 12806 W System.err: 	at android.os.Looper.loop(Looper.java:154)
01-18 07:02:37.423 12764 12806 W System.err: 	at android.os.HandlerThread.run(HandlerThread.java:61)
```

The exception is not caused by this change. Planning to merge on Friday.

NolaDonato:
My controller reform pull requests fix this. I changed the demos to make the gvr.xml file consistent with selectController calls. You should not call selectController unless you have specified controller is gvr.xml. I could make it default to the Google controller automatically instead. Should this be the correct behavior?

liaxim:
I would say let's not use a controller unless explicitly requested by the app.

NolaDonato:
Tom would like us to default to using the controller if they specify nothing in gvr.xml so it will always work on Daydream. If they specify useControllerTypes="" this means no controllers and the app will handle everything. In this case they will get the exception if the app calls selectController. I have made this change.

PR #1696, opened 2018-01-08
framework: silence the verbose "DataDescriptor: %s" trace by default

liaxim:
Stuff like ``12-29 11:52:01.229 30675 30751 V gvrf : DataDescriptor: mat4 u_bone_matrix[60]`` will disappear from the logs.

---

### DIFF
--- a/GVRf/Framework/framework/src/main/jni/objects/data_descriptor.cpp
+++ b/GVRf/Framework/framework/src/main/jni/objects/data_descriptor.cpp
@@ -12,13 +12,13 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#define VERBOSE_LOGGING 0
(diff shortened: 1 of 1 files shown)


ragner:
👍

NolaDonato:
If you see that particular message in the logs it means it is trying to create a shader. Seeing it per frame implies a bug since we should not create shaders every frame.

liaxim:
@NolaDonato It looks to me like render data's dirty flag after once being set never gets cleared. This can and does lead to continuous shader regeneration. I posted a question to you in https://github.com/Samsung/GearVRf/pull/1654 which removes some code that is most like necessary.

liaxim:
@ragner gave thumbs up. Enough for such a tiny change. Merging.

PR #1697, opened 2018-01-09
Adding call to get value of system property

apjagdale:
GearVRf-DCO-1.0-Signed-off-by: Abhijit Jagdale
a.jagdale@samsung.com


liaxim:
Looks good but needs DCO.

PR #1662, opened 2017-12-11
backend_oculus: leave vrapi initialized after onPause

liaxim:
Enables taking screenshots and recording videos from the Oculus global menu
Addresses https://github.com/Samsung/GearVRf/issues/1235

---
GearVRf-DCO-1.0-Signed-off-by: Mihail Marinov <m.marinov@samsung.com>

PR #1704, opened 2018-01-18
backend_oculus: fix for vrapi usage so tests run again without crashes

liaxim:
Broke the automated tests in https://github.com/Samsung/GearVRf/pull/1662. Fixing it up here. VrApi is not really friendly to multiple activities per process but this is exactly the case with our tests.

Pending further verification. But tests do run now.

---


liaxim:
Tested with the demos and the automated tests. Planning to merge on Friday.

liaxim:
It is kind of a hack. The most recently created activity becomes the vrapi "owner" and initializes vrapi with its activity instance. If another activity gets subsequently created it takes over the vrapi by uninitializing and initializing it again. If older activity resumes then it checks the "owner" and if it is not the owner - takes over vrapi. I am unable to find another solution.

Uninitializing for the "owner" activity matters since otherwise there would be some leaks reported in the logs and possibly exceptions at a later time.
Initializing needs the correct activity instance; if initialized with older activity later there will be jni crashes due to the activity getting deleted.

PR #1709, opened 2018-01-23
controller: fix a problem due double recycle and a concurrent modification exception

liaxim:
Fixes some of the crashes with Daydream after merging https://github.com/Samsung/GearVRf/pull/1646

Test case: https://github.com/gearvrf/GearVRf-Tests/pull/231

---

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/io/GearCursorController.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/io/GearCursorController.java
@@ -16,6 +16,7 @@
 
 package org.gearvrf.io;
 
+import android.app.Activity;
 import android.graphics.PointF;
 import android.os.Handler;
 import android.os.HandlerThread;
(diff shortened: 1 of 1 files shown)


liaxim:
Previously the same instance of the Array was being iterated over by one thread and modified by the other. The primitive array is not going to throw a ConcurrentModifExc but now that I look at it, it needs more work.

liaxim:
Updated to use ConcurrentLinkedQueue in SendEvents. ConcurrentLinkedQueue could be overkill for our one producer/one consumer case. Apparently it is meant for much heavier concurrency but I find the performance good.

liaxim:
Added test case at https://github.com/gearvrf/GearVRf-Tests/pull/231

liaxim:
@NolaDonato gvr-events and Daydream controller; I really think there is something with the picking||collider. If you enter the quad from the top, GVRViewSO doesn't get onEnter until you reach the listview. Then the reticle jumps and onEnter happens.

liaxim:
Same thing happens on S8 except that the reticle jumps at an earlier point.

liaxim:

On S8 occasionally I still see a double-recycled MotionEvent exception. No idea how it happens yet. Will address subsequently.

PR #1711, opened 2018-01-25
fixed issue with x3d sphere diffuse color

NolaDonato:
The sphere was using the texture shader instead of the phong shader.
- added a constructor to GVRSphereSceneObject
- modify X3D parser to use Phong shader for sphere


NolaDonato:
wrong branch

PR #1712, opened 2018-01-25
fixed issue with x3d sphere diffuse color

NolaDonato:
The sphere was using the texture shader instead of the phong shader.

added a constructor to GVRSphereSceneObject
modify X3D parser to use Phong shader for sphere


mwitchwilliams:
Why is this only a change in GVRSphere when it effects the other primitives (as shown in the demo I produced)?
Does this match what other X3D implementations on the web such as X3Dom?
What is the effect if an emmissive color is also set? Or if a light is in the scene?

NolaDonato:
The phong shader has exactly the same behavior on a sphere as on any other scene object. I changed the sphere to fix the issue with your file. If other primitives have similar issues, we should verify they are using the correct shader. As to whether our phong shader implementation matches the X3Dom, I cannot say. If it matches for regular meshes then it will now also match for the sphere.

PR #1713, opened 2018-01-25
Optimize GL texture parameter setting

NolaDonato:
Do not update the GL texture parameters unless they have changed.


liaxim:
Looks good. Planning to merge on Friday.

PR #1714, opened 2018-01-25
Remove render-texture dependency from the backend

roshanch:
GearVRF DCO signed off by: Roshan Chaudhari roshan.c@samsung.com


liaxim:
Need rebase

GVRf/Framework/backend_oculus/src/main/jni/ovr_activity.cpp
```diff
@@ -112,19 +111,19 @@ namespace gvr {
...
 
-    RenderTextureInfo renderTextureInfo;
-    renderTextureInfo.fboId = fbo.getRenderBufferFBOId(index);
-    renderTextureInfo.fboHeight = fbo.getHeight();
-    renderTextureInfo.fdboWidth = fbo.getWidth();
-    renderTextureInfo.multisamples = mMultisamplesConfiguration;
-    renderTextureInfo.useMultiview = use_multiview;
-    renderTextureInfo.views = use_multiview ? 2 : 1;
-    renderTextureInfo.texId = fbo.getColorTexId(index);
+    dummy::RenderTextureInfo* renderTextureInfo = new dummy::RenderTextureInfo();
```
liaxim: In this particular backend, there is already dependency to the framework. Why not just use directly gvr::RenderTextureInfo?

liaxim:
Planning to merge on Thursday

roshanch:
I think we should wait as GVRRendleBundle needs refactoring and there is dependency on backends.

roshanch:
I removed IRendleBundle interface so we can have backend specific RendleBundle.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRViewManager.java
```diff
@@ -197,11 +197,13 @@ void onSurfaceCreated() {
         }
 
         mPreviousTimeNanos = GVRTime.getCurrentTime();
+
+
+        final GVRScene scene = null == mMainScene ? new GVRScene(GVRViewManager.this) : mMainScene;
+        setMainSceneImpl(scene);
```
liaxim: Why did this section had to move up?

roshanch: because makerendleBundle() has dependency for the mainscene. We create renderTarget in there which needs scene.

liaxim:
Looks good to me. Can be merged when the dependencies are ready.

liaxim:
Planning to merge on Tuesday

Issue #1717, opened 2018-01-26
Shader type is always invalid


ragner:
What is the right way to identify the shader type of a materail?
The following code is always logging `INVALID SHADER`

```
        GVRMaterial material = new GVRMaterial(context, GVRMaterial.GVRShaderType.Phong.ID);

        if (material.getShaderType() != GVRMaterial.GVRShaderType.Phong.ID) {
            Log.w("SHADER ERR", "INVALID SHADER: " + material.getShaderType());
        }
```

bohuang3d:
Looks like in the current version, the shaderId supplied to GVRMaterial constructor is no longer simply assigned but gets sent GVRShaderData's contructor and to GVRShaderManager
https://github.com/Samsung/GearVRf/blob/6c3fd322deb7c6c2ff333b4ef992e78b0b932fa0/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRShaderData.java

If it's copied during the process, then that explains what you observe.

SteveGreatApe:
I wanted to create a function to verify a scene object had the expected shader type, but couldn't work out a way to do this without using reflection. As this was just a debug feature I didn't think it worth raising as an issue, but if yours is a legitimate use case then the ID field of GVRShaderId will need to made publicly accessible, or a utility function within GVRShaderId added to do the check for you.

This is what I came up with to do the compare.
```
    public static void assertEquals(GVRShaderId shader1, GVRShaderId shader2) {
        try {
            Field idField = GVRShaderId.class.getDeclaredField("ID");
            idField.setAccessible(true);
            Object id1 = idField.get(shader1);
            Object id2 = idField.get(shader2);
            if (!id1.equals(id2)) {
                String error = "Error wrong shader type, expecting: " + id1 + ", Found: " + id2;
                Log.e(TAG, error);
                throw new RuntimeException(error);
            }
        } catch (NoSuchFieldException | IllegalAccessException e) {
            e.printStackTrace();
        }
    }
```

PR #1720, opened 2018-01-31
Fix thread-safety problem with dragging

NolaDonato:
Made an internal Dragger class which handles the mechanics of dragging an object with the cursor controller.,
This should fix issue #1717


NolaDonato:
This is superseded by other fixes already in master.

PR #1721, opened 2018-01-31
Use uniform block for light sources

NolaDonato:
Modified the GL renderer to store lights in a single uniform block instead of directly setting individual light properties.


NolaDonato:
This PR still has issues running consecutive shadow mapping tests. I am putting it up so Mihail can help me find out why.

liaxim:
Ok. I ran ShadowTests with this pull request on few phones (s8, note8, note5 and s7). There were no crashes for me. On what phone do you run the tests? What is the crash you see?

NolaDonato:
Really? I am running on S8. I will do git clean -fdx, invalidate caches and try again on S8 and Note 8 too.

NolaDonato:
Sometimes it crashes with a seg fault and sometimes it messes up the shader and it crashes during shader compilation. And many times it just doesn't display the shadow even though it goes thru all the code with no errors.

NolaDonato:
After making the depth shader thread safe, the tests no longer crash. I still get intermittent results - some tests do not show shadows. They are not reproducible, the tests that fail are different every run.

liaxim:
Is this still work in progress?

liaxim:
ShadowTests passed fine for me too.

NolaDonato:
Last night I fixed a major problem with it so the shadow tests pass. The lighting tests have subtle differences that I don't understand. I will investigate more today.

NolaDonato:
I fixed a minor issue involving removing lights. Now most of the asset tests run. We have explained the differences in the light tests as precision issues due to compilation differences in the shader.

liaxim:
gvr-multilight renders 10 fps lower for me on s8. Also toggling the lights didn't work.

liaxim:
```
gvr-multilight, s8, master      - 59-60 fps
gvr-multilight, s8, lightbuffer - 49-51 fps
```

Can you confirm that? I might be doing something wrong.

NolaDonato:
I do not think you are doing anything wrong. I get 45-46 fps on the S7.  I am still messing with GL calls to get the uniform buffer to load faster. We may have to abandon this approach for GL - not sure yet. We have to do it for Vulkan because there is no other way.

NolaDonato:
?

I think this might be due to drawing the right eye twice. When I tried lightbuffer branch with Sushant's change, I get 60 fps for both cases (lights in a uniform buffer vs direct GL calls to update light uniforms).


roshanch:
may be that sample is not giving enough work to gpu.. but yeah i think if we are updating lights per frame then using direct GL calls would give more performaance.

NolaDonato:
I think we should measure the performance of lighting under a variety of circumstances to determine when GL calls are better than uniform blocks. I maintain that if the lights don't move, uniform blocks might be better because it is one GL calls to bind the buffer as opposed to lots of GL calls to update the light uniforms for each object.

liaxim:
The LightTests could crash if run as part of the whole test suite due to what seems like nullptr dereference in LightList::updateLightBlock. We will deal with this separately as we have some other occasional crashes in the tests like that and has to do something with the object cleanup.

GVRf/Framework/framework/src/main/jni/objects/components/render_data.h
```diff
@@ -42,19 +42,6 @@ class TextureCapturer;
 class RenderPass;
 struct RenderState;
 
-template<typename T>
-std::string to_string(T value) {
```
liaxim: Not sure how it built for me first-hand but now it doesn't anymore because glslang/Include/Common.h cannot be found. Since std::to_string is not available through the standard <string> header on Android for some reason, we should rather use our own version of to_string instead of something else non-standard.

liaxim: Probably one more reason to move to clang and libc++. Other than GCC going to be removed from NDK v18.

liaxim:
Double-checked but gvr-simplephysics and gvr-shadows produce black screen and then crash.

NolaDonato:
fixed crashes in physics and shadows

GVRf/Framework/framework/src/main/jni/objects/lightlist.cpp
```diff
@@ -0,0 +1,372 @@
...
+        const std::vector<Light*>& lights = it->second;
+        for (auto it2 = lights.begin(); it2 != lights.end(); ++it2)
+        {
+            Light *light = (*it2);
+            light->setBlockOffset(numFloats);
+            numFloats += light->getTotalSize() / sizeof(float);
+        }
+    }
+    if ((mLightBlock == NULL) ||
+        (numFloats > mLightBlock->getTotalSize()))
```
apjagdale: mLightBlock->getTotalSize() should be divided by sizeof(float)

roshanch:
can we have design overview of this PR

GVRf/Framework/framework/src/main/jni/objects/lightlist.cpp
```diff
@@ -0,0 +1,372 @@
...
+
+ShadowMap* LightList::updateLightBlock(Renderer* renderer)
+{
+    std::lock_guard < std::recursive_mutex > lock(mLock);
+    bool dirty = mDirty != 0;
+    ShadowMap* shadowMap = NULL;
+
+    if (mDirty & REBUILD_SHADERS)
+    {
+        return scanLights();
```
apjagdale: This function should not return from here. There are cases when we add a light, both the flags REBUILD_SHADERS and LIGHT_ADDED are set.

liaxim:
@roshanch So what was the result of the design overview? Are further updates to this PR expected?

GVRf/Framework/framework/src/main/jni/objects/lightlist.cpp
```diff
@@ -0,0 +1,372 @@
...
+}
+
+/*
+ * Removes an existing light from the scene.
+ * Return true if light was removed, false if light was not in the scene.
+ */
+bool LightList::removeLight(Light* light)
+{
+    std::lock_guard < std::recursive_mutex > lock(mLock);
+
```
roshanch: why do we need recursive lock here?

roshanch:
@liaxim code seems okay to me except recursive lock.

GVRf/Framework/framework/src/main/jni/objects/lightlist.cpp
NolaDonato: Because we have a situation where it hangs if I don't use a recursive lock. It has something to do with clear being called in an inconvenient place between tests. If you take the recursive lock out and run the lighting and shadow tests, you will find the hang (unless something else fixed it).

GVRf/Framework/framework/src/main/jni/objects/lightlist.cpp
liaxim: @NolaDonato ?

liaxim:
Demos work. Tests are ok - no crashes but new golden masters have to created for some. VideoTests tests tend to hang but it is not related to this change.

All that is left is the answer to @apjagdale 's comment about mLightBlock->getTotalSize(); it seems like a valid point to me.

GVRf/Framework/framework/src/main/jni/objects/lightlist.cpp
NolaDonato: Yes you are correct, total size is the number of bytes

liaxim:
Needs rebase.

liaxim:
Please rebase

NolaDonato:
I have rebased this and tested it with gvr-multilight and lighting tests.

liaxim:
Still doesn't build without patching for me due to the to_string change. Will be applying the patch after merging.

liaxim:
I see "UniformBlock: ERROR: no uniform block allocated" because of the DepthShader. What should be done about this?

apjagdale:
Since GVRDepthShader does not have a material we can add a check to the LOGE if descriptor string is null, do not print the error message.

PR #1527, opened 2017-10-11
X3D v4 fix for DEF/USE and  Level-of-Detail fixes

mwitchwilliams:
Fix for X3D &lt;IndexedFaceSet&gt; and &lt;ImageTexture&gt; DEF-USE [_where a 3d mesh or texture map can be reused in the same scene_].
Also integrated some Level-of-Detail &lt;LOD&gt; corrections from maint_v3.x

LOD + TouchSensor no longer works.  Waiting for other changes from maint-v3.x to be merged into this branch, such as GVRBaseSensor.java.  See https://github.com/Samsung/GearVRf/pull/1455 for more info.

GearVRF DCO signed off by: Mitch Williams
m1.williams@partner.samsung.com

PR #1425, opened 2017-08-31
Fixes to X3D LOD to support child objects

mwitchwilliams:
Added support for X3D's &lt;LOD&gt; (level of detail) node to enable child and sibling 3d meshes using &lt;Transform&gt; and &lt;Group&gt; nodes.
Fixes a bug identified by T.Flynn.

There is also a pull request of tests and documentation in the GVRf Tests repository at: 
https://github.com/gearvrf/GearVRf-Tests/pull/132

GearVRf-DCO-1.0-Signed-off-by: Mitch Williams
m1.williams@partner.samsung.com

PR #1298, opened 2017-07-18
X3D Inline subdir and render order

mwitchwilliams:
2 fixes here:
- Corrected X3D &lt;Inline&gt; finding .x3d files and texture maps in sub-directories within Android Studio's asset folder.  This addressed Issue #1261, https://github.com/Samsung/GearVRf/issues/1261
- Restored line inside X3D parser 
`gvrRenderData.setAlphaToCoverage(true);`
to correct a rendering order issue involving transparencies.

GearVRf-DCO-1.0-Signed-off-by: Mitch Williams
m1.williams@partner.samsung.com

PR #1724, opened 2018-02-05
x3d: make x3d its own extension

liaxim:
Makes X3D an extension; the only notable part is the addition of the X3DLoader class and its use by the AssetLoader.

Related change: https://github.com/gearvrf/GearVRf-Demos/pull/581

---


mwitchwilliams:
Won't impact any of my current development (on JavaScript interactivity for Text, Movies).

Can't come up with any issues. A few Pull Requests for maint_v3.x and not have been merged into Master v4.x (https://github.com/Samsung/GearVRf/pull/1298, https://github.com/Samsung/GearVRf/pull/1527, https://github.com/Samsung/GearVRf/pull/1425). I anticipated fixing these in Master anyway if they didn't get into v4.x. Easy fixes.

As long as it rebuilds, seems fine to me. May start using this new location as early as this week.

liaxim:
@NolaDonato @thomasflynn Any objections? With the upcoming exoplayer dependency this change will help keep the core framework in check.

NolaDonato:
The asset loader provides a number of ways to load an X3D model. You can put a scene object in the scene and then tell it to load under that scene object in a separate thread and notify you when it is done. You can ask it to load an X3D model and give you the root scene object so you can add it later yourself. I think some of our internal applications use these features.

I think it would be better if the asset loader could be provided with the X3D loader by the user if they want to use the extension. Then we keep all the capabilities but X3D can still be an extension. If you try to load something with ".x3d" and the asset loader doesn't have an X3D loader it can give an exception.

liaxim:
It all boils down to loadX3DModel which has been dealt with. If the extension is not present, exception is thrown. The user provides the loader by including the extension in her build. If the x3d extension is not present the only capability lost is the capability to load x3d files. How does that sound?

NolaDonato:
It sounds better than what I proposed. I didn't know how to check for the presence of an extension.

liaxim:
If there are no objections, planning to merge this on Tuesday.

PR #1726, opened 2018-02-05
arm64: fix jni error; now 64bit build runs

liaxim:
1. fix jni error crashing the 64bit build; armeabi-v7a apparently is not so strict but was still incorrect
2. remove bad call to cameraRig_->setRotation(glm::quat()) and fix onDrawFrame formatting

---


liaxim:

PR #1728, opened 2018-02-06
framework: delete local ref as soon as we are done

liaxim:
Roshan is seeing local ref table overflow in some cases. This change ensures we are deleting them as soon as possible.

---


NolaDonato:
I will look around to see if there are other places we do this.

liaxim:

PR #1729, opened 2018-02-07
gradle: make sure Messages.properties is included in the final apk

liaxim:
gvr-javascript if built with the aars (and presumably from maven) crashes due to missing Messages.properties. The file is needed by the mozilla js interpreter. Apparently something changed after the gradle update in December.

Tried to find a better solution but nothing else worked.

---


liaxim:

PR #1730, opened 2018-02-07
Silence per frame log trace from gl_vertex_buffer

liaxim:
Fix compiler warning due to a C-cast

---


liaxim:
Planning to merge on Monday

Issue #1734, opened 2018-02-09
Inward facing cylinder wall has normals pointing the wrong way


SteveGreatApe:
I'm using a GVRCylinderSceneObject with FacingOut set to false to use as a circular room. Inside the cylinder I've placed a GVRPointLight to light the walls. I find if I set the light to mainly diffuse then the walls are very poorly lit.

If I run this function to reverse the normals then walls are lit up properly.

```
    public static void mirrorNormals(GVRSceneObject sceneObject) {
        GVRMesh mesh = sceneObject.getRenderData().getMesh();
        float[] norms = mesh.getNormals();
        for(int index = 0; index < norms.length; index ++) {
            norms[index] = -norms[index];
        }
        mesh.setNormals(norms);
    }
```

bohuang3d:
Which constructor are you using? The one with signature

    public GVRCylinderSceneObject(GVRContext gvrContext, boolean facingOut,
            ArrayList<GVRTexture> textureList) 

is not assigning its facingOut parameter to anything

bohuang3d:
Add
params.FacingOut = facingOut;
to after the line
CylinderParams params = new CylinderParams();

to the definition of this constructor should fix

SteveGreatApe:
I'm using this signature:

`public GVRCylinderSceneObject(GVRContext gvrContext, GVRCylinderSceneObject.CylinderParams params) `

The FacingOut field is working okay with this function except for the normals. It looks like what you've found is a different bug.

bohuang3d:
ok. In 
private void createBody(CylinderParams params)

the normals are actually not conditionally generated. We should add something like

if (params.FacingOut) {
nx0 = -nx0;
nx1 = -nx1;
...
...
//so on and so forth before copying to the normal buffer
}

mwitchwilliams:
I ran a similar test using X3D which calls the same Cylinder object and Cylinder Params.

It looks like the Normals inside the Cylinder are not flipping, except on the top and bottom of the Cylinder.
The left Cylinder has a texture map (same texture applied to a plane in the background) and a blue Cylinder on the right.

I attached  screen shot as well as the X3D file plus texture map.

mitch

Mitch Williams
Samsung GearVR X3D development
310-809-4836 (outside line)

mwitchwilliams:
Just another screen shot.
I rotated both cylinders to show the texture mapping on the inner top of Cylinder, but not on the sides.


Mitch Williams
Samsung GearVR X3D development
[...]

PR #1736, opened 2018-02-10
Fix normal generation for cylinder

NolaDonato:
The cylinder code did not flip the normals if the "facing in" option was requested. It also had the number of texture coordinates wrong.

This patch fixes issue #1734, GearVRf-Tests PR #238 has a test for it.


NolaDonato:
replaced by #1741

PR #1737, opened 2018-02-10
framework: get rid of the reference to the java scene object that the native code maintains

liaxim:
Not pretty but does remove the need to keep constantly a reference to the java object.

RenderData presents a similar reference problem that will take a different solution.

---


liaxim:
GVRRenderData now uses an intermediate object that holds a weak reference to the GVRRenderData instance. The C++ code keeps a global reference to this intermediate object. The bindShader call goes through this intermediate object. Have verified that both GVRRenderData and intermediate objects are getting gc-ed.

It would be nice to keep the JNI stuff in a JNI layer but with this bindShader requirement it is not that easy any more.

liaxim:
The OomTests used to crash every other time for me but do not crash with this change in.

liaxim:
JNI error Nola mentioned fixed. But now a lot of new crashes in the tests due to the detachAllComponents from Scene.clear (nothing wrong with doing that but..) added by the resonanceaudio pull request. Fixed some but null ptr checks everywhere is no solution. They have to propagate to a lot of places and what crashed I fixed but I am sure that is not all. A Component maybe shouldn't have to reference the scene object and have an owner concept.

NolaDonato:
I have run the tests with and without this PR. They crash sporadically without it.  Let's merge tomorrow if no more comments.

PR #1740, opened 2018-02-12
Fix regression in X3D touch sensor

NolaDonato:
The code to attach it was incorrectly commented out (my bad)


thomasflynn:
fixes the problem i was seeing with x3d sensors

PR #1741, opened 2018-02-13
Fix normal generation for cylinder

NolaDonato:
The cylinder code did not flip the normals if the "facing in" option was requested. It also had the number of texture coordinates wrong.


liaxim:

PR #1744, opened 2018-02-14
Fixed X3D emission & asset load exceptions

NolaDonato:
Recent changes to the asset loader produce the error message "X3D extension no available" when a model fails to load. This patch now produces that message only if the X3D extension is not available and a message with "FileNotFound" if the file cannot be loaded

Changed the Phong shader to use the emissive color instead of the diffuse color (if it is nonzero) when there is no lighting. This matches the X3D spec.


liaxim:
Looks good. Thank you for fixing the x3d loader I didn't properly test.

@mwitchwilliams Please verify the emissive color part.

NolaDonato:
replaced by #1752. Emission will be fixed with separate X3D shader.

PR #1745, opened 2018-02-14
framework: fix the way the IContextEvents.onInit is being emitted

liaxim:


liaxim:
Merging since there is a crash otherwise.

PR #1752, opened 2018-02-15
Fix X3D error messages

NolaDonato:
Recent changes to the asset loader produce the error message "X3D extension no available" when a model fails to load. This patch now produces that message only if the X3D extension is not available and a message with "FileNotFound" if the file cannot be loaded


thomasflynn:
i get a compile error about:

import org.gearvrf.GVRCursorController

not being found.

thomasflynn:
wait, sorry, that was a bug in my app.  it hadn't been updated since the merge of the API changes this morning.

thomasflynn:
ok, this looks good.

PR #1753, opened 2018-02-16
framework: workaround for RenderShaderTests crash

liaxim:
Fixes the crash caused by RenderShaderTests. The tearDown for the tests clears the scene, which detaches all components. It all happens on the test thread. Since there is no way for c++ code to claim ownership of a component, transforms (for example) end up getting deleted concurrently to the tearDown execution.

---


liaxim:
Don't see crashes anymore. The VideoTests tend to hang at the end of the run. Something is probably leaked since I see a lot of undertaker and eventhandler threads lying around when the videotests hang.

liaxim:
Fixed the biggest leak by explicitly stopping the GearCursorController's event thread. Also fixed a rare vrapi-related crash.

liaxim:
Planning to merge on Monday. Definitely makes the automated tests much more stable. There is a change to removeAllSceneObjects that some may find questionable.

PR #1754, opened 2018-02-17
Fixed bounding volume performance regression

NolaDonato:
This should fix the performance problem with x3d-photo


thomasflynn:
tested with my app.  with this change, runs at ~50FPS.  With dirtyHierarchialBoundingVolume() in place of the transform_dirty_ flag, I get a solid 60FPS.

NolaDonato:
I am calling this function more but when I look at the code it seems redundant. I will study this more to figure out why but for now this should fix the performance problem.

thomasflynn:
ok, with this latest commit, my app now runs at 60.

GVRf/Framework/framework/src/main/jni/objects/scene_object.cpp
```diff
@@ -467,7 +473,7 @@ bool SceneObject::intersectsBoundingVolume(SceneObject *scene_object) {
 
 
 BoundingVolume& SceneObject::getBoundingVolume() {
-    if (!bounding_volume_dirty_ && !transform_dirty_) {
```
liaxim: @NolaDonato Any memory of why you had to add the check in first place? How to verify that whatever you fixed is still working correctly?

```
commit 9f3f152be4d9e4a9b37beb2e9da5a57673f33f8a
Author: Nola Donato <nola.donato@samsung.com>
Date:   Mon Jan 23 23:50:53 2017 -0800

    fixed problem with bounding box computation
    
    which caused the cursor to be incorrectly culled in some of the demos
```

NolaDonato:
I did not add the transform_dirty flag - in the code it is only used by batching. As far as I can tell, it is never cleared. That was causing the bounding volume to be computed every frame. The code that was there was wrong to begin with.

I will remove the extra functions I added. They were part of a fix that didn't work.

I still do not understand why dirtyHierarchicalBounds() has to be called for every transform change.  That is what fixed the performance problem but it would seem that you would only need to call it for the object whose transform changes - not all of its children. When I tried that, it did not work.

PR #1766, opened 2018-02-26
Move shader manager from RenderBundle to context

NolaDonato:
The shader manager should not be part of the RenderBundle because that gets deleted on pause and recreated on resume. But scene object remain across pause/resume and they reference shaders. Now the shader manager is part of the context.


NolaDonato:
This change is not necessary

Issue #1777, opened 2018-03-09
Crash when detach the device from Gear when the controller is connected 


felipecs81:
Hi Guys,

I am working on a project for Gear VR and I am using version 4.0.1-SNAPSHOT.
So far so good, I am able to connect the control and perform all operations, but when I disconnect the device from Gear, an exception is thrown into an internal class.
This only occurs when you have a control connected, if I am using only the headset this error does not happens. I have already tried to disable or invalidate the control on the onDestroy method of activity but nothing worked.
This also happens in this module of the GearVRDemo Project. gvr-controller
Help me, please, just need to figure this out to finish my project.

Here is the error.

Thank you very much.

03-09 12:01:32.858 22270-22358/? E/AndroidRuntime: FATAL EXCEPTION: GLThread 8836
                                                   Process: au.com.appspeople.beingvr_android, PID: 22270
                                                   java.lang.NullPointerException: Attempt to invoke virtual method 'org.gearvrf.GVRContext org.gearvrf.io.GVRCursorController.getGVRContext()' on a null object reference
                                                       at org.gearvrf.io.GVRInputManager.addCursorController(GVRInputManager.java:272)
                                                       at org.gearvrf.io.GVRInputManager$SingleControllerSelector.onCursorControllerRemoved(GVRInputManager.java:803)
                                                       at java.lang.reflect.Method.invoke(Native Method)
                                                       at org.gearvrf.GVREventManager.invokeMethod(GVREventManager.java:307)
                                                       at org.gearvrf.GVREventManager.sendEventWithMaskParamArray(GVREventManager.java:147)
                                                       at org.gearvrf.GVREventManager.sendEventWithMask(GVREventManager.java:113)
                                                       at org.gearvrf.GVREventManager.sendEvent(GVREventManager.java:108)
                                                       at org.gearvrf.io.GVRInputManager.removeCursorController(GVRInputManager.java:288)
                                                       at org.gearvrf.io.GVRCursorController.setEnable(GVRCursorController.java:761)
                                                       at org.gearvrf.io.GVRGearCursorController.setEnable(GVRGearCursorController.java:244)
                                                       at org.gearvrf.io.GVRInputManager.removeCursorController(GVRInputManager.java:286)
                                                       at org.gearvrf.io.GVRGearCursorController.onDrawFrame(GVRGearCursorController.java:312)
                                                       at org.gearvrf.OvrViewManager.onDrawFrame(OvrViewManager.java:301)
[...]

PR #1784, opened 2018-03-12
controller: do not fallback to gaze if there is none

liaxim:
fix for issue #1777

---


liaxim:
Small fix. Merging to create new snapshots.

PR #1785, opened 2018-03-12
 Support selecting a backend explicitly

liaxim:
Also brings in couple of commits from the maint_v3.x branch. The goal is to build an apk that includes all possible backends and be able to select any of them via a system property.

---


liaxim:
Ready for review

liaxim:
If there are no comments will merge on Thursday.

PR #1788, opened 2018-03-13
backend_oculus: ensure the camera rig transform is updated prior to everything else

liaxim:
The camera rig transform was previously updated after beforeDrawEyes (which also runs the drawFrameListeners). It can cause visual glitches if somebody modifies the camera rig transform from one of those listeners. And in any case, it is more correct to update the camera rig transform prior to doing anything else.

---


liaxim:
Was raised by an internal user in February.

liaxim:
Planning to merge on Tuesday

PR #1789, opened 2018-03-13
GL: clear stencil buffer correctly

sushantojal:
The useStencilBuffer_ flag in glRenderer was being set from onSurfaceCreated. But it was being reset to false from onSurfaceChanged because of the nativeLeaveVrMode() call. As a result, the stencil buffers were not getting cleared. Setting the useStencilBuffer_ flag from onSurfaceChanged in this PR.


sushantojal:
should probably leave the configureRendering call in onSurfaceCreated in GVRViewManager since nativeLeaveVrMode is called only for oculus.

liaxim:
Looks good to me. Planning to merge on Tuesday.

PR #1794, opened 2018-03-15
build: support switching to clang

liaxim:
Set GVRF_USE_CLANG to true to enable use of clang.

Known issue: apps using libgdx do not work as expected.

---


liaxim:
Gcc's removal is imminent so unless we want to get stuck with an old nkd this has to be done. Also makes compilation faster and we get to enjoy more strict compiler.

liaxim:
All demos (Oculus) work except those that use libgdx.

liaxim:
Planning to merge on Monday.

Issue #1786, opened 2018-03-13
GVRViewSceneObject on click not fired


ageofish:
When running the gvr_events example in the headset, the buttons doesn't seem to be pressed (not highlighted), also OnClick events are not fired.
When running the app on the phone, everything works with tapping on the screen.

Seems like a bug?
I override onTouch and was able to get the event in headset, and onCursorControllerSelected seems to be selecting GAZE controller in the headset correctly, any ideas?

liaxim:
Yes, a bug. Was able to reproduce. Thanks for letting us know. We'll have it fixed.

liaxim:
@ageofish There is a fix in https://github.com/Samsung/GearVRf/pull/1799. There are still some issues but gvr-events should work ok enough.

liaxim:
@ageofish I believe the issue is addressed. Will close this ticket in few days.

PR #1799, opened 2018-03-16
gazecursor: fixes so events work when docked

liaxim:
Fix for issue #1786, "GVRViewSceneObject on click not fired"

There is more to it but makes gvr-events for example usable again (when docked).

---


ragner:
Works beautifully for me 👍

PR #1779, opened 2018-03-09
Vulkan lights only

apjagdale:
Added support for vulkan lights. [shadow map remaining]
Lights for both GL and Vulkan are UBO's.
Material for Vulkan now is UBO instead of Push Constants.
Changed few LOGE to LOGI.
This PR does not have screenshot fixes.
Depends on PR #1721 
Tested with gvr-multilight and few light tests.

GearVRF DCO signed off by: Abhijit Jagdale a.jagdale@samsung.com

PR #1803, opened 2018-03-20
Vulkan: fixes for texture tests and texture transparency tests

sushantojal:
Some of the texture tests have dependency on #1779 and there might be crashes in texture tests because of that. Have tested them with #1779 and they should be fine once that pull request is merged.

With this PR, these are the results of running the framework tests:

Passed:

    - BoundingVolumeTests
    - CullTests
    - RenderConfigTests
    - StencilTests
    - TextureTests
    - TextureTransparencyTests
    - PickerTests
    - MiscTests

Partial (listing the failed sub-tests):

    - RenderTests: testTwoPostEffects
    - SceneObjectTests: fails arbitrarily on some tests
    - RenderPerfTests: fails with tests having cylinder model; quads pass

Crashes / Not Passsed:

    - LightTests
    - ShadowTests
    - VideoTests
    - RenderShaderTests: meshFormatsWithShaderCombinationsTest (depends on lighting)

GearVRF DCO signed off by: Sushant Ojal sushant.o@samsung.com


liaxim:
Needs rebase

liaxim:
@roshanch @apjagdale Please review

apjagdale:
Seems good.

roshanch:
looks good to me.

liaxim:
Waiting on #1779.

liaxim:
Needs rebase

sushantojal:
Texture transparency tests run correctly. Texture tests run individually but some fail randomly when all are run at once. Scene object tests have the same issue. Will do a separate PR for this bug.

GVRf/Framework/framework/src/main/jni/vulkan/vulkan_shader.cpp
```diff
@@ -113,8 +113,11 @@ int VulkanShader::bindTextures(VulkanMaterial* material, std::vector<VkWriteDesc
         const DataDescriptor::DataEntry* e = mTextureDesc.find(texname);
         if ((e == NULL) || e->NotUsed)
         {
-            return;
+            return -1;
```
liaxim: There should be no return value. This is what func is: ``std::function< void(const char* texname, Texture* tex) >``. It actually crashes if compiled with clang.

liaxim:
Transparency tests passed for me. Ran TextureTests and SceneObjectsTests few times without crashes but not all test cases pass.

liaxim:
Needs rebase also

sushantojal:
should be ok now. SceneObject tests, texture tests and texture transparency tests should all run.

GVRf/Framework/framework/src/main/jni/vulkan/vulkanCore.cpp
```diff
@@ -1276,7 +1281,8 @@ void VulkanCore::InitPipelineForRenderData(const GVR_VK_Vertices* m_vertices, Vu
 
         // TODO: add shadowmap descriptor
 
-        vkShader->bindTextures(vkmtl, writes,  descriptorSet);
+         if(vkShader->bindTextures(vkmtl, writes,  descriptorSet) == -1)
```
liaxim: VulkanShader::bindTextures is confusing. It can't actually return -1. Please check it out; ``texIndex`` and ``fail`` are not being used.

liaxim:

Issue #1802, opened 2018-03-19
App exiting while run on DayDream device


spakanati123:
Hi,
I built the app for both Samsung gear device(oculus) and Google daydream device.
The app input events perfectly fine on Samsung gear device(oculus) including swipe, click and back key events.
But on DayDream headset i can see the weird behavior, tapping any Google Controller(remote) key consecutive two times making my app exit to device home screen.

My implementation logic.
I am dispatching received touch events from GVRCursorController(GVRCursorController.getMotionEvent()) to my GVRSceneObject. 

Please suggest me a way to do for DayDream devices if my approach is wrong.

spakanati123:
This is log trace:
```

03-20 01:55:31.481 15066-15066/com.qp.vrdemo D/ViewRootImpl@485f199[VRDemoHomeActivity]: MSG_WINDOW_FOCUS_CHANGED 0
03-20 01:55:31.556 15066-15066/com.qp.vrdemo V/Configuration:  needToUpdateOverlays
03-20 01:55:31.556 15066-15066/com.qp.vrdemo I/chatty: uid=10364(u0_a364) com.qp.vrdemo identical 2 lines
03-20 01:55:31.556 15066-15066/com.qp.vrdemo V/Configuration:  needToUpdateOverlays
03-20 01:55:31.624 15066-15066/com.qp.vrdemo D/SurfaceView: BG show() Surface(name=Background for - SurfaceView - com.qp.vrdemo/com.qp.vrdemo.VRDemoHomeActivity@9b9e936@1) com.google.vr.ndk.base.GvrLayoutImpl$AsyncReprojectionSurfaceView{9b9e936 V.E...... ........ 0,0-2678,1440}
03-20 01:55:31.625 15066-15066/com.qp.vrdemo D/SurfaceView: surfaceDestroyed 1 com.google.vr.ndk.base.GvrLayoutImpl$AsyncReprojectionSurfaceView{9b9e936 V.E...... ........ 0,0-2678,1440}
03-20 01:55:31.638 15066-15066/com.qp.vrdemo D/SurfaceView: BG show() Surface(name=Background for - SurfaceView - com.qp.vrdemo/com.qp.vrdemo.VRDemoHomeActivity@79339a4@1) android.opengl.GLSurfaceView{79339a4 V.E...... ........ 0,0-2678,1440}
[...]

PR #1811, opened 2018-03-23
controller: add GestureDetector to the framework and new events to GVRMain

liaxim:
Replace the sendEventsToActivity feature with onSwipe event

Demos pull request: https://github.com/gearvrf/GearVRf-Demos/pull/592
Fix for https://github.com/Samsung/GearVRf/issues/1802

### DIFF
--- a/GVRf/Extensions/3DCursor/3DCursorLibrary/src/main/java/org/gearvrf/io/cursor3d/CursorManager.java
+++ b/GVRf/Extensions/3DCursor/3DCursorLibrary/src/main/java/org/gearvrf/io/cursor3d/CursorManager.java
@@ -18,31 +18,25 @@ package org.gearvrf.io.cursor3d;
 
 import android.view.MotionEvent;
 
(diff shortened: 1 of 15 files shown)


liaxim:
I believe the generated event coordinates are not correct screen-space coordinates so the possibility of fixing it by correcting the coordinates exists. I'd like to understand why we need sendEventsToActivity(true) in first place though.

liaxim:
@NolaDonato Please let me know what do you think about this. Have only updated gvr-simplephysics thus far but the rest will be similar.

1. Channel the events fed into the gesture detector through IActivityEvents.dispatchTouchEvent
Not calling directly GVRActivity.dispatch* avoids the problem with Daydream
2. Since this is a common requirement, move the implementation to GVRActivity and deliver onSwipe events to GVRMain if requested by calling enableSwipeEvents
3. There is some overhead this is why swipe events have to explicitly enabled first

liaxim:

PR #1817, opened 2018-04-06
GVR_VK_CHECK: replace the use of assert

liaxim:


liaxim:
Planning to merge on Monday.

assert is noop in release builds which doesn't look appropriate to me for GVR_VK_CHECK.

PR #1825, opened 2018-04-09
Anchor / Inline X3D filenames can be surrounded by extra quotes.

mwitchwilliams:
Fixed an issue with X3D &lt;Inline&gt; and &lt;Anchor&gt; nodes where a new URL could be surrounded by an additional set of single or double quotes.
Also fixed issue related to bindings created in AnimationInteractivityManager.java now called just once after ALL &lt;Inline&gt; files loaded.  Previously, this task occurred after every &lt;Inline&gt; node.

**ISSUE**: Replacing X3D files using &lt;Anchor&gt; node not showing lights.

Test files attached.  Each of the two files have a Sphere when clicked loads the other .x3d file.  It doesn't matter which of the two files to begin demo, but the 'secondary file' includes an animated Cone plus Cylinder with green emissiveColor.  
After the first file displays, none of the lights work when going to the second file - the light only appears for the original file.

[X3DAnchor_LightsProblem.zip](https://github.com/Samsung/GearVRf/files/1891419/X3DAnchor_LightsProblem.zip)

Issue first researched based on a demo from Tom.


liaxim:
Needs rebase

mwitchwilliams:
Request this PR be merged as it is holding up some critical fixes and the outstanding issue of the lights not being turned on when loading a new file/scene is not an X3D issue.  I can be added to the Issues list on the Git site.

NolaDonato:
I agree the lights are not an X3D issue. I tried a fix which worked for other apps but unfortunately not for X3D.

liaxim:
Ok. Planning to merge on Thursday.

PR #1827, opened 2018-04-10
Transform and state sorting reform

NolaDonato:
1. Added capability to calculate matrices in a general way and put them in a transform blocfk
2. Accumulated state sorting culling and transform code into a new RenderSorter class
3. Custom state sorting for main scene, post effects and shadow maps
4. Eliminated redundant GL state calls
5. Accumulated all render state into a single 64 bit number
6. Store render state in render pass
7. Added transparent flag to ShaderData
8. Added GVRRenderer class

Not for merge yet - Vulkan is not working.


liaxim:
There are some commits adding pictures that do not belong here.

GVRf/Framework/framework/src/main/jni/objects/components/render_data.cpp
```diff
@@ -149,161 +112,20 @@ void RenderData::bindShader(JNIEnv* env, jobject localSceneObject, bool isMultiv
...
-        render_data_string.append(std::to_string(stencilFuncMask_));
-        render_data_string.append(std::to_string(stencilOpSfail_));
-        render_data_string.append(std::to_string(stencilOpDpfail_));
-        render_data_string.append(std::to_string(stencilOpDppass_));
-        render_data_string.append(std::to_string(source_alpha_blend_func_));
-        render_data_string.append(std::to_string(dest_alpha_blend_func_));
         render_data_string.append(std::to_string(mesh_->getVertexBuffer()->getDescriptor()));
         hash_code = render_data_string;
-        hash_code_dirty_ = false;
+        pass(0)->render_modes().clearDirty();
```
roshanch: not for all the passes?

Caprica666: We have to talk about the hash code. All of the render states now fit in a single 64 bit word, one for each pass. The render passes are rendered separately, not together anymore. OpenGL doesn't use this hash code. We should discuss its significance in Vulkan.

Caprica666:
I removed the one picture that was left in the repo. How do I remove commits in the middle? I was in a hurry to finish Siggraph registration and they needed photos of a specific size for the badge. I used Git to transfer stuff between PC and Linux...

liaxim:
You could "git revert" the unwanted commits. Or use git interactive rebase - "git rebase -i HEAD~15" to drop those commits.

liaxim:
And needs a rebase

GVRf/Framework/framework/src/main/jni/engine/renderer/render_state.h
```diff
@@ -0,0 +1,712 @@
...
+            offset_units = 0;
+        }
+
+        RenderModes& operator=(const RenderModes& src)
+        {
+            render_flags = src.render_flags;
+            offset_factor = src.offset_factor;
+            offset_units = src.offset_units;
+            render_order = src.render_order;
+            sample_coverage = src.sample_coverage;
```
liaxim: Missing return; please compile with GVRF_USE_CLANG=true and address the warnings in the new and changed code.

liaxim:
gvr-accessibility doesn't look right:

liaxim:
gvr-controls:

liaxim:
gvr-immersivepedia

liaxim:
gvr-keyboard launches into a black screen

liaxim:
gvr-multilight - z-fighting?

liaxim:
gvr-modelviewer2 - looks weird, giant reticle

liaxim:
gvr-shadows doesn't look right as you look up and down

liaxim:
gvr-sceneobjects

liaxim:
gvr-simplephysics, gvr-switch, gvr-tutorial-lesson6, gvr-x3d-demo all don't work

liaxim:
Such a giant change needs to be backed up by automated tests proving correctness; or at least some level of. The previous thing was quite thoroughly tested over an extended period of time; I seriously doubt going through all the demo covers everything and can be considered enough for a change of this magnitude.

liaxim:
Ok, is there a Demos pull request that goes along with this one that is missing?

GVRf/Framework/framework/src/main/java/org/gearvrf/shaders/GVRPhongLayeredShader.java
```diff
@@ -101,5 +111,48 @@ protected void setMaterialDefaults(GVRShaderData material)
...
+     * calculate the inverse of the model view matrix.
+     * @param inputMatrices input matrices - projection, left view, right view and model.
+     * @param outputMatrices left model view projection, right model view projection
+     *                       (for lighting) left model view inverse transpose,
+     *                       right model view inverse transpose
+     * @param numMatrices    number of output matrices expected for stereo
+     * @param isStereo       true for stereo rendering, false for mono
+     * @returns number of matrices actually calculated and stored in the output buffer
+     */
+    public static int calcMatrix(float[] inputMatrices, float[] outputMatrices, int numMatrices, boolean isStereo)
```
liaxim: The calcMatrix methods on the Java side show up as unused. Are they needed?

NolaDonato:
PR #597 in GearVRf-Demos addresses some of these issues. I am debugging the others. I think there is an issue with render mask which is why gvr-sceneobjects is broken. I also took out the default of making everything use alpha blending. This is why some of the others are broken. We should not have this on by default.

GVRf/Extensions/gvrf-particlesystem/src/main/res/raw/particle_vert.vsh
```diff
@@ -35,14 +35,7 @@ void main() {
...
 
-    #ifdef HAS_MULTIVIEW
-           bool render_mask = (u_render_mask & (gl_ViewID_OVR + uint(1))) > uint(0) ? true : false;
-           mat4 mvp = u_mvp_[gl_ViewID_OVR];
-           gl_Position = mvp  * posn;
-    #else
-       	    gl_Position =  u_mvp * posn;
-    #endif
-
+    gl_Position =  u_mvp * posn;
```
liaxim: The HAS_MULTIVIEW ifdef block is really not needed anymore?

NolaDonato:
There were some shader generation bugs causing some issues with the examples. I have fixed these. Now the immersivepedia, accessibility, shadows, controls, simplephysics, modelviewer2, keyboard, switch, tutorial , and x3ddemo samples work now.

GVRf/Extensions/gvrf-particlesystem/src/main/res/raw/particle_vert.vsh
NolaDonato: No, this is no longer necessary because the Java shader generation code takes care of it now.

GVRf/Framework/framework/src/main/jni/engine/renderer/render_state.h
```diff
@@ -0,0 +1,713 @@
...
+        void setSampleCoverage(float f)
+        {
+            if (sample_coverage!= f)
+            {
+                markDirty();
+                sample_coverage = f;
+            }
+        }
+
+        uint64_t getRenderFlags() const
```
roshanch: what is intention of this function?

NolaDonato: This function is for sorting on render modes. It puts all the render flags in a single 64 bit integer. The only things not in there are 3 float values (polygon offset, ..)

Look at MainSceneSorter::mergeByMode to see its usage. It can also be used as a hash code.

roshanch: why last 3 values are not considered for render modes? we use them in setRenderStates(). shouldn't the hash code contain all the values?

roshanch:
complexscene freezes.

NolaDonato:
This PR has been merged with master and passes all the framework tests on OpenGL. Vulkan  is still in development.

Asset tests have missing objects and sometimes only show for one eye.

The merge with master needs work.

roshanch:
@NolaDonato @liaxim I think, we should merge OpenGL first. Already there are many commits in this PR, so it will be better to have Vulkan support as separate PR.

NolaDonato:
If we do that Vulkan will be completely broken until that second pull request. In any case, take a look at the merge. Something broke in lighting and I could use some help evaluating whether I merged correctly.

liaxim:
It is unfortunate that this work cannot be done in more incremental way.

roshanch:
the reason I am saying it because efforts needed to rebase it. everyday we have to rebase it..

liaxim:
@roshanch I absolutely agree. This is why this work needs to be broken down and done in increments. Also this pull is of such size that it is effectively un-reviewable. Increments we can reason about and easier to test since you can tell what to target.

NolaDonato:
I just did the rebase. Since you have been doing it every day perhaps you can spot what I did wrong. Sushant and I are going to try to do the Vulkan part.

NolaDonato:
I have done a lot of testing and the problem is not a merge failure. The code works fine on a Note8 with Android O. It fails on a Note8 with Android N. It also fails on an S8 with Android O. It appears to corrupt uniform blocks. The lighting anomalies are caused by the view matrix being incorrect. Printing it right before the uniform block is loaded into the GPU shows no errors. Still investigating.

sushantojal:
The issue is with the GL_MAX_VERTEX_UNIFORM_COMPONENTS query. Looks like the value returned by the driver is not reliable. For gvr-multilight, any block of more than 18 matrices (288 floats) does not work for the S8+ (adreno, Android N) I am using; even though the query returns 1024 (64 mats). Works fine for values 18 and below.
[Similar](https://stackoverflow.com/questions/8586543/what-is-the-actual-number-of-vertex-uniform-components-for-glsl-shader-on-ati-gr) issues have existed for other drivers, but I wasnt able to find any documentation confirming if this is a known issue on this particular driver version.
We'll need to come up with a reliable workaround for this.

liaxim:
Should we close?

NolaDonato:
No I am going to revive this one

liaxim:
Revive against GearVRf instead of SXRSDK?!

Issue #1535, opened 2017-10-13
gvr-360photo, vulkan, s8: crash after pause/resume


liaxim:
```
10-13 14:42:45.829 16668 16668 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
10-13 14:42:45.832 16668 16668 F DEBUG   : Build fingerprint: 'samsung/dreamqltesq/dreamqltesq:7.0/NRD90M/G950USQU1AQFJ:eng/test-keys'
10-13 14:42:45.833 16668 16668 F DEBUG   : Revision: '12'
10-13 14:42:45.833 16668 16668 F DEBUG   : ABI: 'arm'
10-13 14:42:45.835 16668 16668 F DEBUG   : pid: 16495, tid: 16539, name: GLThread 6466  >>> org.gearvrf.gvr360Photo <<<
10-13 14:42:45.836 16668 16668 F DEBUG   : signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x0
10-13 14:42:45.836 16668 16668 F DEBUG   :     r0 db165d10  r1 00000000  r2 00000000  r3 e588f3d5
10-13 14:42:45.837 16668 16668 F DEBUG   :     r4 00000000  r5 db165d10  r6 c7ef6b58  r7 dadf86e8
10-13 14:42:45.837 16668 16668 F DEBUG   :     r8 12c68440  r9 db144000  sl 12c42698  fp 12d061a0
10-13 14:42:45.837 16668 16668 F DEBUG   :     ip e5455968  sp dadf86b0  lr e588f43f  pc 00000000  cpsr 40000010
10-13 14:42:45.888 16668 16668 F DEBUG   : 
10-13 14:42:45.888 16668 16668 F DEBUG   : backtrace:
10-13 14:42:45.888 16668 16668 F DEBUG   :     #00 pc 00000000  <unknown>
10-13 14:42:45.888 16668 16668 F DEBUG   :     #01 pc 0000f43d  /system/lib/libvulkan.so
[...]

PR #1831, opened 2018-04-12
vulkan: monoscopic fixes

sushantojal:
Recreate swapChain when surface is changed.
Added vkDeviceComponent class for cleaning up resources created from logical device.
Track rendertextures when using monoscopic backend. (necessary to stop gc from using resources currently in use).
Draw thread synchronization in MonoscopicViewManager for vulkan.


liaxim:
Needs rebase

GVRf/Framework/framework/src/main/jni/vulkan/vk_Framebuffer.cpp
```diff
@@ -17,14 +17,32 @@
 #include "vulkanCore.h"
 
 namespace gvr {
+
     VKFramebuffer::~VKFramebuffer() {
-        delete mAttachments[COLOR_IMAGE];
-        delete mAttachments[DEPTH_IMAGE];
+
```
apjagdale: Delete multisample image as well.

GVRf/Framework/framework/src/main/jni/vulkan/vulkanCore.cpp
```diff
@@ -1327,12 +1336,104 @@ void VulkanCore::InitPipelineForRenderData(const GVR_VK_Vertices* m_vertices, Vu
...
+
+        VulkanRenderer* vk_renderer= static_cast<VulkanRenderer*>(Renderer::getInstance());
+        //clear the handles to the swapChain images and imageviews. This is necessary because the driver
+        //tries to reuse these handles when a new surface is created. When the garbage collector eventually runs
+        //it might accidentally clear up the very same handles we are reusing.
+
+        for(auto tex: vk_renderer->mOnScreenTextures)
+        {
+            vkImageBase ** attachments = tex->getFBO()->getAttachments();
+            delete attachments[COLOR_IMAGE];
```
apjagdale: Same for multisampled image

GVRf/Framework/backend_monoscopic/src/main/java/org/gearvrf/MonoscopicViewManager.java
```diff
@@ -215,7 +215,7 @@ public void surfaceCreated(SurfaceHolder surfaceHolder) {
                         mInputManager.setScene(scene);
                         mRotationSensor.onResume();
                         long vulkanCoreObj;
-                        vulkanCoreObj = NativeVulkanCore.getInstance(vulkanSurfaceView.getHolder().getSurface());
+                        vulkanCoreObj = NativeVulkanCore.getInstance(vulkanSurfaceView.getHolder().getSurface(), true);
```
liaxim: Instead of this surfaceInit flag why not directly call NativeVulkanCore_recreateSwapchain from vulkanSurfaceView...surfaceCreated? Seems to me like it would clearer and simpler.

GVRf/Framework/framework/src/main/jni/vulkan/vulkan_vertex_buffer.cpp
```diff
@@ -179,15 +179,15 @@ namespace gvr {
                     VkVertexInputAttributeDescription binding;
                     binding.binding = GVR_VK_VERTEX_BUFFER_BIND_ID;
                     binding.location = e.Index;
-                    LOGE("location %d attrMapping[i].offset %d , name %s", entry->Index, entry->Offset, entry->Name);
+                    LOGI("location %d attrMapping[i].offset %d , name %s", entry->Index, entry->Offset, entry->Name);
```
liaxim: Maybe this should be LOGV

GVRf/Framework/framework/src/main/jni/vulkan/vulkan_vertex_buffer.cpp
```diff
@@ -179,15 +179,15 @@ namespace gvr {
...
                     binding.format = getDataType(entry->Type); //float3
                     binding.offset = entry->Offset;
                     vertices->vi_attrs.push_back(binding);
                     i++;
                 }
                 else                                // mesh uses attribute but shader does not
                 {
-                    LOGE("entry is not present %s", e.Name);
+                    LOGI("entry is not present %s", e.Name);
```
liaxim: And this maybe LOGW

GVRf/Framework/framework/src/main/jni/vulkan/vulkanCore.cpp
```diff
@@ -1334,12 +1337,106 @@ void VulkanCore::InitPipelineForRenderData(const GVR_VK_Vertices* m_vertices, Vu
     }
 
     VulkanCore::~VulkanCore() {
-        vkDestroyDevice(getDevice(), nullptr);
+
+        //destroy all the resources associated with the current device.
```
liaxim: I don't think this ever gets called. See issue https://github.com/Samsung/GearVRf/issues/1535.

liaxim:
Btw, TextureTests with monoscopic seem to hang on my s8.

GVRf/Framework/framework/src/main/jni/engine/renderer/vulkan_renderer.cpp
```diff
@@ -120,8 +120,11 @@ RenderTexture* VulkanRenderer::createRenderTexture(int width, int height, int sa
                                                    int jcolor_format, int jdepth_format, bool resolve_depth,
                                                    const TextureParameters* texture_parameters, int number_views, bool monoscopic)
 {
-    if(monoscopic)
-        return new VkRenderTextureOnScreen(width, height, sample_count);
+    if(monoscopic) {
+        VkRenderTextureOnScreen * tex = new VkRenderTextureOnScreen(width, height, sample_count);
+        mOnScreenTextures.push_back(tex);
```
liaxim: Let's find a way to avoid expanding upon the monoscopic-specific hack we already have in here. The MonoscopicViewManager already tracks the renderTextures via the mRenderTarget array. Why can't we recreate/reset the old renderTextures from the Java side from surfaceCreated?

GVRf/Framework/backend_monoscopic/src/main/java/org/gearvrf/MonoscopicViewManager.java
sushantojal: yes, in java we could check if a vulkan instance exists before the getInstance call and call recreateswapchain if the instance exists. I did this in the c++ side right now.

sushantojal:
@liaxim  texture tests will work this PR along with PR #1841 and PR 264 in tests repo. Screenshots are not a part of this pull request.

liaxim:
Using mOnScreenTextures specifically for the monoscopic backend will have to be fixed before we can merge this. We can't add backend-specifics to the framework. Or expand on the very few things that managed to escape in the past.

GVRf/Framework/framework/src/main/jni/engine/renderer/vulkan_renderer.cpp
sushantojal: working on this. Instead of destroying the renderTargets on termination of draw thread, as we are doing right now, we could possibly retain them and just switch the swapChain image handles in recreateSwapChain.

sushantojal:
Removed mOnScreenTextures  completely. To track the renderTargets (from which references to swapchain image handles should be removed), now adding them to the renderBundle, and fetching them in vulkanCore.
Still not able to determine when to use resetInstance. Using them in onDestroy works only for apps with one activity and crashes in tests (because onCreate of second activity is called before onDestroy of first).

PR #1835, opened 2018-04-17
Vulkan : Shadowmap

apjagdale:
Implemented similar to that of GL in which shadow is packed into color buffer and then unpacked in later pass to get the depth data.

Demo used: gvr-shadows


apjagdale:
I will free the cascade framebuffers when #1831 is merged.

liaxim:
@apjagdale Do we **have to** wait on #1831?

apjagdale:
@liaxim Yes.

liaxim:
Needs rebase

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRRenderTarget.java
```diff
@@ -138,9 +138,22 @@ public GVRRenderTexture getTexture()
...
+    private GVRDrawFrameListener  mDrawListener = new GVRDrawFrameListener()
     {
-        getGVRContext().getActivity().getViewManager().cullAndRender(this, mScene);
+        public void onDrawFrame(float frameTime)
+        {
+            getGVRContext().getActivity().getViewManager().cullAndRender(GVRRenderTarget.this, mScene);
+        }
+    };
+
+    public void startRendering()
```
liaxim: Public APIs need JavaDoc.

GVRf/Framework/framework/src/main/jni/vulkan/vk_imagebase.cpp
```diff
@@ -33,6 +32,19 @@ int getComponentsNumber(VkFormat format){
...
+    VkImageAspectFlagBits getAspectFlagForFormat(VkFormat format){
+        switch (format){
+            case VK_FORMAT_R8G8B8A8_UNORM:
+                return VK_IMAGE_ASPECT_COLOR_BIT;
+            case VK_FORMAT_D16_UNORM:
+            case VK_FORMAT_D32_SFLOAT:
+            case VK_FORMAT_D24_UNORM_S8_UINT:
+                return VK_IMAGE_ASPECT_DEPTH_BIT;
+            default:
+                LOGE("format not found");
```
liaxim: Missing return

liaxim:
GL's render target no longer works as before - see problems in the gvr-blurfilter and gvr-rendertotexture demos.

Also is it just me or is the trooper's shadow missing in gvr-shadows really (GL)? Have noticed this some weeks ago.

liaxim:
gvr-shadows works and look fine with vulkan. gvr-simplephysics works too though doesn't quite look as well as when with gl (I added back the lights for vulkan to check it out).

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRRenderTarget.java
apjagdale: This is the reason why blurfilter and rendertotexture not working. With this new API, we need to call startRendering() from app who create render target. Will discuss with @NolaDonato how it should be done.

apjagdale:
Cleanup regarding Start and End rendering added.
The gvr-shadows issue for GL exist priop my pull request. Its random, debugging it.
The gvr-simplephysics also debugging it.

liaxim:
Ok, I assume you plan to add more commits to this pull request.

NolaDonato:
mHasFrameCallback indicates whether or not the subclass has implemented onDrawFrame. The default behavior of the parent class GVRBehavior is to start a draw frame listener when the behavior is enabled. For render targets, calling setEnable will start a draw frame listener that does rendering. For the picker, calling setEnable will start a draw frame listener that does picking.

For shadow maps, calling setEnable should NOT do any rendering - the rendering is done underneath in the renderer. Currently, render targets made by the user to do render to texture are not handled underneath by the renderer - the application has to call setEnable to start rendering. In the future we would like to handle scheduling of all render targets by the Renderer and move away from using the draw frame listener. We are not at that point yet so right now we have this hybrid implementation where shadow maps are handled internally and application render targets are not.

Setting mHasFrameCallback to null disables automatic rendering for the shadow maps. Eventually we will eliminate onDrawFrame from GVRRenderTarget completely.

apjagdale:
@liaxim Thats the only change I wanted to add to this pull request. I will look into fixing of gvr-shadows for GL and the gvr-simplephysics for Vulkan.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRShaderTemplate.java
```diff
@@ -65,6 +65,7 @@
 public class GVRShaderTemplate extends GVRShader
 {
     private final static String TAG = "GVRShaderTemplate";
+    private final int shadowmapStartLocation = 20;
```
liaxim: Why is it 20 and not any other number?

GVRf/Framework/framework/src/main/java/org/gearvrf/shaders/GVRPhongShader.java
```diff
@@ -48,7 +48,7 @@ public GVRPhongShader(GVRContext gvrcontext)
        {
             super("float4 ambient_color; float4 diffuse_color; float4 specular_color; float4 emissive_color; float3 u_color; float u_opacity; float specular_exponent; float line_width; float2 u_lightmap_offset; float2 u_lightmap_scale",
                    "sampler2D diffuseTexture; sampler2D ambientTexture; sampler2D specularTexture; sampler2D opacityTexture; sampler2D lightmapTexture; sampler2D normalTexture; sampler2D emissiveTexture",
-                   "float3 a_position float2 a_texcoord float2 a_texcoord1 float2 a_texcoord2 float2 a_texcoord3 float3 a_normal float4 a_bone_weights int4 a_bone_indices float4 a_tangent float4 a_bitangent",
+                   "float3 a_position float2 a_texcoord float2 a_texcoord1 float2 a_texcoord2 float2 a_texcoord3 float3 a_normal float4 a_bone_weights int4 a_bone_indices float3 a_tangent float3 a_bitangent",
```
liaxim: Should a similar change be applied to GVRPhongLayeredShader?

liaxim:
I can confirm the demos with GL are fine; except gvr-vuforia because of the aforementioned viewport thing.

liaxim:
Please pay attention to the compiler warnings in the new code and fix them. Thanks.

liaxim:

liaxim:
@apjagdale Daydream and monoscopic backends no longer work. Looks like something in this pull request caused it. Please take a look.

apjagdale:
@liaxim okay will look into this.

liaxim:
@apjagdale Looks like it will be a good idea to correctly default-initialize RenderTextureInfo::viewport somehow.

PR #1838, opened 2018-04-18
gradle: support "discovery" of extra_settings.gradle

liaxim:
Automatically include modules from other repos cloned in the same directory as GearVRf. Needed for the internal build system.


liaxim:

PR #1839, opened 2018-04-18
monoscopic: support fixed resolution

liaxim:
If eye-buffer-parms.resolutionWidth/Height is set in gvr.xml use that; if not default to the screen resolution.


liaxim:
@sushantojal This should let you configure the tests with one resolution (FHD+?) and have them produce the same resolution screenshots across different devices. Have not verified it since the screenshotting feature doesn't work for me with monoscopic.

sushantojal:
Yes, this works. Thanks.

liaxim:
Planning to merge on Monday.

Issue #1848, opened 2018-04-23
ShaderData::bindTexture texture diffuseTexture at loc=0 not ready


SteveGreatApe:
Bit of a strange one this, I've got a scene object where I can set it's material to either be a texture created from a bitmap, or I set the texture to null and the ambient and diffuse colours to a selected colour instead.

Setting the texture from a bitmap works without a problem, but when I try and switch from a texture to a solid colour by setting the diffuseTexture to null:

```
            material.setTexture("diffuseTexture", null);
```

I then get this error rapidly repeating in the log:
```
...
2018-04-23 15:20:05.734 27690-28066/com.greatape.avrgallery V/gvrf: ShaderData::bindTexture texture diffuseTexture at loc=0 not ready
2018-04-23 15:20:05.747 27690-28066/com.greatape.avrgallery V/gvrf: ShaderData::bindTexture texture diffuseTexture at loc=0 not ready
2018-04-23 15:20:05.751 27690-28066/com.greatape.avrgallery V/gvrf: ShaderData::bindTexture texture diffuseTexture at loc=0 not ready
2018-04-23 15:20:05.763 27690-28066/com.greatape.avrgallery V/gvrf: ShaderData::bindTexture texture diffuseTexture at loc=0 not ready
...
```

And bizarrely the object now appears to be showing a texture from another object in the scene. If I then later try and change the colour, which also happens to set the texture to null again, this seems to fix the problem and it changes to showing the fixed colour as expected.

NolaDonato:
I have tried to reproduce this with the PhongShader but I cannot. What shader are you using? Could you send me a code snippet?

NolaDonato:
I think you were using the Texture shader, I reproduced it with that. PR #1850 should fix it, please check.

SteveGreatApe:
Thanks, it's the Phong Shader I'm using, I've tried your fix and doesn't seem to have worked for me, I've added some logging to it to make sure I'm going through that code and it's definitely running your change.

I can reproduce it in the multilight demo app, see below for changes. When the texture is set to null I see it replaced by the texture being used for the 3D character.

```
diff --git a/gvr-multilight/app/src/main/java/org/gearvrf/multilight/MultiLightMain.java b/gvr-multilight/app/src/main/java/org/gearvrf/multilight/MultiLightMain.java
index 68d831f..faf8aa6 100644
--- a/gvr-multilight/app/src/main/java/org/gearvrf/multilight/MultiLightMain.java
[...]

PR #1850, opened 2018-04-23
bug fix for removing texture

NolaDonato:
When a texture is set to null, GearVRf was not always regenerating the shader. This patch fixes that - issue #1848.


GVRf/Framework/framework/src/main/jni/objects/shader_data.cpp
```diff
@@ -68,7 +68,7 @@ void ShaderData::setTexture(const char* key, Texture* texture)
         if (temp.compare(key) == 0)
         {
             Texture* oldtex = mTextures[i];
-            makeDirty(oldtex ? MOD_TEXTURE : NEW_TEXTURE);
+            makeDirty(((oldtex != nullptr) && (texture != nullptr)) ? MOD_TEXTURE : NEW_TEXTURE);
```
liaxim: What is the difference between MOD_TEXTURE and NEW_TEXTURE? Also I don't see obvious use of MOD_TEXTURE.

NolaDonato:
MOD_TEXTURE means you are changing which texture the material uses. NEW_TEXTURE means that a) the material had a texture previously and you are removing it or b) the material did not have a texture previously and you are adding it. Cases A and B require the shader to be changed. MOD_TEXTURE does not.

NolaDonato:
This PR is not finished yet. I am working with a user and there is still a case that fails.

Issue #1853, opened 2018-04-25
unable to scroll listview 


spakanati123:
i can feel very difficult to scroll the listview when using in GVRF.
 hasFocus() returns true for listview when hover and swipe using the controller. So i can say scroll functionality is missing here.
i can observe issue in gve-events sample app as well.

thomasflynn:
just ran gvr-events in the headset and the list scrolled just fine using the headset trackpad, though picking in the list seemed a little off.  I tried to add controller support (adding useControllerTypes="gaze,controller" in vr-app-settings in the gvr.xml) to see if there is a problem with the controller, but i'm having some pairing issues with my phone.  once i sort that out (maybe new batteries?), i'll test with the controller.

spakanati123:
@thomasflynn i guess u tried on latest gvr-events code branch, even using headset trackpad also i need to struggle many number of times to scroll to end of the list.
i am using the latest model of samsung VR headset.

spakanati123:
Yeah...my Galaxy S8 runs on Android 'O'.

thomasflynn:
ok, thanks for confirming.  we're looking into it.

liaxim:
This may have some relation to https://github.com/Samsung/GearVRf/issues/1829.

liaxim:
It is related to #1829. After adding ``super.requestLayout();`` near the end of ``GVRViewSceneObject.RootViewGroup.dispatchDraw`` scrolling works as expected. @ragner Do you have any advice on the correct approach for fixing this refresh problem?

ragner:
@liaxim  I will take a look

ragner:
@spakanati123  We have a fix for this.

spakanati123:
Thanks @ragner , @thomasflynn, @liaxim

liaxim:
Fixed by https://github.com/Samsung/GearVRf/pull/1854. Planning to close in few days.

spakanati123:
@liaxim it seems the build not triggered by this change, can u please trigger build.

liaxim:
@spakanati123 Snapshots have been updated.

spakanati123:
I am unable to increase my scene size even changing the frame layout width and height , it always fixed size(small rectangle box) irrespective of the width and height i gave. 
Can u please suggest to change the size of scene for my layout. i tried scale as well.
```
        frameLayout = new FrameLayout(this);
        frameLayout.setBackgroundColor(Color.WHITE);
        registerView(frameLayout);
[...]

Issue #1829, opened 2018-04-11
gvr-keyboardviewer no longer echoes keypresswes


Caprica666:
This demo used to echo the keys pressed on the virtual keyboard on the selected line in the box. Now it never prints the keys.

liaxim:
I believe this is one more Android O specific problem. Not reproducing on my Android N devices.

NolaDonato:
I will test on another phone

liaxim:
@ragner By adding ``mRootGroup.findFocus().requestLayout()`` to ``InputMethodHandler.handleCharacter`` characters start to show up but I am not convinced this is the right fix. Could you please take a look at this issue too and let us know what you think? Thanks.

liaxim:
Fixed by https://github.com/Samsung/GearVRf/pull/1854.

PR #1854, opened 2018-04-28
Fix rendering android views at Android O

ragner:
Should fix:
https://github.com/Samsung/GearVRf/issues/1853
https://github.com/Samsung/GearVRf/issues/1829 


ragner:
@liaxim Should fix all of that issues.

ragner:
Yes, We need keep compatibility with API 24. 
We may need keep both

ragner:
I can fix this on Wednesday because we are having a holiday here in Brazil  because of labor day.

ragner:
@liaxim  done.

PR #1856, opened 2018-04-30
monoscopic: support monoscopic-specific fov-y

liaxim:
An app can have a special fov-y for when using the monoscopic backend by adding this to gvr.xml:eye-buffer-parms: ``monoscopic-fov-y="50"``. If set, takes precedence over the standard ``fov-y`` parameter.


liaxim:
@thomasflynn Does this do what you have in mind?

liaxim:
@thomasflynn Would this be adequate solution for the monoscopic fov?

I think it might be better to have gvr.xml for the monoscopic backend that applies after the default one but that would be a bit more work. Like if you say use "gvr.xml" then you could also supply "gvr_monoscopic.xml" which applies after the main one. Or "gvr_daydream", etc.

Here I am just adding a new special attribute for monoscopic which is not as neat. Let me know what you think. Thanks.

liaxim:
Thanks. Planning to merge on Thursday.

I'd discourage backend-specific code. We can provide some flexibility in a generic way but that should be it.

Having said that nothing can stop you from calling gvrContext.getClass().getName() and knowing what to look for there, do some things in a backend-specific manner.

PR #1860, opened 2018-05-07
Fix action up dispatched twice

ragner:
* ACTION_UP event is dispatched at onTouchEnd() only


liaxim:
Where did you notice that?

ragner:
You can see this logging `e.getAction()` at RootViewGroup::dispatchPickerInputEvent().

You should be able to reproduce with any touchable renderable view.

ragner:
But there is other way to fix this bug at GVRPicker:generatePickEvents() but I am not sure:
See:
`
else
            {
                propagateOnInside(collision); 
                if (prevHit.touched && !mTouched)
                {
                    propagateOnNoTouch(collision);
                }
                else if (!prevHit.touched && mTouched)
                {
                    propagateOnTouch(collision);
                }
            }

`

liaxim:
Looks good. Now gvr-keyboardview works well too - previously you had to click many and many times in the text field to bring up the keyboard.

liaxim:
Planning to merge on Friday.

PR #1863, opened 2018-05-09
add GVRPicker.HasPickListChanged

NolaDonato:
1. GVRPicker.onPick is called every frame. This function tells if the pick list actually changed or not so we 
   can avoid redundant processing.
2. Added null check for GVRSceneObject.setName
3. Throw exception if app tries to add a component that is already there

This PR is required for the widget library event processing.


liaxim:
Looks good to me but due to the new restriction gvr-3dcursor no longer starts up.

NolaDonato:
This has been rolled into #1874. I tested with the gvr-cursor3D sample and it works fine. There must have been something else wonky in this branch.

PR #1866, opened 2018-05-11
Fixing monoscopic gl and day dream backend 

apjagdale:
No render textures are created for monoscopic GL and day dream backend therefore adding checks at begin and end rendering.


liaxim:

Monoscopic+vulkan looks like this:


I don't remember seeing it before. Is this something related to the vulkan shadowmap changes?

apjagdale:
@liaxim I will look into this. It is irrespective of shadowmap changes.

apjagdale:
@liaxim This is due to device returning a different resolution for android and vulkan. In phone if we hide the navigation bar it works as expected.

Issue #1846, opened 2018-04-23
PROJECT_CURSOR_ON_SURFACE doesn't work


SteveGreatApe:
I've updated my GearVRf version from the master version for the first time in a little while and found PROJECT_CURSOR_ON_SURFACE doesn't work anymore.

The easiest way to reproduce this is in the gvr-controller demo. Set the cursor depth to a larger number such as:
```
//                newController.setCursorDepth(DEPTH);
                newController.setCursorDepth(DEPTH * 2);
```
You can see the cursor depth is not affected by the objects it's pointing at and just goes straight through them.

NolaDonato:
I was not able to reproduce this. I changed the cursor depth in gvr-controller and I still see the cursor on top of the bunnies and boxes. This is with the latest master.

SteveGreatApe:
In my own app where I see the problem I'm using the latest master. In the gvr-controller demo where I reproduced it, it's set to build against '4.0.1-SNAPSHOT'

From the logging I added I can see that in GVRCursorController.java updateCursor() it calculates dist correctly, but it only runs that code when I actually click on an object. The rest of the time it sets the cursor position in moveCursor() where it uses mCursorDepth to set the depth.

When trying to reproduce it in gvr-controller make sure you're not setting DEPTH to the new value, as that will also move the bunnies and boxes so they're at the same depth as the cursor.

NolaDonato:
pull request #1867 should fix this issue

SteveGreatApe:
Thanks, looks like it's working okay now.

liaxim:
#1867 was merged. @SteveGreatApe Can we close the issue? Thanks.

SteveGreatApe:
The length of the pointer ray is correct now, but with the latest code I've lost the cursor on the end of the ray. The initial fix on it's own didn't cause that problem, so there's some other recent change that's done this.

PR #1867, opened 2018-05-11
Fixed cursor projection

NolaDonato:
Even if we don't process the pick event, we should examine the list of picked objects to determine how to update the cursor

This fixes issue #1846 


liaxim:
Planning to merge on Friday.

PR #1870, opened 2018-05-16
Vulkan: Fixing invert shader and posteffect add or remove

apjagdale:
Demo used to test is gvr-accessibility


liaxim:
Looks good. Planning to merge on Monday.

One thing, if the invert colors post-effect is enabled there is a lot of jitter with monoscopic. I assume the post-effect is a performance problem for our vulkan impl right now.

PR #1872, opened 2018-05-16
multiple controller support

roshanch:
numControllers can be set into gvr.xml. By default it is 1.


liaxim:
Nope

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRViewManager.java
```diff
@@ -63,13 +64,17 @@
...
         mScriptManager = new GVRScriptManager(this);
         mEventManager = new GVREventManager(this);
-        mInputManager = new GVRInputManager(this, appSettings.getCursorControllerTypes());
+        mInputManager = new GVRInputManager(this, appSettings.getCursorControllerTypes(),appSettings.getNumControllers());
         mInputManager.scanDevices();
     }
 
-    void onPause() {}
+    void onPause() {
+        mControllerReader.onPause();
```
liaxim: mControllerReader can be null here if the Oculus backend is in use.

GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrViewManager.java
```diff
@@ -285,12 +278,11 @@ protected void afterDrawEyes() {
     @Override
     void onSurfaceCreated() {
         super.onSurfaceCreated();
-        mGearController = mInputManager.getGearController();
-        if (mGearController != null)
-        {
-            mGearController.attachReader(
-                    new OvrControllerReader(mActivity.getActivityNative().getNative()));
-        }
+        mControllerReader = new OvrControllerReader(mActivity.getActivityNative().getNative());
```
liaxim: I think the controller reader could be instantiated from the ctor.

liaxim:
I guess I should've tested more. @roshanch Please look into the crash I mentioned. If anybody experiences problems comment out the method bodies of GVRViewManager.onPause and GVRViewManager.onResume. I am sure it will be fixed quickly so I will not revert the commits. Apps using Oculus are affected.

PR #1873, opened 2018-05-17
support PBR for gltf 2.0 

sushantojal:


GVRf/Framework/framework/src/main/java/org/gearvrf/GVRJassimpAdapter.java
```diff
@@ -743,13 +796,21 @@ private void loadTexture(GVRAssetLoader.AssetRequest assetRequest,
         }
         if (texIndex > 0)
         {
+            if (usingPBR)
```
liaxim: Why does this have to special-cased for GVRPBRShader?

NolaDonato: Because assimp puts multiple textures of the same kind into the file and they are redundant. It causes our code to make layered textures (because that is how they are normally expressed by assimp). The support for GLTF in assimp is in development. We should rewrite JassimpAdapter.java and jassimp.cpp in C++ and have it create GearVRF objects directly. Then we could filter out the crud before it gets to the asset loader instead of make generic Java objects from it and filtering out the crud after.

liaxim:
Looks good to me. The demos are all right. Planning to merge on Thursday.

@NolaDonato Any comments?

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRJassimpAdapter.java
```diff
@@ -743,13 +799,21 @@ private void loadTexture(GVRAssetLoader.AssetRequest assetRequest,
...
+            {
+                assetRequest.onTextureError(mContext, textureKey + " duplicate ignored for PBR renderer" + mFileName, mFileName);
```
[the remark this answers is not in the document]
NolaDonato: This is why we have the test for PBR. The GLTF2 importer generates two diffuse maps in that specific case and we ignore one. In all other cases, we don't ignore that map. We have to call the error callback so the texture is "counted" by the asset loader. Otherwise it will be waiting for that texture to be loaded.

I think a future version of jassimp may handle this differently because I looked at their newer source code. I could try pulling out the newest GLTF 2 importer from jassimp and see if it works with the rest of 4.1.

Issue #271, opened 2015-10-09
glError 0x502 on app relaunch


rahul27:
I've noticed this bug while running apps with the Gear VR developer mode ON. I am able to recreate it quite consistently.

Recreation:
1) Turn on Gear VR Developer mode in Settings.
2) Open a sample app with textures (I've tried EyePicker and Solar System).
3) Once launched (past the splash sceen), hit the back button. This will take you to the "Exit to Oculus Home?" Screen.
4) Now hit back again and all the mesh objects disappear and you get the following errors on logcat:

```
10-09 15:35:46.143: I/gvrf(23413): after TextureShader::render() glError (0x502)
10-09 15:35:46.163: I/gvrf(23413): after TextureShader::render() glError (0x502)
10-09 15:35:46.163: I/gvrf(23413): after TextureShader::render() glError (0x502)
10-09 15:35:46.173: I/gvrf(23413): after TextureShader::render() glError (0x502)
10-09 15:35:46.173: I/gvrf(23413): after TextureShader::render() glError (0x502)
10-09 15:35:46.193: I/gvrf(23413): after TextureShader::render() glError (0x502)
```

followed by

```
10-09 15:35:52.173: I/gvrf(23413): after CustomShader::render() glError (0x502)
10-09 15:35:52.173: I/gvrf(23413): after CustomShader::render() glError (0x502)
10-09 15:35:52.173: I/gvrf(23413): after CustomShader::render() glError (0x502)
10-09 15:35:52.173: I/gvrf(23413): after CustomShader::render() glError (0x502)
10-09 15:35:52.193: I/gvrf(23413): after CustomShader::render() glError (0x502)
10-09 15:35:52.193: I/gvrf(23413): after CustomShader::render() glError (0x502)
10-09 15:35:52.193: I/gvrf(23413): after CustomShader::render() glError (0x502)
10-09 15:35:52.193: I/gvrf(23413): after CustomShader::render() glError (0x502)
```

I am developing on a Note 4 (SM N9100).

rahul27:
Looks like someone has already reported this one:

https://github.com/Samsung/GearVRf/issues/257

PR #1874, opened 2018-05-17
Fix ColliderGroup picking

NolaDonato:
Picking colliders within a group was not working because the child colliders do not have SceneObject owners. This works if the child Colliders have explicitly been initialized to not use the SceneObject (the MeshCollider has a mesh attached, the SphereCollider has a radius). 
Some apps added a ColliderGroup and then a MeshCollider with no initialization. This didn't work because the child colliders are not attached to a SceneObject. The correct fix is to pass the owner of the ColliderGroup to the child colliders. If they are not initialized, they will now behave as if they are directly attached to the SceneObject of the ColliderGroup.


liaxim:
GVRSceneObject.attachComponent no longer has the change from #1863 to throw an exception. This is why gvr-3dcursor and the rest work with this pull request.

ragner:
Hi all,
Why we don't add support to add multiple components of same type to a SceneObject?
We are facing some issues that looks necessary this support.

At our Rag doll  implementation we figured out that we need support to multiple constraints attached to a "rigidbody". 

We have spatial audio source implemented as component also. It is not possible add multiple audio sources to the same SceneObject instance.

So we could have components of type unique and others of type multiple.

I am not sure whether looks good have a group for every component type that needs multiple instance attached to a scene object.

NolaDonato:
Components are attached to the scene object. The way you get a component from an object is to request it by type. For example, GVRSceneObject.getRenderData() is really GVRSceneObject.getComponent(GVRRenderData.getComponentType()).
With this method, users can define their own components and attach them to the object and get them back.

The way GearVRf handles the multiple case for the GVRCollider type is to have a GVRColliderGroup which is itself a collider. It has children that are also colliders. This approach allows multiple types of colliders as well as allowing multiple colliders to be attached to the same scene object. I think this is pretty much how you would have to implement your "multiple" type.

I suggest we define a GVRComponentGroup interface that implements the "multiple" type in a more general way. GVRColliderGroup would implement this interface to add components of the same type as children (which is what it does now). Your rag doll could have a GVRConstraintGroup component that also implements this interface. 

In the case of a single component, getComponent() returns that component. In the case of a group component, it would return the group. Then applications could ask for a specific child, or add children, etc.

Does this work for your use case?

ragner:
Whether you don't agree with my suggestion
In my opinion the class GVRComponentGroup could be a Generic class 
to avoid we have to implement a "group class type" 
like GVRColliderGroup and GVRConstraintGroup 
for each "multiple"  type.


```
public class GVRComponentGroup <T extends GVRComponent> extends GVRComponent {

    private final List<T> components = new ArrayList<T>();

    protected GVRComponentGroup(GVRContext gvrContext) {
        super(gvrContext, NativeComponentGroup.ctor());
    }

    public void addComponent(T component) {
        components.add(component);
        NativeComponentGroup.addComponent(getNative(), component.getNative());
    }

    public void removeComponent(T component) {
        components.remove(component);
        NativeComponentGroup.removeComponent(getNative(),
                component.getNative());
    }

    public int getChildCount() {
        return components.size();
    }

    public T getChildAt(int i ) {
        return components.get(i);
    }
}

class NativeComponentGroup {
    static native long ctor();

    static native float[] getHit(long group);

    static native void addComponent(long group, long item);

    static native void removeComponent(long group, long item);
}
```

NolaDonato:
The issue with this is that GVRColliderGroup needs to inherit from GVRCollider so it will be accessible with GVRSceneObject.getComponent(GVRCollider.getComponentType()). This is why GVRComponentGroup should be an INTERFACE. Java does not support multiple inheritance.

ragner:
Hi Nola,
Whether ColliderGroup **needs** to inherit from GVRCollider I see your point.

But for my case a just need a container that allows multiple component of same type. I don't need some "specialized" group.

About GVRSceneObject.getComponent(componentType)) we could fix it just with something like:

```
public class GVRComponentGroup <T extends GVRComponent> extends GVRComponent {
...
    protected GVRComponentGroup(GVRContext gvrContext, long type) {
        super(gvrContext, NativeComponentGroup.ctor(type));
    }
...
    public void addComponent(T component) {
        if (component.getType() != mType) {
            throw new UnsupportedOperationException("Invalid type");
        }

        components.add(component);
        NativeComponentGroup.addComponent(getNative(), component.getNative());
    }
}

class NativeComponentGroup {
    static native long ctor(long type);
...
}
```

NolaDonato:
If we are going to change the architecture and the API,  we should do so with conceptual integrity. If you can successfully implement ColliderGroup on top of NativeComponentGroup in C++ **without using multiple inheritance**, this change will fit with the existing model and allow some extensibility.

Note that the native implementation of the component group must be different for each type of component. GVRComponentGroup<GVRCollider> has to construct a different class of C++ object than GVRComponentGroup<Constraint> even though they call the same Java template. A group of colliders is NOT the same class a group of constraints in C++.  I suggest a C++ class hierarchy like this:
ColliderGroup: 
    inherits from ComponentGroup, component type COLLIDER, overrides addCollider
ComponentGroup: (your new class)
    inherits from Component, component type - whatever you pass the constructor
Collider:
    inherits from Component, component type COLLIDER (same as it is now - no changes)

I think you can just  implement addCollider and removeCollider on top of ComponentGroup. You will need to do this in Java as well.

NolaDonato:
@ragner, you can make these changes on top of this pull request. I will add you as a collaborator. I have added a test for the collider group in PR #271 in GearVRf-Tests. I have tested these changes and all picking tests pass.

ragner:
Ok.

NolaDonato:
Dont merge until Ragner finishes and makes sure the new changes work for constraints.

ragner:
Dear @NolaDonato,
I have finished for colliders.

I am looking forward for your review.

NolaDonato:
Your change completely eliminated the ability for collider groups to do hit testing and modified the API so any app that used collider groups would not work. I have fixed it.

ragner:
Hi @NolaDonato,
My change doesn't remove the ability for collider groups to do hit testing.
Collider groups should continue working after my changes.

Do you have any test case where it does not work as expected!?

NolaDonato:
I added a collider group picking test in https://github.com/gearvrf/GearVRf-Tests/pull/271
In order for ColliderGroup to act as a collider to the picker, it must inherit from Collider so its hit testing functions will be called.

In order for GVRColliderGroup to be returned when GVRSceneObject.getComponent(GVRCollider.getComponentType()) it has to have a type of GVRCollider.getComponentType()

liaxim:
Is this ready for review and test?

ragner:
Hi @liaxim yes it is ready.
My idea of a generic GVRComponentGroup doesn't look good. We just should remove it from this pull request. 

Will try a new pull request today with physics constraints. I will make constraint not exclusive component type so will be possible add more than one constraint to the same scene object. 

So GVRf's API will have a behavior like Unit3D.

NolaDonato:
This PR has been scaled back to no longer include component groups. The basic building blocks for them are still there: NativeComponent has functions to add and remove child components. Only component groups will ever need to call these functions. ColliderGroup is a component group, it overrides addChildComponent and removeChildComponent to implement grouping. Other groups can do the same in C++. The default implementation of these functions is to do nothing.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRComponent.java
```diff
@@ -144,7 +143,15 @@ public void setEnable(boolean flag) {
             onDisable();
         }
     }
-    
+
+    /**
```
liaxim: There is no GVRComponentGroup. should this method stay?

NolaDonato: no, i will remove it

NolaDonato:
I have removed SelectableGroup. As far as I can tell, it no longer does anything after our extensive changes to the 3D cursor library.

liaxim:

PR #1875, opened 2018-05-18
Regenerate shader when texture is moved, fix spotlight

NolaDonato:
Set ShaderData dirty bits to NEW_TEXTURE when texture is removed to force regeneration of the shader.
Apply attenuation to spotlights  (shader fix).

This fixes some failures in the RenderShaderTests


liaxim:

PR #1876, opened 2018-05-18
Constraints API refactor

crbozz:
In order to make it easier to add multiple constraints in the same rigid body the constraints API was changed. Now the constructors receive both the rigid bodies and constraints are no longer derived from GVRComponent. Thus constraints are not attached to scene objects, instead they are viewed simply as rigid bodies links.

Some other changes contained in this pull request:

- slider constraint construction now receives rotation for both rigid bodies;
- physics loader now automatically adds mesh collider to scene objects that do not have this component;
- physics loader can now load bullet file not only from assets folder.

### DIFF
--- a/GVRf/Extensions/gvrf-physics/src/main/java/org/gearvrf/physics/GVRConeTwistConstraint.java
+++ b/GVRf/Extensions/gvrf-physics/src/main/java/org/gearvrf/physics/GVRConeTwistConstraint.java
@@ -22,16 +22,17 @@ import org.gearvrf.GVRContext;
  */
 
 /**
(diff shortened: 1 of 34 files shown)


crbozz:
We conclude these changes to be needed when we start working in a ragdoll sample. The limitation of one constraint per scene object would require some extra efforts mainly from the artists.

NolaDonato:
Please read the discussion in PR #1874 (ColliderGroup). We are going to implement GVRComponentGroup so you can collect multiple constraints together more easily. Please have GVRConstraint inherit from GVRComponent so you can use this new feature.

PR #1884, opened 2018-05-24
support new Bitmap configs 

sushantojal:
Android API 26 introduces new Bitmap configs, RGBA_16 and HARDWARE. After we create a bitmap, we pass it down to the native side and give the pixel data to GL using glTexImage2D. The format info required for this call was being extracted from AndroidBitmap_getInfo. This was returning an unknown format if the original bitmap format was RGBA_16, since the NDK does not expose these formats.
This issue was not present in Android N because any bitmap for format RGBA_16 was being converted automatically to RGBA_8888 (most probably during decodeStream()). 
This PR passes the bitmap format from java to the native side, and select the corresponding GL formats there, if it exists. 
Also we check if the formats we are passing through to the native side, are one of the supported formats. If not, we attempt to create a new bitmap with a supported format.


GVRf/Framework/framework/src/main/java/org/gearvrf/GVRBitmapImage.java
```diff
@@ -125,9 +129,27 @@ public GVRBitmapImage(GVRContext gvrContext, int width, int height, byte[] grays
...
+        NativeBitmapImage.updateFromBitmap(getNative(), bmap, bmap.hasAlpha(), bmap.getConfig().name());
+    }
+
+    private Bitmap getBitmapSupported(Bitmap orig)
+    {
+        Bitmap supBitmap = Bitmap.createBitmap( orig.getWidth(), orig.getHeight(), Bitmap.Config.ARGB_8888 );
+        Canvas c = new Canvas(supBitmap);
+        Paint p = new Paint();
+        c.drawBitmap(orig,0,0,p);
+        orig.recycle();
```
liaxim: Please don't recycle the original immediately. It won't be recycled if conversion is not necessary so lets not introduce special case side-effects.

liaxim:
Looks good to me. Planning to merge on Thursday.

PR #1885, opened 2018-05-24
Improvements to physics loader

crbozz:
- Bullet file now can be loaded from anywhere in the device;
- loader automatically creates default collider when the scene object does not have it.


liaxim:
Looks good to me. Planning to merge on Friday.

PR #1892, opened 2018-05-31
backend_oculus: fix controller reader npe

liaxim:
Looks like the NPE caused by the recent controller change is easy enough to fix. @roshanch Please review and verify though.


roshanch:
if ovrControllerReader is instantiated in constructor, it will cause timing issues. I think @NolaDonato or @dnvasilev can comment here. I would say adding null check in viewManager would be proper fix, though i couldn't reproduce the timing issue now.

liaxim:
Let's identify the timing issues. Null check in ViewManager may not be enough since onSurfaceCreated and onResume can happen in different order; so you may never call the reader from onResume.

liaxim:
Now for all backends the reader is instantiated from the ctor; oculus was the exception for unknown reason. Also whatever OvrControllerReader's ctor does is pretty innocent allocation and initialization. I will deal with these timing issues if they are there.

liaxim:
Looked back at the history. Looks like whatever race there is, it is about when a reader is attachReader-ed to the controller and not when the reader is instantiated.

PR #1893, opened 2018-05-31
Multiple constraints to same scene object

ragner:
Change Component API to support multiple component of same type(not exclusive):
 1) GVRComponent::isExclusive() - To check if the component is of type exclusive. 
            It is not possible add more than one component of same type when it is exclusive. Whether the the component is not exclusive you can add as much as you need. 
2) GVRComponent::next() -   Returns the next component os same type attached to the same scene object. Will be always null for exclusive component.

Makes all constraints not exclusive type:
3) GVRConstraint::isExclusive() - Returns false.
  
Test to this pull request:
https://github.com/gearvrf/GearVRf-Tests/pull/276

After this PR the attach/detach of components  will work almost like Unity3D api.

### DIFF
--- a/GVRf/Extensions/gvrf-physics/src/main/java/org/gearvrf/physics/GVRConeTwistConstraint.java
+++ b/GVRf/Extensions/gvrf-physics/src/main/java/org/gearvrf/physics/GVRConeTwistConstraint.java
@@ -41,8 +41,10 @@ public class GVRConeTwistConstraint extends GVRConstraint {
      */
     public GVRConeTwistConstraint(GVRContext gvrContext, GVRRigidBody rigidBodyB, final float vortex[],
                            final float bodyRotation[], final float coneRotation[]) {
-        super(gvrContext, Native3DConeTwistConstraint.ctor(rigidBodyB.getNative(), vortex,
+        this(gvrContext, Native3DConeTwistConstraint.ctor(rigidBodyB.getNative(), vortex,
                 bodyRotation, coneRotation));
+
+        mBodyB = rigidBodyB;
     }
 
     /** Used only by {@link GVRPhysicsLoader} */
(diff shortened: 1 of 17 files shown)


NolaDonato:
I do not see any reason for the concept of "exclusive". Any group of components is "exclusive", there is no need to have a group of components of different types. In PR #1874 addChildComponent checks to see if the type of the child is the same as the type of the group. That should be all you need.

ragner:
Hi @NolaDonato,
I am just proposing for the framework the same behavior than Unit3D for components attachment.

I am adding the concept of "exclusive" just to identify components like GVRTransform that allows just a single instance attached to same scene object. 

Colliders, Constraints, spatial AudioSource, etc are "not exclusive".

NolaDonato:
So you want a flag to identify groups vs single components? This is why I proposed to make GVRComponentGroup an interface. Then it would be easy to identify groups. I think this change adds little to no overhead in Java and uses the language facilities rather than a separate flag to make the distinction.

NolaDonato:
OK I added an interface GVRComponent.IComponentGroup to PR #1874. It contains addChildComponent and removeChildComponent. If your component type implements this interface, it is "non-exclusive". You can check this with (mycomponent instanceof GVRComponent.IComponentGroup) in Java. GVRColliderGroup implements this interface. Your constraint groups can as well. This is the initial solution I proposed a few weeks ago. It is cleaner than adding special flags or having C++ multiple inheritance. It lets the C++ classes implement adding the components any way they want.

liaxim:
Needs rebase

ragner:
Hi @liaxim, it
doesn't look good to me the ideia of different Group class for every "non-exclusive" component like ConstraintGroup, ColliderGroup, SpatialAudioGroup, etc. It looks hard to mantain a lot of groups.

In my opinion should be possible attache multiple "non-exclusive" components to a scene object or we should have a "Generic" ComponentGroup that works for all "non-exclusive" components and we should handle the special cases with specialized group types. 

But it is my opinion and probably I may be wrong.

liaxim:
@ragner It is a reasonable request. Would something like https://github.com/Samsung/GearVRf/pull/1911 help? I don't want to go into the collider group now so there will be some overlap.

ragner:
Hi @liaxim, #1911 looks nice to me.
Thanks

ragner:
After #1911 this PR looks not necessary anymore.
I will try other on top of #1911 that looks good to me,
thanks

PR #1894, opened 2018-05-31
Makes collider a not exclusive component

ragner:
Whether constraints PR  https://github.com/Samsung/GearVRf/pull/1893 looks good we could make the collider a not exclusive type also.

To test this just add multiple colliders to the same scene object. 

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRComponent.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRComponent.java
@@ -40,7 +40,11 @@ public class GVRComponent extends GVRHybridObject {
 
     protected boolean mIsEnabled;
     protected long mType = 0;
-    protected GVRComponent mParent = null;
+    protected boolean mExclusive = true;
+    /* mNex is used by scene object to handle the collision of not
+       exclusive components attached to the same scene object.
+     */
+    protected GVRComponent mNext = null;
 
     /**
      * Constructor for a component that is not attached to a scene object.
@@ -186,6 +190,24 @@ public class GVRComponent extends GVRHybridObject {
     }
 
     /**
+     * @return true whether it is a exclusive component type, otherwise returns false.
+     */
+    public boolean isExclusive() {
+        if (getNative() != 0)
+            return NativeComponent.isExclusive(getNative());
+
+        return mExclusive;
+    }
+
+    /**
+     * @return The next component of same type attached to the same scene object
+     *      if it is not a component of type exclusivej, otherwise returns null.
+     */
+    public GVRComponent next() {
+        return mNext;
+    }
+
+    /**
      * Get the transform of the scene object this component is attached to.
      * 
      * @return GVRTransform of scene object
@@ -253,4 +275,5 @@ class NativeComponent {
     static native void setEnable(long component, boolean flag);
     static native void addChildComponent(long component, long child);
     static native void removeChildComponent(long component, long child);
+    static native boolean isExclusive(long component);
 }
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRSceneObject.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRSceneObject.java
@@ -273,9 +273,10 @@ public class GVRSceneObject extends GVRHybridObject implements PrettyPrint, IScr
         {
             for (GVRComponent component : mComponents.values())
             {
-                if (component != null)
+                while (component != null)
                 {
                     component.setOwnerObject(null);
+                    component = component.mNext;
                 }
             }
             mComponents.clear();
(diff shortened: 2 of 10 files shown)


NolaDonato:
What is an "exclusive" component and why do we need this in our API?

ragner:
Hi @NolaDonato, this is just my proposal to support multiple component of same type at same scene object.

After  PR  https://github.com/Samsung/GearVRf/pull/1893 a "exclusive" component is a component that doesn't allow multiple of same type attached to the same scene object. So a "not exclusive" component allows you attach as much as you need of same type to the same scene object.

Collider is a not exclusive component (GVRCollider::isExclusive() returns false). After this PR you could add as much collider as you need to same scene object.

I know this PR may conflict with ColliderGroup logic but I would like to try a solution that fits all component type.

NolaDonato:
So your idea of an "exclusive" component is a regular GVRComponent subclass that does not have children. A "non-exclusive" component is a GVRComponent which allows chiildren of the same type. Unfortunately, this does not fit the collider model or the component model. If you attach multiple components of the same type to a scene object, what is returned by GVRSceneObject.getComponent()?

ragner:
Hi @liaxim,
as you can see at my sample code above this PR is an alternative solution to the "group" because
in my opinion the framework should support "non-exclusive" components attached to the same scene object. With this PR the groups are not necessary anymore.

In my opinion the framework should support something like:

```
obj.attachComponent(colliderA)
obj.attachComponent(colliderB)
obj.attachComponent(colliderN)
```

But @NolaDonato rejected this proposal. :(

ragner:
I want to add different instance of same component that is why I cannot give them different type.
For example, different instances of spatial audio component playing different sources attached to same scene object.
  
With this PR GVRSceneObject.getComponent(componentType) returns the first component attached to scene object. To go through all components of same type attached to scene object you need go to its next:

```
obj.attachComponent(constraintA)
obj.attachComponent(constraintB)
obj.attachComponent(constraintN)

c = obj.getComponent(constraint type);
// go through all constraints attached to obj
while(c != null) {
   c.setImpulse(1.0f);
  c = c.next();
}
```

The advantage is that this solution cover all components
not is necessary implement a group for each type of "non-exclusive" component  anymore.

But in my opinion the main point is if we are going to implement a specialized group for each  "non-exclusive" component or if we are going to implement a common code that works for all components. 

Anyway I have not found any physics SDK with constraint group but  I am working on ConstraintGroup to GVRf.

NolaDonato:
The reason we don't have multiple components on a scene object is performance. In the C++ layer, SceneObject::getComponent is used in innermost loops. We expect the number of components on a scene object to be small and we put the Transform and RenderData components first in the list because they are used by the renderer.

Originally GearVRf did not have components at all. It was actually faster because each "component" was directly tied to the scene object. This approach was not very extensible so we added a "lite" version of components which sacrifices a bit of performance but provides extensibility. The GearVRF model is simpler than Unity but still allows the user to write their own components.

Multiple components on a scene object is one of the more confusing things in Unity. Allowing multiple components means you have to prohibit the user from doing it (adding more API complexity) in the cases where it doesn't make sense. Transform, RenderData, RenderTarget, Camera, Light are all components used by the renderer which assumes there is only one per scene object. The user will have to write code to handle this complexity. An extra exception that hardly every occurs means they write an exception handler that is rarely called.

The "common code" we are talking duplicating is a list of components with the same type. There are good reasons for keeping the list of constraints separate from the list of colliders. The picker is more efficient if all of its collider pointers are kept together for cache coherency. Having pointers to other stuff in between is inefficient. Picking is done every frame. The same goes for rendering. It is done every frame and getTransform(), getRenderData() are called repeatedly and skipping over 4 colliders and 3 constraints before we get to the light is not performant. For the same reason, the ColliderGroup keeps the colliders for a single scene object together. I suspect keeping the constraints grouped together in the way they will be accessed in the per-frame loop might apply to the physics code as well.

The user has to write different code for the different cases anyway. You have to know it is a group if you are going to write a loop. You have to know it is not a group to just call getComponent.

IComponentGroup group;
group.addChildComponent(constraintA)
group.addChildComponent(constraintB)
group.addChildComponent(constraintN)
obj.attachComponent(group);

GVRComponent c;
// go through all constraints attached to obj
while (c != group.next()) {
   c.setImpulse(1.0f);
   c = group.next();
}

ragner:
Dear @NolaDonato  and @liaxim,
my code doesn't increase the list of components in C++ or Java layer.

So about "skipping over 4 colliders and 3 constraints before we get to the light is not performant."
that doesn't happen.

Whether you take a look at my code you will see that.

ragner:
About "The picker is more efficient if all of its collider pointers are kept together for cache coherency. ", yes my code does this.

NolaDonato:
Take a look at PR #1904. It make component groups iterable from Java. It requires no changes to C++. Does this work for your constraints?

PR #1895, opened 2018-05-31
framework: can't load libjav8.so; uses static standard lib

liaxim:
Can't load; the static standard lib breaks other modules.

@mwitchwilliams Until @xcaostagit rebuilds jav8 to use the shared standard library it cannot be used. Meaning x3d scripting will be broken.


mwitchwilliams:
OK, so I just retested master that I pulled this morning (Thurs., May 31), and the JavaScript did not work, which sounds consistent with your note above, @xcaostagit .  Other non-JavaScript demo's seem to work.

PR #1896, opened 2018-06-04
Do not modify loaded rigid bodies

crbozz:
Do not change collision form and other settings from loaded rigid bodies, increasing accuracy when authoring physics.


liaxim:
Planning to merge on Thursday

PR #1902, opened 2018-06-12
Call updateTexImage() only after new frame available

ragner:
Register OnFrameAvailableListener to call updateTexImage() only  after a new frame is available.


liaxim:
@ragner What is the problem this fixes? How can we test and verify? Thanks.

ragner:
Hi @liaxim,
I am changing our code to use OnFrameAvailableListener.  OnFrameAvailableListener is notified by SurfaceTexture for every new frame available so we can call `mSurfaceTexture.updateTexImage()` properly.

This change should avoid some unnecessary call to `mSurfaceTexture.updateTexImage()` but I didn't figure out how to test it :(

https://developer.android.com/reference/kotlin/android/graphics/SurfaceTexture.OnFrameAvailableListener

liaxim:

PR #1903, opened 2018-06-12
Fixing travis build issue

apjagdale:
Automated build failed since only_daydream option tried building platformsdk_support.


liaxim:
You could add a comment why the check is necessary as it may not be obvious. Or you could simply check for the OVR_PLATFORM_SDK variable instead.

apjagdale:
@liaxim Added check for OVR_PLATFORM_SDK. So for travis we build framework and extensions with daydream backend. There was a check before for not building platform_sdk when build is set for only_daydream.

liaxim:
I mistakenly removed that check when made the change to use backend_xxx instead of the other flags.

PR #1864, opened 2018-05-09
X3D PlaneSensor implementation

mwitchwilliams:
Added X3D &lt;PlaneSensor&gt; implementation, including able to be used with JavaScript in the &lt;Script&gt; node.

Tests with documentation is available at https://github.com/gearvrf/GearVRf-Tests/pull/269

GearVRf-DCO-1.0-Signed-off-by: Mitch Williams
m1.williams@partner.samsung.com

PR #1904, opened 2018-06-12
make component groups iterable

NolaDonato:
Allow Java to iterate over the components of a group, make IComponentGroup inherit from Iterable

This PR addresses the issues brought up in PR #1864 without having to change the C++ code at all. Class GVRComponent.Group supplies the Java implementation for an iterable group which all component groups can share. This uses Java's language facilities instead of inventing our own scheme for iterating the groups.

Now you can do as follows:

IComponentGroup<XXX> group;
group.addChildComponent(child1);
group.addChildComponent(child2);
for (XXX c : group) {  do something }


NolaDonato:
Merged into PR #1911

PR #1913, opened 2018-06-19
moving updatePosData to GVRViewManager

roshanch:

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRViewManager.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/GVRViewManager.java
@@ -508,6 +508,7 @@ abstract class GVRViewManager extends GVRContext {
     protected void beforeDrawEyes() {
         GVRNotifications.notifyBeforeStep();
         mFrameHandler.beforeDrawEyes();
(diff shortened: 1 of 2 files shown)


roshanch:
i think putting it on back-end side would be better. But there might be better place for it. @liaxim what do you think about it.

liaxim:
I believe there should be a generic facility to query a backend adapter (backend_oculus, backend_daydream) for a specific model asset so the appropriate model can be shown if you use Oculus or Daydream. The actual assets will be bundled with the backend adapter (backend_oculus, backend_daydream).

So the backend_oculus adapter will bundle assets for the Oculus controller and the backend_daydream adapter for the Daydream controller. In that sense your pull request looks good to me.

liaxim:
It means gear_vr_controller.obj and associated assets will have to move to the backend_oculus adapter.

NolaDonato:
We should test it with a model coming from the back end. I am not sure if it will work.

roshanch:
i have requested models from Mike. I will update PR once I get them.

liaxim:

We can add the models in a subsequent pull request. Btw, I am not sure updatePosData is really needed. Will look into it as part of another pull request that is in progress.

PR #1917, opened 2018-06-22
Adding support in renderer for blend shapes

NolaDonato:
1. Added GVRMeshMorph component which accepts a set of blend shapes and constructs a texture from them to use with the morphing shader
2. Added morphing support to GVRPhongShader
3. Added GVRFloatAnimator to interpolate through a set of floating point keys
4. Add GVRMorphAnimator to support animation of morphing blend weights

Added gvr-morph demo and morphing tests in separate pull requests


liaxim:
- Javadocs now or later?
- Is GVRMorphController needed?

There is some duplication of GVRAnimationChannel but I guess it is unavoidable.

Vulkan demos that used to work no longer do due to the layout changes (gvr-simplephysics for example). We can deal with this later; we need something that is not so fragile.

NolaDonato:
You are right. We are working on optimizing GVRAnimationChannel. We will try to get rid of GVRMorphController.

PR #1906, opened 2018-06-14
Add new methods to  apply impulses on rigid body

ragner:
This PR add new API to apply impulses on rigid body.

`GearVRf-DCO-1.0-Signed-off-by: Ragner Magalhaes <ragner.n@samsung.com>`

PR #1918, opened 2018-06-22
Improvements to rag doll demo

ragner:
* Improve Component group to forward its owner to the children.
* Add physics dragging support
* Fix for bitwise warning 
* It is on top of https://github.com/Samsung/GearVRf/pull/1906


ragner:
Hi @NolaDonato,
could you comment in the code line where that inconsistency happens?

Because I am not setting the owner without adding the component to a scene object.

NolaDonato:
From GVRComponentGroup:

public void addChildComponent(T child) {
        synchronized (mComponents) {
            mComponents.add(child);
            if (owner != null) {
                child.setOwnerObject(owner);
            }
        }
    }
You are setting the owner of the child component to be the same as the component group. The component group has been added to the scene object. The child object has only been added to the component group, not the scene object. If you want the owner to be propagated FOR CONSTRAINTS, then put that code in GVRConstraint instead of putting it in a general place where it can cause unforeseen side effects.

ragner:
I see your point,
 onAttach should happens when the component is attached to a scene object as you know
but the Component group is forwarding the onAttach() to its children.

ragner:
I don't would like to implement some "special" code in the components to handle group support.

crbozz:
@liaxim I've made some changes as required by @NolaDonato

liaxim:
Looks good to me. @NolaDonato ?

liaxim:

Issue #1155, opened 2017-04-19
Antialiasing on Views inside GVRSceneObject


felislynx-silae:
If i use approach from GVRTextViewSceneObject to "draw" my view which is drawing on canvas for example ring with text inside.
There is absolutely no antialiasing. Flags on paint object have no effect at all Paint(Paint.ANTI_ALIAS_FLAG); or setAntiAlias(true); or even setFilterBitmap(true);.
So main question is this an issue with rendering/implementation which can be fixed or should i discontinue approach and use gl textures etc instead of Android Views?

liaxim:
It is kind of known problem we have not found a solution for. Let me have a look again but I don't expect much..

In the meantime, if the approach GVRConsole is using works for you (https://github.com/Samsung/GearVRf/blob/master/GVRf/Framework/framework/src/main/java/org/gearvrf/debug/GVRConsole.java) then you can try that. I know anti aliasing is on there - you can examine the overlay text shown by the gvr-nonglthreadupdate sample.

felislynx-silae:
@liaxim I'll give "gvrconsole" a try. To be honest previous approach was much more AndroidDev friendly to create customized experience with UI (like menus). I think it would be nice feature for those who choose GearVRf instead of Unity/Unreal ;). I'll Let you know about progress with converting GVRConsole to UI elements (i'll work on sample app which we used to detect bug before)

liaxim:
How do you find the https://github.com/gearvrf/GearVRf-Demos/tree/master/gvr-renderableview sample? Is there a lot of aliasing for you? The console is just an option; we will try to figure out something for the views.

liaxim:
I actually think antialiasing is always on. Tried few different Android versions and phones. It might be useful if you could share an example app with us. Screenshots would be helpful too. Thanks.

We are aware of some problems with text not being legible but preferably we should look at your specific case.

felislynx-silae:
@liaxim https://github.com/felislynx-silae/GearBugSample/tree/develop <-- here you have sample view with AA issues. I've checked GVRConsole but it won't help in this case. GVR-Renderableview sample was something that i've tried before current approach but it's difficult to set every view to be less cubic more 2D (you will see what i mean in sample)

thomasflynn:
[...]

PR #1929, opened 2018-06-29
Fixing two post effect

apjagdale:
Wrong number of parameters were passed.

### DIFF
--- a/GVRf/Framework/framework/src/main/jni/view_manager_jni.cpp
+++ b/GVRf/Framework/framework/src/main/jni/view_manager_jni.cpp
@@ -43,7 +43,6 @@ extern "C" {
                                                       jlong jrenderTarget, jlong jscene,
                                                       jobject javaSceneObject,
                                                       jlong jshader_manager,
-                                                      jlong jpost_effect_shader_manager,
                                                       jlong jpost_effect_render_texture_a,
                                                       jlong jpost_effect_render_texture_b)
     {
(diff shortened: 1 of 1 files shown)


ragner:
It is working for me. 👍 
Thanks

ragner:
My test at gvr-blurfilter from https://github.com/ragner/GearVRf-Demos/tree/blurfilter2Pass
may be showing some side effect of this PR.

It is showing the blurred image in front of the camera instead of on the scene object.

liaxim:
@ragner You need to update pos_tex.vsh in similar fashion:
```
#extension GL_ARB_separate_shader_objects : enable
#extension GL_ARB_shading_language_420pack : enable

@MATRIX_UNIFORMS

precision mediump float;
layout ( location = 0 ) in  vec3 a_position;
layout ( location = 1 ) in vec2 a_texcoord;

layout ( location = 0 ) out vec2 vTextureCoord;

void main()
{
  vTextureCoord = a_texcoord.xy;
  gl_Position = u_mvp * vec4(a_position, 1.0);
}
```

liaxim:
@ragner Bear in mind that you will be rendering the same scene-object three times with different settings on top of each other. You are not going to combine the vertical and the horizontal blur using GVRRenderPass-es.

You might want to examine the https://github.com/Samsung/GearVRf/issues/1155 issue and the convolution shader @thomasflynn added there. It only needs a small patch to pass the mvp to the shader. I had a note to make the solution to that issue built-in "feature" for android views but looks like you are doing something like that already. Please consider making it part of the core framework. Thanks.

liaxim:
Planning to merge on Friday, it restores the gvr-blurfilter to its 3.x functionality. @ragner I think you have other, separate issues too we could deal with in additional PRs.

ragner:
Hi @liaxim about https://github.com/thomasflynn/GearVRf-Demos/blob/blur/gvr-events/app/src/main/res/raw/convolution_fragment.fsh

The convolutionMatrix is not been used in the shader. Should I just remove it?

liaxim:
It has been a while since I looked at this code but if that matrix is unused, then go ahead and remove it.

thomasflynn:
yeah, i originally intended to be able to update the convolution matrix via uniform, but had troubles with it and just ended up hard-coding it in the shader.  if that's fine with you too, then go ahead and remove the convolution matrix.

ragner:
Hi @thomasflynn and @liaxim ,
I have got some nice result after some small adjustments ... nice shader!
I will make a PR soon,
thank you.

PR #1930, opened 2018-07-02
Add GVRApplication allowing making apps not derived from GVRActivity

liaxim:
Demos PR: https://github.com/gearvrf/GearVRf-Demos/pull/626

Doesn't change the fact GVRf's backend adapters create full screen SurfaceViews/GLSurfaceViews. So it doesn't make GVRf more of a library than before. Just removes the requirement to derive from GVRActivity.

Provide GVRActivity implementation on top of GVRApplication that preserves API compatibility for older apps.

Small API changes since use of GVRActivity is no longer mandatory.

---


liaxim:
Pending javadocs. Works with all backends, tests run the same.

liaxim:
Thanks. Planning to merge on Friday.

PR #1932, opened 2018-07-03
gradle: include the support for vulkan validation layers only in the debug builds

liaxim:
Removes a bunch of libVk shared libraries from release builds.


liaxim:
Planning to merge on Friday. Validation libVk binaries will be removed in release builds.

PR #1935, opened 2018-07-06
Fix for view's anti-aliasing issue

ragner:
Anti-aliasing issue at form of gvr-keyboardview demo should be fixed.

GVRKeyboardS.O. will be fixed by other PR.

Use GVRViewSceneObject#setTextureBufferSize(int size) to ajust according to your application
layout to fix the anti-aliasing issue. The default buffer size should cover almost all issues.


liaxim:
Looks good and works as expected. Planning to merge on Thursday.

PR #1936, opened 2018-07-09
X3D: Fixed errors with mesh having normals; new cursor setting.

mwitchwilliams:
Corrected two issues: Mesh with no normals (see demo below); and set the gaze cursor from default to be on surface.

[NoNormalsDemo.zip](https://github.com/Samsung/GearVRf/files/2177992/NoNormalsDemo.zip)

Mitch Williams <m1.williams@partner.samsung.com>

### DIFF
--- a/GVRf/Extensions/x3d/src/main/java/org/gearvrf/x3d/X3Dobject.java
+++ b/GVRf/Extensions/x3d/src/main/java/org/gearvrf/x3d/X3Dobject.java
@@ -555,7 +555,7 @@ public class X3Dobject {
             {
                 descriptor += " float2 a_texcoord";
             }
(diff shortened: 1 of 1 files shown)


liaxim:
This is what I see when I run the attached demo:

shot

Is that expected?

mwitchwilliams:
Interesting, because it worked ok for me, including the slider bar interactivity.
screenshot_20180713-101020
Not sure where this leaves us.

liaxim:
I will try again on different phones.

liaxim:
Ok, not a problem with this pull request. Planning to merge this one on Monday.

The problem is fixed by https://github.com/Samsung/GearVRf/pull/1943.

PR #1938, opened 2018-07-13
import blend shapes from gltf 2.0 file

sushantojal:
The vertex data (position, normal, tangents)  of all the blend shapes of a mesh are being imported from the gltf file and being saved in a GVRVertexBuffer List in the corresponding GVRMesh.
The changes in the native assimp code are in a pull request [here](https://github.com/NolaDonato/assimp/pull/2)

GearVRf DCO signed off by: Sahithi Podila s.pod@samsung.com 
Nola Donato nola.donato@samsung.com
Sushant Ojal sushant.o@samsung.com


sushantojal:
Still having some issues building the 64 bit libassimp.so. Will commit that as soon as I figure it out (along with compressed prebuilt .so with symbols)

liaxim:
Not seeing side-effects. We could merge after the 64bit binary is included.

NolaDonato:
If you made API changes to assimp, please put the new include files in framework/src/main/jni/contrib/assimp.include.assimp

sushantojal:
I enabled the mesh morph component to be attached to the scene object at the time of object import.
Also, no API changes were made to assimp. The only header files changed were gltf2asset.h and gltf2importer.h

sushantojal:
Merged changes belonging to @NolaDonato and @spodila1 changes as well.

sushantojal:
yes this is ready for review, should be tested with the PR for the gvr-morph demo.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRJassimpAdapter.java
```diff
@@ -147,13 +161,22 @@ public GVRMesh createMesh(GVRContext ctx, AiMesh aiMesh, EnumSet<GVRImportSettin
...
-            FloatBuffer bitangentBuffer = aiMesh.getBitangentBuffer();
+
             tangentsArray = new float[tangentBuffer.capacity()];
             tangentBuffer.get(tangentsArray, 0, tangentBuffer.capacity());
-            bitangentsArray = new float[bitangentBuffer.capacity()];
-            bitangentBuffer.get(bitangentsArray, 0, bitangentBuffer.capacity());
+            bitangentsArray = new float[tangentsArray.length];
+            for(int i = 0; i < tangentsArray.length; i += 3)
+            {
+                Vector3f tangent = new Vector3f(tangentsArray[i], tangentsArray[i + 1], tangentsArray[i + 2]);
```
liaxim: Not a big deal but:

1. tangent, normal and bitagent could be allocated outside of the loop once and initialized accordingly in the loop

or

2. skip the vector3fs and directly calculate the cross product 

All that to avoid unnecessary new small object allocations. The GC is though optimized for exactly that so it is more of a style issue. Still, if avoiding the allocations is not overly complicated, why not..

liaxim:
Have you tried S9? I get a black screen and FPS of about 10fps nevertheless (on my eng build S9+). No obvious errors in logcat.

Note8, S8 - ~60fps, works as expected
Note5 - 45-50fps, works as expected

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRShader.java
```diff
@@ -51,7 +51,7 @@
  */
 public class GVRShader
 {
-    protected boolean mWriteShadersToDisk = false;
+    protected boolean mWriteShadersToDisk = true;
```
liaxim: Please set back to false

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRShaderData.java
```diff
@@ -157,7 +157,9 @@ public GVRShaderId getShaderType()
      * </p>
      * @return string with texture descriptor from shader
      */
-    public String getTextureDescriptor() { return mTextureDescriptor; }
+    public String getTextureDescriptor() {
+        Log.i("desctexture","dzs"+mTextureDescriptor);
```
liaxim: Is this trace necessary?

liaxim:
@apjagdale Please review the shader updates. Thanks.

liaxim:
@NolaDonato Should JOIN_IDENTICAL_VERTICES be included in the getRecommendedMorphSettings? The demo doesn't work as expected otherwise: https://github.com/gearvrf/GearVRf-Demos/pull/627.

NolaDonato:
No, if JOIN_IDENTICAL_VERTICES is included the blend shapes will not have the same topology.

liaxim:
Ok, then why https://github.com/gearvrf/GearVRf-Demos/pull/627 doesn't work with the existing morph settings?

sushantojal:
@liaxim didnt find any issues in the CPU profiler when using S9 yet, still looking into it.

liaxim:
@sushantojal Does the app run as expected on your S9?

sushantojal:
no it doesnt, its similar to what you described, black screen ~10 fps, there is no obvious issue on the cpu side

liaxim:
getRecommendedMorphSettings now works.

On S9 frames are rendered at close to 60fps but then vrapi_SubmitFrame blocks for about 100ms. Not that many reasons why it would block, especially for that long. Maybe this:
```
/// This will block until the textures from the previous vrapi_SubmitFrame() have been
/// consumed by the background thread, to allow one frame of overlap for maximum
/// GPU utilization, while preventing multiple frames from piling up variable latency.
```

Interestingly enough, all other backend adapters work but monoscopic. Maybe it would be easier to diagnose with monoscopic since it is simpler?

thomasflynn:
ok, tested on my S9+.  works fine.  the issue with the prod build we should still investigate internally, but i don't think we need to hold up this check-in for that.  anyone disagrees, you can tackle me tomorrow to discuss it.

liaxim:
One todo for me - the 64bit assimp binary.

PR #1948, opened 2018-07-17
Fix receiving of IActivityEvents.onControllerEvent

### LINKED ISSUE #1947: Problem receiving IActivityEvents.onControllerEvent
This is a problem that has come in from the recent GVRActivity change.  I'm registering my event reciever like this:
```

        mEventListener = new GVREventListeners.ActivityEvents() {
            @Override
            public void onControllerEvent(Vector3f position, Quaternionf orientation, PointF touchPadPosition, boolean touched, Vector3f angularAcceleration, Vector3f angularVelocity) {
                ...
                }
            }
        };
        mEventReceiver = gvrContext.getApplication().getEventReceiver();
        mEventReceiver.addListener(mEventListener);

```

I used to use gvrContext.getActivity().getEventReceiver(), but it looks the right thing to do now should be switch to getApplication().getEventReceiver(), instead.

But with this change the events weren't coming through, so I debugged inside sendEventWithMaskParamArray and found the reason they aren't getting delivered is that they are getting rejected here due to having the wrong owner:
```

        if ((sendMask & SEND_MASK_LISTENERS) != 0) {
               ......
                for (IEvents listener : listeners) {
                    // Skip the listener due to different type, or has been removed
                    if (!eventsClass.isInstance(listener) || receiver.getOwner() != target)
                        continue;
               ......
```
The receivers owner is now the GVRApplication instance, but the code in GVRGearCursorController is still trying to send to the Activity:

```
    private void handleControllerEvent(final ControllerEvent event)
    {
        context.getEventManager().sendEvent(context.getActivity(), IActivityEvents.class,
                                            "onControllerEvent",

```
I've tried changing the above to use context.getApplication() and it seems to fix the problem.

liaxim:
Fixes #1947 

### DIFF
--- a/GVRf/Framework/framework/src/main/java/org/gearvrf/io/GVRGearCursorController.java
+++ b/GVRf/Framework/framework/src/main/java/org/gearvrf/io/GVRGearCursorController.java
@@ -518,7 +518,7 @@ public final class GVRGearCursorController extends GVRCursorController
 
     private void handleControllerEvent(final ControllerEvent event)
     {
-        context.getEventManager().sendEvent(context.getActivity(), IActivityEvents.class,
+        context.getEventManager().sendEvent(context.getApplication(), IActivityEvents.class,
                                             "onControllerEvent",
                                             event.position, event.rotation, event.pointF,
                                             event.touched, event.angularAcceleration,event.angularVelocity);


liaxim:
Planning to merge on Thursday

PR #1949, opened 2018-07-19
Fix ActivityDelegateStubs, setMain returns true by default

liaxim:


liaxim:
Some backend adapters may not implement setMain. If it returns always false it means an error occurred.

PR #1950, opened 2018-07-19
GVRf Javascript Extension

xcaostagit:
Javascript, particularly V8, has large footprint (19M + 16M). This change moves most
of the Javascript related code into the "script" extension. Note the current
support of Javascript has serious issues in itself but such code are not
modified by this change.


xcaostagit:
GVRScriptBehavior was deriving from GVRBehavior. However GVRScene needs to send a script event (c.f. GVRScene.java line 565) and it doesn't know the existence of GVRScriptBehavior. So something has to be created inside the framework in order for GVRScene to be happy. Hence GVRScriptBehaviorBase.

xcaostagit:
Regarding GVRScriptManagerBase, it was IScriptManager as an interface. However interfaces will not allow default actions to happen. Although I don't see the need for default actions in certain methods so far, I would leave that option open for now.

xcaostagit:
The NPE crash @liaxim found is now fixed. Please re-test.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRScene.java
```diff
@@ -556,7 +556,7 @@ public void onInit(GVRContext gvrContext, GVRScene scene) {
         private void recursivelySendOnInit(GVRSceneObject sceneObject) {
             getGVRContext().getEventManager().sendEvent(
                     sceneObject, ISceneObjectEvents.class, "onInit", getGVRContext(), sceneObject);
-            GVRScriptBehavior script = (GVRScriptBehavior) sceneObject.getComponent(GVRScriptBehavior.getComponentType());
+            GVRScriptBehaviorBase script = (GVRScriptBehaviorBase) sceneObject.getComponent(GVRScriptBehaviorBase.getComponentType());
```
liaxim: The scripting is not quite completely decoupled from the core framework due to all the support base classes. This PR is a great improvement and allows for footprint reduction. But if you'd like we could explore ways of decoupling completely later. For example, the scripting module could "listen" for the "onInit" event and internally handle it by propagating to scripting consumers - instead of the GVRScene class doing it.

ovr_sdk_mobile
```diff
@@ -0,0 +1 @@
+../ovr_sdk_mobile
```
liaxim: Not needed?

liaxim:
Please add javadocs to new public classes and apis.

ovr_sdk_mobile
xcaostagit: Sorry this is an accident. Will remove in the next update

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRScene.java
xcaostagit: Agreed. The  listener idea is also a great suggestion. The main purpose of this change is for size reduction. Much more improvements in the scripting area need to be implemented.

liaxim:
gvr-x3d-demo needs a dependency to the script extension. Otherwise GVRScriptManager cannot be found.

xcaostagit:
Working on the update for gvr-x3d-demo now. Will have an update soon.

liaxim:
libjav8.so is not loaded. Might have to be dlopen-ed explicitly or System.loadLibrary-ed. Though the latter ends with this error:
```
07-26 11:06:27.114 21721 21721 E AndroidRuntime: FATAL EXCEPTION: main
07-26 11:06:27.114 21721 21721 E AndroidRuntime: Process: org.gearvrf.x3ddemo, PID: 21721
07-26 11:06:27.114 21721 21721 E AndroidRuntime: java.lang.UnsatisfiedLinkError: dlopen failed: cannot locate symbol "__emutls_get_address" referenced by "/data/app/org.gearvrf.x3ddemo-1/lib/arm/libjav8.so"...
07-26 11:06:27.114 21721 21721 E AndroidRuntime: 	at java.lang.Runtime.loadLibrary0(Runtime.java:989)
07-26 11:06:27.114 21721 21721 E AndroidRuntime: 	at java.lang.System.loadLibrary(System.java:1567)
07-26 11:06:27.114 21721 21721 E AndroidRuntime: 	at org.gearvrf.script.GVRScriptManager.<clinit>(GVRScriptManager.java:49)
07-26 11:06:27.114 21721 21721 E AndroidRuntime: 	at java.lang.Class.classForName(Native Method)
07-26 11:06:27.114 21721 21721 E AndroidRuntime: 	at java.lang.Class.forName(Class.java:400)
07-26 11:06:27.114 21721 21721 E AndroidRuntime: 	at java.lang.Class.forName(Class.java:326)
07-26 11:06:27.114 21721 21721 E AndroidRuntime: 	at org.gearvrf.GVRViewManager.initScriptManager(GVRViewManager.java:819)
07-26 11:06:27.114 21721 21721 E AndroidRuntime: 	at org.gearvrf.GVRViewManager.<init>(GVRViewManager.java:65)
```

xcaostagit:
libjav8.so was built with NDK16b and the rest of the code (framework, extensions and demos) are built with NDK17+. They don't work well together causing the UnsatisfiedLinkError. I will fix it in the next update.

mwitchwilliams:
Ran a complete series of JavaScript tests - control Movie (slider bar to frame, pause, re-start), drag objects, X3D anchor to change files, change camera position &lt;Viewpoint&gt;, change Texture map, planeSensor with JS, and scale texture maps - and all worked great.
**Looks Good To Me**

liaxim:
Looks good. If there are no comment will merge on Tuesday.

PR #1954, opened 2018-07-23
Dirty flag of lights not cleared at all

apjagdale:
LIGHT_REMOVED flag when set is never cleared. Causes render data to be dirty every time. 
Issue found by gvr-cubemap demo which clears the scene which sets  LIGHT_REMOVED flag and never cleared.

GearVRf DCO signed off by: Abhijit Jagdale a.jagdale@samsung.com


apjagdale:
@NolaDonato Tested with framework test. They pass. There are few failures but those are irrespective of my change.

liaxim:
Should this section clear the flag too?

```
    if (mDirty & REBUILD_SHADERS)
    {
        return scanLights();
    }
```

apjagdale:
@liaxim No. That flag is cleared from shadersRebuilt() in render_target.cpp

PR #1955, opened 2018-07-27
Fixed Animation bug in X3D

spodila1:
Use an array of floats for setting positions, rotations and scale values.

Tested with asset-tests

GearVRf DCO signed off by: Sahithi Podila s.pod@partner.samsung.com


liaxim:
Picked only the change to x3d/AnimationInteractivityManager.java.

PR #1958, opened 2018-08-03
Stickers

apjagdale:
Added screenshot capture feature by using 3 Pixel Buffer objects.
New class introduced : GVRSticker(GVRContext gvrContext, String tag, long interval)
 - tag for outputting PNG's to the SD Card along with appended frame ID in folder GearVRFScreenshots
 - (interval) Time in milliseconds to take screenshot at that interval

Methods:
 - start/stopCapturing() 

GearVRf DCO signed off by: Abhijit Jagdale a.jagdale@samsung.com


roshanch:
We are using horizontal/ vertical flip shaders, or post effects just for inverting a scene. we should be modifying mvp instead to rotate or flip the scene instead of all that overhead which is not efficient.

liaxim:
Couple of tests I believe would be useful.
There is a backend adapter that no longer compiles.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRSticker.java
```diff
@@ -0,0 +1,150 @@
...
+     */
+    public GVRSticker(GVRContext gvrContext, String tag, long interval){
+        mGVRContext = gvrContext;
+        mTag = tag;
+        mInterval = interval;
+
+        File sdcard = Environment.getExternalStorageDirectory();
+        mDirectory = sdcard.getAbsolutePath() + "/GearVRFScreenshots/";
+        File d = new File(mDirectory);
+        d.mkdirs();
```
liaxim: Probably the fact that WRITE_EXTERNAL_STORAGE is required should be mentioned. Also I'd recommend checking the return value of mkdirs and throwing an exception immediately if needed.

liaxim:
Have you verified screenshotting with all backend adapters?

GVRf/Framework/backend_oculus/src/main/java/org/gearvrf/OvrViewManager.java
```diff
@@ -236,6 +233,7 @@ void onDrawEye(int eye, int swapChainIndex, boolean use_multiview) {
 
                     renderTarget.cullFromCamera(mMainScene, mainCameraRig.getCenterCamera(), mRenderBundle.getShaderManager());
                     captureCenterEye(renderTarget, false);
+                    captureSticker();
```
liaxim: What about other backend adapters?

liaxim:
What about generating stickers in "headless mode"? Where the framework is only used to render into the images as a background "service"? Should there be all new backend adapter for that mode?

liaxim:
Looks pretty good overall.

liaxim:
Needs rebase. Are any of the comments valid?

NolaDonato:
Please generalize beyond stickers. This should be a component you attach to the Scene object that owns the render target to capture textures.

NolaDonato:
Yes that would be useful. We should be able to specify it in terms of frames or seconds

liaxim:
Should we close?

ftservice (commented):
Android system webviwver

PR #1968, opened 2018-08-16
Fix animation finish callback being called every iteration

dsazulay:
Small fix on animation callback. The method finished of GVROnFinish is called only after the last iteration of an animation.

GearVRf-DCO-1.0-Signed-off-by: Diego Azulay <diego.a@samsung.com>


liaxim:
Makes sense. Planning to merge on Wednesday.

PR #1978, opened 2018-09-07
Bug fixes for skeletal animation

NolaDonato:
This patch fixes the messed up animations in gvr-meshanimation and the worm not animating in gvr-controls


liaxim:
gvr-rendertotexture: animation still not fixed
gvr-meshanimation: (minor) the trex is facing the camera instead of its past orientation

PR #1985, opened 2018-09-17
Use precompiled libshaderc

liaxim:
Speeds up the build noticeably. Also makes it possible to use Android Studio 3.3 Canary builds which have a problem otherwise.


liaxim:

Issue #1962, opened 2018-08-13
Null pointer reference in onAddedToScene


SteveGreatApe:
I synced to the latest code and started hitting this crash on startup, 90% of the time I'm getting the crash, the other 10% it starts up enough to draw the view, but then the view is frozen.

Reverting back to just before this commit got things working again for me, https://github.com/Samsung/GearVRf/commit/6206cb0a2614b0bd31f6cdbb1df1e57c7623bafe
```
2018-08-13 17:51:22.242 3346-3346/? I//system/bin/tombstoned: received crash request for pid 7831
2018-08-13 17:51:22.244 7947-7947/? I/crash_dump32: performing dump of process 7831 (target tid = 7875)
2018-08-13 17:51:22.245 7947-7947/? A/DEBUG: *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
2018-08-13 17:51:22.245 7947-7947/? A/DEBUG: Build fingerprint: 'samsung/heroltexx/herolte:8.0.0/R16NW/G930FXXU2ERG2:user/release-keys'
2018-08-13 17:51:22.245 7947-7947/? A/DEBUG: Revision: '8'
2018-08-13 17:51:22.245 7947-7947/? A/DEBUG: ABI: 'arm'
2018-08-13 17:51:22.245 7947-7947/? A/DEBUG: pid: 7831, tid: 7875, name: GLThread 16361  >>> com.greatape.avrgallery <<<
2018-08-13 17:51:22.245 7947-7947/? A/DEBUG: signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0xc
2018-08-13 17:51:22.245 7947-7947/? A/DEBUG: Cause: null pointer dereference
2018-08-13 17:51:22.245 7947-7947/? A/DEBUG:     r0 00000022  r1 c24c61a8  r2 345ad82c  r3 00000000
2018-08-13 17:51:22.245 7947-7947/? A/DEBUG:     r4 00000006  r5 d5bb5480  r6 00000000  r7 c24c61c8
2018-08-13 17:51:22.245 7947-7947/? A/DEBUG:     r8 00000056  r9 d2ea5000  sl c24c6230  fp d2ea5000
2018-08-13 17:51:22.245 7947-7947/? A/DEBUG:     ip e66b1d78  sp c24c6098  lr c3ead7cb  pc c3ead7d6  cpsr 600f0030
[...]

PR #1987, opened 2018-09-19
Bvh animation

spodila1:
Loads BVH Animation and plays on avatar. 

Tested with gvr-avatar demo with the following assets

Model -- YBOT
BvhAnimations -- Samba_Dancing_mixamo.com.bvh, Zombie_Stand_Up_mixamo.com, Football_Hike_mixamo.com

https://github.com/gearvrf/GearVRf-Demos/pull/641

GearVRf-DCO-1.0-Signed-off-by: Sahithi Podila sah.pod@partner.samsung.com


GVRf/Framework/framework/src/main/java/org/gearvrf/GVRShader.java
```diff
@@ -51,7 +51,7 @@
  */
 public class GVRShader
 {
-    protected boolean mWriteShadersToDisk = false;
+    protected boolean mWriteShadersToDisk = true;
```
liaxim: Please set back to false.

liaxim:
Please revert the formatting-only changes to jassimp.cpp.

NolaDonato:
I have also tested it. I have a few changes I would like to add to support importing the skeleton from the BVH file.

liaxim:
The updated gvr-avatar doesn't work for me.

spodila1:
I have not added any code changes in jassimp.cpp. 

I tested with gvr-avatar demo (Note8). It is working fine.

NolaDonato:
I made some minor modifications to the BVH parser so it could handle mixed spaces and tabs. I also exposed the ability to read the skeleton independently from the motion.

I also pushed some of the skeleton and pose data from Java to C++ so physics C++ code can access it. This code will be tested by the retargeting demo I am working on.

NolaDonato:
I also pushed some minor changes to the gvr-avatar demo. It works for me with the YBot character and its BVH animations.

liaxim:
Please test with more demo apps and let us know which. Thanks.

mWriteShadersToDisk is still true. jassimp.cpp still has a ton of formatting-only changes.

GVRf/Framework/framework/src/main/java/org/gearvrf/animation/GVRAnimation.java
```diff
@@ -393,6 +439,12 @@ final boolean onDrawFrame(float frameTime) {
         return stillRunning;
     }
 
+    public void animate(float timeInSec)
```
liaxim: Seems like it doesn't have to be public. Please remove the qualifier or add javadoc. Preferably remove the qualifier.

NolaDonato:
?Actually applications can call it to step the animation programmatically


NolaDonato:
The javadocs are added and a few bugs are fixed. Ready to go. Seems to run the demo OK on my S8.

liaxim:
It is great that gvr-avatar works but a meaningful subset of the demo apps have to be tested. Some using animations are already broken so I dunno about them but anything with physics should be sanity checked. Let me know which demo apps have been verified. Thanks.

NolaDonato:
I tested with immersivepedia, meshanimation, controls and morph

GVRf/Extensions/gvrf-physics/src/main/java/org/gearvrf/physics/GVRPhysicsLoader.java
```diff
@@ -46,7 +46,8 @@
      * @param fileName Physics settings file name.
      * @param scene The scene containing the objects to attach physics components.
      */
-    public static void loadPhysicsFile(GVRContext gvrContext, String fileName, GVRScene scene) {
+    public static void loadPhysicsFile(GVRContext gvrContext, String fileName, GVRScene scene) throws IOException
```
liaxim: gvr-physicsload doesn't compile:

```
GearVRf-Demos/gvr-physicsload/app/src/main/java/org/gearvrf/physicsload/MainActivity.java:194: error: unreported exception IOException; must be caught or declared to be thrown
            GVRPhysicsLoader.loadPhysicsFile(gvrContext, "fixed_slider.bullet", mainScene);
                                            ^
1 error
```

GVRf/Framework/framework/src/main/java/org/gearvrf/animation/GVRAnimator.java
```diff
@@ -202,6 +218,35 @@ public void setRepeatMode(int repeatMode)
             anim.setRepeatMode(repeatMode);
         }
     }
+    public void setOffset(float startOffset)
```
liaxim: Bunch of new public APIs need javadoc.

GVRf/Framework/framework/src/main/java/org/gearvrf/animation/GVRAvatar.java
```diff
@@ -9,14 +9,13 @@
 import org.gearvrf.GVRResourceVolume;
```
liaxim: Needs copyright header.

liaxim:
gvr-cubemap doesn't work for me with this PR. On tap I get a black screen.

liaxim:
gvr-morph, gvr-outline, gvr-videoplayer - slower to load
gvr-cubemap - not all cubemaps show up
gvr-physicsload - doesn't compile
gvr-widgetviewer - some models are all black

Is the avatar supposed to tremble like that?

NolaDonato:
Yes, the avatar trembles because these are videos recorded from a body tracking setup. The original BVH files include the trembling.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRShader.java
thomasflynn: this should probably be false, right?

NolaDonato:
Yes, good catch

GVRf/Framework/framework/src/main/java/org/gearvrf/animation/GVRAnimation.java
```diff
@@ -332,13 +388,16 @@ public GVRAnimation start(GVRAnimationEngine engine) {
      * @return {@code true} to keep running the animation; {@code false} to shut
      *         it down
      */
+
     final boolean onDrawFrame(float frameTime) {
         final int previousCycleCount = (int) (mElapsedTime / mDuration);
 
-        mElapsedTime += frameTime;
+//        mElapsedTime += (frameTime*animationSpeed);
+        mElapsedTime += 1 / 30.0f;
```
thomasflynn: is this something that should be hard-coded?

GVRf/Extensions/gvrf-physics/src/main/java/org/gearvrf/physics/GVRPhysicsLoader.java
thomasflynn: gvr-physicsload no longer compiles because this method now "throws IOException" where it didn't before.  Either this methods needs to not throw IOException, or the demo needs to be updated to catch it.

spodila1:
@thomasflynn : I have updated to GVRScene.java and  tested with the demo 
                          gvr-cubemap working

liaxim:
If we are to remove the libgdx extension, we should also remove the gvr-modelviewer2 demo app. Ok?

thomasflynn:
yes, i think we'll have to do that until it can be updated.

liaxim:
gvr-morph is again much slower to load the model. I fixed that master but it doesn't look like a bad merge problem. It is something new.

liaxim:
While gvr-widgetviewer shows the integration of libgdx, it also loads and renders gltf models. So the fact that it shows all black models means something is broken there. I guess we can merge as-is if that is what you want.

thomasflynn:
we should probably take a look before merging.  which models are coming up black?

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
```diff
@@ -431,6 +431,10 @@ private void generateLoadEvent()
                     animator.start();
                 }
             }
+            synchronized (mNumTextures)
+            {
+                mNumTextures = -1;
```
liaxim: Why -1 and not 0. What is the difference?

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRMeshMorph.java
```diff
@@ -320,6 +317,7 @@ public boolean update()
             blendshapeTex.setImage(blendshapeImage);
             mtl.setTexture("blendshapeTexture", blendshapeTex);
         }
+        Log.v("MORPH", "Blend Shape Diffs\n%s", dumpDiffs(20));
```
liaxim: This log must be removed, right?

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRScene.java
```diff
@@ -156,17 +156,27 @@ public int removeSceneObjectsByName(final String name) {
...
 
         NativeScene.removeAllSceneObjects(getNative());
-        for (final GVRSceneObject child : mSceneRoot.getChildren()) {
+        for (final GVRSceneObject child : mSceneRoot.getChildren())
+        {
             child.getParent().removeChildObject(child);
         }
 
-        if (null != head) {
+        mSceneRoot = new GVRSceneObject(getGVRContext());
```
liaxim: What is this change for? I think it might reverting a fix in master.

GVRf/Framework/framework/src/main/java/org/gearvrf/animation/GVRAnimation.java
```diff
@@ -315,11 +350,32 @@ public GVRAnimation setOnFinish(GVROnFinish callback) {
...
         if (sDebug)
         {
             Log.d("ANIMATION", "%s started", getClass().getSimpleName());
         }
-        return this;
+    }
+
+    protected void onFinish()
+    {
+        if (sDebug)
```
liaxim: Make sDebug final so the compiler can optimize away the branch.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRScene.java
thomasflynn: @NolaDonato i believe 'git blame' indicated that you changed this code.  do you have insight on the change?

GVRf/Framework/framework/src/main/java/org/gearvrf/animation/GVRAnimator.java
```diff
@@ -202,6 +218,27 @@ public void setRepeatMode(int repeatMode)
...
+    {
+        for (GVRAnimation anim : mAnimations)
+        {
+            anim.setOffset(startOffset);
+        }
+    }
+    public void setSpeed(float speed)
```
liaxim: Javadocs?

GVRf/Framework/framework/src/main/java/org/gearvrf/animation/GVRAnimationEngine.java
```diff
@@ -114,6 +114,7 @@ public GVRAnimation start(GVRAnimation animation) {
             animation.reset();
             mAnimations.add(animation);
         }
+        animation.onStart();
```
liaxim: What is onStart for? Doesn't seem to be used for anything.

GVRf/Framework/framework/src/main/java/org/gearvrf/animation/GVRSkeleton.java
```diff
@@ -90,12 +113,14 @@
...
-    final protected int[] mParentBones;
-    final protected int[] mBoneOptions;
+    protected GVRSceneObject[] mBones;
+    protected int[] mParentBones;
+    protected int[] mBoneOptions;
     final private Quaternionf mTempQuatA = new Quaternionf();
     final private Quaternionf mTempQuatB = new Quaternionf();
     final private Matrix4f mTempMtx = new Matrix4f();
+    private static int[] sTempBoneParents;
+    private static GVRSceneObject sTempRoot;
```
liaxim: Static references to GVRf objects are not a good idea in general. In particular the scene object referenced will never be garbage collected. So it is a memory leak.

GVRf/Framework/framework/src/main/java/org/gearvrf/animation/keyframe/BVHImporter.java
```diff
@@ -250,70 +393,130 @@ public GVRSkeletonAnimation readMotion(BufferedReader buffreader, GVRSkeleton sk
...
+
+                    }
+
                 }
+
                 q.normalize();
                 bindpose.getLocalRotation(boneIndex, b);
                 q.mul(b);
-                int f = (boneIndex * 5) * frameIndex;
+                f = 5 * frameIndex;
```
liaxim: A lot of hard-coded numbers in this method. Hard to read and maintain.

liaxim:
There are still a ton of whitespace-only/formatting changes to jassimp.cpp which should be reverted.

I still don't get while asset loading APIs are added to the GVRAvatar. They might be avatar-specific but why do we have a GVRAssetLoader in first place if not to collect all apis in one place.

liaxim:
gvr-widgetviewer: the initial model is ok; the rest are black.

thomasflynn:
i can confirm that the widgetviewer models are black.  and changing the light to a different color only seems to help on the monkey.  so, something is messed up there.
i haven't seen a speed difference in gvr-morph though.  seems to come up in ~2.x seconds on both master and this branch.

liaxim:
gvr-morph startup is fixed in the last couple of commits. The logging was slowing it down. If I have to guess, it was ``Log.v("MORPH", "Blend Shape Diffs\n%s", dumpDiffs(20));``.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRMeshMorph.java
```diff
@@ -253,6 +250,7 @@ public void setBlendShape(int index, GVRSceneObject obj)
     public void setBlendShape(int index, GVRVertexBuffer vbuf)
     {
         int shapeDescriptorFlags = 0;
+        float[] vec3dataBase;
```
thomasflynn: perhaps i'm going blind, where is this vec3dataBase being used in this method?

NolaDonato:
I squashed this PR and removed the change made to normalmap.vsh. This fixed the monkey and made the PR more readable.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRMeshMorph.java
NolaDonato: I took this out

thomasflynn:
ok, the models in the widgetviewer now display correctly.

thomasflynn:
there's still a lot of unnecessary white-space changes, but it is apparently difficult to revert them.  i've already given a lecture about it.  is there anything else that needs work in this pull-request?

liaxim:
There is still the change to GVRScene.removeAllSceneObjects that reverts a fix from master.

liaxim:
To revert a file to a previous version: ``git checkout c744982d38f0acfee5b6b37f9fafb40cae7572ee  ./framework/src/main/java/org/gearvrf/GVRScene.java``. This will make GVRScene go back to the version prior to the changes in this pull request.

liaxim:
Same revert operation can be done for jassimp.cpp. The only change there is commented out code removal. "diff -w" ignores whitespace changes.

``git diff -w HEAD upstream/master ./framework/src/main/jni/contrib/jassimp/jassimp.cpp``

```
--- a/GVRf/Framework/framework/src/main/jni/contrib/jassimp/jassimp.cpp
+++ b/GVRf/Framework/framework/src/main/jni/contrib/jassimp/jassimp.cpp
@@ -1454,9 +1454,14 @@ static bool loadMaterials(JNIEnv *env, const aiScene* cScene, jobject& jScene)
 
                for (unsigned int p = 0; p < cMaterial->mNumProperties; p++)
                {
+                       //printf("%s - %u - %u\n", cScene->mMaterials[m]->mProperties[p]->mKey.C_Str(), 
+                       //      cScene->mMaterials[m]->mProperties[p]->mSemantic,
+                       //      cScene->mMaterials[m]->mProperties[p]->mDataLength);
+
                        const aiMaterialProperty* cProperty = cMaterial->mProperties[p];
 
                        lprintf("   converting property %s ...\n", cProperty->mKey.C_Str());
+
                        jobject jProperty = NULL;
                        SmartLocalRef refProperty(env, jProperty);
```

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
liaxim: Still curious to know - what is the meaning of -1 here?

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRMeshMorph.java
liaxim: vec3dataBase seems to be still here.

GVRf/Framework/framework/src/main/java/org/gearvrf/animation/GVRSkeleton.java
liaxim: sTempRoot still around. 

Can be changed to
```
        super(root.getGVRContext(), 0);
        setNative(NativeSkeleton.ctor(makeParentBoneIds(root, boneNames)));
```

Will allow makeParentBoneIds to be instance method and sTempRoot to be removed.

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
NolaDonato: This prevents it from trying to load textures after the asset has been loaded. Theoretically it should never get to generateLoadEvent() unless mTextures is 0.  But it will come here again if it tries to load another texture for that asset. It is basically a guard against having the asset generate a load event twice.

thomasflynn: @spodila1 please include Nola's note above as comments for this line in your next patch.  thanks.

GVRf/Extensions/gvrf-physics/src/main/java/org/gearvrf/physics/GVRPhysicsLoader.java
thomasflynn: it looks like the "throws IOException" is added back in.  does gvr-physicsload still compile?

spodila1: Yes it is working and the fix was made in the demo not in the framework.

spodila1:
1. I reverted GVRScene.java to the one in master. I am not sure whether to add the following piece of code back. I think it should be added.

`mSceneRoot = new GVRSceneObject(getGVRContext());
        NativeScene.setSceneRoot(getNative(), mSceneRoot.getNative());
        GVRSceneObject parent = head.getParent();
        if (parent != null)
        {
            parent.removeChildObject(head);
        }`
2. Revert operation is done for jassimp.cpp.
3. Removed float[] vec3dataBase;

GVRf/Framework/framework/src/main/java/org/gearvrf/GVRAssetLoader.java
spodila1: I have included.

liaxim:
GVRScene.removeAllSceneObjects should be the same as in master unless you intentionally want to change it. It took a while to get it right (see issue https://github.com/Samsung/GearVRf/issues/1962) and if you want to change it somehow - let me know how you verify.

liaxim:
Demo apps check out. Last thing is to restore GVRScene.removeAllSceneObjects.

spodila1:
@liaxim restored

Issue #648, opened 2016-06-29
Can't go back to the Activity


mohan6feet:
In gvr-events i set Clicklistner for to access 360 photo Activiity by adding these codes.
    private OnClickListener clickListener=new OnClickListener() {
        @Override
        public void onClick(View v) {
            startActivity(new Intent(EventsActivity.this, Minimal360PhotoActivity.class));
}
}

After seeing the 360 photo i press the back button in Gear VR, i can't go back previous activity, what code to go back and where it specified?

rahul27:
There are ways to do this, but it is recommended that you be cautious with multiple activities. Let me know if this helps:

https://github.com/Samsung/GearVRf/issues/477

mohan6feet:
Is there any coding examples for that?

liaxim:
We don't have examples for that as it is not expected to "just work".

PR #2005, opened 2018-10-19
Made GVRAnchor, GVRPlane components, simplified hit testing

NolaDonato:
1. GVRAnchor & GVRPlane are now components, since they are both attached to scene objects. (This is why we did components in the first place).
2. Simplified AR hit testing, eliminating the need to know about the pass thru object.
    If not hit testing against the pass thru, use collision hit location directly
3. Fixed box collider hit testing

A pull request for the AR avatar demo will follow shortly


ragner:
Hi @NolaDonato,
ARCoreAnchor is going to change the model matrix of sceneObject at every frame step.
It is a problem, for example, if the scene object is a 3d model loaded and it was scaled by the application.  So the 3d model will loose its scale after it has been attached to the anchor.

In my opinion, ARCoreAnchor could have a internal sceneObject and be attached to it. So we could add the sceneObject from the application as child of the  ARCoreAnchor's scene object. 

The same comment to ARCorePlane.

My point is: ARCoreAnchor and ARCorePlane don't change the model matrix of the scene object "attached" to them. Otherwise the application will have to create a sceneObject just to attach the ancho to it and then add the "visible 3d model" as child of this sceneObject to avoid changes at model matrix of the "visible 3d model"

NolaDonato:
The asset loader already puts an extra scene object on most models. The only case for which this is not true is a single mesh with no children (your "andy" model falls in this category). If the application has to add a scene object anyway, why make it internal to MixedReality? I agree that if we could keep it totally internal and hidden from the application we should do this. But we can't actually do that because the application chooses where it wants the model to go in the scene.

ragner:
@dsazulay Could you test this PR with gvr-arcore demo, doing the proper adjusts to the demo?

NolaDonato:
?I have already done this work. I just made a pull request for GearVRf-Demos.


NolaDonato:
I will try to merge them

NolaDonato:
Merge PR #2016 with this PR

NolaDonato:
Verified working with gvr-arcore, gvr-avatar from PR #648 from GearVRf-Demos

NolaDonato:
@ragner, @dsazulay please check my work. I tried to add you as reviewers and failed miserably.

liaxim:
Should we close?

ragner:
We have some fix for this. 
But we are going to make it at sxrsdk.
The issue is related to the update of plane's size

liaxim:
Closing then

PR #2006, opened 2018-10-19
Support vertex color shader in jassimp

NolaDonato:
If vertex colors are found in a mesh, Jassimp will now use GVRVertexColorShader instead of GVRPhongShader. The vertex color shader has been rewritten to allow lighting if normals are also present in the mesh.

This PR has not been tested yet. It is for Mitch to review.


GVRf/Framework/framework/src/main/java/org/gearvrf/shaders/GVRTextureShader.java
```diff
@@ -29,10 +29,6 @@
 
 import org.gearvrf.R;
 
-/**
```
thomasflynn: why was the javadoc removed for a public class?

GVRf/Framework/framework/src/main/res/raw/vcolor_vertex.vsh
```diff
@@ -1,31 +1,113 @@
+#e
```
thomasflynn: is this "#e" a typo?

thomasflynn:
@mwitchwilliams you marked this as 'needs work'.  can you leave a comment as to why?

GVRf/Framework/framework/src/main/res/raw/vcolor_vertex.vsh
mwitchwilliams: This PR uses a different V & F shaders than I used in PLYdemo (https://github.com/gearvrf/GearVRf-Demos/pull/647).  PLYdemo uses the same V & F shaders used by line drawing.

mwitchwilliams: I removed the "#e" and it seemed to make no difference.  I still didn't see the PLY object.
I even re-built Framework and sync'd gradle again, and same results.

liaxim:
Should we close?

PR #2007, opened 2018-10-22
Motion event

asg024:
Clone motion event to use in GVRPicker. 
GearVRf DCO signed off by: Svetlana Gurenkova s.gurenkova@samsung.com


liaxim:
Ok, so what is the reason for this? Assuming it is some kind of a bug - how to reproduce? Who is recycling the newly obtained event?

asg024:
ControllerPick will clear the event as soon as it is processed. 
I am working on implementation fling/scrolling in widgetLib. I need to process MotionEvent passed to  onInside()  callback.

liaxim:
Still not convinced who and when recycles the new event. Will run some tests.

asg024:
GVRCursorController.java, line 1095:
            if (mEvent != null)
            {
                mEvent.recycle();
                mEvent = null;
            }

NolaDonato:
The issue we were having is that the event that was being passed to the ControllerPick class was being recycled by GVRCursorController. So we need to copy the event and give it to ControllerPick. Then it is owned by the thread that uses it and that thread recycles it.

liaxim:
So it is all a bit of mess I agree but what would recycle the event that is in the processedMotionEvent collection then?

Please regression test with the demos; with and without controller.

NolaDonato:
I guess the other thread was supposed to do it. Maybe copying the event is not the right thing to do here. I am not sure. Maybe GVRCursorController should recycle the event it gets passed but make a copy for the other thread.

PR #2008, opened 2018-10-23
PLY file parsing, set vertexColor shader in framework

mwitchwilliams:
Moved setting the vertex color shader to framework from PLY_demo.
A change to Ply demo PR (https://github.com/gearvrf/GearVRf-Demos/pull/647)
to accompany this change.

Tested with ply, obj, glTF and fbx files to make sure only the ply file is affected by this change.


liaxim:
Should we close?

mwitchwilliams:
PLY file format parsing to be supported in SXR.
In addition, this was not the version that would be implemented in GearVR anyway.  Instead, we will likely use https://github.com/Samsung/GearVRf/pull/2006 if this is to be supported in GearVR.

PR #2009, opened 2018-10-24
[maint_v3.x] gradle: make google's maven repo first to avoid weird build errors

liaxim:


liaxim:
Lots of builds are failing. Some because of this, some because of IT issues..

PR #2018, opened 2018-11-07
Cleans up controllers in onPause()

ejsamsung:
Instead of cleaning up the controllers in response to onDestroy(),
clean them up in response to onPause().  If onDestroy() doesn't get
called for some reason, skipping the clean up seems to cause
intermittent issues for subsequent apps; onPause() is, in some
circumstances, a more reliable place to do this.


ejsamsung:
GearVRf-DCO-1.0-Signed-off-by: Erik Jaesler erik.jaesler@samsung.com

Issue #21, opened 2015-04-17
GVRAccelerateDecelerateInterpolator::mapRatio(float) - Negative infinity 


nathan-almeida:
A map ratio for positive infinity value returns NAN instead of NEGATIVE_INFINITY

```java
public void testMapRatioNegativeInfinity() {
        GVRAccelerateDecelerateInterpolator interpolator = GVRAccelerateDecelerateInterpolator.getInstance();
        assertEquals(interpolator.mapRatio(Float.NEGATIVE_INFINITY), Float.NEGATIVE_INFINITY);
}
```

JonShemitz:
I'm not sure there's any point to testing this parameter: This is _not_ user-facing code; it is only called by the animation engine, and the inputs are always in the range [0, 1].

PR #2022, opened 2018-12-04
Abstract body tracking API

NolaDonato:
Same as PR #21 from sxrsdk
GVRBodyTracker class


liaxim:
gvr-avatar no longer builds

NolaDonato:
Oh yeah it needs changes too. I will do a PR against demos.

PR #2033, opened 2019-03-01
Change MeshCollider onHit to return distance in world coordinate

mpolsz:
`MeshCollider::isHit(SceneObject* owner, const glm::vec3& rayStart, const glm::vec3& rayDir)` returned `ColliderData::Distance` in mesh coordinates instead of world coordinates. Other colliders (box and sphere) returns Distance in world coordinates. This created issues with picking objects with scale other than [1f, 1f, 1f]. This fixes picking order and issue with wrong lenght of pointer for PROJECT_CURSOR_ON_SURFACE


NolaDonato:
Run the picker tests and see if they still work, also test with demos that use the controller picking -  gvr-controller, gvr-accessibility, gvr-keyboard

liaxim:
The sanity checks pass. @mpolsz Please add a GearVRf-DCO and we can merge.

mpolsz:
GearVRf-DCO-1.0-Signed-off-by: Mateusz Olszewski mateusz.olszewski@fream.pl

PR #2038, opened 2019-05-13
 No longer leak external textures

liaxim:
- External textures were leaked - both heap and EGL memory
- Adding the OnFrameAvailableListenerImpl to the view scene object is not required but I consider it nice to have
- ExternalRenderer-related entities are obsolete - removed


liaxim:
Verified the demos. @NolaDonato please review.

liaxim:
Verified. Will create SXRSDK PR in a few days.

